PR #13, opened 2018-05-29
Add retry to http util

YeFan000:


src/main/java/net/snowflake/ingest/example/SnowflakeIngestBasicExample.java
```diff
@@ -36,10 +36,10 @@
   // pre-requesites
   private static String account = "s3testaccount";
   private static String user = "snowman";
-  private static String host = "s3testaccount.snowflakecomputing.com";
+  private static String host = "s3testaccount.reg.local";
```
varunganesh88: This should not change.

src/main/java/net/snowflake/ingest/example/SnowflakeIngestBasicExample.java
```diff
@@ -36,10 +36,10 @@
   // pre-requesites
   private static String account = "s3testaccount";
```
varunganesh88: All values in this file should effectively remain unchanged. This is what customers will see in their examples, so we should not be exposing internal testing elements.

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -4,14 +4,20 @@
 
 package net.snowflake.ingest.utils;
 
+import net.snowflake.ingest.SimpleIngestManager;
```
varunganesh88: We should not import SimpleIngestManager here. I think it happened because of the LOGGER addition. The Logger should use HttpUtil.class

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -42,8 +49,51 @@ private static void initHttpClient()
...
+          public boolean retryRequest(
+              final HttpResponse response, final int executionCount,
+              final HttpContext context)
+          {
+            this.executionCount = executionCount;
+            int statusCode = response.getStatusLine().getStatusCode();
+
+            boolean needNextRetry = (statusCode == 408 || statusCode >= 500 )
+                && executionCount < 4;
+            if(executionCount == 4)
```
varunganesh88: Instead of numeric constants in code, I'd recommend using variables above. So this should read something like 
boolean needNextRetry = ... & executionCount < MAX_RETRIES + 1, and define MAX_RETRIES = 3 above.
Do same for status code 408.

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -42,8 +49,51 @@ private static void initHttpClient()
...
+                  String.valueOf(interval) + " millisecond.");
+            }
+
+            return needNextRetry;
+          }
+
+          @Override
+          public long getRetryInterval()
+          {
+            long interval = (1 << executionCount)  * 1000;
```
varunganesh88: Add a comment above here indicating what exactly you are doing for the retry interval. Also add a small comment indicating that the retryInterval is in milli seconds.

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -29,7 +35,8 @@ static public HttpClient getHttpClient()
 
     return httpClient;
   }
-
+  private static final Logger LOGGER =
+      LoggerFactory.getLogger(SimpleIngestManager.class);
```
howryu: logger name should be HttpUtil.class

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -29,7 +35,8 @@ static public HttpClient getHttpClient()
 
     return httpClient;
```
howryu: I know you did not write this code, but getHttpClient() method should be synchronized since this is static method and httpClient object is shared by multiple thread potentially. This is by design since httpClient is thread-safe and we avoid httpClient per SimpleIngestManager. So if two threads are initializing SimpleIngestManager at the same time, there is a race condition here. You don't need to make the change immediately, but just keep in mind, there is a potential improvement here.

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -42,8 +49,51 @@ private static void initHttpClient()
...
+            if(executionCount == 4)
+            {
+              LOGGER.info("Reach the max retry time.");
+            }
+
+            if(needNextRetry && executionCount < 4)
+            {
+              long interval = (1 << executionCount)  * 1000;
+              LOGGER.info("Will retry in " +
+                  String.valueOf(interval) + " millisecond.");
```
howryu: String concatenation in logger method is not a good practice since no matter info level logger is enabled or not, java will first try to compute the string by concatenating string, (which is expensive since object is immutable)

```
LOGGER.info("Sleep time in millisecond: {}", interval);
```
This way, no string will be concatenated, and it will replace `{}` with interval when info level logging is enabled.

PR #14, opened 2018-06-01
Update token with publicKeyFp in Issuer

YeFan000:


src/main/java/net/snowflake/ingest/connection/SecurityManager.java
```diff
@@ -110,14 +120,26 @@
...
     //create our JWT claim object
     JwtClaims claims = new JwtClaims();
 
     //set the issuer to the fully qualified username
-    claims.setIssuer(account + "." + user);
-    LOGGER.info("Creating Token with Issuer {}", account + "." + user);
+    String publicKeyFPInJwt = null;
+    try
+    {
+      publicKeyFPInJwt = getPublicKeyFp(savePublicKey(keyPair.getPublic()));
```
varunganesh88: Please make sure that your code is within 80 columns on your editor. Usually when we exceed 80 columns, we break into the next line.

src/test/java/net/snowflake/ingest/connection/SecurityManagerTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+    PublicKey pubKey = loadPublicKey(storedPublicKey);
+    PrivateKey priKey = loadPrivateKey(storedPrivateKey);
+
+    KeyPair keypair = new KeyPair(pubKey, priKey);
+
+    String accountName = "";
+    String userName = "";
+    SecurityManager securityManager = new SecurityManager(accountName, userName, keypair);
+    securityManager.regenerateToken();
```
varunganesh88: Construction of SecurityManager automatically calls regenerateToken. There is no need for it to be called explicitly here. Library functions that are non essential to users should not be made public.

src/main/java/net/snowflake/ingest/connection/SecurityManager.java
```diff
@@ -176,4 +198,82 @@ String getToken()
...
 
+  /**
+   * Hashes input bytes using SHA-256.
+   *
+   * @param input  input bytes
+   * @param offset offset into the output buffer to begin storing the digest
+   * @param len    number of bytes within buf allotted for the digest
+   * @return SHA-256 hash
+   */
+  public static byte[] sha256Hash(byte[] input, int offset, int len)
```
varunganesh88: There is no reason for all these new methods in this file to be public. Please make them private.

src/main/java/net/snowflake/ingest/connection/SecurityManager.java
```diff
@@ -176,4 +198,82 @@ String getToken()
...
+      MessageDigest md = MessageDigest.getInstance("SHA-256");
+      md.update(input, offset, len);
+      return md.digest();
+    }
+    catch (NoSuchAlgorithmException e)
+    {
+      throw new RuntimeException(e);
+    }
+  }
+  /**
```
varunganesh88: Please leave a newline after the finish of a function.

src/main/java/net/snowflake/ingest/connection/SecurityManager.java
```diff
@@ -110,14 +120,26 @@
...
+      publicKeyFPInJwt = getPublicKeyFp(savePublicKey(keyPair.getPublic()));
+      publicKeyFingerPrint = publicKeyFPInJwt;
+    }
+    catch(Exception e)
+    {
+      LOGGER.info("Failed to Generate public key fingerprint");
+    }
+
+    //set the subject to the fully qualified username
+    claims.setSubject(account + "." + user);
```
varunganesh88: I am a little confused here. Where exactly are we adding the publicKeyFP to the token? We don't seem to be doing that.

src/main/java/net/snowflake/ingest/connection/SecurityManager.java
```diff
@@ -176,4 +198,82 @@ String getToken()
...
+      throw new RuntimeException(e);
+    }
+  }
+  /**
+   * Hashes input bytes using SHA-256.
+   *
+   * @param input input bytes
+   * @return SHA-256 hash
+   */
+  public static byte[] sha256Hash(byte[] input)
```
varunganesh88: Where is this function used in the code?

src/test/java/net/snowflake/ingest/connection/SecurityManagerTest.java
```diff
@@ -0,0 +1,130 @@
...
+    String publicKeyFp = securityManager.getPublicKeyFingerPrint();
+    boolean valid = publicKeyFp.equals(expectedPublicKeyFp);
```
howryu: Use junit Assert. https://junit.org/junit4/javadoc/4.12/org/junit/Assert.html

Otherwise it will not show tests failed.

src/main/java/net/snowflake/ingest/connection/SecurityManager.java
```diff
@@ -110,14 +120,26 @@
...
-    LOGGER.info("Creating Token with Issuer {}", account + "." + user);
+    String publicKeyFPInJwt = null;
+    try
+    {
+      publicKeyFPInJwt = getPublicKeyFp(savePublicKey(keyPair.getPublic()));
+      publicKeyFingerPrint = publicKeyFPInJwt;
+    }
+    catch(Exception e)
+    {
+      LOGGER.info("Failed to Generate public key fingerprint");
```
howryu: Just logging info level probably not enough. Generating public key fingerprint is pretty important, without public key fingerprint, the whole key pair authentication will fail. So just catching the exception and let it go through is probably not a good idea. Maybe throw some exception is better. Adding a class called SnowflakeException and indicating a `INTERNALL_ERROR`

src/main/java/net/snowflake/ingest/connection/SecurityManager.java
```diff
@@ -176,4 +198,82 @@ String getToken()
...
+    return new String(org.apache.commons.codec.binary.Base64.encodeBase64(sha256Hash(input)));
+  }
+  /**
+   * Given a publickey
+   *
+   * @return the fingerprint of public key
+   */
+  private String getPublicKeyFp(String publicKey)
+  {
+    byte[] rawBytes = org.apache.commons.codec.binary.Base64.decodeBase64(publicKey);
```
howryu: You did not specify appache commons package as the dependency in pom files. I don't know how it can work. Anyway, don't use apache commons. It is adding a dependency to the library. I suggest avoiding adding dependency unless absolutely necessary. Since this package requires java8, and java8 include `java.util.Base64`, you should use that class to do base64 encode/decode. 

https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html

src/main/java/net/snowflake/ingest/connection/SecurityManager.java
```diff
@@ -176,4 +198,82 @@ String getToken()
...
+  public static String sha256HashBase64(byte[] input)
+  {
+    return new String(org.apache.commons.codec.binary.Base64.encodeBase64(sha256Hash(input)));
+  }
+  /**
+   * Given a publickey
+   *
+   * @return the fingerprint of public key
+   */
+  private String getPublicKeyFp(String publicKey)
```
howryu: I suggest using name `calculatePublicKeyFp`, `get...` is more like a getter, which just return the private member.

src/main/java/net/snowflake/ingest/connection/SecurityManager.java
```diff
@@ -110,14 +120,26 @@
...
+      publicKeyFingerPrint = publicKeyFPInJwt;
+    }
+    catch(Exception e)
+    {
+      LOGGER.info("Failed to Generate public key fingerprint");
+    }
+
+    //set the subject to the fully qualified username
+    claims.setSubject(account + "." + user);
+    LOGGER.info("Creating Token with subject {}", account + "." + user);
```
howryu: `LOGGER.info("Creating Token with subject {}.{}", account, user);`

src/main/java/net/snowflake/ingest/connection/SecurityManager.java
```diff
@@ -176,4 +198,82 @@ String getToken()
...
+   * Converts PublicKey object to encodedBase64String using RSA scheme.
+   *
+   * @param publicKey
+   * @return
+   * @throws NoSuchAlgorithmException
+   * @throws InvalidKeySpecException
+   */
+  public static String savePublicKey(PublicKey publicKey)
+      throws NoSuchAlgorithmException, InvalidKeySpecException
+  {
```
howryu: You probably don't need a method here. Public key's raw bytes can be calculated using `publicKey.getEncoded()`

src/main/java/net/snowflake/ingest/connection/SecurityManager.java
```diff
@@ -176,4 +198,82 @@ String getToken()
...
+    KeyFactory fact = KeyFactory.getInstance("RSA");
+    X509EncodedKeySpec spec = fact.getKeySpec(publicKey,
+        X509EncodedKeySpec.class);
+    byte[] publicKeyEncode = Base64.getEncoder().encode(spec.getEncoded());
+
+    return new String(publicKeyEncode);
+  }
+
+  public String getPublicKeyFingerPrint()
```
howryu: The only place you probably need this is tests. So package private should be good enough.

src/main/java/net/snowflake/ingest/connection/SecurityManager.java
```diff
@@ -176,4 +198,82 @@ String getToken()
...
+  }
+
+  /**
+   * Hashes input bytes using SHA-256 and converts hash into a string using
+   * Base64 encoding.
+   *
+   * @param input input bytes
+   * @return Base64-encoded SHA-256 hash
+   */
+  public static String sha256HashBase64(byte[] input)
```
howryu: Method sha256HashBase64/sha256Hash is better moved to a separate utility class (maybe under util package namespace), like `Cryptor` so that it can be reused by other classes in the future.

src/main/java/net/snowflake/ingest/utils/Sha256Hash.java
```diff
@@ -0,0 +1,54 @@
+package net.snowflake.ingest.utils;
+
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Base64;
+
+public class Sha256Hash
```
howryu: Rename to `Cryptor`, so that later all cryptography related functions can all be added into this class. 

Also since this is a util function, it is better to make all method static, so that you don't need to instantiate the class.
When calling the static method, you can do something like
```
Cryptor.sha256HashBase64(publicKeyEncoded);
```

You can take a look at GS code, `com.snowflake.security.Cryptor.java`, (I actually copied sha256Hash function from that class)

howryu:
Looks good to me except for that one suggestion. Make sure you can actually ingest file (authentication succeed).

howryu:
LGTM

howryu:
@varunganesh88 Will you merge it?

PR #16, opened 2018-07-12
SimpleIngestManager should not prevent safe shutdown of application

varunganesh88:
Added support for SecurityManager to spawn daemon threads so as to not prevent application from shutting down safely.

### DIFF
--- a/src/main/java/net/snowflake/ingest/utils/ThreadFactoryUtil.java
+++ b/src/main/java/net/snowflake/ingest/utils/ThreadFactoryUtil.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright (c) 2012-2017 Snowflake Computing Inc. All rights reserved.
+ */
+
+package net.snowflake.ingest.utils;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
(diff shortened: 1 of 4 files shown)


howryu:
LGTM

src/main/java/net/snowflake/ingest/utils/ThreadFactoryUtil.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * This class provides support functions for writing a thread factory wrapper
+ */
+public class ThreadFactoryUtil
+{
+  private static final Logger LOGGER =
+      LoggerFactory.getLogger(ThreadFactoryUtil.class);
+
+  public static ThreadFactory poolThreadFactory(final String threadBaseName,
```
drobert: This feels like a slightly less functional version of `Executors.DefaultThreadFactory`. It's largely the same but does not take into consideration a security manager nor priority (although it *does* supply additional uncaught excception handling).

Curious if it makes sense to start with a full copy-paste of the default implementation and just customize the parts in question (esp. `isDaemon`), though thread priority might be interesting as well.

varunganesh88: Hello Robert. You make a good point regarding us having to do more if we were using the thread factory. However, we plan to get rid of this approach of using a thread factory altogether as we feel the better way to do it would be to use a background cache loader to update the security token whenever we find it going out of scope. We have added this thread factory temporarily to facilitate smooth shutdown but will subsequently remove it.

PR #18, opened 2018-07-17
update jdbc version

YeFan000:
When writing kafka connector, it need the latest jdbc driver to build connection using key pair.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -155,7 +155,7 @@
         <dependency>
             <groupId>net.snowflake</groupId>
             <artifactId>snowflake-jdbc</artifactId>
-            <version>3.0.0</version>
+            <version>3.6.6</version>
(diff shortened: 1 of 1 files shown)


howryu:
In theory this should not matter since ingest-sdk-java did not require JDBC as a dependency since in this project JDBC is not really used except for testing purpose. So this dependency should be in test scope. You should define jdbc dependency in your kafka project dependency.

varunganesh88:
JDBC will be used for more than testing since we are not simply going to talk to the ingest service, but also have to create tables, create pipes dynamically per kafka topic.

YeFan000:
Ok, I will test it.

varunganesh88:
I misread this change to the wrong code repository. I think Haowei's comment is a valid one.

PR #19, opened 2018-08-06
move jdbc lib to test scope, and shade all dependencies

binglihub:


pom.xml
```diff
@@ -38,6 +38,7 @@
     <properties>
         <maven.compiler.source>1.8</maven.compiler.source>
         <maven.compiler.target>1.8</maven.compiler.target>
+        <shadeBase>net.snowflake.client.jdbc.internal</shadeBase>
```
varunganesh88: We agreed this has to be changed.

pom.xml
```diff
@@ -48,6 +49,66 @@
             </resource>
         </resources>
         <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
```
varunganesh88: Please add a comment telling what the repackaging is for

varunganesh88:
Looks good to me.

PR #21, opened 2018-10-19
Remove RSA Public Key

binglihub:
1. remove RSA public key requirement
2. upgrade Jackson libs tp 2.9.6
3. create integration test

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -13,6 +13,7 @@
     <description>Snowflake Ingest SDK</description>
     <url>https://www.snowflake.net/</url>
 
+
     <licenses>
       <license>
         <name>The Apache Software License, Version 2.0</name>
@@ -39,6 +40,8 @@
         <maven.compiler.source>1.8</maven.compiler.source>
         <maven.compiler.target>1.8</maven.compiler.target>
         <shadeBase>net.snowflake.ingest.internal</shadeBase>
+        <jacoco.skip.instrument>true</jacoco.skip.instrument>
+        <jacoco.version>0.8.1</jacoco.version>
     </properties>
 
     <build>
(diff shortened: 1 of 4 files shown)


src/main/java/net/snowflake/ingest/SimpleIngestManager.java
```diff
@@ -232,30 +240,116 @@ public SimpleIngestManager(String account, String user, String pipe,
...
+   *                                  RSA algorithm
+   * @throws InvalidKeySpecException  if private key or public key is invalid
+   */
+  private KeyPair createKeyPairFromPrivateKey(PrivateKey privateKey) throws
+      NoSuchAlgorithmException, InvalidKeySpecException
+  {
+    KeyFactory kf = KeyFactory.getInstance("RSA");
+
+    //generate public key from private key
+    RSAPrivateCrtKey privk = (RSAPrivateCrtKey) privateKey;
```
howryu: Throw proper user error in case user is passing in a PrivateKey not instancesof RSAPrivateCrtKey

src/test/java/net/snowflake/ingest/SimpleIngestSuite.java
```diff
@@ -0,0 +1,167 @@
...
+    TestUtils.executeQuery(
+        "drop table if exists " + tableName
+    );
+  }
+
+  /**
+   * ingest test
+   */
+  @Test
+  public void test1() throws Exception
```
howryu: Maybe a better name? testSimpleIngest?

src/test/java/net/snowflake/ingest/SimpleIngestSuite.java
```diff
@@ -0,0 +1,167 @@
...
+
+  }
+
+  /**
+   * Remove test table and pipe
+   */
+  @After
+  public void afterAll()
+  {
+    //System.out.println(111);
```
howryu: remove

test_files/test1.csv
```diff
@@ -0,0 +1,3 @@
+"a",1
```
howryu: Maybe add move test data to test resource directory? `src/test/resource`

PR #27, opened 2018-10-26
Move gpg.keyname to settings.xml because of maven 3.5 cli parse bug

Preston4tw:
This is an attempt to fix an issue identified when attempting to update Maven. It looks like Maven 3.5 may have an issue parsing complex command line options, see https://issues.apache.org/jira/browse/MNG-6385. gpg.keyname appears to be the only option that requires extra quoting so I've moved that into the settings.xml file.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -5,9 +5,6 @@ THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 if [ -z "$GPG_KEY_ID" ]; then
   echo "[ERROR] Key Id not specified!"
   exit 1
-else
-# name can include spaces
-  MVN_GPG_OPTIONS+=("\"-Dgpg.keyname=$GPG_KEY_ID\"")
 fi
 
 if [ -z "$GPG_KEY_PASSPHRASE" ]; then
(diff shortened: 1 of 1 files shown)


howryu:
Thanks! I will try to do another push tomorrow.

howryu:
I think it is the best to put everything into the pom file. I will try that.

howryu:
@Preston4tw I moved those config into pom.xml in #28
And here is output. I still cannot sign the package. 

```
2018-10-26 15:50:15 can't connect to `/home/jenkins/.gnupg/S.gpg-agent': No such file or directory
2018-10-26 15:50:15 Error opening terminal: unknown.
2018-10-26 15:50:16 gpg-agent[26500]: command get_passphrase failed: End of file
2018-10-26 15:50:20 gpg: problem with the agent: IPC write error
2018-10-26 15:50:20 gpg: Invalid passphrase; please try again ...
2018-10-26 15:50:20 gpg: problem with the agent: Not supported
2018-10-26 15:50:20 gpg: no default secret key: General error
2018-10-26 15:50:20 gpg: signing failed: General error
```

howryu:
Nvm, seems working for me now

Issue #31, opened 2019-01-22
RequestBuilder class does not properly construct the URL of the host account. Needs to account for the account name.


rtempleton:
Line 226. It should append the accountName to hostName

howryu:
Closed since this is not an issue.

PR #32, opened 2019-01-22
Fix for issue: #31

rtempleton:
https://github.com/snowflakedb/snowflake-ingest-java/issues/31


howryu:
I don't think we should do this. Host name is used for dns resolution. So let's say you have an account `testaccount`. The host name you specified should be `testaccount.us-east-1.snowflakecomputing.com`. So hostname specified by user should contain account name.

Issue #11, opened 2018-01-19
Is IngestResponse.requestId always null?


ml10:
After making an `ingestFiles` request, the `requestId` of `IngestResponse` always appears `null`.

This means that if you call `getRequestUUID()`, it results in an NPE because `java.util.UUID.fromString` will throw on null input.

Here's how I ran across this in my code, trying to get the `requestUUID` out from the `response`. 
```
      }.flatMap{ (response: IngestResponse) =>
        println(s"ingestFiles response: $response")
        val optResponse = Option(response)
        println(s"optResponse: $optResponse")
        optResponse.map(_.getRequestUUID)
      }
```

Sorry this is not java code, but hopefully the intent is clear. Here's the result of a run:

```
[run-main-0] INFO net.snowflake.ingest.SimpleIngestManager - Attempting to unmarshall insert response - HttpResponseProxy{HTTP/1.1 200 OK [Content-Type: application/json, Date: Fri, 19 Jan 2018 22:58:35 GMT, Server: nginx, Strict-Transport-Security: max-age=31536000, X-Frame-Options: deny, Content-Length: 88, Connection: keep-alive] ResponseEntityProxy{[Content-Type: application/json,Content-Length: 88,Chunked: false]}}
ingestFiles response: null
optResponse: Some(null)
java.lang.NullPointerException
  at java.util.UUID.fromString(UUID.java:192)
  at net.snowflake.ingest.connection.IngestResponse.getRequestUUID(IngestResponse.java:27)
  at io.healthgrades.dataload.evented.SnowpipeRestClient$.$anonfun$ingestResponse$5(SnowpipeRestClient.scala:77)
  ...
```

sziegler:
This sounds like a bug.  We will look into this.  Which endpoint were you calling?

Just FYI the REST endpoints are simply including the requestId given by the REST request in the response for convenience.  While the Java API may use UUIDs, this is not required by the REST endpoint.  At  first glance, it appears that here the server is not echoing the requestId properly.

ml10:
This was with the `/v1/data/pipes/{pipeline}/insertFiles` endpoint. 

I can see how the UUID isn't strictly needed, but as `net.snowflake.ingest.SimpleIngestManager` is now, if I don't provide one, the `SimpleIngestManager` generates one for me before sending the request.

Hartland-3:
[...]

PR #39, opened 2019-11-15
Modify IngestResponse POJO to support responseCode and avoid null returns

jaympatel-sf:
Fixes: https://github.com/snowflakedb/snowflake-ingest-java/issues/11

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -74,4 +74,8 @@ If you would just like to build the jar in the source directory, you can run
 
 However, for general usage, pulling a pre-built jar from maven is recommended.
 
+Testing (SimpleIngestIT Test)
+=====================================
 
+- Modify ``TestUtils.java`` file and replace *PROFILE_PATH* with ``profile.json.example`` for testing.
+- Use an unencrypted version(Only for testing) of private key while generating keys(private and public pair) using OpenSSL.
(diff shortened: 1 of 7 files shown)


README.rst
```diff
@@ -74,4 +74,8 @@ If you would just like to build the jar in the source directory, you can run
 
 However, for general usage, pulling a pre-built jar from maven is recommended.
 
+Testing (SimpleIngestIT Test)
+=====================================
 
+- Modify ``TestUtils.java`` file and replace *PROFILE_PATH* with ``profile.json.example`` for testing.
+- Use an unencrypted version(Only for testing) of private key while generating keys(private and public pair) using OpenSSL.
```
gjorgji: Can you give an example or pointer to a website that describes how to do this?

src/test/java/net/snowflake/ingest/TestUtils.java
```diff
@@ -20,7 +20,7 @@
 
 public class TestUtils
 {
-  //profile path
+  // profile path, follow readme for the format
   private final static String PROFILE_PATH = "profile.json";
```
gjorgji: We can just refer to profile.json.example here directly, so whoever runs the test can modify the parameters in this file.

jaympatel-sf: I don't think we should do this. There is a file called ```profile.json.enc``` which is decrypted into ```profile.json```

The file ```.travis.yml```  decrypts the encrypted file using openssl to hide the parameters used to call an endpoint. If we modify this, it will not run the travis tests successfully.

README.rst
jaympatel-sf: The links are already mentioned above in readme file in the section RSA key pair, but I can add it here for more clarity.

PR #40, opened 2019-11-15
Updating snowflake-ingest-sdk, jackson-core, jackson-databind, nimbus-jose-jwt

gjorgji:
Updating snowflake-ingest-sdk, jackson-core, jackson-databind to compliant versions.


jaympatel-sf:
Throws error at: 
https://github.com/snowflakedb/snowflake-ingest-java/blob/master/src/main/java/net/snowflake/ingest/connection/ServiceResponseHandler.java#L36

PR #43, opened 2019-11-16
Fips Compliance Test with Bouncy castle

sfc-gh-japatel:


jaympatel-sf:
The previous commits were part of bug fix, this new commit introduces test which will confirm this SDK works with FIPS BC implementation.

src/test/java/net/snowflake/ingest/FipsCompliantIngestIT.java
```diff
@@ -0,0 +1,328 @@
...
+   * Creates the stages and files we'll use for this test
+   */
+  private static void setup(Set<String> files, String filesLocation)
+      throws Exception
+  {
+    //use the right database
+    IngestExampleHelper.doQuery(conn, "use database " + DATABASE);
+
+    //use the right schema
+    IngestExampleHelper.doQuery(conn, "use schema " + SCHEMA);
```
gjorgji: Why dont you do "create or replace" or "create if not exists"?

src/test/java/net/snowflake/ingest/FipsCompliantIngestIT.java
```diff
@@ -0,0 +1,328 @@
...
+ * <b>Prerequisites:</b>
+ * You must have a valid set of Snowflake credentials
+ * We use encrypted key pair so follow the instructions written on the
+ * documentation and also in readme of this repo.
+ */
+public class FipsCompliantIngestIT
+{
+  // Details required to connect to Snowflake over jdbc and establish
+  // pre-requesites
+  private static final String ACCOUNT = "s3testaccount";
```
gjorgji: Will this test now run as part of the Travis tests?

src/test/java/net/snowflake/ingest/FipsCompliantIngestIT.java
```diff
@@ -0,0 +1,328 @@
...
+      {
+        this.filesWatchList = files;
+      }
+      String beginMark = null;
+
+      public HistoryResponse call()
+          throws Exception
+      {
+        HistoryResponse filesHistory = null;
+        while (true)
```
gjorgji: You probably want to have a failswitch here. ex. if after 10 iteration we still haven't received the expected history response, fail the test.

src/test/java/net/snowflake/ingest/FipsCompliantIngestIT.java
```diff
@@ -0,0 +1,328 @@
...
+        {
+          Thread.sleep(500);
```
gjorgji: Nit - use constants of this an other numbers, paths, ect. you directly use in the code.

src/test/java/net/snowflake/ingest/FipsCompliantIngestIT.java
```diff
@@ -0,0 +1,328 @@
...
+    setup(files, filesLocationUrl);
+  }
+
+  /**
+   * Remove test table and pipe
+   */
+  @After
+  public void afterAll()
+  {
+    IngestExampleHelper.doQuery(conn, "drop pipe if exists " + FQ_PIPE);
```
gjorgji: Drop db would be easier?

Also, I dont think we will be able to support concurrent executions of two such tests, as one might be dropping the objects during the test cleanup that the other one is still using them.

src/test/java/net/snowflake/ingest/FipsCompliantIngestIT.java
```diff
@@ -0,0 +1,328 @@
...
+                                      FQ_PIPE, keypair, SCHEME,
+                                      HOST, PORT);
+    setup(files, filesLocationUrl);
+  }
+
+  /**
+   * Remove test table and pipe
+   */
+  @After
+  public void afterAll()
```
gjorgji: Shouldn't you also be doing whatever opposite of BouncyCastleFIPSInit.init() is?

src/test/java/net/snowflake/ingest/FipsCompliantIngestIT.java
```diff
@@ -0,0 +1,328 @@
...
+          .sum();
+
+      final Set<String> loadedStatus = historyRangeResponse.files
+          .stream()
+          .map(HistoryResponse.FileEntry::getStatus)
+          .map(IngestStatus::toString)
+          .collect(Collectors.toSet());
+
+      // Sum of all rows = 45
+      Assert.assertEquals(45, actualSumOfRowsInserted);
```
gjorgji: Why 45? Would be better to use constants and derive this number, rather than have a "magic value" for it :)

src/test/java/net/snowflake/ingest/FipsCompliantIngestIT.java
jaympatel-sf: Yes, thanks Gjorgji. I am running into similar issue now. My Travis tests are failing, and are probably due to this reason. I will figure out how to run in sequentially.

src/test/java/net/snowflake/ingest/FipsCompliantIngestIT.java
jaympatel-sf: Yes, much better, will do that!

src/test/java/net/snowflake/ingest/FipsCompliantIngestIT.java
jaympatel-sf: Apparently yes and is is failing. I am investigating that. Also, I don't think this test should be part of Travis. (If you think otherwise), I will figure out how to exclude this.

src/test/java/net/snowflake/ingest/FipsCompliantIngestIT.java
jaympatel-sf: Agreed, will add few more asserts.

src/test/java/net/snowflake/ingest/FipsCompliantIngestIT.java
gjorgji: Agree - it is a very specific test that we dont need to run frequently.

jaympatel-sf:
I am still having issues running both the tests on my local machine due to usage of two bouncycastle libraries. I tried undo-ing init and there are still problems. 
We have two tests, SimpleIngestIT and one I added in this PR - FipsComplianceIngestIT

- SimpleIngestIT uses the SecurityProvider from _**bcprov-ext-jdk15on**_ while.
- FipsComplianceIngestIT uses _**bc-fips**_

I had a brief discussion with Damien, (As we already knew) the purpose of this test was to verify client can run the application using this SDK with modifications in JVM to use BC fips libraries, that is being verified. 

Unless you feel otherwise, I will hence abstain from pushing this code, because only one of the tests can be added and run successfully.

If we want, we can add a note in readme saying our APIs will work with fips-compliant mode on.

Issue #17, opened 2018-07-12
History FileEntry getLastInsertTime() and getTimeReceived() should return time as LocalDateTime and not LocalDate


varun-ganesh:
Because we are returning LocalDate instead of LocalDateTime, the time details are lost in the conversion

PR #49, opened 2020-04-22
SNOW-153193 snowflake-ingest-java bug fix

sfc-gh-aalmahmood:
ingestManager.getHistory.getLastInsertTime returns a date, it should return a timestamp.


sfc-gh-japatel:
LGTM! Thank you!
Lets also link issue #17 as a fix while committing and close that issue.

PR #59, opened 2020-07-22
add semgrep check

sfc-gh-jfan:
Semgrep is a command-line tool for static analysis. We can use pre-built or custom rules to enforce code and security standards in codebase. When PR is raised, semgrep will check using the rule defined in .semgrep.yml file to detect potential risky code which causes security flaws.


sfc-gh-japatel:
LGTM!

PR #66, opened 2020-11-07
Disable SL4j shading to enable custom properties

sfc-gh-japatel:
This will not bring the slf4j package into internal package. Shading is not needed for slf4j, it stops clients using the SDK not use their default/custom properties. 
https://maven.apache.org/plugins/maven-shade-plugin/

Test:
Generate a jar and use it on my own sample app. 

`log4j.logger.net.snowflake.ingest.connection=ERROR, <Any Appender>
log4j.additivity.net.snowflake.ingest.connection=false`


sfc-gh-jliang (approved):
LGTM, good finding

PR #69, opened 2021-01-07
SNOW-246819 Modify Ingest JAVA SDK and add showSkippedFiles and unmatchedPatternFiles

sfc-gh-rramachandran:
Description
New properties that were added to the ingest api's response are now added to the SDK.

Testing
Added an end-to-end test that uses pattern and returns an unmatched file.


src/test/java/net/snowflake/ingest/SimpleIngestIT.java
```diff
@@ -114,61 +138,105 @@ public void testSimpleIngest() throws Exception
...
+    );
+
+    //create ingest manager
+    SimpleIngestManager manager = TestUtils.getManager(pipeWithPatternName);
+    Set<String> files = new HashSet<>();
+    files.add(TEST_FILE_NAME);
+    files.add(TEST_FILE_NAME_2);
+
+    //get an insert response after we submit
+    IngestResponse insertResponse = manager.ingestFiles(SimpleIngestManager.wrapFilepaths(files), null);
```
sfc-gh-japatel, line 174: Should we also verify if showskippedfiles is not present in the response object? https://snowflakecomputing.atlassian.net/browse/SNOW-75404
(Based on this JIRA discussion)

sfc-gh-japatel: i.e if the response should include two newly added fields?
(if they are null/empty we might as well ignore them for being backward compatible?)

sfc-gh-rramachandran: Currently they are null if the 2 new properties are not returned from the service. I can add an assert for it. One way of ignoring them (for backward compatible reasons) is by versioning the response object and adding new APIs to return the new versioned object, but that seems too heavy and unnecessary. Is there any other way of ignoring them in Java? 

I have a clarifying question regarding back-compat: This change is introducing new behavior but is not changing any existing behavior, isn't it backward compatible already? Or are we considering addition of new properties to the response as a breaking change?

sfc-gh-japatel: I think we can ignore it using jackson annotation. https://www.baeldung.com/jackson-ignore-null-fields

It wont be a breaking change if we release a new version. But by backward compatibility, I meant if customers adopt this new released version, are they expected to change anything if they dont want to utilize new fields? (If my understanding is correct for B/W compatibility)

sfc-gh-rramachandran: The ignore null in jackson ignores it during serialization from what I can gather, whereas, this will be during de-serialization. 

I don't think the customers have to change anything if they adopt the new version but don't want any of the new fields as these are additional new fields on the response and are read-only. IMO, this shouldn't break back-compat (The basic test worked without any changes needed in them after I added these 2 new fields).

sfc-gh-japatel: I think we are still fine as this doesnt qualify for "no B/W compatible change".
https://bambielli.com/til/2018-01-12-what-is-a-breaking-change/
But, I still feel we should remove null fields.

sfc-gh-japatel: Ahh true for only serialization. Are we ensuring that the response from GS will not contain null fields?

sfc-gh-rramachandran: Yup, GS is ensuring only non empty fields are returned
https://github.com/snowflakedb/snowflake/blob/a2d9148b9478b1164f6bde1fdfc01eea5089b723/GlobalServices/src/main/java/com/snowflake/ingest/PushFileResponse.java#L12

sfc-gh-japatel: Sounds good, thanks for confirming!

sfc-gh-japatel (approved):
lgtm! Thank you.

PR #70, opened 2021-01-14
Add http proxy username and password support in httpclient for Kafka Connector

sfc-gh-japatel:
SNOW-260640

Background
- Kafka connector added configuration which allows to setup proxy host. 
- This proxy host can have authentication based on username and password. 
- The proxy host is able to call snowpipe api when there are not user/pwd, but fails to do it if proxy enforces one. 

This change allows snowpipe KC to call snowpipe API from KC cluster with proxy on. 

This is a naive implementation of proxy user/password at the moment since KC implements it at JVM level instead of KC level. 
Because this version is already released in KC, we would tackle this naive implementation first and then set proxy parameter which are unique per kafka connector. 

Future update:
- Instead of fetching it from JVM env, we would like to take it in a request. (SimpleIngestManager ctor)

### DIFF
--- a/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
+++ b/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
@@ -4,15 +4,21 @@
 
 package net.snowflake.ingest.utils;
 
+import com.google.common.base.Strings;
 import org.apache.http.HttpHost;
 import org.apache.http.HttpRequest;
 import org.apache.http.HttpResponse;
 import org.apache.http.NoHttpResponseException;
+import org.apache.http.auth.AuthScope;
+import org.apache.http.auth.Credentials;
+import org.apache.http.auth.UsernamePasswordCredentials;
+import org.apache.http.client.CredentialsProvider;
 import org.apache.http.client.HttpClient;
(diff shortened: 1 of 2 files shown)


sfc-gh-jliang (approved):
lgtm, is it possible to add a test for this?

sfc-gh-japatel:
> lgtm, is it possible to add a test for this?

Sure, let me add some test.

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -34,6 +40,10 @@
   private static String USE_PROXY = "http.useProxy";
   private static String PROXY_HOST = "http.proxyHost";
   private static String PROXY_PORT = "http.proxyPort";
+
+  public static final String HTTP_PROXY_USER = "http.proxyUser";
```
sfc-gh-rramachandran, line 44: private?

sfc-gh-rramachandran (approved):
LGTM

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-japatel: Ahh yes.

PR #74, opened 2021-01-21
Fix slf4j shading

sfc-gh-japatel:
Another one of Shading issue. 
Although I removed shaded jar from relocated (internal) package, slf4j was still part of shaded jar. 
Screen Shot 2021-01-20 at 8 33 29 PM
Screen Shot 2021-01-20 at 8 33 49 PM

This change, reduces scope of base slf4j to provided and removing implementation (slf4j-simple) from shaded plugin. 

This is what it looks like now:
Screen Shot 2021-01-21 at 12 31 48 AM

## Root cause:
- KC was not able to use custom log4j properties which used to generate file before 0.9.12 (snowpipe sdk where we release unshaded slf4j jar)
- Priority:
- - Urgent

## Testing
- Tested local jar in KC and tests are passing. sf.log file is also getting created.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -7,7 +7,7 @@
     <!-- Arifact name and version information -->
     <groupId>net.snowflake</groupId>
     <artifactId>snowflake-ingest-sdk</artifactId>
-    <version>0.10.0</version>
+    <version>0.10.1</version>
     <packaging>jar</packaging>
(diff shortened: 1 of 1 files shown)


sfc-gh-giyer (approved):
lgtm

PR #77, opened 2021-02-02
Fail github actions on mvn verify failure

sfc-gh-japatel:
- Remove travis related code.


src/test/java/net/snowflake/ingest/SimpleIngestIT.java
```diff
@@ -233,7 +233,7 @@ private void getHistoryAndAssertLoad(SimpleIngestManager manager, String test_fi
     //try to wait until the future is done
     try {
       //wait up to 2 minutes to load
```
sfc-gh-rramachandran, line 235: NIT: Fix comment. Is it taking more than 2 minutes for the load to complete? Should we reduce the number of files?

.github/workflows/End2EndTest.yml
```diff
@@ -30,5 +30,6 @@ jobs:
         env:
           JACOCO_COVERAGE: true
           WHITESOURCE_API_KEY: ${{ secrets.WHITESOURCE_API_KEY }}
+        continue-on-error: false
```
sfc-gh-rramachandran, line 33: Surprised this is not the default behavior :)

sfc-gh-rramachandran (approved):
LGTM!

src/test/java/net/snowflake/ingest/SimpleIngestIT.java
sfc-gh-japatel: Will fix. We are just adding 1 file, shouldnt take too long. I will revert to 2 mins.

PR #78, opened 2021-02-03
Trim account name if it contains a '.'

sfc-gh-japatel:
SNOW-257844

### DIFF
--- a/src/main/java/net/snowflake/ingest/connection/SecurityManager.java
+++ b/src/main/java/net/snowflake/ingest/connection/SecurityManager.java
@@ -86,7 +86,7 @@ final class SecurityManager implements AutoCloseable
     {
       throw new IllegalArgumentException();
     }
-    account = accountname.toUpperCase();
+    account = parseAccount(accountname);
     user = username.toUpperCase();
 
     //create our automatic reference to a string (our token)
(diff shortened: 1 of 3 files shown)


src/main/java/net/snowflake/ingest/connection/SecurityManager.java
```diff
@@ -120,6 +120,33 @@
...
+   * @return initial part of account name if originally it contained a ".", otherwise return the same accountName.
+   * @see <a href="https://github.com/snowflakedb/snowflake-connector-python/blob/master/src/snowflake/connector/util_text.py#L227">Python Connector</a>
+   */
+  private String parseAccount(final String accountName)
+  {
+    String parseAccount = null;
+    if (accountName.contains("."))
+    {
+      final String[] accountParts = accountName.split("[.]");
+      if (accountParts.length > 1)
```
sfc-gh-rramachandran, line 138: NIT: Is there a valid scenario where this is false? Can we remove it?

sfc-gh-rramachandran (approved):
LGTM!

src/main/java/net/snowflake/ingest/connection/SecurityManager.java
sfc-gh-japatel: Very rare. In cases like they pass an accountname which ends in a dot may be?

sfc-gh-hchaturvedi (approved):
LGTM. Can you please ensure it is tested against a prod env before we merge / release it?

sfc-gh-japatel:
> LGTM. Can you please ensure it is tested against a prod env before we merge / release it?

I have tested it on local machine. but not on an actual account which has a "." in its name. Do you have recommendation or a test account which has dot in name?

sfc-gh-japatel:
Tested this on `testaccount.reg.snow...com`

PR #79, opened 2021-02-09
Bump up Java ingest sdk

sfc-gh-japatel:
Release notes:
- Fixes issue where JWT token invalid was returned from Snowflake servers when accountNames had `.` in them. 
  - We will trim account name's suffix after `.`
- Made it functionally similar to python connector.


sfc-gh-rramachandran (approved):
LGTM

PR #80, opened 2021-03-02
Change slf4j-simple to test scope

sinogermany:
`slf4j-simple` should not be a runtime dependency. Log provider should be specified by the user of the library.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -258,6 +258,7 @@
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-simple</artifactId>
             <version>1.7.21</version>
+            <scope>test</scope>
         </dependency>
 
         <!-- java.lang.NoClassDefFoundError: javax/xml/bind/JAXBException -->


sfc-gh-japatel:
Thanks for this PR, currently we dont allow merging non-contributors in this repo. 
This seems a legit change, I will try to merge this.

sfc-gh-japatel:
On a side note, what issues do you see if this is not a test dependency? Did you mean consumers of this SDK wont be able to set up/override their own logging?

sinogermany:
> On a side note, what issues do you see if this is not a test dependency? Did you mean consumers of this SDK wont be able to set up/override their own logging?

Thanks @sfc-gh-japatel . So `slf4j-api` contains the _contract_ per se, and every single application is using different bindings, like `slf4j-simple` or `slf4j-log4j`. As a library, it's better not to introduce any bindings as runtime dependencies so that the applications that use this library can choose which slf4j binding they want to use.

In our case, we are using this library in one of our applications and we are using `slf4j-log4j` but because the library specifies `slf4j-simple` as a runtime dependency, it's becoming indeterministic which binding `slf4j` will choose. It ended up choosing `slf4j-simple` in our case which invalidated all our log4j configurations. 

As stated in the PR it's better to require an `slf4j` binding on the app itself than forcing `slf4j-simple` binding as a runtime requirement for this library.

Hope that makes sense :) 

Thanks for reviewing this PR.

sfc-gh-japatel (approved):
lgtm! Thank you!

sfc-gh-japatel:
Merged: https://github.com/snowflakedb/snowflake-ingest-java/pull/83

PR #84, opened 2021-03-30
Update README to include instructions on running the examples

sfc-gh-tzhang:
Run into this issue during my setup, so updating the README to make sure others know what to do if they need to run examples in the example folder

The change looks nasty, but the result looks good since I really want to have overline text but it doesn't work with ```
image


sfc-gh-tjones (approved):

sfc-gh-japatel (approved):
lgtm

PR #86, opened 2021-04-01
SNOW-331292 Add GoogleJavaFormat as a GH Action

sfc-gh-afarra:
SNOW-331292 Add GoogleJavaFormat as a GH Action

Runs on pushes to master and all PRs


sfc-gh-tzhang:
Thanks for the help!

PR #90, opened 2021-05-12
SNOW-225155 add custom user agent in snowpipe requests

sfc-gh-japatel:
**Background:** 
- This is required for Kafka connector. 
- KC will use additionInformation which is passed in by the configuration parameter, we will use this information to differentiate whether a KC is self hosted, confluent hosted or potentially by AWS as MSK. 

**More Info:**
- Modify existing user agent string since that is not matching with current standards.
- We cannot have validation on SDK side since there is no validation for snowpipe dispatcher as of yet (Server side)
- Added test
- Technically you would require the user agent information from only single Snowpipe API (ideally `insertFiles`) since the subsequent calls should all happen from same environment, it can be debatable to just add this into `/insertFiles` endpoint but there is no harm in adding to all requests to be consistent.

### DIFF
--- a/src/main/java/net/snowflake/ingest/SimpleIngestManager.java
+++ b/src/main/java/net/snowflake/ingest/SimpleIngestManager.java
@@ -4,7 +4,7 @@
 
 package net.snowflake.ingest;
 
-import static net.snowflake.ingest.connection.RequestBuilder.DEFAULT_HOST;
+import static net.snowflake.ingest.connection.RequestBuilder.DEFAULT_HOST_SUFFIX;
 
 import com.google.common.base.Strings;
 import java.io.IOException;
@@ -71,6 +71,8 @@ public class SimpleIngestManager implements AutoCloseable {
 
     // Hostname to connect to, default will be RequestBuilder#DEFAULT_HOST
     private String hostName;
+
+    private PrivateKey privateKey;
 
     /**
      * getAccount - returns the name of the account this builder will inject into the IngestManager
@@ -184,7 +186,8 @@ public class SimpleIngestManager implements AutoCloseable {
      */
     public SimpleIngestManager build() {
       if (Strings.isNullOrEmpty(hostName)) {
-        return new SimpleIngestManager(account, user, pipe, DEFAULT_HOST, keypair, userAgentSuffix);
+        return new SimpleIngestManager(
+            account, user, pipe, DEFAULT_HOST_SUFFIX, keypair, userAgentSuffix);
       }
       return new SimpleIngestManager(account, user, pipe, hostName, keypair, userAgentSuffix);
     }
@@ -305,12 +308,16 @@ public class SimpleIngestManager implements AutoCloseable {
    * Using this constructor for Builder pattern. KeyPair can be passed in now since we have
    * made @see {@link SimpleIngestManager#createKeyPairFromPrivateKey(PrivateKey)} public
    *
-   * @param account
-   * @param user
-   * @param pipe
-   * @param hostName
-   * @param keyPair
-   * @param userAgentSuffix
+   * @param account The account into which we're loading Note: account should not include region or
+   *     cloud provider info. e.g. if host is testaccount.us-east-1.azure .snowflakecomputing.com,
+   *     account should be testaccount. If this is the case, you should use the constructor that
+   *     accepts hostname as argument
+   * @param user the user performing this load
+   * @param pipe the fully qualified name of the pipe
+   * @param hostName the hostname
+   * @param keyPair keyPair associated with the private key used for authentication. See @see {@link
+   *     SimpleIngestManager#createKeyPairFromPrivateKey} to generate KP from p8Key
+   * @param userAgentSuffix user agent suffix we want to add.
    */
   public SimpleIngestManager(
       String account,
--- a/src/main/java/net/snowflake/ingest/SimpleIngestManager.java
+++ b/src/main/java/net/snowflake/ingest/SimpleIngestManager.java
@@ -4,6 +4,9 @@
 
 package net.snowflake.ingest;
 
+import static net.snowflake.ingest.connection.RequestBuilder.DEFAULT_HOST;
+
+import com.google.common.base.Strings;
 import java.io.IOException;
 import java.net.URISyntaxException;
 import java.security.KeyFactory;
@@ -62,6 +65,12 @@ public class SimpleIngestManager implements AutoCloseable {
 
     // the key pair we want to use to authenticate
     private KeyPair keypair;
+
+    // Used to add additional information to User-Agent(Http Header)
+    private String userAgentSuffix;
+
+    // Hostname to connect to, default will be RequestBuilder#DEFAULT_HOST
+    private String hostName;
 
     /**
      * getAccount - returns the name of the account this builder will inject into the IngestManager
@@ -144,11 +153,40 @@ public class SimpleIngestManager implements AutoCloseable {
     }
 
     /**
+     * Get the set user agent suffix. (This is in additional to the default one.)
+     *
+     * <p>It can be null or empty.
+     */
+    public String getUserAgentSuffix() {
+      return userAgentSuffix;
+    }
+
+    /* Sets the user agent suffix as part of this SimpleIngestManager Instance */
+    public Builder setUserAgentSuffix(String userAgentSuffix) {
+      this.userAgentSuffix = userAgentSuffix;
+      return this;
+    }
+
+    /* Gets the host name */
+    public String getHostName() {
+      return hostName;
+    }
+
+    /* Sets host name, if not explicitly used, a default will be used. */
+    public Builder setHostName(String hostName) {
+      this.hostName = hostName;
+      return this;
+    }
+
+    /**
      * build - returns a new instance of SimpleIngestManager using the information set in this
      * builder object
      */
     public SimpleIngestManager build() {
-      return new SimpleIngestManager(account, user, pipe, keypair);
+      if (Strings.isNullOrEmpty(hostName)) {
+        return new SimpleIngestManager(account, user, pipe, DEFAULT_HOST, keypair, userAgentSuffix);
+      }
+      return new SimpleIngestManager(account, user, pipe, hostName, keypair, userAgentSuffix);
     }
   }
 
@@ -172,26 +210,7 @@ public class SimpleIngestManager implements AutoCloseable {
   // the request builder who handles building the HttpRequests we send
   private final RequestBuilder builder;
 
-  /**
-   * init - Does the basic work of constructing a SimpleIngestManager that is common across all
-   * constructors
-   *
-   * @param account The account into which we're loading
-   * @param user the user performing this load
-   * @param pipe the fully qualified name of pipe
-   * @param keyPair the KeyPair we'll use to sign JWT tokens
-   */
-  private void init(String account, String user, String pipe, KeyPair keyPair) {
-    // set up our reference variables
-    this.account = account;
-    this.user = user;
-    this.pipe = pipe;
-    this.keyPair = keyPair;
-
-    // make our client for sending requests
-    httpClient = HttpUtil.getHttpClient();
-    // make the request builder we'll use to build messages to the service
-  }
+  // ========= Constructors Begin =========
 
   /**
    * Constructs a SimpleIngestManager for a given user in a specific account In addition, this also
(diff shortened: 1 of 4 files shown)


src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
```diff
@@ -132,17 +138,38 @@ private static Properties loadProperties() {
...
 
-  private static String getUserAgent() {
+  /**
+   * Creates a string for user agent which should always be present in all requests to Snowflake
+   * (Snowpipe APIs) Here is the format we will use SnowpipeJavaSDK/version (platform details)
+   * JAVA/<java-version>
+   *
+   * @return the default agent string
+   */
+  private static String getDefaultUserAgent() {
```
sfc-gh-japatel, line 148: @sfc-gh-stakeda if you can just review this part of the code.

src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
```diff
@@ -132,17 +138,41 @@ private static Properties loadProperties() {
...
-    // {client-name}/{version}/{java-version}/{platform}
-    final String userAgentFormat = "%s/%s/%s/%s";
-    return String.format(userAgentFormat, CLIENT_NAME, clientVersion, javaVersion, platform);
+    defaultUserAgent.append(JAVA_USER_AGENT + "/" + javaVersion);
+
+    return defaultUserAgent.toString();
+  }
+
+  private static String buildCustomUserAgent(String additionalUserAgentInfo) {
+    return USER_AGENT.trim() + " " + additionalUserAgentInfo;
```
sfc-gh-stakeda, line 175: you may want to validate if `additionalUserAgentInfo` can meet the User-Agent RFC?

sfc-gh-japatel: The reason I didnt add was because the additional info is going to be used for monitoring purposes and there is no version and version-number.

src/main/java/net/snowflake/ingest/SimpleIngestManager.java
```diff
@@ -448,18 +453,24 @@ public IngestResponse ingestFiles(List<StagedFileWrapper> files, UUID requestId)
    *     construction failure
    */
   public IngestResponse ingestFiles(
-      List<StagedFileWrapper> files, UUID requestId, boolean showSkippedFiles)
+      List<StagedFileWrapper> files,
+      UUID requestId,
+      boolean showSkippedFiles,
+      String additionalUserAgentInfo)
```
sfc-gh-rramachandran, line 459: Can we have an overload to the constructor taking the user agent info instead of adding it to all the public apis? Are there any disadvantages that I am missing?

src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
```diff
@@ -374,11 +404,18 @@ private String generateFilesJSON(List<StagedFileWrapper> files) {
...
+   * Add user agent to the request Header for passed Http request
    *
    * @param request the URI request
+   * @param customUserAgent adds the user agent header to a request along with the default one. If
+   *     it is null or empty, only default one is used.
    */
-  private static void addUserAgent(HttpUriRequest request) {
+  private static void addUserAgent(HttpUriRequest request, String customUserAgent) {
+    if (!Strings.isNullOrEmpty(customUserAgent)) {
+      final String userAgent = buildCustomUserAgent(customUserAgent);
```
sfc-gh-rramachandran, line 415: Another advantage of passing it via the constructor - we can prevent constructing this every time

src/test/java/net/snowflake/ingest/SimpleIngestIT.java
```diff
@@ -212,4 +229,61 @@ private void getHistoryAndAssertLoad(SimpleIngestManager manager, String test_fi
...
+  }
+
+  private void verifyDefaultUserAgent(
+      final Header[] headers,
+      final boolean verifyAdditionalUserAgentInfo,
+      final String httpUserAgentInformation) {
+    for (Header h : headers) {
+      if (h.getName().equalsIgnoreCase(HttpHeaders.USER_AGENT)) {
+        if (verifyAdditionalUserAgentInfo) {
+          assertTrue(h.getValue().contains(httpUserAgentInformation));
```
sfc-gh-rramachandran, line 272: Maybe we can assert that it endsWith the custom agent info?

sfc-gh-rramachandran (commented):
Left some comments. Changes mostly look good

src/test/java/net/snowflake/ingest/SimpleIngestIT.java
sfc-gh-japatel: Will do..

src/main/java/net/snowflake/ingest/SimpleIngestManager.java
sfc-gh-japatel: Passing it as constructor makes sense as well. (might be a smaller change)
Reason I did go with this approach was because it was being set at request level.

src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
sfc-gh-japatel: Umm, that would work as well, but I thought it would be good to have it  request level granularity instead of SimpleIngestManager ctor level. Thoughts?

src/main/java/net/snowflake/ingest/SimpleIngestManager.java
sfc-gh-japatel: But, I kinda agree with having at manager level instead of changing the apis. Let me do that! Thank you!

src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
sfc-gh-stakeda: ok, fine. As I mentioned in the slack, just keep in mind that there is a chance that some library or validator may reject an invalid user-agent. Btw if no version exists, you could make up.

src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
sfc-gh-stakeda: This method looks ok to me.

src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
sfc-gh-japatel: Sounds good, I will keep this as is and once we finalize the suffix from KC side, will add checkers here.

sfc-gh-stakeda (approved):
composing the default user agent string looks good to me.

src/main/java/net/snowflake/ingest/SimpleIngestManager.java
```diff
@@ -170,25 +210,30 @@ public SimpleIngestManager build() {
   // the request builder who handles building the HttpRequests we send
   private final RequestBuilder builder;
 
+  // ========= Constructors Begin =========
```
sfc-gh-japatel, line 213: Just rearranged some deprecated constructors.

src/main/java/net/snowflake/ingest/SimpleIngestManager.java
```diff
@@ -322,7 +411,7 @@ public SimpleIngestManager(
    * @throws NoSuchAlgorithmException if can't create key factory by using RSA algorithm
    * @throws InvalidKeySpecException if private key or public key is invalid
    */
-  private KeyPair createKeyPairFromPrivateKey(PrivateKey privateKey)
+  public static KeyPair createKeyPairFromPrivateKey(PrivateKey privateKey)
```
sfc-gh-rramachandran, line 414: Was this made public for testing? Can we make it @ VisibleForTesting instead?

src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
```diff
@@ -40,27 +41,35 @@
...
   private final String host;
 
+  private final String userAgentSuffix;
+
+  /* Member variables End */
+
+  /* Static constants Begin */
+
+  // the default host is snowflakecomputing.com
+  public static final String DEFAULT_HOST = "snowflakecomputing.com";
```
sfc-gh-rramachandran, line 65: NIT: DEFAULT_HOST_SUFFIX maybe?

sfc-gh-rramachandran (approved):
LGTM! Thanks for making the change! Left some non-blocking cosmetic comments

src/main/java/net/snowflake/ingest/SimpleIngestManager.java
```diff
@@ -61,6 +66,12 @@
     // the key pair we want to use to authenticate
     private KeyPair keypair;
 
+    // Used to add additional information to User-Agent(Http Header)
+    private String userAgentSuffix;
+
+    // Hostname to connect to, default will be RequestBuilder#DEFAULT_HOST
+    private String hostName;
+
```
sfc-gh-rwang, line 74: When do we use this inner Builder class and why these two fields are necessary here?

src/main/java/net/snowflake/ingest/SimpleIngestManager.java
```diff
@@ -199,20 +244,31 @@ private void init(String account, String user, String pipe, KeyPair keyPair) {
...
+   *     accepts hostname as argument
    * @param user the user performing this load
    * @param pipe the fully qualified name of the pipe
    * @param keyPair the KeyPair we'll use to sign JWT tokens
+   * @param schemeName http or https
+   * @param hostName the hostname
+   * @param port the port number
    */
   @Deprecated
-  public SimpleIngestManager(String account, String user, String pipe, KeyPair keyPair) {
```
sfc-gh-rwang, line 210: Will this be an API change if we remove this constructor?

src/main/java/net/snowflake/ingest/SimpleIngestManager.java
```diff
@@ -199,20 +244,31 @@ private void init(String account, String user, String pipe, KeyPair keyPair) {
...
+   *     account should be testaccount If this is the case, you should use the constructor that
+   *     accepts hostname as argument
    * @param user the user performing this load
    * @param pipe the fully qualified name of the pipe
    * @param keyPair the KeyPair we'll use to sign JWT tokens
+   * @param schemeName http or https
+   * @param hostName the hostname
+   * @param port the port number
    */
   @Deprecated
```
sfc-gh-rwang, line 258: Should we remove this deprecated annotation?

src/main/java/net/snowflake/ingest/SimpleIngestManager.java
```diff
@@ -246,38 +302,28 @@ public SimpleIngestManager(String account, String user, String pipe, PrivateKey
...
-   * @param keyPair the KeyPair we'll use to sign JWT tokens
-   * @param schemeName http or https
-   * @param hostName the hostname
-   * @param port the port number
+   * @param account
+   * @param user
+   * @param pipe
+   * @param hostName
+   * @param keyPair
+   * @param userAgentSuffix
```
sfc-gh-rwang, line 313: Would be good if add some comments for these fields in each constructor? Might be useful when users want to show method signatures in IDE.

src/main/java/net/snowflake/ingest/SimpleIngestManager.java
```diff
@@ -246,38 +302,28 @@ public SimpleIngestManager(String account, String user, String pipe, PrivateKey
...
   public SimpleIngestManager(
       String account,
       String user,
       String pipe,
-      KeyPair keyPair,
-      String schemeName,
       String hostName,
-      int port) {
+      KeyPair keyPair,
+      String userAgentSuffix) {
```
sfc-gh-rwang, line 321: Same question here, would this be a breaking change for customers who are using this API?

src/main/java/net/snowflake/ingest/SimpleIngestManager.java
sfc-gh-japatel: Builder patterns makes it easier to construct an object. Having multi argument constructor can easily fade into confusions regarding what each parameter accepts. https://www.vogella.com/tutorials/DesignPatternBuilder/article.html#:~:text=The%20builder%20pattern%20simplifies%20the,to%20create%20an%20immutable%20class.

src/main/java/net/snowflake/ingest/SimpleIngestManager.java
sfc-gh-japatel: I just rearranged this deprecated constructors here. We can still keep this deprecated.

src/main/java/net/snowflake/ingest/SimpleIngestManager.java
sfc-gh-japatel: Sorry for confusion, I rearranged some ctor's. No Behavior change.

src/main/java/net/snowflake/ingest/SimpleIngestManager.java
sfc-gh-japatel: Technically yes, but I thought it could be used as well hence made it public static.

src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
sfc-gh-japatel: Will do.

src/main/java/net/snowflake/ingest/SimpleIngestManager.java
sfc-gh-japatel: This is a bit confusing since I moved around ctors. 
So this is a new flavor of API. It wont be breaking existing customers.

sfc-gh-rwang (approved):
LGTM, thanks!

src/main/java/net/snowflake/ingest/SimpleIngestManager.java
sfc-gh-japatel: Also, I dont know why all keypair APIs were deprecated but it seems like they can be used. Builder has setKeyPair which would mean they can use this public API.

src/main/java/net/snowflake/ingest/SimpleIngestManager.java
sfc-gh-japatel: Done

src/main/java/net/snowflake/ingest/SimpleIngestManager.java
sfc-gh-japatel: Here is what I propose:
1. I will keep the code as is. (I know it is all over the place)
2. I will change keypair extraction logic and generting it inside SecurityManager.. 
3. We can provide two methods: generateP8Key from privateKey bytes and also generateKPFromP8Key - Both of them can be useful for clients. (Helper methods)

src/main/java/net/snowflake/ingest/SimpleIngestManager.java
sfc-gh-japatel: Will do..

PR #93, opened 2021-06-01
Rename file after we used gh actions to pass deploy script

sfc-gh-japatel:
- Jenkins job for releasing this SDK was failing in last step for whitesource check. (It can still release but says file not found since I changed the file name few months back)


sfc-gh-rramachandran (approved):
LGTM

PR #95, opened 2021-08-20
new semgrep integration

sfc-gh-spandey:


sfc-gh-japatel:
LGTM! May I know what this does?

sfc-gh-japatel (approved):

sfc-gh-japatel:
Also, should I make this required? Also, we already have semgrep! Is there a difference?

sfc-gh-spandey:
Hey @sfc-gh-japatel this is a new version of the semgrep check we're using with a new semgrep backend
We'll work towards deprecating the older semgrep check that was running and switch to only having the new one running
Once this is done, we'll ask you to make the check required

PR #96, opened 2021-09-03
Shade Guava and exclude guava compile-only dependencies

ryanstrat:
Guava was left unshaded, which was causing dependency conflicts for downstream projects that include a different version of Guava.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
-    <!-- Arifact name and version information -->
+    <!-- Artifact name and version information -->
     <groupId>net.snowflake</groupId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -342,6 +360,10 @@
                                     <pattern>org.apache</pattern>
                                     <shadedPattern>${shadeBase}.apache</shadedPattern>
                                 </relocation>
+                                <relocation>
```
rzukow, line 363: I was just facing dependency issue and wanted to propose this change. Thanks!

sfc-gh-rzukowski, line 363: I was just facing dep issue and wanted to propose this change. Thanks!

sfc-gh-rzukowski:
> Guava was left unshaded, which was causing dependency conflicts for downstream projects that include a different version of Guava.

Thanks @ryanstrat! This is valid change IMO, but I just proposed getting rid of Guava completely: https://github.com/snowflakedb/snowflake-ingest-java/pull/100/files

ryanstrat:
That's a much better solution (and will also solve my problem). Thanks.

PR #100, opened 2021-10-04
Get rid of Guava dependency.

sfc-gh-rzukowski:
The only method from Guava that was in use is Strings.isNullOrEmpty. I think it is not necessary to introduce dependency for that purpose.

### DIFF
--- a/src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
+++ b/src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
@@ -5,7 +5,6 @@
 package net.snowflake.ingest.connection;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.google.common.base.Strings;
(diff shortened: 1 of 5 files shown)


src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
```diff
@@ -461,7 +462,7 @@ private String generateFilesJSON(List<StagedFileWrapper> files) {
    *     default one. If it is null or empty, only default one is used.
    */
   private static void addUserAgent(HttpUriRequest request, String userAgentSuffix) {
-    if (!Strings.isNullOrEmpty(userAgentSuffix)) {
+    if (isNullOrEmpty(userAgentSuffix)) {
```
sfc-gh-japatel, line 465: nit: `!`

sfc-gh-japatel (approved):
lgtm! will send out PR since it requires gfmt

PR #102, opened 2021-11-16
SNOW-500221 Client Api SDK changes

sfc-gh-anwang:
Make changes to Ingest SDK for both post api client/configure and get api client/status

### DIFF
--- a/src/main/java/net/snowflake/ingest/SimpleIngestManager.java
+++ b/src/main/java/net/snowflake/ingest/SimpleIngestManager.java
@@ -4,25 +4,6 @@
 
 package net.snowflake.ingest;
 
-import static net.snowflake.ingest.connection.RequestBuilder.DEFAULT_HOST_SUFFIX;
-import static net.snowflake.ingest.utils.StringsUtils.isNullOrEmpty;
-
-import java.io.IOException;
-import java.net.URISyntaxException;
-import java.security.KeyFactory;
-import java.security.KeyPair;
-import java.security.NoSuchAlgorithmException;
-import java.security.PrivateKey;
-import java.security.PublicKey;
-import java.security.interfaces.RSAPrivateCrtKey;
-import java.security.spec.InvalidKeySpecException;
-import java.security.spec.RSAPublicKeySpec;
-import java.util.Collections;
-import java.util.List;
-import java.util.Set;
-import java.util.UUID;
-import java.util.stream.Collectors;
-
 import net.snowflake.ingest.connection.ClientStatusResponse;
 import net.snowflake.ingest.connection.ConfigureClientResponse;
 import net.snowflake.ingest.connection.HistoryRangeResponse;
@@ -40,6 +21,25 @@ import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.methods.HttpPost;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.security.KeyFactory;
+import java.security.KeyPair;
+import java.security.NoSuchAlgorithmException;
+import java.security.PrivateKey;
+import java.security.PublicKey;
+import java.security.interfaces.RSAPrivateCrtKey;
+import java.security.spec.InvalidKeySpecException;
+import java.security.spec.RSAPublicKeySpec;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+import java.util.UUID;
+import java.util.stream.Collectors;
+
+import static net.snowflake.ingest.connection.RequestBuilder.DEFAULT_HOST_SUFFIX;
+import static net.snowflake.ingest.utils.StringsUtils.isNullOrEmpty;
 
 /**
  * This class provides a basic, low-level abstraction over the Snowflake Ingest Service REST api
@@ -495,7 +495,7 @@ public class SimpleIngestManager implements AutoCloseable {
    *     construction failure
    */
   public IngestResponse ingestFile(StagedFileWrapper file, UUID requestId)
-      throws URISyntaxException, IOException, Exception {
+      throws Exception {
     return ingestFiles(Collections.singletonList(file), requestId);
   }
 
@@ -512,7 +512,7 @@ public class SimpleIngestManager implements AutoCloseable {
    *     construction failure
    */
   public IngestResponse ingestFile(StagedFileWrapper file, UUID requestId, boolean showSkippedFiles)
-      throws URISyntaxException, IOException, Exception {
+      throws Exception {
     return ingestFiles(Collections.singletonList(file), requestId, showSkippedFiles);
   }
 
@@ -562,7 +562,7 @@ public class SimpleIngestManager implements AutoCloseable {
     HttpResponse response = httpClient.execute(httpPostForIngestFile);
 
     LOGGER.info("Attempting to unmarshall insert response - {}", response);
-    return ServiceResponseHandler.unmarshallIngestResponse(response);
+    return ServiceResponseHandler.unmarshallIngestResponse(response, requestId);
   }
 
   /**
@@ -592,7 +592,7 @@ public class SimpleIngestManager implements AutoCloseable {
     HttpResponse response = httpClient.execute(httpGetHistory);
 
     LOGGER.info("Attempting to unmarshall history response - {}", response);
-    return ServiceResponseHandler.unmarshallHistoryResponse(response);
+    return ServiceResponseHandler.unmarshallHistoryResponse(response, requestId);
   }
 
   /**
@@ -624,7 +624,7 @@ public class SimpleIngestManager implements AutoCloseable {
                 requestId, pipe, startTimeInclusive, endTimeExclusive));
 
     LOGGER.info("Attempting to unmarshall history range response - {}", response);
-    return ServiceResponseHandler.unmarshallHistoryRangeResponse(response);
+    return ServiceResponseHandler.unmarshallHistoryRangeResponse(response, requestId);
   }
 
   /**
@@ -644,7 +644,7 @@ public class SimpleIngestManager implements AutoCloseable {
     HttpResponse response =
         httpClient.execute(builder.generateConfigureClientRequest(requestId, pipe));
     LOGGER.info("Attempting to unmarshall configure client response - {}", response);
-    return ServiceResponseHandler.unmarshallConfigureClientResponse(response);
+    return ServiceResponseHandler.unmarshallConfigureClientResponse(response, requestId);
   }
 
   /**
@@ -664,7 +664,7 @@ public class SimpleIngestManager implements AutoCloseable {
     HttpResponse response =
         httpClient.execute(builder.generateGetClientStatusRequest(requestId, pipe));
     LOGGER.info("Attempting to unmarshall get client status response - {}", response);
-    return ServiceResponseHandler.unmarshallGetClientStatus(response);
+    return ServiceResponseHandler.unmarshallGetClientStatus(response, requestId);
   }
 
   /**
--- a/src/main/java/net/snowflake/ingest/SimpleIngestManager.java
+++ b/src/main/java/net/snowflake/ingest/SimpleIngestManager.java
@@ -22,12 +22,8 @@ import java.util.List;
 import java.util.Set;
 import java.util.UUID;
 import java.util.stream.Collectors;
-import net.snowflake.ingest.connection.HistoryRangeResponse;
-import net.snowflake.ingest.connection.HistoryResponse;
-import net.snowflake.ingest.connection.IngestResponse;
-import net.snowflake.ingest.connection.IngestResponseException;
-import net.snowflake.ingest.connection.RequestBuilder;
-import net.snowflake.ingest.connection.ServiceResponseHandler;
+
+import net.snowflake.ingest.connection.*;
 import net.snowflake.ingest.utils.BackOffException;
 import net.snowflake.ingest.utils.HttpUtil;
 import net.snowflake.ingest.utils.StagedFileWrapper;
@@ -597,7 +593,7 @@ public class SimpleIngestManager implements AutoCloseable {
    * Pings the service to see the current ingest history for this table
    *
    * @param requestId a UUID we use to label the request, if null, one is generated for the user
-   *     * @param startTimeInclusive Start time inclusive of scan range, in ISO-8601 format. Missing
+   * @param startTimeInclusive Start time inclusive of scan range, in ISO-8601 format. Missing
    *     millisecond part in string will lead to a zero milliseconds. This is a required query
    *     parameter, and a 400 will be returned if this query parameter is missing
    * @param endTimeExclusive End time exclusive of scan range. If this query parameter is missing or
@@ -623,6 +619,44 @@ public class SimpleIngestManager implements AutoCloseable {
 
     LOGGER.info("Attempting to unmarshall history range response - {}", response);
     return ServiceResponseHandler.unmarshallHistoryRangeResponse(response);
+  }
+
+  /**
+   * Register a snowpipe client and returns the client sequencer
+   *
+   * @param requestId  a UUID we use to label the request, if null, one is generated for the user
+   * @return
+   * @throws URISyntaxException
+   * @throws IOException
+   * @throws IngestResponseException
+   */
+  public ConfigureClientResponse configureClient(
+          UUID requestId) throws URISyntaxException, IOException, IngestResponseException {
+    if (requestId == null) {
+      requestId = UUID.randomUUID();
+    }
+    HttpResponse response = httpClient.execute(builder.generateConfigureClientRequest(requestId, pipe));
+    LOGGER.info("Attempting to unmarshall configure client response - {}", response);
+    return ServiceResponseHandler.unmarshallConfigureClientResponse(response);
+  }
+
+  /**
+   * Get client status for snowpipe which contains offset token and client sequencer
+   *
+   * @param requestId  a UUID we use to label the request, if null, one is generated for the user
+   * @return
+   * @throws URISyntaxException
+   * @throws IOException
+   * @throws IngestResponseException
+   */
+  public ClientStatusResponse getClientStatus(
+          UUID requestId) throws URISyntaxException, IOException, IngestResponseException {
+    if (requestId == null) {
+      requestId = UUID.randomUUID();
+    }
+    HttpResponse response = httpClient.execute(builder.generateGetClientStatusRequest(requestId, pipe));
+    LOGGER.info("Attempting to unmarshall get client status response - {}", response);
+    return ServiceResponseHandler.unmarshallGetClientStatus(response);
   }
 
   /**
--- a/src/main/java/net/snowflake/ingest/SimpleIngestManager.java
+++ b/src/main/java/net/snowflake/ingest/SimpleIngestManager.java
@@ -624,18 +624,19 @@ public class SimpleIngestManager implements AutoCloseable {
   /**
    * Register a snowpipe client and returns the client sequencer
    *
-   * @param requestId  a UUID we use to label the request, if null, one is generated for the user
+   * @param requestId a UUID we use to label the request, if null, one is generated for the user
    * @return
    * @throws URISyntaxException
    * @throws IOException
    * @throws IngestResponseException
    */
-  public ConfigureClientResponse configureClient(
-          UUID requestId) throws URISyntaxException, IOException, IngestResponseException {
+  public ConfigureClientResponse configureClient(UUID requestId)
+      throws URISyntaxException, IOException, IngestResponseException {
     if (requestId == null) {
       requestId = UUID.randomUUID();
     }
-    HttpResponse response = httpClient.execute(builder.generateConfigureClientRequest(requestId, pipe));
+    HttpResponse response =
+        httpClient.execute(builder.generateConfigureClientRequest(requestId, pipe));
     LOGGER.info("Attempting to unmarshall configure client response - {}", response);
     return ServiceResponseHandler.unmarshallConfigureClientResponse(response);
   }
@@ -643,18 +644,19 @@ public class SimpleIngestManager implements AutoCloseable {
   /**
    * Get client status for snowpipe which contains offset token and client sequencer
    *
-   * @param requestId  a UUID we use to label the request, if null, one is generated for the user
+   * @param requestId a UUID we use to label the request, if null, one is generated for the user
    * @return
    * @throws URISyntaxException
    * @throws IOException
    * @throws IngestResponseException
    */
-  public ClientStatusResponse getClientStatus(
-          UUID requestId) throws URISyntaxException, IOException, IngestResponseException {
+  public ClientStatusResponse getClientStatus(UUID requestId)
+      throws URISyntaxException, IOException, IngestResponseException {
     if (requestId == null) {
       requestId = UUID.randomUUID();
     }
-    HttpResponse response = httpClient.execute(builder.generateGetClientStatusRequest(requestId, pipe));
+    HttpResponse response =
+        httpClient.execute(builder.generateGetClientStatusRequest(requestId, pipe));
     LOGGER.info("Attempting to unmarshall get client status response - {}", response);
     return ServiceResponseHandler.unmarshallGetClientStatus(response);
   }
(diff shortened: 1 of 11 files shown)


src/test/java/net/snowflake/ingest/SimpleIngestIT.java
```diff
@@ -65,9 +65,6 @@ public void beforeAll() throws Exception {
 
     stageWithPatternName = "ingest_sdk_test_stage_pattern" + num;
 
-    TestUtils.executeQuery("use database SNOWPIPE_SDK_DB");
```
sfc-gh-anwang, line 68: Since we are passing the database and schema in the profile.json, not sure if we need to hard code it here.

src/main/java/net/snowflake/ingest/SimpleIngestManager.java
```diff
@@ -22,12 +22,8 @@
...
 import java.util.UUID;
 import java.util.stream.Collectors;
-import net.snowflake.ingest.connection.HistoryRangeResponse;
-import net.snowflake.ingest.connection.HistoryResponse;
-import net.snowflake.ingest.connection.IngestResponse;
-import net.snowflake.ingest.connection.IngestResponseException;
-import net.snowflake.ingest.connection.RequestBuilder;
-import net.snowflake.ingest.connection.ServiceResponseHandler;
+
+import net.snowflake.ingest.connection.*;
```
sfc-gh-japatel, line 26: nit: Let's not use `*` import.

sfc-gh-japatel: https://stackoverflow.com/questions/147454/why-is-using-a-wild-card-with-a-java-import-statement-bad

src/main/java/net/snowflake/ingest/SimpleIngestManager.java
```diff
@@ -625,6 +621,46 @@ public HistoryRangeResponse getHistoryRange(
...
+   *
+   * @param requestId a UUID we use to label the request, if null, one is generated for the user
+   * @return
+   * @throws URISyntaxException
+   * @throws IOException
+   * @throws IngestResponseException
+   */
+  public ConfigureClientResponse configureClient(UUID requestId)
+      throws URISyntaxException, IOException, IngestResponseException {
+    if (requestId == null) {
```
sfc-gh-japatel, line 635: Lets check for emptiness too.. 
```suggestion
    if (requestId == null || requestId.toString().isEmpty()) {
```

src/main/java/net/snowflake/ingest/connection/ClientStatusResponse.java
```diff
@@ -0,0 +1,33 @@
...
+        + "clientSequencer='"
+        + clientSequencer
+        + '\''
+        + ", offsetToken='"
+        + offsetToken
+        + '\''
+        + '}';
+  }
+
+  public Long getClientSequencer() {
```
sfc-gh-japatel, line 26: nit: Javadocs on public setter.

src/main/java/net/snowflake/ingest/connection/ClientStatusResponse.java
```diff
@@ -0,0 +1,33 @@
...
+    return clientSequencer;
+  }
+
+  public String getOffsetToken() {
```
sfc-gh-japatel, line 30: nit: javadoc.

src/main/java/net/snowflake/ingest/connection/ConfigureClientResponse.java
```diff
@@ -0,0 +1,20 @@
...
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class ConfigureClientResponse {
+  private Long clientSequencer;
+
+  @Override
+  public String toString() {
+    return "IngestResponse{" + "clientSequencer='" + clientSequencer + '\'' + '}';
+  }
+
+  public Long getClientSequencer() {
```
sfc-gh-japatel, line 17: Javadoc.

src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
```diff
@@ -426,6 +433,45 @@ private URI makeHistoryRangeURI(
     // build the final URI
     return builder.build();
   }
+
+  /**
+   * makeConfigureClientURI - Given a request UUID, and a fully qualified pipe name make a URI for
+   * the configure snowpipe client
```
sfc-gh-japatel, line 439: Can you add how would the URI look like. (Just copying from server side)
```suggestion
   * http://snowflakeURL{:PORT}/v1/data/pipes/{pipeName}/client/configure
   *
   * <p>Where snowflake URL can be an old url or new regionless URL.
   *
   * <p>Request Body is Empty
   *
   * <p>And our response looks like:
   *
   * <pre>
   *   {
   *      'clientSequencer': LONG
   *   }
   * </pre>

```

sfc-gh-japatel: Food for thought, should we return requestId from clientSDK? (I know we dont pass it down server side)
But since SDK is adding it, it would be good to return back requestId in response too.

src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
```diff
@@ -426,6 +433,45 @@ private URI makeHistoryRangeURI(
...
+      throw new IllegalArgumentException();
+    }
+    URIBuilder builder = makeBaseURI(requestId);
+    builder.setPath(String.format(CONFIGURE_CLIENT_ENDPOINT_FORMAT, pipe));
+    LOGGER.info("Final Configure Client URIBuilder - {}", builder);
+    return builder.build();
+  }
+
+  /**
+   * makeGetClientURI - Given a request UUID, and a fully qualified pipe name make a URI for getting
```
sfc-gh-japatel, line 457: lets add an example of how a URI looks like here.

src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
```diff
@@ -566,6 +612,38 @@ public HttpGet generateHistoryRangeRequest(
     return get;
   }
 
+  /**
+   * generateConfigureClientRequest - given a requestId and a pipe, make a configure client request
```
sfc-gh-japatel, line 616: nit: remove `generateConfigureClientRequest`

src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
```diff
@@ -566,6 +612,38 @@ public HttpGet generateHistoryRangeRequest(
...
+  public HttpPost generateConfigureClientRequest(UUID requestID, String pipe)
+      throws URISyntaxException {
+    URI configureClientURI = makeConfigureClientURI(requestID, pipe);
+    HttpPost post = new HttpPost(configureClientURI);
+    addHeaders(post, securityManager.getToken(), this.userAgentSuffix);
+    return post;
+  }
+
+  /**
+   * generateGetClientStatusRequest - given a requestId and a pipe, make a get client status request
```
sfc-gh-japatel, line 632: remove `generateGetClientStatusRequest`

src/main/java/net/snowflake/ingest/connection/ServiceResponseHandler.java
```diff
@@ -128,49 +106,98 @@ public static HistoryRangeResponse unmarshallHistoryRangeResponse(HttpResponse r
...
-        LOGGER.error("Status code {} found in response from service", statusLine.getStatusCode());
-        String blob = EntityUtils.toString(response.getEntity());
-        throw new IngestResponseException(
-            statusLine.getStatusCode(),
-            IngestResponseException.IngestExceptionBody.parseBody(blob));
+    StatusLine statusLine = response.getStatusLine();
+    if (!isStatusOK(statusLine)) {
+      // A network issue occurred!
+      LOGGER.warn(
+          "Exceptional Status Code found in " + "unmarshallHistoryRangeResponse - {}",
```
sfc-gh-japatel, line 184: nit: Since this is used by all responses. should we remove `unmarshallHistoryRangeResponse`? and may be pass from caller?

src/main/java/net/snowflake/ingest/connection/ServiceResponseHandler.java
```diff
@@ -128,49 +106,98 @@ public static HistoryRangeResponse unmarshallHistoryRangeResponse(HttpResponse r
...
   /**
-   * handleExceptionStatusCode - throws the correct error for a status
+   * handleExceptionStatusCode - throws the correct error when response status is not OK
    *
-   * @param statusLine the status line we want to check
-   * @throws BackOffException -- if we have a 503 exception
-   * @throws IOException - if we don't know what it is
+   * @param response HttpResponse
+   * @throws BackOffException if we have a 503 exception
+   * @throws IOException if we don't know what it is
```
sfc-gh-japatel, line 176: and `IngestResponseException`

src/main/java/net/snowflake/ingest/connection/ServiceResponseHandler.java
```diff
@@ -128,49 +106,98 @@ public static HistoryRangeResponse unmarshallHistoryRangeResponse(HttpResponse r
...
-      // A network issue occurred!
-      LOGGER.warn(
-          "Exceptional Status Code found in " + "unmarshallHistoryRangeResponse - {}",
-          line.getStatusCode());
+    // handle the exceptional status code
+    handleExceptionalStatus(response);
 
-      // handle the exceptional status code
-      handleExceptionalStatus(line, response);
-      return null;
```
sfc-gh-japatel, line 141: Was this unreachable here? Would we have sent `IngestResponseException` in all non ok responses?

src/test/java/net/snowflake/ingest/SimpleIngestIT.java
```diff
@@ -315,4 +312,60 @@ private void verifyDefaultUserAgent(
...
+  @Test
+  public void testConfigureClientNoPipeFound() throws Exception {
+    final String userAgentSuffix = "kafka-provider/NONE";
+    SimpleIngestManager manager = TestUtils.getManager("nopipe", userAgentSuffix);
+    try {
+      manager.configureClient(null);
+    } catch (IngestResponseException exception) {
+      assertEquals(
+          exception.getMessage(),
+          "HTTP Status: 404 ErrorBody: {\n"
```
sfc-gh-japatel, line 333: can we use object mapper to read this as a tree? It will give us JsonNode and it will be much easier to assert. 
```suggestion
JsonNode node = mapper.reader().readTree(exception.getMessage());
Assert node.get(code).getText()
.. and few more
```

src/test/java/net/snowflake/ingest/SimpleIngestIT.java
```diff
@@ -315,4 +312,60 @@ private void verifyDefaultUserAgent(
...
+    final String userAgentSuffix = "kafka-provider/NONE";
+    SimpleIngestManager manager = TestUtils.getManager(pipeName, userAgentSuffix);
+    manager.configureClient(null);
+    ClientStatusResponse clientStatusResponse = manager.getClientStatus(null);
+    assertEquals(0L, clientStatusResponse.getClientSequencer().longValue());
+    assertEquals(null, clientStatusResponse.getOffsetToken());
+  }
+
+  @Test
+  public void testGetClientStatusNoPipeFound() throws Exception {
```
sfc-gh-japatel, line 354: https://github.com/snowflakedb/snowflake-ingest-java/runs/4231421925?check_suite_focus=true

The APIs wont work now because they are not enabled for this account. 

Lets ignore(@Ignore annotation from Junit) this tests now.. But lets verify if they are successful against a dev/reg environment.

src/test/java/net/snowflake/ingest/SimpleIngestIT.java
```diff
@@ -13,8 +13,8 @@
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
-import net.snowflake.ingest.connection.HistoryResponse;
-import net.snowflake.ingest.connection.IngestResponse;
+
+import net.snowflake.ingest.connection.*;
```
sfc-gh-japatel, line 17: nit: no `*` import.

src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
sfc-gh-anwang: we can add them in the response. But in my understanding the requestId is only for debug purpose. However, when non 200 status code occurred, we don't return response object, instead we throw an error.  Should we add the requestId to IngestResponseException? If we want to return the requestId back to users, it makes sense we return them at both response and exception.

src/main/java/net/snowflake/ingest/connection/ServiceResponseHandler.java
sfc-gh-anwang: yes, we sent Exception for all non ok responses.

src/main/java/net/snowflake/ingest/SimpleIngestManager.java
```diff
@@ -544,13 +551,15 @@ public IngestResponse ingestFiles(
...
 
     // the request id we want to send with this payload
-    UUID request = requestId == null ? UUID.randomUUID() : requestId;
+    if (requestId == null || requestId.toString().isEmpty()) {
+      requestId = UUID.randomUUID();
+    }
 
     // We're about to send this request number
-    LOGGER.info("Sending Request UUID - {}", request);
+    LOGGER.info("Sending Request UUID - {}", requestId);
```
sfc-gh-japatel, line 559: actually, lets remove this log line. We should have minimum log info or club multiple of them together. more loggers means more space on client machine.

src/main/java/net/snowflake/ingest/SimpleIngestManager.java
```diff
@@ -575,7 +584,7 @@ public IngestResponse ingestFiles(
   public HistoryResponse getHistory(UUID requestId, Integer recentSeconds, String beginMark)
       throws URISyntaxException, IOException, IngestResponseException {
     // if we have no requestId generate one
-    if (requestId == null) {
+    if (requestId == null || requestId.toString().isEmpty()) {
```
sfc-gh-japatel, line 587: 👍

src/test/java/net/snowflake/ingest/SimpleIngestIT.java
```diff
@@ -321,51 +333,40 @@ public void testConfigureClientHappyCase() throws Exception {
...
+  @Ignore
   @Test
   public void testGetClientStatusHappyCase() throws Exception {
     final String userAgentSuffix = "kafka-provider/NONE";
     SimpleIngestManager manager = TestUtils.getManager(pipeName, userAgentSuffix);
     manager.configureClient(null);
     ClientStatusResponse clientStatusResponse = manager.getClientStatus(null);
     assertEquals(0L, clientStatusResponse.getClientSequencer().longValue());
-    assertEquals(null, clientStatusResponse.getOffsetToken());
+    assertNull(clientStatusResponse.getOffsetToken());
```
sfc-gh-japatel, line 359: umm, wont this be `null` string? or the `ClientStatusResponse` deserializes it into null?

sfc-gh-anwang: yah, assertNull makes sure it is null

src/main/java/net/snowflake/ingest/connection/ServiceResponseHandler.java
```diff
@@ -60,20 +60,8 @@ public static IngestResponse unmarshallIngestResponse(HttpResponse response)
...
     }
 
-    // Grab the status line from the response
-    StatusLine statusLine = response.getStatusLine();
-
-    // If we didn't get a good status code, handle it
-    if (!isStatusOK(statusLine)) {
-
-      // Exception status
-      LOGGER.warn(
```
sfc-gh-japatel, line 70: So this warn error message is kind of useful to find out which API has non ok status.. 

Can you create an Enum with API names and pass it down to handleExceptionalStatus and print the enum along with logger.warn in switch case??

sfc-gh-japatel: It will be hard to debug right now on which API has failed from Client SDK side. and we will be calling multiple APIs from KC. It would be better to distinguish between APIs and failures. Lets also print requestId in warn if possible. 

Unless this error is surfaced upwards and its being logged at top level.

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -109,17 +110,19 @@ private static void initHttpClient() {
...
       @Override
       public boolean retryRequest(
           final HttpResponse response, final int executionCount, final HttpContext context) {
         this.executionCount = executionCount;
         int statusCode = response.getStatusLine().getStatusCode();
-        LOGGER.info(
-            "In retryRequest for service unavailability with statusCode:{} and uri:{}",
-            statusCode,
-            getRequestUriFromContext(context));
+        if (statusCode != 200) {
```
sfc-gh-japatel, line 120: can we move this to the block where we check `if (needNextRetry)`?

sfc-gh-japatel: We only want to print in retryRequest if we satisfy the criteria for retry..

src/main/java/net/snowflake/ingest/connection/ServiceResponseHandler.java
```diff
@@ -61,7 +63,7 @@ public static IngestResponse unmarshallIngestResponse(HttpResponse response)
     }
 
     // handle the exceptional status code
-    handleExceptionalStatus(response);
+    handleExceptionalStatus(response, requestId, "InsertFiles");
```
sfc-gh-japatel, line 68: can we replace magic strings with Enum?

src/main/java/net/snowflake/ingest/connection/ServiceResponseHandler.java
```diff
@@ -99,8 +103,19 @@ public static HistoryResponse unmarshallHistoryResponse(HttpResponse response, U
     return mapper.readValue(blob, HistoryResponse.class);
   }
 
+  /**
```
sfc-gh-japatel, line 106: Thank you!

src/main/java/net/snowflake/ingest/SimpleIngestManager.java
```diff
@@ -486,13 +489,14 @@ public String getPipe() {
...
-   * @throws IOException - if we have some other network failure
    * @throws URISyntaxException - if the provided account name was illegal and caused a URI
    *     construction failure
+   * @throws IOException - if we have some other network failure
+   * @throws IngestResponseException - if snowflake encountered error during ingest
+   * @throws BackOffException - if we have a 503 response
    */
   public IngestResponse ingestFile(StagedFileWrapper file, UUID requestId)
-      throws URISyntaxException, IOException, Exception {
+      throws URISyntaxException, IOException, IngestResponseException, BackOffException {
```
sfc-gh-japatel, line 499: Where is BackOffException coming from?

src/main/java/net/snowflake/ingest/connection/ServiceResponseHandler.java
```diff
@@ -149,28 +134,96 @@ public static HistoryRangeResponse unmarshallHistoryRangeResponse(HttpResponse r
...
+          throw new BackOffException();
+
+        default:
+          String blob = EntityUtils.toString(response.getEntity());
+          throw new IngestResponseException(
+              statusLine.getStatusCode(),
+              IngestResponseException.IngestExceptionBody.parseBody(blob));
+      }
+    } else {
+      LOGGER.info("OK status code from {}, requestId: {}", apiName, requestId.toString());
```
sfc-gh-japatel, line 226: I dont think this is needed.

sfc-gh-japatel (changes requested):
added some comments.

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-anwang: Ok,I thought it was more like a debug log.

src/main/java/net/snowflake/ingest/connection/ServiceResponseHandler.java
sfc-gh-anwang: Ok

sfc-gh-anwang:
I did run the default GoogleJavaFormat on the files, is there additional script that I need to run to pass the GoogleJavaFormat check?

src/main/java/net/snowflake/ingest/connection/ServiceResponseHandler.java
```diff
@@ -24,7 +24,9 @@
 public final class ServiceResponseHandler {
   // Create a logger for this class
   private static final Logger LOGGER = LoggerFactory.getLogger(ServiceResponseHandler.class);
-
+  private enum ApiName{
```
sfc-gh-japatel, line 27: javadocs please.

src/main/java/net/snowflake/ingest/connection/ServiceResponseHandler.java
```diff
@@ -24,7 +24,9 @@
...
+    InsertFiles, GetHistory, GetHistoryRange, ConfigureClient, GetClientStatus
```
sfc-gh-japatel, line 28: capitalised names since they are constants.

sfc-gh-japatel: One liners for each enum types. 
Lets remove verb from APIs. 
INSERT_FILES

INSERT_REPORT

LOAD_HISTORY_SCAN

CLIENT_CONFIGURE

CLIENT_STATUS

https://docs.snowflake.com/en/user-guide/data-load-snowpipe-rest-apis.html
Lets also have a private constructor. 
We can add various fields in it but it is not needed at the moment. (One of that I can think of is VERB -> GET, POST etc)

sfc-gh-japatel:
> I did run the default GoogleJavaFormat on the files, is there additional script that I need to run to pass the GoogleJavaFormat check?

https://github.com/google/google-java-format

Here is what I use:

```
java -jar ~/path-to/google-java-format-1.10.0-all-deps.jar  -i $(find . -type f -name "*.java" | grep ".*/src/.*java")
```

sfc-gh-japatel:
Can you also add the guide to README?

src/test/java/net/snowflake/ingest/TestUtils.java
```diff
@@ -18,7 +18,7 @@
 
 public class TestUtils {
   // profile path, follow readme for the format
-  private static final String PROFILE_PATH = "profile.json";
+  private static final String PROFILE_PATH = "profile.json.example";
```
sfc-gh-japatel: lets not change this!

src/test/java/net/snowflake/ingest/TestUtils.java
```diff
@@ -18,7 +18,7 @@
 
 public class TestUtils {
   // profile path, follow readme for the format
-  private static final String PROFILE_PATH = "profile.json.example";
+  private static final String PROFILE_PATH = "profile.json";
```
sfc-gh-japatel: ok

src/main/java/net/snowflake/ingest/connection/ServiceResponseHandler.java
```diff
@@ -149,28 +150,95 @@ public static HistoryRangeResponse unmarshallHistoryRangeResponse(HttpResponse r
...
+      LOGGER.error(
+          "Exceptional Status Code from {}: {}, requestId:{}",
+          apiName,
+          statusLine.getStatusCode(),
+          requestId.toString());
+      // if we have a 503 exception throw a backoff
+      switch (statusLine.getStatusCode()) {
+          // If we have a 503, BACKOFF
+        case HttpStatus.SC_SERVICE_UNAVAILABLE:
+          LOGGER.warn("503 Status hit, backoff");
```
sfc-gh-japatel, line 233: Lets also print `apiName,
          requestId.toString());`

here

src/main/java/net/snowflake/ingest/connection/ServiceResponseHandler.java
```diff
@@ -23,6 +24,23 @@
   // Create a logger for this class
   private static final Logger LOGGER = LoggerFactory.getLogger(ServiceResponseHandler.class);
 
+  /** Enum for every public api, used in handleExceptionalStatus for logging purpose */
```
sfc-gh-japatel, line 27: nit: https://docs.snowflake.com/en/user-guide/data-load-snowpipe-rest-apis.html

sfc-gh-japatel (approved):
minor comments. LGTM. 
Please address couple of small comments before pushing.

PR #109, opened 2022-01-12
Snowpipe Streaming V1

sfc-gh-tzhang:
This PR contains all the commits related to Snowpipe Streaming V1 that we have been working on last year, exciting time will come soon!

Snowpipe Streaming Ingest

Custom Java applications can interface with the Snowpipe streaming ingest API to prompt low-latency loads of rowsets using the Snowflake Ingest SDK. The API writes rowsets of data to Snowflake tables, unlike bulk data loads or Snowpipe, which write data from staged files. This architecture results in lower load latencies, with equal or lower costs for loading similar volumes of data.

The streaming ingest API ("API") is intended to complement Snowpipe, not replace it. Try this preview version of the API in streaming scenarios where data is streamed in rowsets (e.g. Apache Kafka topics) rather than in files. The API fits into an ingest workflow that includes an existing custom Java application that produces or receives records. The API removes the need to create files to load data into Snowflake tables.


sfc-gh-tjones (approved):
Love it! Go, go, go!

sfc-gh-japatel (approved):
Yayyy! Great stuff. Excited for what's coming ahead. Can't wait to pull from Maven.

chrisVia:
@sfc-gh-tzhang  I'm sorry to bump an old MR, but can the author clarify if it is an intentional decision to promote the `snowflake-jdbc` dependency in this project to _compile_ scope instead of _provided_ scope? This blows up the artifact size published on maven central from 7.5Mb to **48Mb**. Since the published pom is the `public_pom.xml` instead of the real `pom.xml`, we can't use exclusions rules to bring the artifact size back under control when we try to depend on this artifact in our own projects. Thanks!

sfc-gh-tzhang:
NP, at least for now we don't have plan to change the current behavior, which is the same as JDBC, but it could be revisited if we hear more feedback for this

PR #111, opened 2022-01-19
Add nullability to the field  metadata

sfc-gh-mnaides:
Add column nullability info to the field metadatada


sfc-gh-tzhang (approved):
LGTM!

PR #112, opened 2022-01-20
SNOW-356153 Move streaming ingest to use profile.json

sfc-gh-mnaides:
Moved streaming ingest from using profile.properties to profile.json with existing keys

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClientFactory.java
+++ b/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClientFactory.java
@@ -40,10 +40,21 @@ public class SnowflakeStreamingIngestClientFactory {
       Utils.assertNotNull("connection properties", this.prop);
 
       if (!this.prop.containsKey(Constants.ACCOUNT_URL)) {
-        throw new SFException(ErrorCode.MISSING_CONFIG, "Account URL");
+        if (this.prop.containsKey(Constants.HOST)
+            && this.prop.containsKey(Constants.SCHEME)
+            && this.prop.containsKey(Constants.PORT)) {
+          this.prop.put(
+              Constants.ACCOUNT_URL,
+              Utils.constructAccountUrl(
+                  this.prop.get(Constants.SCHEME).toString(),
+                  this.prop.get(Constants.HOST).toString(),
+                  Integer.parseInt(prop.get(Constants.PORT).toString())));
+        } else {
+          throw new SFException(ErrorCode.MISSING_CONFIG, "Account URL");
+        }
       }
 
-      if (!this.prop.containsKey(Constants.ROLE_NAME)) {
+      if (!this.prop.containsKey(Constants.ROLE)) {
         throw new SFException(ErrorCode.MISSING_CONFIG, "Role Name");
       }
 
--- a/src/main/java/net/snowflake/ingest/streaming/example/SnowflakeStreamingIngestExample.java
+++ b/src/main/java/net/snowflake/ingest/streaming/example/SnowflakeStreamingIngestExample.java
@@ -4,8 +4,11 @@
 
 package net.snowflake.ingest.streaming.example;
 
-import java.io.FileInputStream;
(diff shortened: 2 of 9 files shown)


sfc-gh-tjones (approved):
Love it!

src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClientFactory.java
```diff
@@ -40,10 +40,21 @@ public SnowflakeStreamingIngestClient build() {
...
+            && this.prop.containsKey(Constants.SCHEME)
+            && this.prop.containsKey(Constants.PORT)) {
+          this.prop.put(
+              Constants.ACCOUNT_URL,
+              Utils.constructAccountUrl(
+                  this.prop.get(Constants.SCHEME).toString(),
+                  this.prop.get(Constants.HOST).toString(),
+                  Integer.parseInt(prop.get(Constants.PORT).toString())));
+        } else {
+          throw new SFException(ErrorCode.MISSING_CONFIG, "Account URL");
```
sfc-gh-tzhang, line 53: Could you add "url" to the json file? Or return a different error? Otherwise customer doesn't know the information that they're missing

src/main/java/net/snowflake/ingest/streaming/example/SnowflakeStreamingIngestExample.java
```diff
@@ -15,15 +18,18 @@
 import net.snowflake.ingest.streaming.SnowflakeStreamingIngestChannel;
 import net.snowflake.ingest.streaming.SnowflakeStreamingIngestClient;
 import net.snowflake.ingest.streaming.SnowflakeStreamingIngestClientFactory;
+import net.snowflake.ingest.utils.Utils;
 
 /** Examples on how to use the Streaming Ingest client APIs */
 public class SnowflakeStreamingIngestExample {
-  private static String PROFILE_PATH = "profile.properties";
+  private static String PROFILE_PATH = "profile.json";
```
sfc-gh-tzhang, line 25: We should remove profile.properties.example

src/main/java/net/snowflake/ingest/utils/Constants.java
```diff
@@ -8,14 +8,21 @@
...
+  public static final String SCHEMA = "schema";
+  public static final String CONNECT_STRING = "connect_string";
+  public static final String HOST = "host";
+  public static final String WAREHOUSE = "warehouse";
+  public static final String SCHEME = "scheme";
   public static final String ACCOUNT_URL = "snowflake.url.name";
-  public static final String PRIVATE_KEY = "snowflake.private.key";
-  public static final String ROLE_NAME = "snowflake.role.name";
+  public static final String PRIVATE_KEY = "private_key";
+  public static final String ROLE = "role";
```
sfc-gh-tzhang, line 23: Could you remove `snowflake.` prefix in ACCOUNT_URL and PRIVATE_KEY_PASSPHRASE as well to make it consistent?

src/test/java/net/snowflake/ingest/TestUtils.java
```diff
@@ -62,18 +79,20 @@
...
+    port = profile.get(PORT).asInt();
+    ssl = profile.get(SSL).asText();
+    database = profile.get(DATABASE).asText();
+    connectString = profile.get(CONNECT_STRING).asText();
+    schema = profile.get(SCHEMA).asText();
+    warehouse = profile.get(WAREHOUSE).asText();
+    host = profile.get(HOST).asText();
+    scheme = profile.get(SCHEME).asText();
+
+    role = Optional.ofNullable(profile.get(ROLE)).map(r -> r.asText()).orElse("DEFAULT_ROLE");
```
sfc-gh-tzhang, line 93: ah, could we do this in the SDK logic instead of in the test, seems like a good logic to have? (if I understand correctly, we will use the DEFAULT_ROLE is there is no role input?)

sfc-gh-tzhang (commented):
late for the review, but please take a look at the comments if you have a chance, thanks!

PR #114, opened 2022-01-21
SNOW-533760: include client name and key during upload

sfc-gh-tzhang:
include client name and key during file upload, which will be used for streaming ingest billing


sfc-gh-tjones (approved):
LGTM!

sfc-gh-mnaides (approved):

PR #116, opened 2022-01-22
NO-SNOW: small fixes on configuration file behaviors

sfc-gh-tzhang:
This is a small follow-up fix on top of https://github.com/snowflakedb/snowflake-ingest-java/pull/112 to correct the behaviors

### DIFF
--- a/profile.json.example
+++ b/profile.json.example
@@ -1,6 +1,7 @@
 {
   "user": "user name",
-  "account": "account name",
+  "url": "https://account_name.snowflakecomputing.com:443"
+  "account": "account_name",
   "private_key": "PEM Private Key",
   "port": 443,
   "host": "account_name.snowflakecomputing.com",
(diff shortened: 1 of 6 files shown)


sfc-gh-mnaides (approved):

profile.json.example
```diff
@@ -1,6 +1,7 @@
 {
   "user": "user name",
-  "account": "account name",
+  "url": "https://account_name.snowflakecomputing.com:443"
```
sfc-gh-tjones, line 3: Nit, I think we can drop the 443

sfc-gh-tjones (approved):

profile.json.example
sfc-gh-tzhang: At least in the KC configuration doc https://docs.snowflake.com/en/user-guide/kafka-connector-install.html, the url in the example contains a port number, is it possible that we use a different port?

PR #117, opened 2022-01-24
NO-SNOW: simplify SnowflakeStreamingIngestClient interface 

sfc-gh-tzhang:
This PR contains a few changes related to simplify the client interface:
1. Add AutoCloseable support
2. Simplify the logic in `close` so it doesn't check for uncommitted channels anymore, our recommendation should be always close the channel or check for uncommitted rows after using it

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClient.java
+++ b/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClient.java
@@ -26,4 +26,7 @@ public interface SnowflakeStreamingIngestClient extends AutoCloseable {
    * @return the client name
    */
   String getName();
+
+  /** @return a boolean to indicate whether the client is closed */
+  boolean isClosed();
 }
--- a/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClient.java
+++ b/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClient.java
@@ -4,15 +4,13 @@
 
 package net.snowflake.ingest.streaming;
 
-import java.util.concurrent.CompletableFuture;
-
 /**
  * A class that is the starting point for using the Streaming Ingest client APIs, a single client
  * maps to exactly one account in Snowflake; however, multiple clients can point to the same
  * account. Each client will contain information for Snowflake authentication and authorization, and
  * it will be used to create one or more {@link SnowflakeStreamingIngestChannel}
  */
-public interface SnowflakeStreamingIngestClient {
+public interface SnowflakeStreamingIngestClient extends AutoCloseable {
 
   /**
    * Open a channel against a Snowflake table using a {@link OpenChannelRequest}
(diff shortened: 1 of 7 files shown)


src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClient.java
```diff
@@ -22,21 +20,10 @@
...
   SnowflakeStreamingIngestChannel openChannel(OpenChannelRequest request);
 
-  /**
-   * Close all the channels in the client, which will make sure all the data in all channels is
-   * committed before closing
-   *
-   * @return a completable future which will be completed when all the channels are closed
-   */
-  CompletableFuture<Void> close();
-
```
sfc-gh-tjones, line 32: Should we keep this in case customers use it in a non-Autocloseable context?

src/main/java/net/snowflake/ingest/streaming/example/SnowflakeStreamingIngestExample.java
```diff
@@ -25,17 +23,14 @@
...
-  public static void main(String[] args)
-      throws ExecutionException, InterruptedException, IOException {
+  public static void main(String[] args) throws Exception {
     ObjectNode profile =
         (ObjectNode) mapper.readTree(new String(Files.readAllBytes(Paths.get(PROFILE_PATH))));
     Properties prop = Utils.getPropertiesFromJson(profile);
 
     // Create a streaming ingest client
-    SnowflakeStreamingIngestClient client =
-        SnowflakeStreamingIngestClientFactory.builder("CLIENT").setProperties(prop).build();
```
sfc-gh-tjones, line 36: Is this still possible? Or must it always be wrapped with `try`?

src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClient.java
sfc-gh-tzhang: if it's non-Autocloseable, they can still call the `close()` function explicitly, right? Did I understand correct what you mean?

src/main/java/net/snowflake/ingest/streaming/example/SnowflakeStreamingIngestExample.java
sfc-gh-tzhang: Yes, you can still initialize that this, and call close explicitly when you finish

src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClient.java
sfc-gh-tjones: Yeah.

src/main/java/net/snowflake/ingest/streaming/example/SnowflakeStreamingIngestExample.java
sfc-gh-tjones: Got in - in that case should we keep the close call above?

sfc-gh-tzhang: hmmm, why? I add a try block so close will be automatically called?

src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClient.java
```diff
@@ -22,21 +20,10 @@
...
   /**
    * Get the client name
    *
    * @return the client name
    */
   String getName();
-
-  /** @return a boolean to indicate whether the client is closed */
-  boolean isClosed();
```
sfc-gh-japatel, line 41: I think we should keep isClosed() just so as to not get client is not initialised error while calling open channel. 

Today I use isClosed() to do that.

sfc-gh-tzhang:
bump, can I get a review on this?

sfc-gh-tjones (approved):

PR #118, opened 2022-01-24
SNOW-436804 Let Snowflake server set parameter values on client configure

sfc-gh-mnaides:
Support letting the Snowflake server set some client side constants.


src/main/java/net/snowflake/ingest/streaming/internal/ParameterProvider.java
```diff
@@ -0,0 +1,77 @@
...
+
+  // Default values in the event no value is set by the server
+  static final long MAX_CHUNK_SIZE_IN_BYTES_DEFAULT = 16000000L;
+  static final long MAX_BLOB_SIZE_IN_BYTES_DEFAULT = 256000000L;
+  static final long BUFFER_FLUSH_INTERVAL_IN_MILLIS_DEFAULT = 1000;
+  static final long BUFFER_FLUSH_CHECK_INTERVAL_IN_MILLIS_DEFAULT = 100;
+  static final long INSERT_THROTTLE_INTERVAL_IN_MILLIS_DEFAULT = 500;
+  static final long INSERT_THROTTLE_THRESHOLD_IN_PERCENTAGE_DEFAULT = 5;
+
+  private Map<String, Long> parameterMap = new HashMap<>();
```
sfc-gh-tjones, line 25: Nit, docstring.

src/main/java/net/snowflake/ingest/streaming/internal/ParameterProvider.java
```diff
@@ -0,0 +1,77 @@
...
+
+  public ParameterProvider() {}
```
sfc-gh-tjones, line 27: Docstring

sfc-gh-tjones (approved):

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -74,6 +74,9 @@
   // Counter to generate unique request ids per client
   private final AtomicLong counter = new AtomicLong(0);
 
+  // Provides constant values set by Snowflake server
+  private static final ParameterProvider parameterProvider = new ParameterProvider();
```
sfc-gh-tzhang, line 78: why is this a static variable? What if the value changed between multiple clients creation?

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
```diff
@@ -226,6 +235,18 @@ synchronized SnowflakeFileTransferMetadataWithAge refreshSnowflakeMetadata(boole
       }
 
       JsonNode responseNode = mapper.valueToTree(response);
+
+      // Update parameter map with the latest values from server
+      Map<String, Long> parameterMap =
+          mapper.convertValue(
+              responseNode.get("parameter_map"), new TypeReference<Map<String, Long>>() {});
```
sfc-gh-tzhang, line 242: I think we want these for some boolean parameters as well, for example, BLOB_NO_HEADER, could we make it more generalize?

sfc-gh-tzhang (commented):
This is good stuff! One major comment I have is to related to making the interface more generalize, so it could be used for different data types, especially boolean because I would imagine this will be a easier way to turn off a certain feature if something went extremely wrong....

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
sfc-gh-tzhang: And could you add a log line so we know what are those values if we need to look up customer's log during debugging issues

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestStageTest.java
```diff
@@ -55,7 +55,8 @@
           + " \"endPoint\": null}}}";
 
   String exampleRemoteMetaResponse =
-      "{\"src_locations\": [\"foo/\"],\"status_code\": 0, \"message\": \"Success\", \"prefix\":"
```
sfc-gh-tzhang, line 58: do we verify somewhere that the value gets read correctly?

src/main/java/net/snowflake/ingest/streaming/internal/ParameterProvider.java
```diff
@@ -0,0 +1,77 @@
...
+  final String INSERT_THROTTLE_THRESHOLD_IN_PERCENTAGE_MAP_KEY =
+      "INSERT_THROTTLE_THRESHOLD_IN_PERCENTAGE";
+
+  // Default values in the event no value is set by the server
+  static final long MAX_CHUNK_SIZE_IN_BYTES_DEFAULT = 16000000L;
+  static final long MAX_BLOB_SIZE_IN_BYTES_DEFAULT = 256000000L;
+  static final long BUFFER_FLUSH_INTERVAL_IN_MILLIS_DEFAULT = 1000;
+  static final long BUFFER_FLUSH_CHECK_INTERVAL_IN_MILLIS_DEFAULT = 100;
+  static final long INSERT_THROTTLE_INTERVAL_IN_MILLIS_DEFAULT = 500;
+  static final long INSERT_THROTTLE_THRESHOLD_IN_PERCENTAGE_DEFAULT = 5;
```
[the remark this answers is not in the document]
sfc-gh-mnaides: I made them package private so I could test against them.  I thought about leaving them in Constants.java but I wanted to bring them out so there wouldn't be confusion about what's an actual constant and what may be overridden by the server

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
sfc-gh-mnaides: Each client should have a single parameterProvider.  ParameterProvider can change its values on each client/configure call without reassigning the Object on the client

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
sfc-gh-mnaides: I initially had the parameterMap be `Map<String, Object>` to be more generalizable but I ran into the issue where Jackson was creating all the `Long` values as `Integer`.  I didn't want there to be any confusion and type casting so I went with making all values `Long` and figured we could encode booleans with 0 or 1.  What do you think?

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestStageTest.java
```diff
@@ -225,6 +231,8 @@ public void testRefreshSnowflakeMetadataRemote() throws Exception {
         "foo/streaming_ingest/", metadataWithAge.fileTransferMetadata.getStageInfo().getLocation());
     Assert.assertEquals(
         "placeholder", metadataWithAge.fileTransferMetadata.getPresignedUrlFileName());
+
+    Assert.assertEquals(123, parameterProvider.getMaxChunkSizeInBytes());
```
sfc-gh-mnaides, line 235: https://github.com/snowflakedb/snowflake-ingest-java/pull/118#discussion_r791276629 @sfc-gh-tzhang here's the test

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
sfc-gh-tzhang: I would still prefer to use Object instead of Long, I believe there is a way to tell Jackson to deserialize to Long instead of int, do you want to search around? If it doesn't work, please create a JIRA and I can try to see if there is anything i can do.

sfc-gh-tzhang (approved):
Ship it!

PR #122, opened 2022-01-27
SNOW-535803 Support TIMESTAMP_TZ in streaming ingest

sfc-gh-mnaides:
Adds support to streaming ingest for TIMESTAMP_TZ columns as well as adds string parsing to TIMESTAMP_NTZ columns

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
@@ -815,7 +815,7 @@ class ArrowRowBuffer {
                                   new SFException(
                                       ErrorCode.INVALID_ROW,
                                       value,
-                                      "Unable to parse timezone for TIMESTAMP_TZ column"))
+                                      "Unable to parse timestamp for TIMESTAMP_TZ column"))
                           .toBinary(Integer.parseInt(field.getMetadata().get(COLUMN_SCALE)), true);
                   stats.addIntValue(timeInBinary);
                   break;
@@ -829,7 +829,7 @@ class ArrowRowBuffer {
                       (IntVector) structVector.getChild(FIELD_FRACTION_IN_NANOSECONDS);
                   IntVector timezoneVector = (IntVector) structVector.getChild(FIELD_TIME_ZONE);
 
-                  rowBufferSize += 0.50; // for children vector's null value
+                  rowBufferSize += 0.375; // for children vector's null value
                   structVector.setIndexDefined(curRowIndex);
 
                   TimestampWrapper timestampWrapper =
@@ -855,7 +855,7 @@ class ArrowRowBuffer {
                                   new SFException(
                                       ErrorCode.INVALID_ROW,
                                       value,
-                                      "Unable to parse timezone for TIMESTAMP_TZ column"))
+                                      "Unable to parse timestamp for TIMESTAMP_TZ column"))
                           .toBinary(Integer.parseInt(field.getMetadata().get(COLUMN_SCALE)), true);
                   stats.addIntValue(timeInBinary);
                   break;
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
@@ -319,35 +319,6 @@ class ArrowRowBuffer {
     return response;
   }
 
-  //  private void transferVectors(VectorSchemaRoot source, VectorSchemaRoot target) {
-  //    for (FieldVector vector : this.vectorsRoot.getFieldVectors()) {
-  //      vector.setValueCount(this.rowCount);
-  //      if (vector instanceof DecimalVector) {
-  //        // DecimalVectors do not transfer FieldType metadata when using
-  //        // vector.getTransferPair. We need to explicitly create the new vector to transfer to
-  //        // in order to keep the metadata.
-  //        ArrowType arrowType =
-  //            new ArrowType.Decimal(
-  //                ((DecimalVector) vector).getPrecision(),
-  //                ((DecimalVector) vector).getScale(),
-  //                DECIMAL_BIT_WIDTH);
-  //        FieldType fieldType =
-  //            new FieldType(
-  //                vector.getField().isNullable(), arrowType, null,
-  // vector.getField().getMetadata());
-  //        Field f = new Field(vector.getName(), fieldType, null);
-  //        DecimalVector newVector = new DecimalVector(f, this.allocator);
-  //        TransferPair t = vector.makeTransferPair(newVector);
-  //        t.transfer();
-  //        oldVectors.add((FieldVector) t.getTo());
-  //      } else {
-  //        TransferPair t = vector.getTransferPair(this.allocator);
-  //        t.transfer();
-  //        oldVectors.add((FieldVector) t.getTo());
-  //      }
-  //    }
-  //  }
-
   /**
    * Flush the data in the row buffer by taking the ownership of the old vectors and pass all the
    * required info back to the flush service to build the blob
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
@@ -518,17 +518,20 @@ class ArrowRowBuffer {
             arrowType = Types.MinorType.BIGINT.getType();
             break;
           case SB16:
-            arrowType = Types.MinorType.STRUCT.getType();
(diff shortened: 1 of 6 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
```diff
@@ -489,65 +489,68 @@ Field buildField(ColumnMetadata column) {
...
-                new FieldType(true, Types.MinorType.BIGINT.getType(), null, metadata);
-            FieldType fieldTypeFraction =
-                new FieldType(true, Types.MinorType.INT.getType(), null, metadata);
-            Field fieldEpoch = new Field(FIELD_EPOCH_IN_SECONDS, fieldTypeEpoch, null);
-            Field fieldFraction = new Field(FIELD_FRACTION_IN_NANOSECONDS, fieldTypeFraction, null);
-            children = new LinkedList<>();
-            children.add(fieldEpoch);
-            children.add(fieldFraction);
-            break;
+            {
```
sfc-gh-tjones, line 492: There was no change here right? Just moved inside brackets?

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
```diff
@@ -489,65 +489,68 @@ Field buildField(ColumnMetadata column) {
...
+              arrowType = Types.MinorType.STRUCT.getType();
+              FieldType fieldTypeEpoch =
+                  new FieldType(true, Types.MinorType.BIGINT.getType(), null, metadata);
+              FieldType fieldTypeTimezone =
+                  new FieldType(true, Types.MinorType.INT.getType(), null, metadata);
+              Field fieldEpoch = new Field(FIELD_EPOCH_IN_SECONDS, fieldTypeEpoch, null);
+              Field fieldTimezone = new Field(FIELD_TIME_ZONE, fieldTypeTimezone, null);
+
+              children = new LinkedList<>();
```
sfc-gh-tjones, line 523: Does this have to be a linked list? Can you do something like `children = Arrays.of(fieldEpoch, fieldTimezone)`?

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
```diff
@@ -779,6 +782,89 @@ private float convertRowToArrow(
...
+                      curRowIndex,
+                      timestampWrapper
+                          .getTimeZoneIndex()
+                          .orElseThrow(
+                              () ->
+                                  new SFException(
+                                      ErrorCode.INVALID_ROW,
+                                      value,
+                                      "Unable to parse timezone for TIMESTAMP_TZ column")));
+                  rowBufferSize += 12;
```
sfc-gh-tjones, line 810: Where is 12 coming from?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -109,13 +109,21 @@ static TimestampWrapper validateAndParseTimestampNtzSb16(
...
-          l == 0 ? 0 : Integer.parseInt(items[1]) * (l < 9 ? Power10.intTable[9 - l] : 1);
+      long epoch;
+      int fraction;
+
+      SFTimestamp timestamp = snowflakeDateTimeFormatter.parse(valueString);
+      if (timestamp != null) {
+        epoch = timestamp.getSeconds().longValue();
+        fraction = getFractionFromTimestamp(timestamp);
+      } else {
+        String[] items = valueString.split("\\.");
```
sfc-gh-tjones, line 120: Should we perform a length check here before indexing into items below?

sfc-gh-tjones, line 120: It might help to have an example here to guide why we're doing this.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -130,6 +138,66 @@ static TimestampWrapper validateAndParseTimestampNtzSb16(
...
+            new TimestampWrapper(
+                timestamp.getSeconds().longValue(),
+                fraction,
+                getTimeInScale(getStringValue(epochTimeAtSecondsScale), scale),
+                timestamp.getTimeZoneOffsetMillis(),
+                timestamp);
+        return wrapper;
+      } else {
+        throw new SFException(ErrorCode.INVALID_ROW, input, "Cannot be converted to a time value");
+      }
```
sfc-gh-tjones, line 195: Nit, would it be cleaner to change the above condition to throw if not validate and remove the else?

sfc-gh-tjones (approved):
Some minor comments but LGTM otherwise. Thanks for adding the timestamps back in!

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
sfc-gh-mnaides: correct

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
sfc-gh-mnaides: The Field [constructor](https://github.com/apache/arrow/blob/master/java/vector/src/main/java/org/apache/arrow/vector/types/pojo/Field.java#L102) expects a List, I can make it an ArrayList though.

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
sfc-gh-mnaides: Summing the size of the [BigIntVector](https://arrow.apache.org/docs/java/reference/org/apache/arrow/vector/BigIntVector.html) (8) and the [IntVector](https://arrow.apache.org/docs/java/reference/org/apache/arrow/vector/IntVector.html) (4)

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
```diff
@@ -779,6 +782,89 @@ private float convertRowToArrow(
...
+                  BigInteger timeInBinary =
+                      timestampWrapper
+                          .getSfTimestamp()
+                          .orElseThrow(
+                              () ->
+                                  new SFException(
+                                      ErrorCode.INVALID_ROW,
+                                      value,
+                                      "Unable to parse timezone for TIMESTAMP_TZ column"))
```
sfc-gh-tzhang, line 818: Should the error message be a different one?

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
```diff
@@ -779,6 +782,89 @@ private float convertRowToArrow(
...
+              case SB16:
+                {
+                  StructVector structVector = (StructVector) vector;
+                  BigIntVector epochVector =
+                      (BigIntVector) structVector.getChild(FIELD_EPOCH_IN_SECONDS);
+                  IntVector fractionVector =
+                      (IntVector) structVector.getChild(FIELD_FRACTION_IN_NANOSECONDS);
+                  IntVector timezoneVector = (IntVector) structVector.getChild(FIELD_TIME_ZONE);
+
+                  rowBufferSize += 0.50; // for children vector's null value
```
sfc-gh-tzhang, line 832: this should be 0.375

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
```diff
@@ -779,6 +782,89 @@ private float convertRowToArrow(
...
+                  rowBufferSize += 16;
+                  BigInteger timeInBinary =
+                      timestampWrapper
+                          .getSfTimestamp()
+                          .orElseThrow(
+                              () ->
+                                  new SFException(
+                                      ErrorCode.INVALID_ROW,
+                                      value,
+                                      "Unable to parse timezone for TIMESTAMP_TZ column"))
```
sfc-gh-tzhang, line 858: same comment here

src/main/java/net/snowflake/ingest/streaming/internal/TimestampWrapper.java
```diff
@@ -6,18 +6,35 @@
 
 import java.math.BigInteger;
 import java.util.Objects;
+import java.util.Optional;
+import net.snowflake.client.jdbc.internal.snowflake.common.core.SFTimestamp;
```
sfc-gh-tzhang, line 10: could you add a comment about which this class used for?

sfc-gh-tzhang (approved):
Thanks!

PR #124, opened 2022-02-01
SNOW-518618: remove BC-FIPS dependency

sfc-gh-tzhang:
Remove the bcpkix-fips dependency in the SDK, and use non-fips provider instead of fips provider, hopefully it will fix the DEW issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,15 +83,6 @@ you would need to remove the following scope limits in pom.xml
 &lt;/dependency&gt;
 </pre>
 
-<pre>
-&lt;!-- Remove provided scope from bcpkix-fips --&gt;
-&lt;dependency&gt;
-    &lt;groupId&gt;org.bouncycastle&lt;/groupId&gt;
-    &lt;artifactId&gt;bcpkix-fips&lt;/artifactId&gt;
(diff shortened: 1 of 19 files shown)


sfc-gh-tjones (approved):

sfc-gh-gdoci (approved):
Thank you Toby, I believe it won't be a problem for DEW! 
If I want to always use the latest clientSDK in our tests, is there a way I can do it without having to manually pull master branch, build the latest jar and push it to Nexus? 
E.g. is the latest jar pushed somewhere else automatically every time the version in pom.xml is updated in master?

sfc-gh-tzhang:
> Thank you Toby, I believe it won't be a problem for DEW! If I want to always use the latest clientSDK in our tests, is there a way I can do it without having to manually pull master branch, build the latest jar and push it to Nexus? E.g. is the latest jar pushed somewhere else automatically every time the version in pom.xml is updated in master?

@sfc-gh-gdoci @sfc-gh-mnaides is verifying this change before merging, do you want to work with him to make sure there won't cause the same problem again?

> is there a way I can do it without having to manually pull master branch, build the latest jar and push it to Nexus?

I don't think we have such automation

PR #126, opened 2022-02-03
Allow client to configure parameters

sfc-gh-mnaides:


src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -314,7 +315,7 @@ void distributeFlushTasks() {
       }
 
       // Distribute work at table level, create a new blob if reaching the blob size limit
-      while (itr.hasNext() && totalBufferSize <= MAX_BLOB_SIZE_IN_BYTES) {
+      while (itr.hasNext() && totalBufferSize <= maxBlobSizeInBytes) {
```
sfc-gh-tjones, line 318: Why this change if `maxBlobSizeInByes` is always initialized to the constant? Did you mean to use something from `owningClient` instead?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelFactory.java
```diff
@@ -26,6 +26,7 @@ static SnowflakeStreamingIngestChannelBuilder builder(String name) {
     private String encryptionKey;
     private Long encryptionKeyId;
     private OpenChannelRequest.OnErrorOption onErrorOption;
+    private ParameterProvider parameterProvider;
```
sfc-gh-tjones, line 29: Is this initialized anywhere?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -73,6 +73,9 @@
   // Counter to generate unique request ids per client
   private final AtomicLong counter = new AtomicLong(0);
 
+  // Provides constant values set by Snowflake server
```
sfc-gh-tjones, line 76: These aren't set by the server, right?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -192,7 +201,7 @@ public SnowflakeStreamingIngestClientInternal(
    * @param name the name of the client
    */
   SnowflakeStreamingIngestClientInternal(String name) {
-    this(name, null, null, null, true, null);
+    this(name, null, null, null, true, null, null);
```
sfc-gh-tjones, line 204: Maybe just use an empty map here instead?

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-mnaides: I repurposed my branch that took parameters from the server (including this one) and made some mistakes in the cleanup, including this

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
sfc-gh-mnaides: Right, fixing comment

src/main/java/net/snowflake/ingest/streaming/internal/ParameterProvider.java
```diff
@@ -0,0 +1,106 @@
...
+      this.parameterMap.put(
+          key, parameterOverrides.getOrDefault(key, props.getOrDefault(key, defaultValue)));
+    } else if (parameterOverrides != null) {
+      this.parameterMap.put(key, parameterOverrides.getOrDefault(key, defaultValue));
+    } else if (props != null) {
+      this.parameterMap.put(key, props.getOrDefault(key, defaultValue));
+    }
+  }
+  /**
+   * Sets parameter values by first checking 1. parameterOverrides 2. props 3. default value
```
sfc-gh-tzhang, line 45: can you make it clear that props is the value from the profile.json file?

src/main/java/net/snowflake/ingest/streaming/internal/ParameterProvider.java
```diff
@@ -0,0 +1,106 @@
...
+
+  /**
+   * Construct empty ParameterProvider that will supply default values until updated by
+   * client/configure API call
+   */
+  public ParameterProvider() {}
+
+  private void updateValue(
+      String key, Object defaultValue, Map<String, Object> parameterOverrides, Properties props) {
+    if (parameterOverrides != null && props != null) {
```
sfc-gh-tzhang, line 35: props should be always be not null, right? So we should be checking
1. parameterOverrides is null or not
2. parameterOverrides.get(key) is null or not
3. props.get(key) is null or not

src/main/java/net/snowflake/ingest/streaming/internal/ParameterProvider.java
```diff
@@ -0,0 +1,106 @@
...
+  static final long INSERT_THROTTLE_THRESHOLD_IN_PERCENTAGE_DEFAULT = 5;
+
+  /** Map of parameter name to parameter value. This will be set by client/configure API Call. */
+  private Map<String, Object> parameterMap = new HashMap<>();
+
+  /**
+   * Construct empty ParameterProvider that will supply default values until updated by
+   * client/configure API call
+   */
+  public ParameterProvider() {}
```
sfc-gh-tzhang, line 31: we should just pass `parameterOverrides` and `props` to the construct, so both the `updateValue` and `setParameterMap` function could be simplified

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
```diff
@@ -90,6 +94,11 @@
     this.role = role;
     this.requestBuilder = requestBuilder;
     this.clientName = clientName;
+    /*
+    All integer numbers will be deserialized as longs.
+    If this is false Jackson will deserialize to int or long based on size
+    */
+    this.mapper.configure(DeserializationFeature.USE_LONG_FOR_INTS, true);
```
sfc-gh-tzhang, line 101: everything in this file could be reverted.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -167,6 +171,8 @@
       SharedMetricRegistries.add("Metrics", metrics);
     }
 
+    this.parameterProvider.setParameterMap(parameterOverrides, prop);
```
sfc-gh-tzhang, line 174: Could we add a log after this so we could know the actual values being used in the end?

src/main/java/net/snowflake/ingest/utils/Utils.java
```diff
@@ -74,6 +74,7 @@ public static void assertNotNull(String name, Object value) throws SFException {
    */
   public static Properties createProperties(Properties inputProp, boolean sslEnabled) {
     Properties properties = new Properties();
+    properties.putAll(inputProp);
```
sfc-gh-tzhang, line 77: I think we would still want to use the switch case below to add to the properties, it does more things like toLowerCase and log a entry if the key is not useful.

src/main/java/net/snowflake/ingest/streaming/internal/ParameterProvider.java
sfc-gh-mnaides: The profile.json only really exists in our example, since the client factory just takes a Properties object

src/main/java/net/snowflake/ingest/utils/Utils.java
sfc-gh-mnaides: Please double check my logic, but i think I kept all of the reformatting we do below and figured it was simpler to just pass along all the props even if we don't use them

sfc-gh-tzhang: hmmm, IIUC, you remove a few cases in the switch statement below, so let's say if the input prop is `USER` instead of `user`, we won't able to get it later anymore with our change, right? Also now we will log an entry saying user property is ignored which is not correct. I feel like the old logic is more clear, could we change it back?

PR #130, opened 2022-02-10
SNOW-534105: Enable encryption by default

sfc-gh-tzhang:
After the [fix](https://github.com/snowflakedb/snowflake/pull/46339) at server side, encryption is working properly, now it's the time to remove any non-encrypted code path in the client SDK


sfc-gh-tzhang:
@sfc-gh-tjones this needs to wait for your server side change to enable encryption by default

sfc-gh-tjones (approved):
Thanks Toby!

sfc-gh-mnaides (approved):

PR #131, opened 2022-02-17
Pass in row index corresponding to input rows in response

sfc-gh-japatel:
- Useful to map Kafka offsets with insertErrors so as to send specific records to DLQ. 
- It is possible that rows 1, 5, 9 are bad in on error = continue given 10 rows. 
- We will respond with 0 based indices in insertError.
Included a test.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/InsertValidationResponse.java
+++ b/src/main/java/net/snowflake/ingest/streaming/InsertValidationResponse.java
@@ -49,9 +49,15 @@ public class InsertValidationResponse {
     private final Object rowContent;
     private final SFException exception;
 
-    public InsertError(Object row, SFException exception) {
+    // Used to map this error row with original row in the insertRows Iterable.
+    // i.e the rowIndex can be index 9 in the list of 10 rows.
+    // index is 0 based so as to match with incoming Iterable
+    private final long rowIndex;
+
+    public InsertError(Object row, SFException exception, long rowIndex) {
       this.rowContent = row;
       this.exception = exception;
+      this.rowIndex = rowIndex;
     }
 
     /** Get the row content */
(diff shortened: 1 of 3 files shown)


src/test/java/net/snowflake/ingest/streaming/internal/RowBufferTest.java
```diff
@@ -602,6 +604,31 @@ public void testStringLength() {
     testStringLengthHelper(this.rowBufferOnErrorAbort);
   }
 
+  @Test
+  public void testRowIndexWithMultipleRowsWithErrorr() {
```
sfc-gh-tzhang, line 608: Could you just update existing tests to verify the rowIndex? We have a few, search for `.getInsertErrors()`

sfc-gh-tzhang (approved):
LGTM, one comment about testing to update existing tests

src/main/java/net/snowflake/ingest/streaming/InsertValidationResponse.java
```diff
@@ -68,5 +74,13 @@ public String getMessage() {
     public SFException getException() {
       return this.exception;
     }
+
+    /**
+     * Get the rowIndex. Please note, this index is 0 based so it can be used in fetching nth row
+     * from list.
```
sfc-gh-tzhang, line 80: nit
```suggestion
     * from the input.
```

src/test/java/net/snowflake/ingest/streaming/internal/RowBufferTest.java
sfc-gh-japatel: The existing tests verify other behaviors, I prefer having each test does its own thing. 

Also, we needed a good row and a bad row for that, hence I kept it in a different test.

src/main/java/net/snowflake/ingest/streaming/InsertValidationResponse.java
sfc-gh-japatel: Will do!

PR #134, opened 2022-02-23
SNOW-515458 Fix min/max for binary datatype

sfc-gh-mnaides:
Add correctly encoded min/max info for BINARY type


sfc-gh-tjones (approved):
Debugged with Matt over Slack and saw the fix work, ship it!

PR #135, opened 2022-02-28
SNOW-552292 fix max string ep info when the string is truncated

sfc-gh-mnaides:
We were not correctly truncating the max string EP information in most cases


sfc-gh-tjones (approved):
Nice find and fix!

PR #137, opened 2022-03-03
Debug submit flush task only when isNeedFlush is true to avoid loggin…

sfc-gh-japatel:
…g too many times in DEBUG mode

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
@@ -207,7 +207,7 @@ class FlushService {
         () -> {
           if (this.isNeedFlush) {
             logger.logDebug(
-                "Submit flush task on client={}, isForce={}, isNeedFlush={}, timeDiffMillis={},"
+                "Submit forced or ad-hoc flush task on client={}, isForce={}, isNeedFlush={}, timeDiffMillis={},"
                     + " currentDiffMillis={}",
                 this.owningClient.getName(),
                 isForce,
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
@@ -205,10 +205,10 @@ class FlushService {
   private CompletableFuture<Void> distributeFlush(boolean isForce, long timeDiffMillis) {
     return CompletableFuture.runAsync(
         () -> {
-          if (this.isNeedFlush) {
+          if (this.isNeedFlush || isForce) {
             logger.logDebug(
-                "Submit forced or ad-hoc flush task on client={}, isForce={}, isNeedFlush={}, timeDiffMillis={},"
-                    + " currentDiffMillis={}",
+                "Submit forced or ad-hoc flush task on client={}, isForce={}, isNeedFlush={},"
+                    + " timeDiffMillis={}, currentDiffMillis={}",
                 this.owningClient.getName(),
                 isForce,
                 this.isNeedFlush,
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
@@ -205,22 +205,32 @@ class FlushService {
   private CompletableFuture<Void> distributeFlush(boolean isForce, long timeDiffMillis) {
     return CompletableFuture.runAsync(
         () -> {
-          if (this.isNeedFlush || isForce) {
-            logger.logDebug(
-                "Submit forced or ad-hoc flush task on client={}, isForce={}, isNeedFlush={},"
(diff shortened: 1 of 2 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -205,15 +205,16 @@ String getFilePath() {
...
-          logger.logDebug(
-              "Submit flush task on client={}, isForce={}, isNeedFlush={}, timeDiffMillis={},"
-                  + " currentDiffMillis={}",
-              this.owningClient.getName(),
-              isForce,
-              this.isNeedFlush,
-              timeDiffMillis,
-              System.currentTimeMillis() - this.lastFlushTime);
-
+          if (this.isNeedFlush) {
```
sfc-gh-mnaides, line 208: This will cause us to not debug log flushes triggered by channel closing and the minimum time interval.  Is there a specific issue you're running into here filtering out time based flushes?

sfc-gh-japatel: I think the only reason is the noisy nature/frequency of this. 
Many times when we ask customer to enable debug logging for debugging issues. 

In case of closing I think `isForce`= true, I can add a check if isForce is also true so if either of them is true, we should log. WDYT?

sfc-gh-mnaides: I lean towards keeping it as it is.  This change would remove one debug log line every second which doesn't seem super noisy to me for a debug line

sfc-gh-japatel: I see one more reason is:
1. I deleted the connector and the API it invokes is same when the rebalance tries to close all partition. During KC partition close, lets say I dont need to close client but during connector close(Delete), I do need to close client. 
2. Not closing a client will continue to see this debug log which might be confusing to customers. 

One other work around is to close client also during rebalance..

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -205,15 +205,16 @@ String getFilePath() {
...
+            logger.logDebug(
+                "Submit flush task on client={}, isForce={}, isNeedFlush={}, timeDiffMillis={},"
```
sfc-gh-tzhang, line 210:
```suggestion
                "Submit forced or ad-hoc flush task on client={}, isForce={}, isNeedFlush={}, timeDiffMillis={},"
```

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-mnaides: For a compromise, could we add a parameter that would allow power users to suppress this log line in the way you've outlined but leave the log as is by default?

sfc-gh-tzhang: No objection on removing it, as it may not be very useful, the only thing we will miss is the ability to determine whether the worker has stopped or not

sfc-gh-tzhang (approved):

sfc-gh-tzhang:
Please make sure to conclude with @sfc-gh-mnaides before merging, thanks!

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-japatel: This is how I have set the logger for snowflake ingest
log4j.logger.net.snowflake=DEBUG
We can also set at class level
log4j.logger.net.snowflake.ingest.streaming.internal.FlushService=DEBUG


Or on a completely different level, should I just trace it?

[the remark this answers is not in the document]
sfc-gh-japatel: Yes, https://github.com/snowflakedb/snowflake-ingest-java/pull/137/files#r818974931

sfc-gh-mnaides: I'm fine with dropping it to Trace but logging every flush

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -222,6 +214,25 @@ String getFilePath() {
         this.flushWorker);
   }
 
+  /** If tracing is enabled, print always else, check if it needs flush or is forceful. */
+  private void logFlushTask(boolean isForce, long timeDiffMillis) {
```
sfc-gh-japatel, line 218: Hi @sfc-gh-mnaides let me know if this sounds good! trace everything, if DEBUG, only during isNeedFlush and isForce..

sfc-gh-mnaides (approved):

PR #140, opened 2022-03-07
@snow SNOW-554959 Align chunk encryption with block size and count for the whole interleaved file

sfc-gh-azagrebin:
[SNOW-554959](https://snowflakecomputing.atlassian.net/browse/SNOW-554959)

The decryption happens block-wise on the query side. The block size is 16 bytes for AES/CTR. Hence the SfFile downloads and decrypts file sections aligned to the block size, starting at the beginning of the whole file. The IV is also incremented for each block starting with zero at the beginning of the whole file.

The problem is that the interleaved chunks are not aligned or padded to the block size in the client. The IV starts from zero for each chunk. This breaks the decryption on the query side.

This PR changes the encryption algorithm to use padding to get the file-wide alignment.
The PR also adds block counting starting from the beginning of the file to use it as the IV for each chunk encryption.

The change is tested by ingesting a file with chunks from three tables in devvm.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
@@ -41,6 +41,7 @@ import javax.crypto.IllegalBlockSizeException;
 import javax.crypto.NoSuchPaddingException;
 import net.snowflake.client.jdbc.SnowflakeSQLException;
 import net.snowflake.client.jdbc.internal.google.common.util.concurrent.ThreadFactoryBuilder;
+import net.snowflake.ingest.utils.Constants;
 import net.snowflake.ingest.utils.Cryptor;
 import net.snowflake.ingest.utils.ErrorCode;
 import net.snowflake.ingest.utils.Logging;
@@ -479,16 +480,28 @@ class FlushService {
         byte[] compressedChunkData = BlobBuilder.compress(filePath, chunkData);
 
         // Encrypt the compressed chunk data, the encryption key is derived using the key from
-        // server with the full blob path
+        // server with the full blob path.
+        // We need to maintain IV as a block counter for the whole file, even interleaved,
+        // to align with decryption on the Snowflake query path.
+        // TODO: address alignment for the header SNOW-557866
+        long iv = curDataSize / Constants.ENCRYPTION_ALGORITHM_BLOCK_SIZE_BYTES;
         byte[] encryptedCompressedChunkData =
-            Cryptor.encrypt(compressedChunkData, firstChannel.getEncryptionKey(), filePath);
+            Cryptor.encrypt(compressedChunkData, firstChannel.getEncryptionKey(), filePath, iv);
+
+        // Encryption adds padding to the ENCRYPTION_ALGORITHM_BLOCK_SIZE_BYTES
+        // to align with decryption on the Snowflake query path starting from this chunk offset.
+        // The padding does not have arrow data and not compressed.
+        // Hence, the actual chunk size is smaller by the padding size.
+        // The compression on the Snowflake query path needs the correct size of the compressed
+        // data.
+        int compressedChunkLength = compressedChunkData.length;
 
         // SNOW-514965 Do not default to 0 once server side code to send back TMK ID makes it in
         Long encryptionKeyIdToUse =
             firstChannel.getEncryptionKeyId() != null ? firstChannel.getEncryptionKeyId() : 0L;
 
         // Compute the md5 of the chunk data
-        String md5 = BlobBuilder.computeMD5(encryptedCompressedChunkData);
+        String md5 = BlobBuilder.computeMD5(encryptedCompressedChunkData, compressedChunkLength);
         int encryptedCompressedChunkDataSize = encryptedCompressedChunkData.length;
 
         // Create chunk metadata
@@ -498,7 +511,10 @@ class FlushService {
                 // The start offset will be updated later in BlobBuilder#build to include the blob
                 // header
                 .setChunkStartOffset(curDataSize)
-                .setChunkLength(encryptedCompressedChunkDataSize)
+                // The compressedChunkLength is used because it is the actual data size used for
+                // decompression
+                // and md5 calculation on server side.
+                .setChunkLength(compressedChunkLength)
                 .setChannelList(channelsMetadataList)
                 .setChunkMD5(md5)
                 .setEncryptionKeyId(encryptionKeyIdToUse)
@@ -513,13 +529,14 @@ class FlushService {
 
         logger.logDebug(
             "Finish building chunk in blob={}, table={}, rowCount={}, uncompressedSize={},"
-                + " compressedSize={}, encryptedCompressedSize={}",
+                + " compressedChunkLength={}, encryptedCompressedSize={}",
             filePath,
             firstChannel.getFullyQualifiedTableName(),
             rowCount,
             chunkData.size(),
-            compressedChunkData.length,
-            encryptedCompressedChunkDataSize);
+            compressedChunkLength,
+            encryptedCompressedChunkDataSize,
+            compressedChunkLength);
       }
     }
 
--- a/src/main/java/net/snowflake/ingest/utils/Cryptor.java
+++ b/src/main/java/net/snowflake/ingest/utils/Cryptor.java
@@ -2,6 +2,7 @@ package net.snowflake.ingest.utils;
 
 import static net.snowflake.ingest.utils.Constants.ENCRYPTION_ALGORITHM;
 
+import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.security.InvalidAlgorithmParameterException;
 import java.security.InvalidKeyException;
@@ -18,9 +19,6 @@ import javax.crypto.spec.SecretKeySpec;
 import org.apache.arrow.util.VisibleForTesting;
 
 public class Cryptor {
-
-  /** Initialize 0 IV */
-  private static final IvParameterSpec zeroIv = new IvParameterSpec(new byte[16]);
 
   /**
    * Hashes input bytes using SHA-256.
@@ -95,9 +93,11 @@ public class Cryptor {
    * @param compressedChunkData bytes to encrypt
    * @param encryptionKey symmetric encryption key
    * @param diversifier diversifier for the encryption key
+   * @param iv IV to use for encryption
    * @return encrypted bytes, padded, prefixed with initialization vector
    */
-  public static byte[] encrypt(byte[] compressedChunkData, String encryptionKey, String diversifier)
+  public static byte[] encrypt(
+      byte[] compressedChunkData, String encryptionKey, String diversifier, long iv)
       throws NoSuchPaddingException, NoSuchAlgorithmException, InvalidAlgorithmParameterException,
           InvalidKeyException, IllegalBlockSizeException, BadPaddingException {
     // Generate the derived key
@@ -105,7 +105,8 @@ public class Cryptor {
 
     // Encrypt with zero IV
     Cipher cipher = Cipher.getInstance(ENCRYPTION_ALGORITHM);
-    cipher.init(Cipher.ENCRYPT_MODE, derivedKey, zeroIv);
+    byte[] ivBytes = ByteBuffer.allocate(2 * Long.BYTES).putLong(0).putLong(iv).array();
+    cipher.init(Cipher.ENCRYPT_MODE, derivedKey, new IvParameterSpec(ivBytes));
     return cipher.doFinal(compressedChunkData);
   }
 
@@ -115,10 +116,12 @@ public class Cryptor {
    *
    * @param input bytes to encrypt
    * @param encryptionKey symmetric encryption key
+   * @param diversifier diversifier for the decryption key
+   * @param iv IV to use for decryption
    * @return decrypted ciphertext
    */
   @VisibleForTesting
-  public static byte[] decrypt(byte[] input, String encryptionKey, String diversifier)
+  public static byte[] decrypt(byte[] input, String encryptionKey, String diversifier, long iv)
       throws NoSuchPaddingException, NoSuchAlgorithmException, InvalidAlgorithmParameterException,
           InvalidKeyException, IllegalBlockSizeException, BadPaddingException {
     // Generate the derived key
@@ -126,7 +129,8 @@ public class Cryptor {
 
     // Decrypt with zero IV
     Cipher cipher = Cipher.getInstance(ENCRYPTION_ALGORITHM);
-    cipher.init(Cipher.DECRYPT_MODE, derivedKey, zeroIv);
+    byte[] ivBytes = ByteBuffer.allocate(2 * Long.BYTES).putLong(0).putLong(iv).array();
+    cipher.init(Cipher.DECRYPT_MODE, derivedKey, new IvParameterSpec(ivBytes));
     return cipher.doFinal(input);
   }
 }
--- a/src/test/java/net/snowflake/ingest/streaming/internal/FlushServiceTest.java
+++ b/src/test/java/net/snowflake/ingest/streaming/internal/FlushServiceTest.java
@@ -605,8 +605,8 @@ public class FlushServiceTest {
         Base64.getEncoder().encodeToString("encryption_key".getBytes(StandardCharsets.UTF_8));
     String diversifier = "2021/08/10/blob.bdec";
 
-    byte[] encryptedData = Cryptor.encrypt(data, encryptionKey, diversifier);
-    byte[] decryptedData = Cryptor.decrypt(encryptedData, encryptionKey, diversifier);
+    byte[] encryptedData = Cryptor.encrypt(data, encryptionKey, diversifier, 0);
+    byte[] decryptedData = Cryptor.decrypt(encryptedData, encryptionKey, diversifier, 0);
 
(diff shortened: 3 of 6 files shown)


sfc-gh-azagrebin:
I tried to run the tests for the projects as well but many fail with DB connection.
Here is my profile.json that works if I run the example:
```
{
  "user": "admin",
  "url": "http://snowflake.dev.local:8080",
  "account": "snowflake",
  "private_key": "......",
  "port": 8080,
  "host": "snowflake.dev.local",
  "scheme": "http",
  "connect_string": "jdbc:snowflake://snowflake.dev.local:8080",
  "role": "accountadmin"
}
```

The encryption algorithm with padding again works if I run the example but fails if I run tests with `Cannot find any provider supporting AES/CTR/PKCS7Padding`.

any ideas what I am missing?
cc @sfc-gh-tzhang @sfc-gh-mnaides

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -20,13 +13,7 @@
...
 import java.security.InvalidKeyException;
 import java.security.NoSuchAlgorithmException;
-import java.util.ArrayList;
-import java.util.Calendar;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.TimeZone;
+import java.util.*;
```
sfc-gh-tjones, line 16: Nit, please expand this. We avoid doing wildcard imports as per our style guide.

sfc-gh-tjones (commented):
This looks good to me, but how is IV initialized on the Snowflake side? Don't we need to pass that information to Snowflake?

sfc-gh-azagrebin:
> This looks good to me, but how is IV initialized on the Snowflake side? Don't we need to pass that information to Snowflake?

No, Snowflake calculates the IV the same way as we here, based on the same offset and block size.

sfc-gh-tjones (approved):

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-azagrebin: ups, ide, done

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -479,17 +480,29 @@ BlobMetadata buildAndUpload(String filePath, List<List<ChannelData>> blobData)
...
+
+        // Encryption adds padding to the ENCRYPTION_ALGORITHM_BLOCK_SIZE_BYTES
+        // to align with decryption on the Snowflake query path starting from this chunk offset.
+        // The padding does not have arrow data and not compressed.
+        // Hence, the actual chunk size is smaller by the padding size.
+        // The compression on the Snowflake query path needs the correct size of the compressed
+        // data.
+        int padding = encryptedCompressedChunkData.length - compressedChunkData.length;
+        int encryptedCompressedChunkDataSize = encryptedCompressedChunkData.length;
+        int chunkLength = encryptedCompressedChunkDataSize - padding;
```
sfc-gh-tzhang, line 498: i'm a little confused, so `chunkLength` should just be `compressedChunkData.length`?

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -479,17 +480,29 @@ BlobMetadata buildAndUpload(String filePath, List<List<ChannelData>> blobData)
...
 
         // SNOW-514965 Do not default to 0 once server side code to send back TMK ID makes it in
         Long encryptionKeyIdToUse =
             firstChannel.getEncryptionKeyId() != null ? firstChannel.getEncryptionKeyId() : 0L;
 
         // Compute the md5 of the chunk data
-        String md5 = BlobBuilder.computeMD5(encryptedCompressedChunkData);
-        int encryptedCompressedChunkDataSize = encryptedCompressedChunkData.length;
+        String md5 = BlobBuilder.computeMD5(encryptedCompressedChunkData, chunkLength);
```
sfc-gh-tzhang, line 505: When comparing MD5 at server side, instead of comparing MD5 of the entire file/chunk, we will exclude the padding and compute MD5?

sfc-gh-tzhang: I guess i'm confusing that why for MD5, we're using `chunkLength`, but for CRC checksum at line 525, we're still using `encryptedCompressedChunkDataSize`

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-tzhang: And could you add a log for `chunkLength` at line 527?

src/test/java/net/snowflake/ingest/streaming/internal/FlushServiceTest.java
```diff
@@ -605,8 +605,8 @@ public void testEncryptionDecryption()
         Base64.getEncoder().encodeToString("encryption_key".getBytes(StandardCharsets.UTF_8));
     String diversifier = "2021/08/10/blob.bdec";
 
-    byte[] encryptedData = Cryptor.encrypt(data, encryptionKey, diversifier);
-    byte[] decryptedData = Cryptor.decrypt(encryptedData, encryptionKey, diversifier);
+    byte[] encryptedData = Cryptor.encrypt(data, encryptionKey, diversifier, 0);
```
sfc-gh-tzhang, line 608: could you add a IT test in `StreamingIngestIT` that covers multiple channels for multiple tables?

sfc-gh-tzhang (changes requested):
Left a few questions/comments to make sure the behavior is expected

sfc-gh-mnaides:
Pushed a change to fix the FlushServiceTest issue.  It wasn't a PowerMock issue (but we didn't need PowerMock in that file anyways so I removed it).  The problem was that we weren't registering BouncyCastle as a security provider.  We do register BouncyCastle in several other test utils which is why the tests pass on github and when you run the entire suite, but failed when you just ran FlushServiceTests.

sfc-gh-tzhang:
> Pushed a change to fix the FlushServiceTest issue. It wasn't a PowerMock issue (but we didn't need PowerMock in that file anyways so I removed it). The problem was that we weren't registering BouncyCastle as a security provider. We do register BouncyCastle in several other test utils which is why the tests pass on github and when you run the entire suite, but failed when you just ran FlushServiceTests.

I just ran FlushServiceTest.java locally and didn't see any failure, any idea why?

sfc-gh-mnaides:
> > Pushed a change to fix the FlushServiceTest issue. It wasn't a PowerMock issue (but we didn't need PowerMock in that file anyways so I removed it). The problem was that we weren't registering BouncyCastle as a security provider. We do register BouncyCastle in several other test utils which is why the tests pass on github and when you run the entire suite, but failed when you just ran FlushServiceTests.
> 
> I just ran FlushServiceTest.java locally and didn't see any failure, any idea why?

@sfc-gh-tzhang Did you run with my commit https://github.com/snowflakedb/snowflake-ingest-java/pull/140/commits/cb8a5735c0556e0fa07994d6a54dd399ba301ffd because that should fix the issue.  If you ran it without that commit did you run just the entire FlushServiceTest suite or just FlushServiceTest.testEncryptionDecryption()?

sfc-gh-tzhang:
> > > Pushed a change to fix the FlushServiceTest issue. It wasn't a PowerMock issue (but we didn't need PowerMock in that file anyways so I removed it). The problem was that we weren't registering BouncyCastle as a security provider. We do register BouncyCastle in several other test utils which is why the tests pass on github and when you run the entire suite, but failed when you just ran FlushServiceTests.
> > 
> > 
> > I just ran FlushServiceTest.java locally and didn't see any failure, any idea why?
> 
> @sfc-gh-tzhang Did you run with my commit [cb8a573](https://github.com/snowflakedb/snowflake-ingest-java/commit/cb8a5735c0556e0fa07994d6a54dd399ba301ffd) because that should fix the issue. If you ran it without that commit did you run just the entire FlushServiceTest suite or just FlushServiceTest.testEncryptionDecryption()?

I didn't have your commit, and I tried both the suite and testEncryptionDecryption only, both of them passed

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-azagrebin: True, I did not notice it. I think we should use `chunkLength` for both if they are to be checked/calculated on server side because server side will have only `chunkLength` w/o padding. This is needed for correct decompression.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-azagrebin: good point, leftover

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -479,17 +480,29 @@ BlobMetadata buildAndUpload(String filePath, List<List<ChannelData>> blobData)
         byte[] compressedChunkData = BlobBuilder.compress(filePath, chunkData);
 
         // Encrypt the compressed chunk data, the encryption key is derived using the key from
-        // server with the full blob path
+        // server with the full blob path.
+        // We need to maintain IV as a block counter for the whole file, even interleaved,
+        // to align with decryption on the Snowflake query path.
+        long iv = curDataSize / Constants.ENCRYPTION_ALGORITHM_BLOCK_SIZE_BYTES;
```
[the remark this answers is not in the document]
sfc-gh-azagrebin: This is true, given that the metadata looks to have a dynamic length that is known only after we have built the blob.
Arrow file format puts the meta at the end (not stream that we use). That could be a solution.

src/test/java/net/snowflake/ingest/streaming/internal/FlushServiceTest.java
sfc-gh-azagrebin: I will try. I had a problem with Snowflake connection in devvm with tests.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-azagrebin: though, crc belongs to the file, right?
then this should be all the bytes I guess
where do we check the crc?

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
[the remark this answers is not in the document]
sfc-gh-tzhang: for now, could you create a JIRA to track this and add a TODO here with the JIRA number? thanks!

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -479,16 +480,27 @@ BlobMetadata buildAndUpload(String filePath, List<List<ChannelData>> blobData)
...
-            Cryptor.encrypt(compressedChunkData, firstChannel.getEncryptionKey(), filePath);
+            Cryptor.encrypt(compressedChunkData, firstChannel.getEncryptionKey(), filePath, iv);
+
+        // Encryption adds padding to the ENCRYPTION_ALGORITHM_BLOCK_SIZE_BYTES
+        // to align with decryption on the Snowflake query path starting from this chunk offset.
+        // The padding does not have arrow data and not compressed.
+        // Hence, the actual chunk size is smaller by the padding size.
+        // The compression on the Snowflake query path needs the correct size of the compressed
+        // data.
+        int chunkLength = compressedChunkData.length;
```
sfc-gh-tzhang, line 496:
```suggestion
        int compressedChunkLength = compressedChunkData.length;
```

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-tzhang: > because server side will have only chunkLength w/o padding

Interesting, then how would we figure out the size before decryption? We only use the startOffset?

> where do we check the crc?

hmmm, we don't really use it for now, i think we could leave this as it is. I was just confusing that why we're passing the chunkLength before encryption

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -498,7 +510,7 @@ BlobMetadata buildAndUpload(String filePath, List<List<ChannelData>> blobData)
                 // The start offset will be updated later in BlobBuilder#build to include the blob
                 // header
                 .setChunkStartOffset(curDataSize)
-                .setChunkLength(encryptedCompressedChunkDataSize)
+                .setChunkLength(chunkLength)
```
sfc-gh-tzhang, line 513: could you add a comment explain why we're using the compressedChunkLength, but not encryptedCompressedChunkDataSize

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -513,13 +525,14 @@ BlobMetadata buildAndUpload(String filePath, List<List<ChannelData>> blobData)
 
         logger.logDebug(
             "Finish building chunk in blob={}, table={}, rowCount={}, uncompressedSize={},"
-                + " compressedSize={}, encryptedCompressedSize={}",
+                + " compressedSize={}, encryptedCompressedSize={}, chunkLength={}",
```
sfc-gh-tzhang, line 528: My bad, i believe compressedSize = chunkLength, so no need to add it

src/main/java/net/snowflake/ingest/utils/Cryptor.java
```diff
@@ -97,15 +95,17 @@ private static SecretKey deriveKey(String secretKey, String diversifier) {
    * @param diversifier diversifier for the encryption key
    * @return encrypted bytes, padded, prefixed with initialization vector
    */
-  public static byte[] encrypt(byte[] compressedChunkData, String encryptionKey, String diversifier)
+  public static byte[] encrypt(
```
sfc-gh-tzhang, line 98: please update the javadoc

src/main/java/net/snowflake/ingest/utils/Cryptor.java
```diff
@@ -118,15 +118,16 @@ private static SecretKey deriveKey(String secretKey, String diversifier) {
    * @return decrypted ciphertext
    */
   @VisibleForTesting
-  public static byte[] decrypt(byte[] input, String encryptionKey, String diversifier)
+  public static byte[] decrypt(byte[] input, String encryptionKey, String diversifier, long iv)
```
sfc-gh-tzhang, line 121: here as well

sfc-gh-tzhang (approved):
LGTM, please make sure to add a IT before merging. Thanks!

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-azagrebin: > Interesting, then how would we figure out the size before decryption? We only use the startOffset?
Decryption picks up the padding automatically because of the block alignment.
e.g. if the block size is 4 and compressed size is 7, the encrypted size will be padded to 8.
if server side asks for file section of 7 bytes, the decryption will assume the same padding, download and decrypt 8 bytes.

decompression has to know the exact size that is 7.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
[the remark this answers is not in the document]
sfc-gh-azagrebin: [SNOW-557866](https://snowflakecomputing.atlassian.net/browse/SNOW-557866)

src/test/java/net/snowflake/ingest/streaming/internal/FlushServiceTest.java
sfc-gh-azagrebin: done

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -513,13 +529,14 @@ BlobMetadata buildAndUpload(String filePath, List<List<ChannelData>> blobData)
...
+                + " compressedChunkLength={}, encryptedCompressedSize={}",
             filePath,
             firstChannel.getFullyQualifiedTableName(),
             rowCount,
             chunkData.size(),
-            compressedChunkData.length,
-            encryptedCompressedChunkDataSize);
+            compressedChunkLength,
+            encryptedCompressedChunkDataSize,
+            compressedChunkLength);
```
sfc-gh-tzhang, line 539: @sfc-gh-azagrebin i believe you missed removing this one

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-tzhang: thanks, i think i understand why we need the size before padding now and that's been passed as chunk length, but I'm still confused why we're using `compressedChunkLength` to compute the MD5? i assume we just compare MD5 before decompression and decryption?

sfc-gh-tzhang: nvm, that saw your comment in the code, it's interesting that we're using the `compressedChunkLength` to compute MD5, instead of `encryptedCompressedChunkDataSize`

PR #142, opened 2022-03-08
SNOW-556588 Make timestamp test more robust

sfc-gh-mnaides:
Different test accounts have returned different date string formats.  This makes our test robust against different returned formats


sfc-gh-tzhang (approved):
lgtm, would be good to add a short description in the PR to explain why this is needed

PR #144, opened 2022-03-09
SNOW-557106 enable jmx metrics for streaming

sfc-gh-japatel:
- Default is false.
- Removed thread and jvm metrics.
- Verified the metrics and its mbean names from VisualVM
- Modified the name.

Adding two screenshots
Screen Shot 2022-03-08 at 4 37 18 PM
Screen Shot 2022-03-08 at 4 37 30 PM

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClientFactory.java
+++ b/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClientFactory.java
@@ -47,17 +47,6 @@ public class SnowflakeStreamingIngestClientFactory {
       return this;
     }
 
-    /**
-     * Optional builder to enable JMX metrics for streaming client.
-     *
-     * @param enableJmxMetrics true if metrics per client needs to be emitted
-     * @return this builder instance
-     */
-    public Builder enableJmxMetrics(final boolean enableJmxMetrics) {
-      this.enableJmxMetrics = enableJmxMetrics;
-      return this;
-    }
-
     public SnowflakeStreamingIngestClient build() {
       Utils.assertStringNotNullOrEmpty("client name", this.name);
       Utils.assertNotNull("connection properties", this.prop);
@@ -88,7 +77,7 @@ public class SnowflakeStreamingIngestClientFactory {
       SnowflakeURL accountURL = new SnowflakeURL(this.prop.getProperty(Constants.ACCOUNT_URL));
       Properties prop = Utils.createProperties(this.prop, accountURL.sslEnabled());
       return new SnowflakeStreamingIngestClientInternal(
-          this.name, accountURL, prop, parameterOverrides, this.enableJmxMetrics);
+          this.name, accountURL, prop, parameterOverrides);
     }
   }
 }
--- a/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClientFactory.java
+++ b/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClientFactory.java
@@ -30,9 +30,6 @@ public class SnowflakeStreamingIngestClientFactory {
     // Allows client to override some default parameter values
     private Map<String, Object> parameterOverrides;
 
-    // emits metrics for latency, throughput only if this is enabled. Default is false
-    private boolean enableJmxMetrics = false;
-
     private Builder(String name) {
       this.name = name;
     }
--- a/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClientFactory.java
+++ b/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClientFactory.java
@@ -30,6 +30,9 @@ public class SnowflakeStreamingIngestClientFactory {
     // Allows client to override some default parameter values
     private Map<String, Object> parameterOverrides;
 
+    // emits metrics for latency, throughput only if this is enabled. Default is false
+    private boolean enableJmxMetrics = false;
+
     private Builder(String name) {
       this.name = name;
     }
@@ -41,6 +44,17 @@ public class SnowflakeStreamingIngestClientFactory {
 
     public Builder setParameterOverrides(Map<String, Object> parameterOverrides) {
       this.parameterOverrides = parameterOverrides;
+      return this;
+    }
+
+    /**
+     * Optional builder to enable JMX metrics for streaming client.
+     *
+     * @param enableJmxMetrics true if metrics per client needs to be emitted
+     * @return this builder instance
+     */
+    public Builder enableJmxMetrics(final boolean enableJmxMetrics) {
+      this.enableJmxMetrics = enableJmxMetrics;
       return this;
     }
 
@@ -74,7 +88,7 @@ public class SnowflakeStreamingIngestClientFactory {
       SnowflakeURL accountURL = new SnowflakeURL(this.prop.getProperty(Constants.ACCOUNT_URL));
       Properties prop = Utils.createProperties(this.prop, accountURL.sslEnabled());
       return new SnowflakeStreamingIngestClientInternal(
-          this.name, accountURL, prop, parameterOverrides);
+          this.name, accountURL, prop, parameterOverrides, this.enableJmxMetrics);
     }
   }
 }
--- a/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
@@ -16,6 +16,7 @@ import static net.snowflake.ingest.utils.Constants.REGISTER_BLOB_ENDPOINT;
 import static net.snowflake.ingest.utils.Constants.RESPONSE_SUCCESS;
 import static net.snowflake.ingest.utils.Constants.ROW_SEQUENCER_IS_COMMITTED;
 import static net.snowflake.ingest.utils.Constants.SNOWPIPE_STREAMING_JMX_METRIC_PREFIX;
+import static net.snowflake.ingest.utils.Constants.SNOWPIPE_STREAMING_JVM_AND_THREAD_METRICS_REGISTRY;
 import static net.snowflake.ingest.utils.Constants.SNOWPIPE_STREAMING_SHARED_METRICS_REGISTRY;
 import static net.snowflake.ingest.utils.Constants.USER;
 
@@ -24,8 +25,11 @@ import com.codahale.metrics.Meter;
 import com.codahale.metrics.MetricFilter;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.SharedMetricRegistries;
+import com.codahale.metrics.Slf4jReporter;
 import com.codahale.metrics.Timer;
 import com.codahale.metrics.jmx.JmxReporter;
+import com.codahale.metrics.jvm.MemoryUsageGaugeSet;
+import com.codahale.metrics.jvm.ThreadStatesGaugeSet;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.io.IOException;
 import java.security.KeyPair;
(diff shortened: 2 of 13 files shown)


src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClientFactory.java
```diff
@@ -44,6 +47,17 @@ public Builder setParameterOverrides(Map<String, Object> parameterOverrides) {
...
+     * Optional builder to enable JMX metrics for streaming client.
+     *
+     * @param enableJmxMetrics true if metrics per client needs to be emitted
+     * @return this builder instance
+     */
+    public Builder enableJmxMetrics(final boolean enableJmxMetrics) {
+      this.enableJmxMetrics = enableJmxMetrics;
+      return this;
+    }
+
```
sfc-gh-tzhang, line 60: could you add everything side `parameterOverrides`?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -157,21 +168,34 @@
 
     this.flushService = new FlushService(this, this.channelCache, this.isTestMode);
 
-    if (ENABLE_PERF_MEASUREMENT) {
+    if (this.enableJmxMetrics) {
```
sfc-gh-tzhang, line 171: we still want memory and threads information for JVM, it's not useful for customer, but it's useful when we debugging memory issues. And with your change, are we still enable to see the results in the log?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -432,9 +459,9 @@ public void close() throws Exception {
     isClosed = true;
     this.channelCache.closeAllChannels();
 
-    // Collect the perf metrics before closing if needed
-    if (metrics != null) {
-      Slf4jReporter.forRegistry(metrics).outputTo(logger.getLogger()).build().report();
+    // unregister jmx metrics
+    if (this.enableJmxMetrics && this.metrics != null) {
```
sfc-gh-tzhang, line 463: nit: we should be checking `this.metrics` only

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -586,4 +618,35 @@ FlushService getFlushService() {
...
+   */
+  static ObjectName getObjectName(String clientName, String jmxDomain, String metricName) {
+    try {
+      StringBuilder sb = new StringBuilder(jmxDomain).append(":clientName=").append(clientName);
+
+      sb.append(",name=").append(metricName);
+
+      return new ObjectName(sb.toString());
+    } catch (MalformedObjectNameException e) {
+      logger.logWarn("Could not create Object name for MetricName:{}", metricName);
```
sfc-gh-tzhang, line 639: We try to match the server side coding standard, so please use =
```suggestion
      logger.logWarn("Could not create Object name for MetricName={}", metricName);
```

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -586,4 +618,35 @@ FlushService getFlushService() {
...
+      throw new SFException(ErrorCode.INTERNAL_ERROR, "Invalid metric name");
+    }
+  }
+
+  /** Unregister all streaming related metrics from registry */
+  private void removeMetricsFromRegistry() {
+    if (metrics.getMetrics().size() != 0) {
+      logger.logDebug("Unregistering all metrics for client:{}", this.getName());
```
sfc-gh-tzhang, line 647: same here

src/main/java/net/snowflake/ingest/utils/Constants.java
```diff
@@ -58,6 +58,10 @@
   // Parameters
   public static final boolean DISABLE_BACKGROUND_FLUSH = false;
   public static final boolean COMPRESS_BLOB_TWICE = false;
-  public static final boolean ENABLE_PERF_MEASUREMENT = false;
   public static final boolean BLOB_NO_HEADER = true;
+
+  // Metrics
+  public static final String STREAMING_JMX_METRIC_PREFIX = "snowflake.ingest.sdk";
+
```
sfc-gh-tzhang, line 65: nit: extra line

src/main/java/net/snowflake/ingest/utils/Constants.java
```diff
@@ -58,6 +58,10 @@
...
+  public static final String STREAMING_SHARED_METRICS_REGISTRY = "StreamingSnowpipeMetrics";
```
sfc-gh-tzhang, line 66:
```suggestion
  public static final String STREAMING_SHARED_METRICS_REGISTRY = "SnowpipeStreamingMetrics";
```

src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClientFactory.java
sfc-gh-japatel: I am a bit hesitant about that because that is not a good interface for interaction from client perspective. 

`enableJmxMetrics` is descriptive enough to use. 

Infact I would rather not keep `setParameterOverrides` since it is inviting customers to reset defaults. I thought we were opposed to that. (Althought it is open source)

sfc-gh-tzhang: This [doc](https://docs.google.com/document/d/1m-vMJjFwd0_uAWidk1tv0DDp-HUdZqBCe5rZjHdFUzA/edit) that @sfc-gh-mnaides wrote keeps track the reason about why we are doing this, we won't document this option. But basically all the parameters/constants override should happen inside that class, feel free to suggest if you have a better idea

sfc-gh-japatel: I see. I wasnt aware of this. thank you! 

Lets just go with setParameterOverrides then without creating going through more changes.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
sfc-gh-japatel: if param is false, we wont have those metrics?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
sfc-gh-japatel: sure will do!

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
sfc-gh-japatel: We wont be able to see results in log. Should we keep memory and thread related info in separate metrics object then?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -113,6 +120,9 @@
   Meter uploadThroughput; // Throughput for uploading blobs
   Meter inputThroughput; // Throughput for inserting into the Arrow buffer
 
+  // JVM and thread related metrics
+  MetricRegistry jvmAndThreadMetrics;
```
sfc-gh-tzhang, line 124: can we change it to `jvmMetrics` or `jvmMemoryAndThreadMetrics`?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -432,9 +432,12 @@ public void close() throws Exception {
...
     this.channelCache.closeAllChannels();
 
-    // Collect the perf metrics before closing if needed
-    if (metrics != null) {
-      Slf4jReporter.forRegistry(metrics).outputTo(logger.getLogger()).build().report();
+    // unregister jmx metrics
+    if (this.metrics != null) {
+      removeMetricsFromRegistry();
+
+      // LOG jvm and thread metrics at the end
```
sfc-gh-tzhang, line 439: There're a few other places in this PR also have this problem
```suggestion
      // LOG jvm memory and thread metrics at the end
```

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -157,21 +169,8 @@
...
-      flushLatency = metrics.timer(MetricRegistry.name(name, "flush", "latency"));
-      buildLatency = metrics.timer(MetricRegistry.name(name, "build", "latency"));
-      uploadLatency = metrics.timer(MetricRegistry.name(name, "upload", "latency"));
-      registerLatency = metrics.timer(MetricRegistry.name(name, "register", "latency"));
-      uploadThroughput = metrics.meter(MetricRegistry.name(name, "upload", "throughput"));
-      inputThroughput = metrics.meter(MetricRegistry.name(name, "input", "throughput"));
-      metrics.register(MetricRegistry.name("jvm", "memory"), new MemoryUsageGaugeSet());
-      metrics.register(MetricRegistry.name("jvm", "threads"), new ThreadStatesGaugeSet());
-      SharedMetricRegistries.add("Metrics", metrics);
+    if (this.parameterProvider.hasEnabledSnowpipeStreamingJmxMetrics()) {
```
sfc-gh-tzhang, line 172: It contains both jmx and jvm metrics
```suggestion
    if (this.parameterProvider.hasEnabledSnowpipeStreamingMetrics()) {
```

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -586,4 +589,73 @@ FlushService getFlushService() {
...
+                    getObjectName(this.getName(), jmxDomain, metricName))
+            .build();
+    jmxReporter.start();
+
+    // add JVM and thread metrics too
+    jvmAndThreadMetrics.register(MetricRegistry.name("jvm", "memory"), new MemoryUsageGaugeSet());
+    jvmAndThreadMetrics.register(MetricRegistry.name("jvm", "threads"), new ThreadStatesGaugeSet());
+
+    SharedMetricRegistries.add(
+        SNOWPIPE_STREAMING_JVM_AND_THREAD_METRICS_REGISTRY, jvmAndThreadMetrics);
```
sfc-gh-tzhang, line 628: Could we still output these to logger as well, it will be helpful when we can't use jmx (like in temptest, we're looking at logs uploaded to `workflow_logs_v`

sfc-gh-tzhang (approved):

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
sfc-gh-japatel: oh yes, `jvmMemoryAndThreadMetrics` is better.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
sfc-gh-japatel: will change

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
sfc-gh-japatel: yes we do it in close() just like before

PR #146, opened 2022-03-22
nosnow-run streaming it tests as part of github build

sfc-gh-mnaides:


pom.xml
```diff
@@ -180,11 +180,6 @@
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-failsafe-plugin</artifactId>
                     <version>3.0.0-M5</version>
-                    <configuration>
```
sfc-gh-japatel, line 183: This will run against sfctest0. If this passes, I can run tests from KC too!

sfc-gh-tzhang:
@sfc-gh-mnaides is there any blocker on merging this? We would like to have IT coverage, otherwise I keep getting failures from manual runs and need to investigate which change was causing this...

PR #147, opened 2022-03-24
NO-SNOW: update code owner 

sfc-gh-tzhang:
- Update code owner so the streaming ingest team could be added by default when a PR is opened or an issue is created
- Add myself explicitly for easy PR tracking


sfc-gh-japatel (approved):
lgtm

PR #148, opened 2022-03-25
SNOW-560665: remove usage of instanceof for type checking

sfc-gh-tzhang:
1. Remove usage of instanceof for type checking to fix the Jackson shade dependency issue
2. Always require an encryption_key_id from server side
3. Some minor changes


sfc-gh-tzhang:
@sfc-gh-mnaides Need your input on this fix, PTAL!

sfc-gh-mnaides (approved):
lgtm!

PR #149, opened 2022-03-26
SNOW-564031: Incorrect EP for NULL columns causing BDEC registration failure

sfc-gh-tzhang:
When all the data for a given column in a BDEC file is NULL, the min/max int value expected from server side [should be 0 instead of NULL](https://github.com/snowflakedb/snowflake/blob/dcf9fbfbf6d4933c5aa5013a3a396ee4fdfa8374/GlobalServices/src/main/java/com/snowflake/sql/ep/SqlExpressionPropertiesBuilder.java#L364), this change fixes this issue to avoid BDEC registration failures. Thanks to @sfc-gh-gdoci who identified this bug!

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
@@ -4,8 +4,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import java.math.BigInteger;
 import java.util.Objects;
 
-// TODO https://snowflakecomputing.atlassian.net/browse/SNOW-354886.
-//  Audit register endpoint/FileColumnPorpertyDTO property list.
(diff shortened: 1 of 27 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
```diff
@@ -35,10 +35,19 @@
   // for binary or string columns
   private long maxLength;
 
+  // Default value to use for min/max int when all data in the given column is NULL
+  public static final BigInteger defaultMinMaxIntVal = BigInteger.valueOf(0);
```
sfc-gh-japatel, line 39: Nit: public static final can be all caps. 
```suggestion
  public static final BigInteger DEFAULT_MIN_MAX_INT_VAL_FOR_EP = BigInteger.valueOf(0);
```

sfc-gh-tjones (approved):
Nice!

PR #150, opened 2022-03-26
Release version 1.0.2-beta

sfc-gh-tzhang:
Release note:

1. Fix a blocker issue that would cause data querying failure if a client has multiple channels that belongs to different tables
2. Add JMX metrics
3. Remove usage of instanceof during type checking
4. Fix an issue that file registration would fail for columns with all NULLs
5. Support SDK with proxy server
6. Support Date input as a String


sfc-gh-japatel (approved):
lgtm!

sfc-gh-tjones (approved):

PR #153, opened 2022-03-31
nosnow- support string date formats

sfc-gh-mnaides:
Add support for string Date formats


sfc-gh-tzhang (approved):
LGTM, thanks!

src/test/java/net/snowflake/ingest/streaming/internal/DataValidationUtilTest.java
```diff
@@ -412,6 +412,7 @@ public void testValidateAndParseLong() {
 
   @Test
   public void testValidateAndParseDate() {
```
sfc-gh-tjones, line 414: Should we have some other tests here as well? Like `1967-06-23 01:01:01`?

sfc-gh-tjones (approved):
Nice! See my comment regarding more tests.

PR #154, opened 2022-04-01
SNOW-566797: Support SDK with proxy server 

sfc-gh-tzhang:
Looks like we're not setting proxy properties as part of the `uploadWithoutConnection` call to upload files, this PR adds the support. Will need to see how we would setup the proxy server as part of the GH action if we want to add a real IT tests in the future.

### DIFF
--- a/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
+++ b/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
@@ -7,8 +7,10 @@ package net.snowflake.ingest.utils;
 import static net.snowflake.ingest.utils.Utils.isNullOrEmpty;
 
 import java.security.Security;
+import java.util.Properties;
 import java.util.concurrent.TimeUnit;
 import javax.net.ssl.SSLContext;
+import net.snowflake.client.core.SFSessionProperty;
 import org.apache.http.HttpHost;
 import org.apache.http.HttpRequest;
 import org.apache.http.HttpResponse;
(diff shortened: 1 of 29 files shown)


src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -193,4 +195,40 @@ private static String getRequestUriFromContext(final HttpContext httpContext) {
...
+
+  /**
+   * Helper method to decide whether to add any properties related to proxy server. These properties
+   * are passed on to snowflake JDBC while calling put API.
+   *
+   * @return proxy parameters that could be used by JDBC
+   */
+  public static Properties generateProxyPropertiesForJDBC() {
+    Properties proxyProperties = new Properties();
+    if ("true".equalsIgnoreCase(System.getProperty(USE_PROXY))) {
```
sfc-gh-tjones, line 207: Nit, is there some kind of `BooleanUtils` equivalent we can use here?

sfc-gh-tjones (approved):
Thank you!

sfc-gh-japatel:
lgtm! I have setup proxy on Github Actions, you can take a look at this one: https://github.com/snowflakedb/snowflake-kafka-connector/blob/master/.github/workflows/IntegrationTestAws.yml#L39-L51

Can you file a jira to add IT tests for put files through proxy?

sfc-gh-japatel (approved):

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-tzhang: Yes, updated

sfc-gh-tzhang:
> I have setup proxy on Github Actions, you can take a look at this one: https://github.com/snowflakedb/snowflake-kafka-connector/blob/master/.github/workflows/IntegrationTestAws.yml#L39-L51
> 
> Can you file a jira to add IT tests for put files through proxy?

https://snowflakecomputing.atlassian.net/browse/SNOW-568319, btw, the SAP folks were able to verify the behavior after setting the HOST and PORT, see https://snowflake.slack.com/archives/C02LAQ1HABF/p1648827312148909?thread_ts=1648649091.308319&cid=C02LAQ1HABF

PR #155, opened 2022-04-01
NO-SNOW: Format code

sfc-gh-kbregula:
Hi.

I'm lazy and don't want to install a code formatting tool, and then bother manually running it, so I created a script that installs and formats code with one simple `./format.sh` command.

Best regards,
Kamil Breguła


format.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd "$SCRIPT_DIR"
+
+DOWNLOAD_URL="https://github.com/google/google-java-format/releases/download/v1.15.0/google-java-format-1.15.0-all-deps.jar"
+JAR_FILE="./.cache/google-java-format-1.15.0-all-deps.jar"
```
sfc-gh-japatel, line 8: The google java formatter is actually pinned to version 1.10.0

sfc-gh-japatel: https://github.com/snowflakedb/snowflake-ingest-java/blob/master/.github/workflows/GoogleJavaFormat.yml#L19

We can either upgrade it to 1.15.0 or use the pinned version here. 

I am fine with either.

sfc-gh-kbregula: I updated file `GoogleJavaFormat.yml `to use the same script to make sure it works.

Additionally, I added a git diff display after executing the command, so we can see immediately where the errors are.

Additonally2, I have remove the dependency on thirt-party action which may raise security concerns because it does not have a very good reputation (only 56 stars) and its use did not meet the [Github Action security requirements](https://snowflakecomputing.atlassian.net/wiki/spaces/CLO/pages/1211302564/GH+Actions+Security+Requirements), especially we used tag instead of SHA1 commit hash.

sfc-gh-tzhang:
> Hi.
> 
> I'm lazy and don't want to install a code formatting tool, and then bother manually running it, so I created a script that installs and formats code with one simple `./format.sh` command.
> 
> Best regards, Kamil Breguła

Another option I use is File Watcher or Action On Save tool in the devvm Intellij, you only need to set it up once

format.sh
sfc-gh-tzhang: Could we keep the version at 1.10? I think that's the one used in the devvm Intellij, so anyone using the devvm (like me) will revert the reformat changes you have below

.github/workflows/GoogleJavaFormat.yml
```diff
@@ -11,10 +11,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: GoogleJavaFormat
```
sfc-gh-tzhang, line 14: @sfc-gh-afarra since you made the original change, let us know if you have any concern

format.sh
sfc-gh-japatel: Sounds good! Thanks

sfc-gh-kbregula: @sfc-gh-tzhang Good point. I changed version to 1.10.

sfc-gh-tzhang (approved):

PR #157, opened 2022-04-06
no-snow fix error message printing

sfc-gh-japatel:
- prints the message on KC if parameter for message code 0024 is not
  passed
  `retry for getLatestCommittedOffsetToken. Retry no:1, message:Get channel status request failed: {0}.`
- `IngestResponseException` calls super constructor which puts in the detailedMessage used in `e.getMessage()`

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
@@ -305,7 +305,7 @@ public class SnowflakeStreamingIngestClientInternal implements SnowflakeStreamin
 
       return channel;
     } catch (IOException | IngestResponseException e) {
-      throw new SFException(e, ErrorCode.OPEN_CHANNEL_FAILURE);
+      throw new SFException(e, ErrorCode.OPEN_CHANNEL_FAILURE, e.getMessage());
     }
   }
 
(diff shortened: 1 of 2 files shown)


sfc-gh-tjones (approved):

sfc-gh-tzhang:
> * prints the message on KC if parameter for message code 0024 is not
>   passed
>   `retry for getLatestCommittedOffsetToken. Retry no:1, message:Get channel status request failed: {0}.`
> * `IngestResponseException` calls super constructor which puts in the detailedMessage used in `e.getMessage()`

We pass the exception itself into the SFException, so there should be a CausedBy section which should the exception already?

sfc-gh-tzhang (approved):

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -342,7 +342,7 @@ ChannelsStatusResponse getChannelsStatus(List<SnowflakeStreamingIngestChannelInt
 
       return response;
     } catch (IOException | IngestResponseException e) {
-      throw new SFException(e, ErrorCode.CHANNEL_STATUS_FAILURE);
+      throw new SFException(e, ErrorCode.CHANNEL_STATUS_FAILURE, e.getMessage());
```
sfc-gh-tzhang, line 345: Could you do this for ErrorCode.OPEN_CHANNEL_FAILURE, ErrorCode.REGISTER_BLOB_FAILURE, ErrorCode.CLIENT_CONFIGURE_FAILURE as well? I think it's redundant to pass the message again, but if you want to add it, please do it for all APIs

sfc-gh-japatel: IMO, it is not redundant because it will be used to construct the message which is written in ```0024=Get channel status request failed: {0}.```

sfc-gh-japatel: Will add in other messages too!

PR #158, opened 2022-04-07
SNOW-556750 retry certain streaming ingest specific retries

sfc-gh-mnaides:
Automatically retry streaming ingest API calls for response codes 7 (chunk queue full) and 10 (general retry)


src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestUtils.java
```diff
@@ -0,0 +1,86 @@
...
+        ServiceResponseHandler.unmarshallStreamingIngestResponse(
+            httpClient.execute(
+                requestBuilder.generateStreamingIngestPostRequest(payload, endpoint, message)),
+            targetClass,
+            apiName);
+
+    // Check for Snowflake specific response code
+    if (response.getStatusCode() != RESPONSE_SUCCESS) {
+      if (response.getStatusCode() == RESPONSE_ERR_GENERAL_EXCEPTION_RETRY_REQUEST
+          || response.getStatusCode() == RESPONSE_ERR_ENQUEUE_TABLE_CHUNK_QUEUE_FULL) {
```
sfc-gh-tjones, line 60: FYI we only use `RESPONSE_ERR_ENQUEUE_TABLE_CHUNK_QUEUE_FULL` on a per-blob chunk basis for a table when calling the write API. The response looks like this for that API endpoint:

```
     {
            'blobs': [
                {
                    'chunks': [
                        {
                           'database' : STRING,
                           'schema' : STRING,
                           'table' : STRING,
                           'channels': [
                             {
                               'status_code': long,
                               'channel' : STRING,
                               'client_sequencer' : LONG,
                             },
                             .. (more channels)
                            ]
                        },
                        ... (more chunks)
                    ]
                },
               ... (more blobs
```
`status_code` within the `channels` array will all be set to the chunk full response.

src/main/java/net/snowflake/ingest/streaming/internal/ChannelsStatusResponse.java
```diff
@@ -8,7 +8,7 @@
 import java.util.List;
 
 /** Class used to serialize a response for the channels status endpoint */
-class ChannelsStatusResponse {
+class ChannelsStatusResponse extends StreamingIngestResponse {
```
sfc-gh-tzhang, line 11: what about the client/configure response?

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestResponse.java
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (c) 2021 Snowflake Computing Inc. All rights reserved.
+ */
+
+package net.snowflake.ingest.streaming.internal;
+
+/** Response to the OpenChannelRequest */
```
sfc-gh-tzhang, line 7: please update

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestResponse.java
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (c) 2021 Snowflake Computing Inc. All rights reserved.
```
sfc-gh-tzhang, line 2: nit
```suggestion
 * Copyright (c) 2022 Snowflake Computing Inc. All rights reserved.
```

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestUtils.java
```diff
@@ -0,0 +1,86 @@
...
+import java.io.IOException;
+import java.util.Map;
+import net.snowflake.ingest.connection.IngestResponseException;
+import net.snowflake.ingest.connection.RequestBuilder;
+import net.snowflake.ingest.connection.ServiceResponseHandler;
+import net.snowflake.ingest.utils.ErrorCode;
+import net.snowflake.ingest.utils.SFException;
+import org.apache.http.client.HttpClient;
+
+public class StreamingIngestUtils {
```
sfc-gh-tzhang, line 18: We have very similar retry logic in `HttpUtil.java` for all Ingest APIs, could you add your logic there and consolidate (see `getServiceUnavailableRetryStrategy`)?

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestUtils.java
sfc-gh-tzhang: @sfc-gh-mnaides You could take a look at the end of `registerBlobs` to see how to iterate through each chunk status

@sfc-gh-tjones I believe we discussed offline that the client should only check on `RESPONSE_ERR_GENERAL_EXCEPTION_RETRY_REQUEST`? Because of a few reasons: 
1. Customer/client doesn't need to know the specific error, we should log it at server side instead
2. There is no automated way to keep the status code in sync between client and server (and we don't have tests), so it's in general a good practice to not keep adding unnecessary status code to client

sfc-gh-tzhang (commented):
Thanks for taking over this task, left some comments!

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestUtils.java
sfc-gh-mnaides: I opted to separate out this logic because it interacts with streaming ingest specific status codes.  I didn't want to add unnecessary overhead to every non-streaming ingest api call.  The method `getServiceUnavailableRetryStrategy` is only invoked if the API returns a certain set of exceptional responses.  For the cases I'm dealing with here we already have a 200 response so that method wouldn't be invoked.  In order to make it work with that framework we'd need to add an interceptor that would convert our specific use cases (200 http code, but non-success streaming ingest specific code) into an IOException that would then cause `getServiceUnavailableRetryStrategy` to be invoked.

sfc-gh-tjones (approved):
Nice! Thanks for separating out the chunk queue full case.

sfc-gh-tzhang (changes requested):
I feel like the change in this PR is overcomplicated given that we never saw RESPONSE_ERR_ENQUEUE_TABLE_CHUNK_QUEUE_FULL returned as part of KC or SAP connector testing and we're currently using a very high number for chunk queue limit. On the other hand, it's in general a good practice to keep the client logic simpler, especially this retry logic will be needed in all other language SDKs as well. 

So I'm proposing we do the following:
1. Only check on RESPONSE_ERR_GENERAL_EXCEPTION_RETRY_REQUEST at both blob and chunk level, and make any changes needed at server side to mask RESPONSE_ERR_ENQUEUE_TABLE_CHUNK_QUEUE_FULL (like what we did for ERR_TABLE_SUSPENDED_ON_NODE)
2. Simply retry the entire request and make any changes needed at server side to skip the chunks that are registered already, given the low chance of this error, I don't think performance is a concern right now.
WDYT?

sfc-gh-mnaides:
> I feel like the change in this PR is overcomplicated given that we never saw RESPONSE_ERR_ENQUEUE_TABLE_CHUNK_QUEUE_FULL returned as part of KC or SAP connector testing and we're currently using a very high number for chunk queue limit. On the other hand, it's in general a good practice to keep the client logic simpler, especially this retry logic will be needed in all other language SDKs as well.
> 
> So I'm proposing we do the following:
> 
> 1. Only check on RESPONSE_ERR_GENERAL_EXCEPTION_RETRY_REQUEST at both blob and chunk level, and make any changes needed at server side to mask RESPONSE_ERR_ENQUEUE_TABLE_CHUNK_QUEUE_FULL (like what we did for ERR_TABLE_SUSPENDED_ON_NODE)
> 2. Simply retry the entire request and make any changes needed at server side to skip the chunks that are registered already, given the low chance of this error, I don't think performance is a concern right now.
>    WDYT?

I'd rather see if we can find a solution on the client sdk rather than introduce logic on the server to skip chunks or to mask errors.  Can you say a little more about what you think is overcomplicated?  This PR does two things

1. Will retry streaming ingest api requests if they have a top level status_code that asks for a retry.  
2. The register file api will retry registering only those blob chunks that found a full queue on the server.  I think this is logically the most complicated part of this PR. Current behavior is to invalidate the channel and have the client reopen it.  This will invalidate the channel on the server and cause all of the registration requests in the full queue to be invalidated, forcing the client to replay all the data that was in the commit queue.

I think this complexity is justified and I'd be happy to add more testing if you're worried we lack testing.

sfc-gh-tzhang:
> > I feel like the change in this PR is overcomplicated given that we never saw RESPONSE_ERR_ENQUEUE_TABLE_CHUNK_QUEUE_FULL returned as part of KC or SAP connector testing and we're currently using a very high number for chunk queue limit. On the other hand, it's in general a good practice to keep the client logic simpler, especially this retry logic will be needed in all other language SDKs as well.
> > So I'm proposing we do the following:
> > 
> > 1. Only check on RESPONSE_ERR_GENERAL_EXCEPTION_RETRY_REQUEST at both blob and chunk level, and make any changes needed at server side to mask RESPONSE_ERR_ENQUEUE_TABLE_CHUNK_QUEUE_FULL (like what we did for ERR_TABLE_SUSPENDED_ON_NODE)
> > 2. Simply retry the entire request and make any changes needed at server side to skip the chunks that are registered already, given the low chance of this error, I don't think performance is a concern right now.
> >    WDYT?
> 
> I'd rather see if we can find a solution on the client sdk rather than introduce logic on the server to skip chunks or to mask errors. Can you say a little more about what you think is overcomplicated? This PR does two things
> 
> 1. Will retry streaming ingest api requests if they have a top level status_code that asks for a retry.
> 2. The register file api will retry registering only those blob chunks that found a full queue on the server.  I think this is logically the most complicated part of this PR. Current behavior is to invalidate the channel and have the client reopen it.  This will invalidate the channel on the server and cause all of the registration requests in the full queue to be invalidated, forcing the client to replay all the data that was in the commit queue.
> 
> I think this complexity is justified and I'd be happy to add more testing if you're worried we lack testing.

> I'd rather see if we can find a solution on the client sdk rather than introduce logic on the server to skip chunks or to mask errors.

Why is this? I think in general, the complicated logic should go into server side and the client side should just do an simple back-off retry with a few line of changes?

> Will retry streaming ingest api requests if they have a top level status_code that asks for a retry.

I believe RESPONSE_ERR_GENERAL_EXCEPTION_RETRY_REQUEST could happen on chunk level as well, but please do double check at server side

> The register file api will retry registering only those blob chunks that found a full queue on the server. I think this is logically the most complicated part of this PR.

Yes, I understand that you're trying to do things in a perfect way, but I just feel like it doesn't worth to add such complication given that the low chance of RESPONSE_ERR_ENQUEUE_TABLE_CHUNK_QUEUE_FULL from happening. At least based on what I know, we never hit it in KC or SAP connectors. So why can't we just add a few lines of code to resend the same request again?

> I think this complexity is justified and I'd be happy to add more testing if you're worried we lack testing.

No, I trust you on the logic and it's not about testing :) I just feel like in general this retry logic is complicated to be at client side, with the potential that it will be in all other SDKs as well.

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestUtils.java
```diff
@@ -0,0 +1,114 @@
...
+    public Long apply(T input) {
+      return input.getStatusCode();
+    }
+  }
+
+  private static final DefaultStatusGetter defaultStatusGetter = new DefaultStatusGetter();
+
+  private static final ObjectMapper objectMapper = new ObjectMapper();
+
+  static <T extends StreamingIngestResponse> T executeWithRetries(
```
sfc-gh-tzhang, line 33: Why we have three flavors of `executeWithRetries`, do you think we could simplify it?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -405,21 +427,92 @@ void registerBlobs(List<BlobMetadata> blobs) {
...
-                                        channelCache.invalidateChannelIfSequencersMatch(
-                                            chunkStatus.getDBName(),
-                                            chunkStatus.getSchemaName(),
-                                            chunkStatus.getTableName(),
-                                            channelStatus.getChannelName(),
-                                            channelStatus.getChannelSequencer());
+                                        // If the chunk queue is full, we wait and retry the chunks
+                                        if (channelStatus.getStatusCode()
+                                                == RESPONSE_ERR_ENQUEUE_TABLE_CHUNK_QUEUE_FULL
+                                            && executionCount < MAX_API_RETRY) {
```
sfc-gh-tzhang, line 433: We have a constant `MAX_RETRIES` for general HTTP retries, could you make it clear in the name if this is only for Streaming Ingest APIs?

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestUtils.java
```diff
@@ -0,0 +1,114 @@
...
+      throws IOException, IngestResponseException {
+    T response =
+        ServiceResponseHandler.unmarshallStreamingIngestResponse(
+            httpClient.execute(
+                requestBuilder.generateStreamingIngestPostRequest(payload, endpoint, message)),
+            targetClass,
+            apiName);
+
+    // Check for Snowflake specific response code
+    if (statusGetter.apply(response) == RESPONSE_ERR_GENERAL_EXCEPTION_RETRY_REQUEST) {
```
sfc-gh-tzhang, line 91: Could we add log on this?

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestUtils.java
```diff
@@ -0,0 +1,114 @@
...
+      int retries = 0;
+      while (retries < MAX_API_RETRY) {
+        response =
+            ServiceResponseHandler.unmarshallStreamingIngestResponse(
+                httpClient.execute(
+                    requestBuilder.generateStreamingIngestPostRequest(payload, endpoint, message)),
+                targetClass,
+                apiName);
```
sfc-gh-tzhang, line 99: redundant logic as line 83-88, you could use 
do
{
}
while

src/main/java/net/snowflake/ingest/utils/Constants.java
```diff
@@ -24,6 +24,10 @@ public class Constants {
   public static final String PRIVATE_KEY_PASSPHRASE = "private_key_passphrase";
   public static final String JDBC_PRIVATE_KEY = "privateKey";
   public static final long RESPONSE_SUCCESS = 0L; // Don't change, should match server side
+  public static final long RESPONSE_ERR_GENERAL_EXCEPTION_RETRY_REQUEST =
+      10L; // Don't change, should match server side
+  public static final long RESPONSE_ERR_ENQUEUE_TABLE_CHUNK_QUEUE_FULL =
+      7L; // Don't change, should match server side
```
sfc-gh-tzhang, line 30: Could you add a comment at server side indicating that this error code can't be changed, same thing as what I did in https://github.com/snowflakedb/snowflake/commit/478cbfc305f41f7074bcd52aad5dfd62a7fb96e9

src/test/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientTest.java
```diff
@@ -47,13 +57,81 @@
...
+            "key",
+            1234L,
+            OpenChannelRequest.OnErrorOption.CONTINUE,
+            true);
+    channel4 =
+        new SnowflakeStreamingIngestChannelInternal(
+            "channel4",
+            "db",
+            "schemaName",
+            "tableName",
```
sfc-gh-tzhang, line 124: Could you add a test to verify if there're channels belong to different table so we could verify the multiple chunks scenario?

src/main/java/net/snowflake/ingest/streaming/internal/ChannelsStatusResponse.java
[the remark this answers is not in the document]
sfc-gh-mnaides: It overrides `getStatusCode` I just didn't add the `@Override` annotation.  I'll add it

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -405,21 +427,92 @@ void registerBlobs(List<BlobMetadata> blobs) {
...
+                                              channelStatus.getChannelName(),
+                                              channelStatus.getChannelSequencer());
+                                        }
                                       }
                                     })));
+
+    if (!queueFullChunks.isEmpty()) {
+      List<BlobMetadata> retryBlobs = this.getRetryBlobs(queueFullChunks, blobs);
+      if (retryBlobs.isEmpty()) {
+        throw new SFException(ErrorCode.INTERNAL_ERROR, "Failed to retry queue full chunks");
```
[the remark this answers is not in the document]
sfc-gh-mnaides: This is an exception case where we weren't able to match any of the chunks to retry to the initial blob request

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestUtils.java
sfc-gh-mnaides: Two of the methods are because we pass both Map and String payloads, to this function so it was easy to overload the method to handle both inputs.  The third version is because client configure response is not well represented by a pojo in the client so I have a method that can handle getting the status code from a Map.  Note that there is no duplicated logic here, just different signatures for the method that call eachother

src/main/java/net/snowflake/ingest/streaming/internal/RegisterBlobResponse.java
```diff
@@ -8,7 +8,7 @@
 import java.util.List;
 
 /** Response to the RegisterBlobRequest */
-class RegisterBlobResponse {
+class RegisterBlobResponse extends StreamingIngestResponse {
```
sfc-gh-tzhang, line 11: @Override annotation for this one as well?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -405,21 +433,97 @@ void registerBlobs(List<BlobMetadata> blobs) {
...
-                                            channelStatus.getChannelName(),
-                                            channelStatus.getChannelSequencer(),
-                                            channelStatus.getStatusCode());
-                                        channelCache.invalidateChannelIfSequencersMatch(
-                                            chunkStatus.getDBName(),
-                                            chunkStatus.getSchemaName(),
-                                            chunkStatus.getTableName(),
-                                            channelStatus.getChannelName(),
-                                            channelStatus.getChannelSequencer());
+                                        // If the chunk queue is full, we wait and retry the chunks
```
sfc-gh-tzhang, line 436: please update the comment

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestUtils.java
sfc-gh-tzhang: I think we should create a pojo instead of adding a third function here?

sfc-gh-tzhang (approved):
LGTM if you could address the new comments, thanks!

PR #159, opened 2022-04-08
@snow SNOW-572229 Streaning Ingest: Manual padding of compressed chunk for encryption

sfc-gh-azagrebin:
File scan code in XP requires encrypted data block alignment. We change the encryption to use PKCS7Padding for that. This causes unclear problems that this type of encryption is not available in some cases, e.g. when we try to build and push release version:
```
java.security.NoSuchAlgorithmException: Cannot find any provider supporting AES/CTR/PKCS7Padding
```

According to this stackoverflow [thread](https://stackoverflow.com/questions/10193567/java-security-nosuchalgorithmexceptioncannot-find-any-provider-supporting-aes-e), PKCS5Padding is usually used with block size of up to 8 bytes. It should be the same as PKCS7Padding in case of AES with block size of 16 bytes. PKCS5Padding does not cause NoSuchAlgorithmException but the padding does not work as excepted.

A potential solution is to look into `BouncyCastleProvider` that is unclear yet.

Alternatively, this PR implements a manual padding of compressed data before encryption in the client.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
@@ -25,6 +25,7 @@ import java.util.List;
 import java.util.zip.GZIPOutputStream;
 import javax.xml.bind.DatatypeConverter;
 import net.snowflake.ingest.utils.Logging;
+import net.snowflake.ingest.utils.Pair;
 
 /**
  * Build a single blob file that contains file header plus data. The header will be a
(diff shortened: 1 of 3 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -52,10 +53,12 @@
    *
    * @param filePath blob file full path
    * @param chunkData uncompressed chunk data
-   * @return compressed chunk data
+   * @param blockSizeToAlignTo block size to align to for encryption
+   * @return padded compressed chunk data with actual length
```
sfc-gh-tzhang, line 57: could you explain here why we need to return a pair? It may not be clear that the size returned is before padding, but the stream is after padding

sfc-gh-tzhang (approved):
LGTM, thanks Andrey! I will check with the JDBC folks again about the BouncyCastle issue.

PR #161, opened 2022-04-08
NO-SNOW: fix regression after the JMX metrics change

sfc-gh-tzhang:
Fixing a NullPointerException when the metrics collection is enabled, hopefully this is the last issue before we could do a new release :)


sfc-gh-tjones (approved):
Nice! thanks!

PR #162, opened 2022-04-11
SNOW-569611 Make channel close future idempotent

sfc-gh-kkloudas:
SNOW-569611

@sfc-gh-tzhang I also left a question as a comment so that we can discuss it either in the PR or we can chat.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
@@ -52,7 +52,9 @@ class ChannelCache {
   void closeAllChannels() {
     this.cache
         .values()
-        .forEach(channels -> channels.values().forEach(channel -> channel.markClosed()));
+        .forEach(
+            channels ->
+                channels.values().forEach(SnowflakeStreamingIngestChannelInternal::markClosed));
   }
 
   /** Remove a channel in the channel cache if the channel sequencer matches */
@@ -65,7 +67,7 @@ class ChannelCache {
           // We need to compare the channel sequencer in case the old channel was already been
           // removed
           return channelInCache != null
-                  && channelInCache.getChannelSequencer() == channel.getChannelSequencer()
+                  && channelInCache.getChannelSequencer() == channel.getChannelSequencer() // TODO: 4/11/22 this is wrong because the long is boxed so == checks refs and not the value
                   && v.remove(channel.getName()) != null
                   && v.isEmpty()
               ? null
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
@@ -67,7 +67,7 @@ class ChannelCache {
           // We need to compare the channel sequencer in case the old channel was already been
           // removed
           return channelInCache != null
-                  && channelInCache.getChannelSequencer() == channel.getChannelSequencer() // TODO: 4/11/22 this is wrong because the long is boxed so == checks refs and not the value
+                  && channelInCache.getChannelSequencer().equals(channel.getChannelSequencer())
                   && v.remove(channel.getName()) != null
                   && v.isEmpty()
               ? null
--- a/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
@@ -12,6 +12,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 import net.snowflake.ingest.streaming.InsertValidationResponse;
 import net.snowflake.ingest.streaming.OpenChannelRequest;
@@ -45,7 +46,7 @@ class SnowflakeStreamingIngestChannelInternal implements SnowflakeStreamingInges
   private volatile boolean isValid;
 
   // Indicates whether the channel is closed
-  private volatile boolean isClosed;
+  private final AtomicReference<Boolean> isClosed;
 
   // Reference to the client that owns this channel
   private final SnowflakeStreamingIngestClientInternal owningClient;
@@ -64,6 +65,8 @@ class SnowflakeStreamingIngestChannelInternal implements SnowflakeStreamingInges
 
   // ON_ERROR option for this channel
   private final OpenChannelRequest.OnErrorOption onErrorOption;
+
+  private final CompletableFuture<Void> terminationFuture;
 
   /**
    * Constructor for TESTING ONLY which allows us to set the test mode
@@ -99,7 +102,7 @@ class SnowflakeStreamingIngestChannelInternal implements SnowflakeStreamingInges
     this.channelSequencer = channelSequencer;
     this.rowSequencer = new AtomicLong(rowSequencer);
     this.isValid = true;
-    this.isClosed = false;
+    this.isClosed = new AtomicReference<>(false);
     this.owningClient = client;
     this.isTestMode = isTestMode;
     this.allocator =
@@ -112,6 +115,8 @@ class SnowflakeStreamingIngestChannelInternal implements SnowflakeStreamingInges
     this.encryptionKey = encryptionKey;
     this.encryptionKeyId = encryptionKeyId;
     this.onErrorOption = onErrorOption;
+    this.terminationFuture = new CompletableFuture<>();
+
     logger.logDebug("Channel={} created for table={}", this.channelName, this.tableName);
   }
 
@@ -255,20 +260,31 @@ class SnowflakeStreamingIngestChannelInternal implements SnowflakeStreamingInges
         rowSequencer);
   }
 
-  /** @return a boolean to indicate whether the channel is closed or not */
+  /**
+   * @return {@code true} if the channel was closed, {@code false} otherwise. A return value of
+   *     {@code true} does not mean that the data is committed. For this, you should call {@link
+   *     #close()} and wait on the returned future.
+   */
   @Override
   public boolean isClosed() {
-    return this.isClosed;
+    return this.isClosed.get();
   }
 
-  /** Mark the channel as closed */
-  void markClosed() {
-    this.isClosed = true;
-    logger.logDebug(
-        "Channel is closed, name={}, channel sequencer={}, row sequencer={}",
-        getFullyQualifiedName(),
-        channelSequencer,
-        rowSequencer);
+  /**
+   * Mark the channel as closed atomically.
+   *
+   * @return {@code true} if the channel was already closed, {@code false} otherwise.
+   */
+  boolean markClosed() {
+    final boolean wasAlreadyClosed = this.isClosed.getAndSet(true);
+    if (!wasAlreadyClosed) {
+      logger.logDebug(
+          "Channel is closed, name={}, channel sequencer={}, row sequencer={}",
+          getFullyQualifiedName(),
+          channelSequencer,
+          rowSequencer);
+    }
+    return wasAlreadyClosed;
   }
 
   /**
@@ -281,7 +297,9 @@ class SnowflakeStreamingIngestChannelInternal implements SnowflakeStreamingInges
     // Skip this check for closing because we need to set the channel to closed first and then flush
     // in case there is any leftover rows
     if (isClosed() && !closing) {
-      throw new SFException(ErrorCode.CLOSED_CHANNEL);
+      final CompletableFuture<Void> res = new CompletableFuture<>();
+      res.completeExceptionally(new SFException(ErrorCode.CLOSED_CHANNEL));
+      return res;
     }
 
     // Simply return if there is no data in the channel, this might not work if we support public
@@ -294,38 +312,51 @@ class SnowflakeStreamingIngestChannelInternal implements SnowflakeStreamingInges
   }
 
   /**
-   * Close the channel (this will flush in-flight buffered data)
+   * Close the channel and flush in-flight buffered data.
    *
-   * @return future which will be complete when the channel is closed
+   * @return a {@link CompletableFuture} which will be completed when the channel is closed and the
+   *     data is successfully committed or has failed to be committed for some reason.
    */
   @Override
   public CompletableFuture<Void> close() {
     checkValidation();
 
-    if (isClosed()) {
-      return CompletableFuture.completedFuture(null);
+    if (!markClosed()) {
+      this.owningClient.removeChannelIfSequencersMatch(this);
+
+      // here we flush even if removal fails because, for some reason,
+      // the client sequencers did not match.
+      // This does not seem like the desirable behavior.
+
+      flush(true)
+          .thenRunAsync(
+              () -> {
+                List<SnowflakeStreamingIngestChannelInternal> uncommittedChannels =
+                    this.owningClient.verifyChannelsAreFullyCommitted(
+                        Collections.singletonList(this));
+
+                this.arrowBuffer.close();
+
+                // Throw an exception if the channel has any uncommitted rows
+                if (!uncommittedChannels.isEmpty()) {
+                  throw new SFException(
+                      ErrorCode.CHANNEL_WITH_UNCOMMITTED_ROWS,
+                      uncommittedChannels.stream()
+                          .map(SnowflakeStreamingIngestChannelInternal::getFullyQualifiedName)
+                          .collect(Collectors.toList()));
+                }
+              })
+          .handle(
+              (ignored, t) -> {
+                if (t != null) {
+                  terminationFuture.completeExceptionally(t);
+                } else {
+                  terminationFuture.complete(null);
+                }
+                return null;
+              });
     }
-
-    markClosed();
-    this.owningClient.removeChannelIfSequencersMatch(this);
-    return flush(true)
-        .thenRunAsync(
-            () -> {
-              List<SnowflakeStreamingIngestChannelInternal> uncommittedChannels =
-                  this.owningClient.verifyChannelsAreFullyCommitted(
-                      Collections.singletonList(this));
-
-              this.arrowBuffer.close();
-
-              // Throw an exception if the channel has any uncommitted rows
-              if (!uncommittedChannels.isEmpty()) {
-                throw new SFException(
-                    ErrorCode.CHANNEL_WITH_UNCOMMITTED_ROWS,
-                    uncommittedChannels.stream()
-                        .map(SnowflakeStreamingIngestChannelInternal::getFullyQualifiedName)
-                        .collect(Collectors.toList()));
-              }
-            });
+    return terminationFuture;
   }
 
   /**
--- a/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
@@ -528,7 +528,7 @@ public class SnowflakeStreamingIngestClientInternal implements SnowflakeStreamin
 
   /** Close the client, which will flush first and then release all the resources */
   @Override
-  public void close() throws Exception {
+  public synchronized void close() throws Exception {
     if (isClosed) {
       return;
     }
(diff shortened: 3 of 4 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
```diff
@@ -65,7 +65,7 @@ void removeChannelIfSequencersMatch(SnowflakeStreamingIngestChannelInternal chan
           // We need to compare the channel sequencer in case the old channel was already been
           // removed
           return channelInCache != null
-                  && channelInCache.getChannelSequencer() == channel.getChannelSequencer()
+                  && channelInCache.getChannelSequencer().equals(channel.getChannelSequencer())
```
sfc-gh-tjones, line 68: Why this change out of curiosity? These should be `long`

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
```diff
@@ -65,6 +66,8 @@ class SnowflakeStreamingIngestChannelInternal implements SnowflakeStreamingInges
   // ON_ERROR option for this channel
   private final OpenChannelRequest.OnErrorOption onErrorOption;
 
+  private final CompletableFuture<Void> terminationFuture;
```
sfc-gh-tjones, line 69: Nit, add one-liner explaining what this is.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
```diff
@@ -112,6 +115,9 @@ class SnowflakeStreamingIngestChannelInternal implements SnowflakeStreamingInges
     this.encryptionKey = encryptionKey;
     this.encryptionKeyId = encryptionKeyId;
     this.onErrorOption = onErrorOption;
+
```
sfc-gh-tjones, line 118: Nit, remove space.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
```diff
@@ -258,17 +264,23 @@ void invalidate() {
...
-    logger.logDebug(
-        "Channel is closed, name={}, channel sequencer={}, row sequencer={}",
-        getFullyQualifiedName(),
-        channelSequencer,
-        rowSequencer);
+  /**
+   * Mark the channel as closed atomically.
+   * @return {@code true} if the channel was already closed, {@code false} otherwise.
+   */
+  boolean markClosed() {
```
sfc-gh-tjones, line 274: Alternatively, could this be a `synchronized` method?

sfc-gh-tjones (commented):
This looks good to me, let's have Toby or Matt give their approval before landing.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
```diff
@@ -302,30 +316,39 @@ CompletableFuture<Void> flush(boolean closing) {
...
-                throw new SFException(
-                    ErrorCode.CHANNEL_WITH_UNCOMMITTED_ROWS,
-                    uncommittedChannels.stream()
-                        .map(SnowflakeStreamingIngestChannelInternal::getFullyQualifiedName)
-                        .collect(Collectors.toList()));
-              }
-            });
+      // here we flush even if removal fails because, for some reason,
+      // the client sequencers did not match.
+      // This does not seem like the desirable behavior.
```
sfc-gh-tzhang, line 324: I don't think I understand what you mean here, could you elaborate? But I notice a potential wrong result issue from your question.... please take a look if you're interested. PR: https://github.com/snowflakedb/snowflake-ingest-java/pull/163/files

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
```diff
@@ -45,7 +46,7 @@ class SnowflakeStreamingIngestChannelInternal implements SnowflakeStreamingInges
   private volatile boolean isValid;
 
   // Indicates whether the channel is closed
-  private volatile boolean isClosed;
+  private final AtomicReference<Boolean> isClosed;
```
sfc-gh-tzhang, line 49: what's the difference between AtomicReference and volatile in this case?

sfc-gh-tzhang (commented):
Thanks for the change, two questions:
1. Could you update SnowflakeStreamingIngestClientInternal.close as well?
2. I think your change makes close() idempotent, but it's still possible that others call `isClosed()` before the `close()` finishes, right? So the same issue would still be possible.

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
sfc-gh-kkloudas: Yes and the reason is that `==` on objects checks if the object is the same while `equals()` (if correctly implemented) checks if the value is the same, irrespective of the object that caries it. So this could lead to bugs.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
sfc-gh-kkloudas: @sfc-gh-tzhang The `removeChannelIfSequencersMatch()` removes the channel from the cache if 1) it was there in the first place and 2) if the client seq matches. If all those conditions are satisfied, then it makes sense to call `flush()`, upload and register the created blob. If any of these conditions does not match, and the removal fails (e.g. the client seq is old), then should we flush and upload? or should we just invalidate the channel and move on? 

I am not sure if this scenario is realistic but I just wanted to point this out.

sfc-gh-kkloudas:
@sfc-gh-tzhang For 1) what do we want the semantics of `SnowflakeStreamingIngestClientInternal.close()` to be? I guess we want for it to wait for all channels to have closed, right?
For 2) 

> Thanks for the change, two questions:
> 
> 1. Could you update SnowflakeStreamingIngestClientInternal.close as well?
> 2. I think your change makes close() idempotent, but it's still possible that others call `isClosed()` before the `close()` finishes, right? So the same issue would still be possible.

Could you please clarify what do you mean by the second point? If someone calls `close()` then everyone else will see the channel as `closed`. The `isClosed()` does not perform any action. It simply reads the value.

sfc-gh-kkloudas:
Actually I think that for the `client.close()` I think that we should chat because I have some questions that I need to clarify before figuring out a solution. For example, we do not seem to close `this.arrowBuffer.close();` for each channel and the javadoc of the method says that it releases resources. Do we risk of memory leaks given that in there we also call `Utils.closeAllocator(this.allocator);`?

Also we do not do any `verifyChannelsAreFullyCommitted()` or remove the channel from the cache. Are these things something we should add?

Finally, should we have for each channel a `close-and-flush()` method which is the `close()` now, and a `close-and-do-nothing` which will be called by all the close methods and gives us the `ChannelData` to put in the blob?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
sfc-gh-tzhang: Yeah, it makes sense, the validation check is actually in `checkValidation()` at the top of this function. If the channel is not in the channel cache, it must be invalid, you could add a logWarn to verify this logic

sfc-gh-tzhang:
> @sfc-gh-tzhang
> 
> > Thanks for the change, two questions:
> > 
> > 1. Could you update SnowflakeStreamingIngestClientInternal.close as well?
> 
> The reason I did not change this is because I do not know what do we want the semantics of `SnowflakeStreamingIngestClientInternal.close()` to be? Judging by the signature, we want it to wait for all channels to have closed before exiting, right? I will implement that and we can discuss later if we want something else.
> 
> > 2. I think your change makes close() idempotent, but it's still possible that others call `isClosed()` before the `close()` finishes, right? So the same issue would still be possible.
> 
> Could you please clarify what do you mean by the second point? If someone calls `close()` then everyone else will see the channel as `closed`. The `isClosed()` does not perform any action. It simply reads the value.

I think we could just make close() synchronized, so any subsequent call to it will wait for the first one to finish. I believed the issue you mentioned to me the other day was that with the current implementation, the first call to `channel.close() ` will mark the channel as closed and flush, then subsequent call to `channel.close()` or `channel.isClose()` will simply return and it's possible that customer does something with it before the first call finish, and that's why we want to make this change, does that sound right?

sfc-gh-kkloudas:
> > @sfc-gh-tzhang
> > > Thanks for the change, two questions:
> > > 
> > > 1. Could you update SnowflakeStreamingIngestClientInternal.close as well?
> > 
> > 
> > The reason I did not change this is because I do not know what do we want the semantics of `SnowflakeStreamingIngestClientInternal.close()` to be? Judging by the signature, we want it to wait for all channels to have closed before exiting, right? I will implement that and we can discuss later if we want something else.
> > > 2. I think your change makes close() idempotent, but it's still possible that others call `isClosed()` before the `close()` finishes, right? So the same issue would still be possible.
> > 
> > 
> > Could you please clarify what do you mean by the second point? If someone calls `close()` then everyone else will see the channel as `closed`. The `isClosed()` does not perform any action. It simply reads the value.
> 
> I think we could just make close() synchronized, so any subsequent call to it will wait for the first one to finish. I believed the issue you mentioned to me the other day was that with the current implementation, the first call to `channel.close() ` will mark the channel as closed and flush, then subsequent call to `channel.close()` or `channel.isClose()` will simply return and it's possible that customer does something with it before the first call finish, and that's why we want to make this change, does that sound right?

You are right that the problem was that the first call to `channel.close() ` will mark the channel as closed and flush, then subsequent call to `channel.close()` will simply return and it's possible that customer does something with it before the first call finish. The `isClosed()` will return always `true` during the process of closing because we first set the state to `closed` and then start. But now, every `channel.close()` will wait for the first one to finish. 

Did you have a chance to see my other questions about the `client.close()`?

sfc-gh-tzhang:
> Actually I think that for the `client.close()` I think that we should chat because I have some questions that I need to clarify before figuring out a solution. For example, we do not seem to close `this.arrowBuffer.close();` for each channel and the javadoc of the method says that it releases resources. Do we risk of memory leaks given that in there we also call `Utils.closeAllocator(this.allocator);`?
> 
> Also we do not do any `verifyChannelsAreFullyCommitted()` or remove the channel from the cache. Are these things something we should add?
> 
> Finally, should we have for each channel a `close-and-flush()` method which is the `close()` now, and a `close-and-do-nothing` which will be called by all the close methods and gives us the `ChannelData` to put in the blob?

Allocator.close should be sufficient since we use it the parent allocator when creating the arrow buffers. We were actually calling `verifyChannelsAreFullyCommitted()` on all the channels before, but I removed it as part of https://github.com/snowflakedb/snowflake-ingest-java/pull/117, and the reason is that we ran into issues like a channel is still in the cache and being inactive for a long time, it will be cleaned up at server side and causes false alarm. Also I feel like checking all the channels in cache during close doesn't make sense, so the recommendation is to always use `channel.close`. I was actual thinking about the same thing like you suggested, like support multiple close function with different functionalities, but got pushed back from folks that let's keep the APIs simple for now. I think it makes sense to add them in longer term (maybe PuPr or GA), based on customer's feedback.

sfc-gh-tzhang:
> > > @sfc-gh-tzhang
> > > > Thanks for the change, two questions:
> > > > 
> > > > 1. Could you update SnowflakeStreamingIngestClientInternal.close as well?
> > > 
> > > 
> > > The reason I did not change this is because I do not know what do we want the semantics of `SnowflakeStreamingIngestClientInternal.close()` to be? Judging by the signature, we want it to wait for all channels to have closed before exiting, right? I will implement that and we can discuss later if we want something else.
> > > > 2. I think your change makes close() idempotent, but it's still possible that others call `isClosed()` before the `close()` finishes, right? So the same issue would still be possible.
> > > 
> > > 
> > > Could you please clarify what do you mean by the second point? If someone calls `close()` then everyone else will see the channel as `closed`. The `isClosed()` does not perform any action. It simply reads the value.
> > 
> > 
> > I think we could just make close() synchronized, so any subsequent call to it will wait for the first one to finish. I believed the issue you mentioned to me the other day was that with the current implementation, the first call to `channel.close() ` will mark the channel as closed and flush, then subsequent call to `channel.close()` or `channel.isClose()` will simply return and it's possible that customer does something with it before the first call finish, and that's why we want to make this change, does that sound right?
> 
> You are right that the problem was that the first call to `channel.close() ` will mark the channel as closed and flush, then subsequent call to `channel.close()` will simply return and it's possible that customer does something with it before the first call finish. The `isClosed()` will return always `true` during the process of closing because we first set the state to `closed` and then start. But now, every `channel.close()` will wait for the first one to finish.
> 
> Did you have a chance to see my other questions about the `client.close()`?

Correct, then it seems like a problem that subsequent close() and isClose() behaves differently? I think we want to keep the consistent, right?

sfc-gh-kkloudas:
Oh now I see your point. 

I think that the best solution to this is to not expose the `isClosed()` method at all and either make the `close()` blocking (which I do not really like), or leave it as it (asynchronous) and only let the user check if the channel is closed by checking if the returned future is completed. From the two, I like more the solution of having the `close()` asynchronous (as is). 

If we still want to expose the `isClosed()` we can always make it:

```
public boolean isClosed() {
    return this.terminationFuture.isDone();
  }
```

I did this in the PR btw but I still believe that we should not expose the `isClosed()` to avoid such confusion. I think that it is better to give users only one way to do things. It saves us some trouble later on.

sfc-gh-kkloudas:
> > Actually I think that for the `client.close()` I think that we should chat because I have some questions that I need to clarify before figuring out a solution. For example, we do not seem to close `this.arrowBuffer.close();` for each channel and the javadoc of the method says that it releases resources. Do we risk of memory leaks given that in there we also call `Utils.closeAllocator(this.allocator);`?
> > Also we do not do any `verifyChannelsAreFullyCommitted()` or remove the channel from the cache. Are these things something we should add?
> > Finally, should we have for each channel a `close-and-flush()` method which is the `close()` now, and a `close-and-do-nothing` which will be called by all the close methods and gives us the `ChannelData` to put in the blob?
> 
> Allocator.close should be sufficient since we use it the parent allocator when creating the arrow buffers. We were actually calling `verifyChannelsAreFullyCommitted()` on all the channels before, but I removed it as part of #117, and the reason is that we ran into issues like a channel is still in the cache and being inactive for a long time, it will be cleaned up at server side and causes false alarm. Also I feel like checking all the channels in cache during close doesn't make sense, so the recommendation is to always use `channel.close`. I was actual thinking about the same thing like you suggested -- support multiple close function with different functionalities, but got pushed back from folks that let's keep the APIs simple for now. I think it makes sense to add them in longer term (maybe PuPr or GA), based on customer's feedback.

I am not suggesting to expose those "close" methods to the user. The only method the user will see is the "close-and-flush". The others will be internal. I have to think a bit about this and come back to you about the client.close().

sfc-gh-tzhang:
> Oh now I see your point.
> 
> I think that the best solution to this is to not expose the `isClosed()` method at all and either make the `close()` blocking (which I do not really like), or leave it as it (asynchronous) and only let the user check if the channel is closed by checking if the returned future is completed. From the two, I like more the solution of having the `close()` asynchronous (as is).
> 
> If we still want to expose the `isClosed()` we can always make it:
> 
> ```
> public boolean isClosed() {
>     return this.terminationFuture.isDone();
>   }
> ```
> 
> I did this in the PR btw but I still believe that we should not expose the `isClosed()` to avoid such confusion. I think that it is better to give users only one way to do things. It saves us some trouble later on.

Makes sense to me, let's do that! @sfc-gh-japatel Are you using the `isClosed()` function in any KC logic?

sfc-gh-tzhang:
> > > Actually I think that for the `client.close()` I think that we should chat because I have some questions that I need to clarify before figuring out a solution. For example, we do not seem to close `this.arrowBuffer.close();` for each channel and the javadoc of the method says that it releases resources. Do we risk of memory leaks given that in there we also call `Utils.closeAllocator(this.allocator);`?
> > > Also we do not do any `verifyChannelsAreFullyCommitted()` or remove the channel from the cache. Are these things something we should add?
> > > Finally, should we have for each channel a `close-and-flush()` method which is the `close()` now, and a `close-and-do-nothing` which will be called by all the close methods and gives us the `ChannelData` to put in the blob?
> > 
> > 
> > Allocator.close should be sufficient since we use it the parent allocator when creating the arrow buffers. We were actually calling `verifyChannelsAreFullyCommitted()` on all the channels before, but I removed it as part of #117, and the reason is that we ran into issues like a channel is still in the cache and being inactive for a long time, it will be cleaned up at server side and causes false alarm. Also I feel like checking all the channels in cache during close doesn't make sense, so the recommendation is to always use `channel.close`. I was actual thinking about the same thing like you suggested -- support multiple close function with different functionalities, but got pushed back from folks that let's keep the APIs simple for now. I think it makes sense to add them in longer term (maybe PuPr or GA), based on customer's feedback.
> 
> I am not suggesting to expose those "close" methods to the user. The only method the user will see is the "close-and-flush". The others will be internal. I have to think a bit about this and come back to you about the client.close().

I might miss something, but if we make the function `synchronized` then we should be good?

sfc-gh-japatel:
> > Oh now I see your point.
> > I think that the best solution to this is to not expose the `isClosed()` method at all and either make the `close()` blocking (which I do not really like), or leave it as it (asynchronous) and only let the user check if the channel is closed by checking if the returned future is completed. From the two, I like more the solution of having the `close()` asynchronous (as is).
> > If we still want to expose the `isClosed()` we can always make it:
> > ```
> > public boolean isClosed() {
> >     return this.terminationFuture.isDone();
> >   }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I did this in the PR btw but I still believe that we should not expose the `isClosed()` to avoid such confusion. I think that it is better to give users only one way to do things. It saves us some trouble later on.
> 
> Makes sense to me, let's do that! @sfc-gh-japatel Are you using the `isClosed()` function in any KC logic?

I am using `channel.isClosed()` to check before reopening a channel.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
```diff
@@ -258,17 +264,23 @@ void invalidate() {
   /** @return a boolean to indicate whether the channel is closed or not */
   @Override
   public boolean isClosed() {
-    return this.isClosed;
+    return this.terminationFuture.isDone();
```
sfc-gh-tzhang, line 267: Looks like KC is using it, I'm fine with both approach depends on whether KC has a workaround

sfc-gh-tzhang (dismissed):
LGTM, please make sure you pass all the merge gate, as well as `StreamingIngestIT.java`

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
sfc-gh-kkloudas: I see, then I can leave it given that people are using it and also now it is safer :)

sfc-gh-kkloudas:
For the tests now, 

> LGTM, please make sure you pass all the merge gate, as well as `StreamingIngestIT.java`

Now that we changed the semantics of the `isClosed()` to only return `true` when the `channel.close()` succeeds, the tests fail because the `markClosed()` simply changes the state of the channel to closed, but does not complete the termination future or anything. I am wondering what should be now the semantics.

sfc-gh-tzhang:
> For the tests now,
> 
> > LGTM, please make sure you pass all the merge gate, as well as `StreamingIngestIT.java`
> 
> Now that we changed the semantics of the `isClosed()` to only return `true` when the `channel.close()` succeeds, the tests fail because the `markClosed()` simply changes the state of the channel to closed, but does not complete the termination future or anything. I am wondering what should be now the semantics.

You could modify the test to call .close() first, it shouldn't be an issue 

> For the tests now,
> 
> > LGTM, please make sure you pass all the merge gate, as well as `StreamingIngestIT.java`
> 
> Now that we changed the semantics of the `isClosed()` to only return `true` when the `channel.close()` succeeds, the tests fail because the `markClosed()` simply changes the state of the channel to closed, but does not complete the termination future or anything. I am wondering what should be now the semantics.

Ah, yes! i think we have a bigger issue (wrong result)! We need isClose to return true right after markClose is called, otherwise we may still insert rows into the buffer even after we mark the channel as closed and start flushing, so we need to change the isClose back. I don't have a good idea now, we could just merge your change to fix part of the idempotent issue, or we could just keep my old logic, both options work for me.

PR #163, opened 2022-04-12
NO_SNOW: fix bug in channel.close 

sfc-gh-tzhang:
1. Fix a bug in channel.close which may cause wrong result issue, the problem is that we remove the channel from the channel cache before calling flush, so it won't flush the buffers in this channel even if it has data in them
2. Add a small check to make sure row count shouldn't go beyond MAX INT value
3. Improve channel related error messages
4. Update streaming ingest IT tests to close the channel which will verify everything should be committed

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
@@ -221,18 +221,17 @@ class ArrowRowBuffer {
    * synchronization
    */
   void close() {
-    // If the channel is invalid, cleanup all the leftover data; otherwise, simply close the
-    // allocator which will verify that everything should be cleaned up already
-    this.fields.clear();
-    if (!this.owningChannel.isValid()) {
-      if (this.vectorsRoot != null) {
-        this.vectorsRoot.close();
-        this.tempVectorsRoot.close();
-      }
-      Utils.closeAllocator(this.allocator);
-    } else {
-      this.allocator.close();
+    // If the channel is valid but still has leftover data, log it
+    long allocated = this.allocator.getAllocatedMemory();
+    if (allocated > 0 && this.owningChannel.isValid()) {
+      logger.logWarn(" Memory leaked={} by allocator={}", allocated, this.allocator.toString());
     }
+    if (this.vectorsRoot != null) {
+      this.vectorsRoot.close();
+      this.tempVectorsRoot.close();
+    }
+    this.fields.clear();
+    Utils.closeAllocator(this.allocator);
   }
 
   /** Reset the variables after each flush. Note that the caller needs to handle synchronization */
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
@@ -286,6 +286,9 @@ class ArrowRowBuffer {
                     row, new SFException(e, ErrorCode.INTERNAL_ERROR, e.getMessage()), rowIndex));
           }
           rowIndex++;
+          if (this.rowCount == Integer.MAX_VALUE) {
+            throw new SFException(ErrorCode.INTERNAL_ERROR, "Row count reaches MAX value");
+          }
         }
       } else {
         // If the on_error option is ABORT, simply throw the first exception
@@ -308,6 +311,9 @@ class ArrowRowBuffer {
           }
         }
         rowSize = tempRowSize;
+        if ((long) this.rowCount + tempRowCount > Integer.MAX_VALUE) {
+          throw new SFException(ErrorCode.INTERNAL_ERROR, "Row count reaches MAX value");
+        }
         this.rowCount += tempRowCount;
         this.bufferSize += rowSize;
         this.statsMap.forEach(
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
@@ -221,17 +221,21 @@ class ArrowRowBuffer {
    * synchronization
    */
   void close() {
-    // If the channel is valid but still has leftover data, log it
     long allocated = this.allocator.getAllocatedMemory();
-    if (allocated > 0 && this.owningChannel.isValid()) {
-      logger.logWarn(" Memory leaked={} by allocator={}", allocated, this.allocator.toString());
-    }
     if (this.vectorsRoot != null) {
       this.vectorsRoot.close();
       this.tempVectorsRoot.close();
     }
     this.fields.clear();
     Utils.closeAllocator(this.allocator);
+
+    // If the channel is valid but still has leftover data, throw an exception because it should be
+    // cleaned up already before calling close
+    if (allocated > 0 && this.owningChannel.isValid()) {
+      throw new SFException(
+          ErrorCode.INTERNAL_ERROR,
+          String.format("Memory leaked=%d by allocator=%s", allocated, this.allocator.toString()));
+    }
   }
 
   /** Reset the variables after each flush. Note that the caller needs to handle synchronization */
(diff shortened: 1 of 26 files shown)


src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
```diff
@@ -798,6 +798,28 @@ public void testAbortOnErrorOption() throws Exception {
...
+    // Open a streaming ingest channel from the given client
+    SnowflakeStreamingIngestChannel channel1 = client.openChannel(request1);
+    for (int val = 0; val < 1000; val++) {
+      Map<String, Object> row = new HashMap<>();
+      row.put("c1", Integer.toString(val));
+      verifyInsertValidationResponse(channel1.insertRow(row, Integer.toString(val)));
+    }
+
+    // Close the channel to make sure everything is committed
+    channel1.close().get();
```
sfc-gh-mnaides, line 820: What's the test case here?  Are we just looking to see if it errors?

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
```diff
@@ -221,12 +221,18 @@ void setupSchema(List<ColumnMetadata> columns) {
    * synchronization
    */
   void close() {
-    if (this.vectorsRoot != null) {
```
sfc-gh-mnaides, line 224: Can you say a little more about what's happening here?  When would we want to clear the fields objects but not the  vectors and allocators?  If the worry is that we close the buffer before flushing it, should we put checks for that case specifically instead of having buffers in a weird state where they have open allocators but near to be reinitialized because their fields are empty?

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
sfc-gh-tzhang: Correct, channel.close internally will check whether everything has committed, and throw an exception if not

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
sfc-gh-tzhang: the fields object is just a map and we will clean it all the time. For vectors and allocators, we want to clean them up if the channel is invalid because we will skip flushing it so the buffer may have leftover data. If there is no issue, there will always be a flush before calling this close, so the vectors and allocators should already be empty, so we want to call allocator.close directly which will verify if there any memory leak or not, does that make more sense now?

sfc-gh-tzhang:
ping, we should probably include this fix in the PrPr release

sfc-gh-tjones (approved):

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
sfc-gh-mnaides: I think I understand, but I don't see why you'd want to call `Utils.closeAllocator()` in one case and `this.allocator.close()` in another case.  Wouldn't you want to close child allocators in both cases?

sfc-gh-tzhang: There is no child allocator for in this case, `Utils.closeAllocator()` is used in other places which we do have child allocators if you check the usage

sfc-gh-mnaides: Why wouldn't we have child allocators in this case?  Also the `Utils.closeAllocator()` handles the case of having no child allocators.  Are you trying to avoid calling `alloc.releaseBytes(alloc.getAllocatedMemory());`?

sfc-gh-tzhang: > Why wouldn't we have child allocators in this case?

Because we don't create it, please take a look at how the child allocators are created in `SnowflakeStreamingIngestChannelInternal` and passed down to `ArrowRowBuffer`

> Are you trying to avoid calling alloc.releaseBytes(alloc.getAllocatedMemory());?

Yes, that's another reason, because if we release it first, we won't be able to verify that there is no memory leak

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
```diff
@@ -221,12 +221,18 @@ void setupSchema(List<ColumnMetadata> columns) {
...
-      this.vectorsRoot.close();
-      this.tempVectorsRoot.close();
-    }
+    // If the channel is invalid, cleanup all the leftover data; otherwise, simply close the
+    // allocator which will verify that everything should be cleaned up already
     this.fields.clear();
-    Utils.closeAllocator(this.allocator);
+    if (!this.owningChannel.isValid()) {
+      if (this.vectorsRoot != null) {
+        this.vectorsRoot.close();
```
sfc-gh-kkloudas, line 229: I do not know the details about what these `vectors.close()` do, but if it is needed, shouldn't we also call them during normal termination? Here we call them when `!owningChannel.isValid()` but what happens when the channel is valid?

sfc-gh-gdoci:
> ping, we should probably include this fix in the PrPr release

I agree with this! Recently I added `channel.close().get()` to wait for all rows to be committed before doing the verification steps in the correctness tests and that call failed with 
```
One or more channels [TEST_DB.TEST_SCHEMA.TEST_TABLE_0.CHANNEL_0] might contain uncommitted rows due to server side errors.
please consider reopening the channels to replay the data loading by using the latest persistent offset token.
```
This PR fixes that issue.

sfc-gh-gdoci:
> > ping, we should probably include this fix in the PrPr release
> 
> I agree with this! Recently I added `channel.close().get()` to wait for all rows to be committed (previously just `while(channel.lastCommittedOffsetToken !=...){sleep}`) before doing the verification steps in the correctness tests and that call failed with
> 
> ```
> One or more channels [TEST_DB.TEST_SCHEMA.TEST_TABLE_0.CHANNEL_0] might contain uncommitted rows due to server side errors.
> please consider reopening the channels to replay the data loading by using the latest persistent offset token.
> ```
> 
> This PR fixes that issue with `channel.close`.
> 
> Update: it fails sporadically, rarely than before, but still the issue is not solved by this PR.

Update2: I think my issue has to do with `Constants.COMMIT_MAX_RETRY_COUNT=10` being low. It seems that increasing it to e.g. `100` helps. Might be worth increasing it as part of this PR as well.

sfc-gh-mnaides (approved):

sfc-gh-tzhang:
> > > ping, we should probably include this fix in the PrPr release
> > 
> > 
> > I agree with this! Recently I added `channel.close().get()` to wait for all rows to be committed (previously just `while(channel.lastCommittedOffsetToken !=...){sleep}`) before doing the verification steps in the correctness tests and that call failed with
> > ```
> > One or more channels [TEST_DB.TEST_SCHEMA.TEST_TABLE_0.CHANNEL_0] might contain uncommitted rows due to server side errors.
> > please consider reopening the channels to replay the data loading by using the latest persistent offset token.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This PR fixes that issue with `channel.close`.
> > Update: it fails sporadically, rarely than before, but still the issue is not solved by this PR.
> 
> Update2: I think my issue has to do with `Constants.COMMIT_MAX_RETRY_COUNT=10` being low. It seems that increasing it to e.g. `100` helps. Might be worth increasing it as part of this PR as well.

Hmmm, I think the issue is something else, we will need to debug it together (maybe Tuesday morning since Monday morning is full now :)). But if you see the logic in `verifyChannelsAreFullyCommitted`, we are NOT blindly wait for 5sec, the retry counter is reset as long as we have some progress.

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
sfc-gh-tzhang: We don't really need them because the vectors should be cleaned up already doing normal termination (we always do a flush first), but I guess the logic looks confusing since @sfc-gh-mnaides had a similar question, I update the code to always do the cleanup, but checking the allocated memory should be already 0 before cleaning up for nomination :)

PR #164, opened 2022-04-15
BUG_FIX Consume the response entity to free up connection

sfc-gh-japatel:
- As mentioned in httpclient docs
- https://hc.apache.org/httpcomponents-client-4.5.x/quickstart.html
- https://stackoverflow.com/questions/4775618/httpclient-4-0-1-how-to-release-connection
```
Please note that if response content is not fully consumed the underlying
connection cannot be safely re-used and will be shut down and discarded
by the connection manager.
```

I am also seeing few NoHttpResponseException in existing snowpipe + KC where load is increased. This could be one of the potential reasons, although I don't have a thread dump. 

- no test
- I will push this to master and release a new version on top of `0.10.3`
- I will also send out pr for doing this in other APIs which are not GA

Steps:
1. Push this on top of 0.10.3. 
2. CP this commit on top of master.
3. Do this for rest of APIs called.

### DIFF
--- a/src/main/java/net/snowflake/ingest/connection/ServiceResponseHandler.java
+++ b/src/main/java/net/snowflake/ingest/connection/ServiceResponseHandler.java
@@ -8,6 +8,7 @@ import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.io.IOException;
 import net.snowflake.ingest.utils.BackOffException;
+import org.apache.http.HttpEntity;
 import org.apache.http.HttpResponse;
 import org.apache.http.HttpStatus;
 import org.apache.http.StatusLine;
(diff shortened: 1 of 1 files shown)


src/main/java/net/snowflake/ingest/connection/ServiceResponseHandler.java
```diff
@@ -76,7 +77,12 @@ public static IngestResponse unmarshallIngestResponse(HttpResponse response)
     }
 
     // grab the response entity
-    String blob = EntityUtils.toString(response.getEntity());
+    HttpEntity responseEntity = response.getEntity();
```
sfc-gh-tzhang, line 80: Looks like all the logic are duplicate, can we do better?

sfc-gh-tzhang: Also, can we add this to all APIs?

sfc-gh-tzhang (commented):
Left some comments, is this issue customer reported?

src/main/java/net/snowflake/ingest/connection/ServiceResponseHandler.java
sfc-gh-japatel: I dont think there is any benefit of doing it in a separate method. 

I am doing this only for 3 snowpipe APIs for which confluent reported concerns. Hence I am pushing this only on top of `0.10.3`

I also plan to do it for all APIs though

sfc-gh-tzhang: ok, given that this will be for all APIs, I think it makes sense to add a helper function to avoid duplication?

sfc-gh-tzhang (approved):

src/main/java/net/snowflake/ingest/connection/ServiceResponseHandler.java
sfc-gh-japatel: Sounds good!

PR #166, opened 2022-04-19
Replacing HttpClient with CloseableHttpClient to close the resource 

sfc-gh-japatel:
This is on top of https://github.com/snowflakedb/snowflake-ingest-java/pull/164

(Only check 8d2c72d5566e41aaf3f681a37d57722305436e06)

Here is what my release branch looks like: 
https://github.com/snowflakedb/snowflake-ingest-java/compare/v0.10.3..v0.10.6-release-branch

Branch: https://github.com/snowflakedb/snowflake-ingest-java/tree/v0.10.6-release-branch

### DIFF
--- a/src/main/java/net/snowflake/ingest/SimpleIngestManager.java
+++ b/src/main/java/net/snowflake/ingest/SimpleIngestManager.java
@@ -31,10 +31,10 @@ import net.snowflake.ingest.connection.ServiceResponseHandler;
 import net.snowflake.ingest.utils.BackOffException;
 import net.snowflake.ingest.utils.HttpUtil;
 import net.snowflake.ingest.utils.StagedFileWrapper;
-import org.apache.http.HttpResponse;
-import org.apache.http.client.HttpClient;
+import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.methods.HttpPost;
+import org.apache.http.impl.client.CloseableHttpClient;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


src/main/java/net/snowflake/ingest/SimpleIngestManager.java
```diff
@@ -196,7 +196,7 @@ public SimpleIngestManager build() {
   // logger object for this class
   private static final Logger LOGGER = LoggerFactory.getLogger(SimpleIngestManager.class);
   // HTTP Client that we use for sending requests to the service
-  private HttpClient httpClient;
+  private CloseableHttpClient httpClient;
```
sfc-gh-tzhang, line 199: Do we need to close the client somewhere?

sfc-gh-tzhang (approved):
LGTM, I assume that the same logic will be added for all other APIs sometimes later?

src/main/java/net/snowflake/ingest/SimpleIngestManager.java
sfc-gh-japatel: I was actually thinking of closing the client when `SimpleIngestManager` is closed, but then since httpClient cant remain static instance. Because atleast in KC, we call close on simplingestmaneger which closes the BG service of keyRenewal. 

If I add httpClient.close(), I will also have to wipe off httpClient from JVM since it is static?

I can do `httpClient = null` may be?

sfc-gh-japatel: But then again defeats the purpose of static instance. I would prefer going without close() on closeable instance.

sfc-gh-tjones (approved):

PR #167, opened 2022-04-20
Use CloseableHttpClient and consume Entity on all APIs

sfc-gh-japatel:
Fixing the bug across all APIs

- Closing the entity. 
- Try with resources for httpResponse. (Closes the response)

TLDR: 
- Same as https://github.com/snowflakedb/snowflake-ingest-java/pull/166
- Without the release bump up but for all APIs. 
- Will also push this to master.

### DIFF
--- a/src/main/java/net/snowflake/ingest/SimpleIngestManager.java
+++ b/src/main/java/net/snowflake/ingest/SimpleIngestManager.java
@@ -32,10 +32,10 @@ import net.snowflake.ingest.utils.BackOffException;
 import net.snowflake.ingest.utils.HttpUtil;
 import net.snowflake.ingest.utils.StagedFileWrapper;
 import net.snowflake.ingest.utils.Utils;
-import org.apache.http.HttpResponse;
-import org.apache.http.client.HttpClient;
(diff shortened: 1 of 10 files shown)


src/main/java/net/snowflake/ingest/SimpleIngestManager.java
```diff
@@ -564,13 +564,13 @@ public IngestResponse ingestFiles(
...
 
     // send the request and get a response....
-    HttpResponse response = httpClient.execute(httpPostForIngestFile);
-
-    LOGGER.info(
-        "Attempting to unmarshall insert response - {}, with clientInfo - {}",
-        response,
-        clientInfo);
-    return ServiceResponseHandler.unmarshallIngestResponse(response, requestId);
+    try (CloseableHttpResponse response = httpClient.execute(httpPostForIngestFile)) {
```
sfc-gh-tzhang, line 567: Could you update the four SS APIs as well?

sfc-gh-japatel: yes, let me do it..

sfc-gh-tzhang (approved):
LGTM, thanks a lot!

PR #171, opened 2022-04-26
Fix logging project properties load exception

sfc-gh-azagrebin:


src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
```diff
@@ -10,25 +10,17 @@
...
 import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
 import java.util.UUID;
 import net.snowflake.ingest.SimpleIngestManager;
-import net.snowflake.ingest.utils.ErrorCode;
-import net.snowflake.ingest.utils.SFException;
-import net.snowflake.ingest.utils.SnowflakeURL;
-import net.snowflake.ingest.utils.StagedFileWrapper;
+import net.snowflake.ingest.utils.*;
```
sfc-gh-gdoci, line 23: nit: replace wildcard import with only what we need

sfc-gh-tzhang (approved):
Thanks Andrey! Could you add a test for this if possible? And please ask @sfc-gh-japatel to take a look :)

sfc-gh-japatel:
LGTM, thanks! I tested this by deleting `project.properties` file.. Shows the exception but is just a warn log.

sfc-gh-japatel (approved):

PR #104, opened 2021-11-23
SNOW-488437 Add connection timeout

sfc-gh-anwang:
Resolves issues related to this [ticket](https://snowflakecomputing.atlassian.net/browse/SNOW-488437). 
Currently, the Java SDK does not timeout on HTTP connections, I added timeout in the RequestConfig.

PR #173, opened 2022-04-28
SNOW-585421 use custom connection pooling manager for Apache httpClient

sfc-gh-japatel:
Adding few configurations for `httpClient` as mentioned here. 

- The default implementation of ConnectionManager has just 20 max connections and 2 max connections per route.
- Adding commit https://github.com/snowflakedb/snowflake-ingest-java/pull/104 on top of 0.10.6 release too.
- Adding a background thread for closingIdleConnections as described in [here](https://hc.apache.org/httpcomponents-client-4.5.x/current/tutorial/html/connmgmt.html#d5e405).

Overall guide for HttpClient
- https://hc.apache.org/httpcomponents-client-4.5.x/current/httpclient/apidocs/org/apache/http/impl/conn/PoolingHttpClientConnectionManager.html

### DIFF
--- a/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
+++ b/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
@@ -6,6 +6,8 @@ package net.snowflake.ingest.utils;
 
 import com.google.common.base.Strings;
 import java.security.Security;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import javax.net.ssl.SSLContext;
 import org.apache.http.HttpHost;
 import org.apache.http.HttpRequest;
@@ -17,7 +19,9 @@ import org.apache.http.auth.UsernamePasswordCredentials;
 import org.apache.http.client.CredentialsProvider;
 import org.apache.http.client.HttpRequestRetryHandler;
 import org.apache.http.client.ServiceUnavailableRetryStrategy;
+import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.protocol.HttpClientContext;
+import org.apache.http.conn.routing.HttpRoute;
 import org.apache.http.conn.ssl.DefaultHostnameVerifier;
 import org.apache.http.conn.ssl.SSLConnectionSocketFactory;
 import org.apache.http.impl.client.BasicCredentialsProvider;
@@ -25,6 +29,7 @@ import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.http.impl.conn.DefaultProxyRoutePlanner;
 import org.apache.http.impl.conn.PoolingHttpClientConnectionManager;
+import org.apache.http.pool.PoolStats;
 import org.apache.http.protocol.HttpContext;
 import org.apache.http.ssl.SSLContexts;
 import org.slf4j.Logger;
@@ -43,10 +48,18 @@ public class HttpUtil {
   private static int MAX_RETRIES = 3;
 
   private static CloseableHttpClient httpClient;
+
+  private static final int DEFAULT_CONNECTION_TIMEOUT_MINUTES = 1;
+  private static final int DEFAULT_HTTP_CLIENT_SOCKET_TIMEOUT_MINUTES = 5;
+
   private static final int DEFAULT_MAX_CONNECTIONS_PER_ROUTE = 100;
   private static final int DEFAULT_MAX_CONNECTIONS = 100;
 
+  private static final int DEFAULT_TTL_CONNECTION = 120;
+
   private static final long MONITOR_THREAD_INTERVAL_MS = TimeUnit.SECONDS.toMillis(5);
+
+  private static final int DEFAULT_IDLE_CONNECTION_TIMEOUT_SECONDS = 30;
 
   public static CloseableHttpClient getHttpClient() {
     if (httpClient == null) {
@@ -75,16 +88,21 @@ public class HttpUtil {
     RequestConfig requestConfig =
         RequestConfig.custom()
             .setConnectTimeout(
-                (int) TimeUnit.MILLISECONDS.convert(DEFAULT_CONNECTION_TIMEOUT, TimeUnit.MINUTES))
+                (int)
+                    TimeUnit.MILLISECONDS.convert(
+                        DEFAULT_CONNECTION_TIMEOUT_MINUTES, TimeUnit.MINUTES))
             .setConnectionRequestTimeout(
-                (int) TimeUnit.MILLISECONDS.convert(DEFAULT_CONNECTION_TIMEOUT, TimeUnit.MINUTES))
+                (int)
+                    TimeUnit.MILLISECONDS.convert(
+                        DEFAULT_CONNECTION_TIMEOUT_MINUTES, TimeUnit.MINUTES))
             .setSocketTimeout(
                 (int)
                     TimeUnit.MILLISECONDS.convert(
-                        DEFAULT_HTTP_CLIENT_SOCKET_TIMEOUT, TimeUnit.MINUTES))
+                        DEFAULT_HTTP_CLIENT_SOCKET_TIMEOUT_MINUTES, TimeUnit.MINUTES))
             .build();
 
-    PoolingHttpClientConnectionManager connectionManager = new PoolingHttpClientConnectionManager();
+    PoolingHttpClientConnectionManager connectionManager =
+        new PoolingHttpClientConnectionManager(DEFAULT_TTL_CONNECTION, TimeUnit.SECONDS);
     connectionManager.setDefaultMaxPerRoute(DEFAULT_MAX_CONNECTIONS_PER_ROUTE);
     connectionManager.setMaxTotal(DEFAULT_MAX_CONNECTIONS);
 
@@ -97,7 +115,8 @@ public class HttpUtil {
             .setConnectionManager(connectionManager)
             .setSSLSocketFactory(f)
             .setServiceUnavailableRetryStrategy(getServiceUnavailableRetryStrategy())
-            .setRetryHandler(getHttpRequestRetryHandler());
+            .setRetryHandler(getHttpRequestRetryHandler())
+            .setDefaultRequestConfig(requestConfig);
 
     // proxy settings
     if ("true".equalsIgnoreCase(System.getProperty(USE_PROXY))) {
@@ -225,11 +244,26 @@ public class HttpUtil {
           while (!shutdown) {
             wait(MONITOR_THREAD_INTERVAL_MS);
 
+            StringBuilder sb = new StringBuilder();
+
+            sb.append(
+                createPoolStatsInfo("Total Pool Stats = ", connectionManager.getTotalStats()));
+            Set<HttpRoute> routes = connectionManager.getRoutes();
+
+            if (routes != null) {
+              for (HttpRoute route : routes) {
+                sb.append(createRouteInfo(connectionManager, route));
+              }
+            }
+
+            LOGGER.debug("[IdleConnectionMonitorThread] Pool Stats:\n" + sb);
+
             // Close expired connections
             connectionManager.closeExpiredConnections();
             // Optionally, close connections
-            // that have been idle longer than 120 sec
-            connectionManager.closeIdleConnections(120, TimeUnit.SECONDS);
+            // that have been idle longer than 30 sec
+            connectionManager.closeIdleConnections(
+                DEFAULT_IDLE_CONNECTION_TIMEOUT_SECONDS, TimeUnit.SECONDS);
           }
         }
       } catch (InterruptedException ex) {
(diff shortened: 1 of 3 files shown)


src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -44,6 +49,18 @@ public class HttpUtil {
...
+  private static final int DEFAULT_HTTP_CLIENT_SOCKET_TIMEOUT_MINUTES = 5;
+
+  private static final int DEFAULT_MAX_CONNECTIONS_PER_ROUTE = 100;
+  private static final int DEFAULT_MAX_CONNECTIONS = 100;
+
+  private static final int DEFAULT_TTL_CONNECTION_MINUTES = 2;
+
+  private static final long MONITOR_THREAD_INTERVAL_MS = TimeUnit.SECONDS.toMillis(5);
+
+  private static final int DEFAULT_IDLE_CONNECTION_TIMEOUT_SECONDS = 30;
```
sfc-gh-tzhang, line 62: Could you explain why we need to override all these values? I thought we only need DEFAULT_MAX_CONNECTIONS_PER_ROUTE and DEFAULT_MAX_CONNECTIONS? Also, is it make more sense to pick a lower DEFAULT_MAX_CONNECTIONS_PER_ROUTE number?

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -103,6 +147,12 @@ private static void initHttpClient() {
     }
 
     httpClient = clientBuilder.build();
+
+    // Monitors in a separate thread where it closes any idle connections
+    // https://hc.apache.org/httpcomponents-client-4.5.x/current/tutorial/html/connmgmt.html
+    IdleConnectionMonitorThread idleConnectionMonitorThread =
```
sfc-gh-tzhang, line 153: where do we shut it down?

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-japatel: I dont have a strong reasoning behind this numbers, except for the fact that I remember we had 100 as max partitions in topic ingesting data to snowflake.

jdbc has this value default to 300 max connections. 

From description: We already have this pushed in master, but I wanted to get this in release branch. 
Adding commit https://github.com/snowflakedb/snowflake-ingest-java/pull/104 on top of 0.10.6 release too.

DEFAULT_IDLE_CONNECTION_TIMEOUT_SECONDS = 30 seconds comes from https://hc.apache.org/httpcomponents-client-4.5.x/current/httpclient/apidocs/org/apache/http/impl/conn/PoolingHttpClientConnectionManager.html

DEFAULT_TTL_CONNECTION_MINUTES = 2 comes from the fact that 2 minutes seems a reasonable time to exhaust an active connection and ask for the new connection from pool. https://stackoverflow.com/questions/39644479/poolingnhttpclientconnectionmanager-what-is-timetolive-attribute-for

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-japatel: Good point, we don't shut it down. We also never shut down the securitymanager's scheduler executor service.
I converted this thread into a deamon thread.

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-tzhang: ok, so seems like we're trying to match what JDBC is doing? If yes, then I would suggest to remove DEFAULT_TTL_CONNECTION_MINUTES, since by default, we're NOT removing it from the connection pool in JDBC (value set to -1)

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-tzhang: > We also never shut down the securitymanager's scheduler executor service.

I don't see any reason why we shouldn't it shut it down, maybe add a fix in master?

sfc-gh-tzhang (approved):
lgtm, left some suggestions, thanks!

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -44,6 +49,18 @@ public class HttpUtil {
 
   private static CloseableHttpClient httpClient;
 
+  private static final int DEFAULT_CONNECTION_TIMEOUT_MINUTES = 1;
+  private static final int DEFAULT_HTTP_CLIENT_SOCKET_TIMEOUT_MINUTES = 5;
+
+  private static final int DEFAULT_MAX_CONNECTIONS_PER_ROUTE = 100;
+  private static final int DEFAULT_MAX_CONNECTIONS = 100;
```
sfc-gh-giyer, line 56: may be add a comment about why 100 is chosen for `DEFAULT_MAX_CONNECTIONS` (because we only have 1 route for ingest)

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -44,6 +49,18 @@ public class HttpUtil {
...
+
+  private static final int DEFAULT_TTL_CONNECTION_MINUTES = 2;
```
sfc-gh-giyer, line 58: can you explain the significance of this parameter in a comment?

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -44,6 +49,18 @@ public class HttpUtil {
...
+
+  private static final long MONITOR_THREAD_INTERVAL_MS = TimeUnit.SECONDS.toMillis(5);
```
sfc-gh-giyer, line 60: can you add a comment about what this is monitoring? also consider making the name more specific.

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-giyer, line 62: It will be good to link appropriate documentation which can help readers understand what idle timeout is vs TTL etc.

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -64,15 +81,42 @@ private static void initHttpClient() {
...
+    // Set connectionTimeout which is the timeout until a connection with the server is established
+    // Set connectionRequestTimeout which is the time to wait for getting a connection from the
+    // connection pool
+    // Set socketTimeout which is the max time gap between two consecutive data packets
+    RequestConfig requestConfig =
+        RequestConfig.custom()
+            .setConnectTimeout(
+                (int)
+                    TimeUnit.MILLISECONDS.convert(
+                        DEFAULT_CONNECTION_TIMEOUT_MINUTES, TimeUnit.MINUTES))
```
sfc-gh-giyer, line 93: technically connect timeout is unrelated to the issue at hand. what is the default value for this?

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-giyer: +1 to @sfc-gh-tzhang 's comment. if we start something explicitly we should close/shutdown

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -174,4 +225,82 @@ private static String getRequestUriFromContext(final HttpContext httpContext) {
...
+                createPoolStatsInfo("Total Pool Stats = ", connectionManager.getTotalStats()));
+            Set<HttpRoute> routes = connectionManager.getRoutes();
+
+            if (routes != null) {
+              for (HttpRoute route : routes) {
+                sb.append(createPoolStatsForRoute(connectionManager, route));
+              }
+            }
+
+            LOGGER.debug("[IdleConnectionMonitorThread] Pool Stats:\n" + sb);
```
sfc-gh-giyer, line 260: is this level of logging appropriate? (every 5 seconds)

```suggestion
            LOGGER.debug("[IdleConnectionMonitorThread] Pool Stats:\n" + sb.toString());
```

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -174,4 +225,82 @@ private static String getRequestUriFromContext(final HttpContext httpContext) {
...
+        notifyAll();
+      }
+    }
+  }
+
+  /** Create Pool stats for a route */
+  private static String createPoolStatsForRoute(
+      PoolingHttpClientConnectionManager connectionManager, HttpRoute route) {
+    PoolStats routeStats = connectionManager.getStats(route);
+    return createPoolStatsInfo(
```
sfc-gh-giyer, line 288: why do we need `createPoolStatsInfo` to append two strings? seems unnecessary. I would just do it inline since a StringBuilder will not give you any benefit here.

sfc-gh-giyer (commented):
left a few comments. also, we don't create JIRAs for every PR?

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-japatel: Will do!

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-japatel: I got this commit from an already merged commit to master #104 

default values are -1, which means undefined, (From the documentation it is not clear what exactly is undefined, https://hc.apache.org/httpcomponents-client-4.5.x/current/httpclient/apidocs/org/apache/http/client/config/RequestConfig.html#getConnectTimeout())

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-japatel: I will need it to print total stats and one per each route in the connection.

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-japatel: Will revert it to `-1`.

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-japatel: 1. I have set this thread to be daemon thread which will shutdown once the jvm shutsdown. When all of the non-daemon threads complete, daemon threads terminates automatically.
2. Or else, I will have to call shutdown() method from `SimpleIngestManager`'s `close()` method and also need to reinitialize it when a static instance of httpClient is fetched.

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-japatel: I think debug level logging is fine and customers don't set debug level for ingest sdk too often.

sfc-gh-japatel:
> left a few comments. also, we don't create JIRAs for every PR?

Thanks @sfc-gh-giyer , resolved the comments(Added JIRA), please take a look again.

sfc-gh-japatel:
> > left a few comments. also, we don't create JIRAs for every PR?
> 
> Thanks @sfc-gh-giyer , resolved the comments(Added JIRA), please take a look again.

ping.. We need to push this and release a new version for confluent. Does the new commits looks good @sfc-gh-giyer and @mageshn ?

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-giyer: still, it does not seem like a StringBuilder is useful here with the current usage

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -174,4 +243,84 @@ private static String getRequestUriFromContext(final HttpContext httpContext) {
...
+            connectionManager.closeIdleConnections(
+                DEFAULT_IDLE_CONNECTION_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+          }
+        }
+      } catch (InterruptedException ex) {
+        LOGGER.debug("Terminating Idle Connection Monitor Thread ");
+      }
+    }
+
+    private void shutdown() {
```
sfc-gh-giyer, line 293: when is this invoked?

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -174,4 +243,84 @@ private static String getRequestUriFromContext(final HttpContext httpContext) {
...
+      LOGGER.debug("Shutdown Idle Connection Monitor Thread ");
+      shutdown = true;
```
sfc-gh-giyer, line 295: is shutdown idempotent? if not then a check `if (!shutdown)` should be used to be safe.

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-japatel: Yes, removed the usage of SB.

sfc-gh-giyer (approved):
lgtm

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-japatel: Shutdown is idempotent but may be notifyall is not needed. Let me put this inside `if (!shutdown)`

PR #174, opened 2022-04-29
NO-SNOW: accept account names with hyphen

sfc-gh-tzhang:
Currently, we will error out when a valid account name contains a hyphen (example: `pm-connectors.snowflakecomputing.com`), this change fixes this issue and the regex logic is from KC

### DIFF
--- a/.github/workflows/GoogleJavaFormat.yml
+++ b/.github/workflows/GoogleJavaFormat.yml
@@ -11,10 +11,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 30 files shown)


sfc-gh-japatel (approved):
lgtm! Do we need to release this?

sfc-gh-tzhang:
> lgtm! Do we need to release this?

It depends on the response from the SAP folks, I'm still not fully understand the issue. I tried to test on QA1, but internal deployments like QA1 is using account_locator to access instead of account_name, so i'm not whether `pm_connectors `and `pm-connectors` are the same or not.

PR #175, opened 2022-05-06
NO-SNOW: A few small fixes on Snowpipe Streaming PrPr

sfc-gh-tzhang:
1. Add a profile example file (profile_streaming.json.example) specific for Snowpipe Streaming, since there was customer confusions that NOT everything from profile.json.example makes sense for Snowpipe Streaming
2. Move the properties file verification logic all inside `createProperties`
3. Remove the usage of `getPropertiesFromJson` from the example since it won't work outside of the SDK due to the jackson dependency is shaded

### DIFF
--- a/profile_streaming.json.example
+++ b/profile_streaming.json.example
@@ -5,5 +5,5 @@
   "port": 443,
   "host": "account_name.snowflakecomputing.com",
   "scheme": "https",
-  "role": "accountadmin"
+  "role": "role_name"
 }
--- a/profile_streaming.json.example
+++ b/profile_streaming.json.example
@@ -1,0 +1,9 @@
+{
+  "user": "user name",
+  "url": "https://account_name.snowflakecomputing.com:443", // This is required, or it can be constructed from host, scheme and port
+  "private_key": "PEM Private Key",
+  "port": 443,
+  "host": "account_name.snowflakecomputing.com",
+  "scheme": "https",
+  "role": "accountadmin"
+}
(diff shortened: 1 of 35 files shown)


profile_streaming.json.example
```diff
@@ -0,0 +1,9 @@
+{
+  "user": "user name",
+  "url": "https://account_name.snowflakecomputing.com:443", // This is required, or it can be constructed from host, scheme and port
+  "private_key": "PEM Private Key",
+  "port": 443,
+  "host": "account_name.snowflakecomputing.com",
+  "scheme": "https",
+  "role": "accountadmin"
```
sfc-gh-tjones, line 8: Nit, maybe use a role here other than `accountadmin` since it's a pretty powerful role. It's like someone recommend sshing as "root" to a prod box :)

sfc-gh-tjones (approved):
Love it!

src/test/java/net/snowflake/ingest/TestUtils.java
```diff
@@ -199,7 +204,6 @@ public static Connection getConnection() throws Exception {
     Properties props = new Properties();
     props.put("user", user);
     props.put("account", account);
-    props.put("ssl", ssl);
```
sfc-gh-japatel, line 202: why remove this? Similarly for above ssl?

sfc-gh-tzhang: I don't think this is being used, does it?

sfc-gh-japatel: if we dont pass ssl in properties, the Connection wont be secure.

sfc-gh-japatel (approved):
lgtm barring ssl removal!

PR #181, opened 2022-05-13
Japatel fidelity 0.10.7 remove creation of IdleConnectionMonitor thread

sfc-gh-japatel:
Fixes on top of 0.10.7 (Which is released)

But this are more of commits on top of 0.10.3 which is what fidelity has.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -7,7 +7,7 @@
     <!-- Arifact name and version information -->
     <groupId>net.snowflake</groupId>
     <artifactId>snowflake-ingest-sdk</artifactId>
-    <version>0.10.3</version>
+    <version>0.10.6</version>
     <packaging>jar</packaging>
     <name>Snowflake Ingest SDK</name>
     <description>Snowflake Ingest SDK</description>
--- a/pom.xml
+++ b/pom.xml
@@ -7,7 +7,7 @@
     <!-- Arifact name and version information -->
     <groupId>net.snowflake</groupId>
     <artifactId>snowflake-ingest-sdk</artifactId>
-    <version>0.10.6</version>
+    <version>0.10.7</version>
     <packaging>jar</packaging>
     <name>Snowflake Ingest SDK</name>
     <description>Snowflake Ingest SDK</description>
--- a/src/main/java/net/snowflake/ingest/connection/ServiceResponseHandler.java
+++ b/src/main/java/net/snowflake/ingest/connection/ServiceResponseHandler.java
@@ -8,6 +8,7 @@ import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.io.IOException;
 import net.snowflake.ingest.utils.BackOffException;
+import org.apache.http.HttpEntity;
 import org.apache.http.HttpResponse;
 import org.apache.http.HttpStatus;
 import org.apache.http.StatusLine;
@@ -75,8 +76,7 @@ public final class ServiceResponseHandler {
       return null;
     }
 
-    // grab the response entity
-    String blob = EntityUtils.toString(response.getEntity());
+    String blob = consumeAndReturnResponseEntityAsString(response.getEntity());
 
     // Read out the blob entity into a class
     return mapper.readValue(blob, IngestResponse.class);
@@ -112,8 +112,7 @@ public final class ServiceResponseHandler {
       return null;
     }
 
-    // grab the string version of the response entity
-    String blob = EntityUtils.toString(response.getEntity());
+    String blob = consumeAndReturnResponseEntityAsString(response.getEntity());
 
     // read out our blob into a pojo
     return mapper.readValue(blob, HistoryResponse.class);
@@ -141,9 +140,7 @@ public final class ServiceResponseHandler {
       return null;
     }
 
-    // grab the string version of the response entity
-    String blob = EntityUtils.toString(response.getEntity());
-
+    String blob = consumeAndReturnResponseEntityAsString(response.getEntity());
     // read out our blob into a pojo
     return mapper.readValue(blob, HistoryRangeResponse.class);
   }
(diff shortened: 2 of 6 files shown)


pom.xml
```diff
@@ -7,7 +7,7 @@
     <!-- Arifact name and version information -->
     <groupId>net.snowflake</groupId>
     <artifactId>snowflake-ingest-sdk</artifactId>
-    <version>0.10.3</version>
+    <version>0.10.7</version>
```
sfc-gh-tzhang, line 10: We should be creating a 0.10.3.PATCH based on standard, right?

sfc-gh-tzhang: Also, does `v0.10.3-cur-branch` the branch we used for `0.10.3`?

src/main/java/net/snowflake/ingest/connection/ServiceResponseHandler.java
```diff
@@ -167,10 +164,31 @@ private static void handleExceptionalStatus(StatusLine statusLine, HttpResponse
...
+   * @return String version of this http response which will be later used to deserialize into
+   *     respective Response Object
+   * @throws IOException if parsing error
+   */
+  private static String consumeAndReturnResponseEntityAsString(HttpEntity httpResponseEntity)
+      throws IOException {
+    // grab the string version of the response entity
+    String responseEntityAsString = EntityUtils.toString(httpResponseEntity);
+
+    EntityUtils.consumeQuietly(httpResponseEntity);
```
sfc-gh-tzhang, line 191: Looks like we will swallow the exception if using `consumeQuietly`?

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -42,9 +44,27 @@ public class HttpUtil {
...
+
+  // Default is 2, but scaling it up to 100 to match with default_max_connections
+  private static final int DEFAULT_MAX_CONNECTIONS_PER_ROUTE = 100;
+
+  // 100 is close to max partition number we have seen for a kafka topic ingesting into snowflake.
+  private static final int DEFAULT_MAX_CONNECTIONS = 100;
+
+  // Interval in which we check if there are connections which needs to be closed.
+  private static final long IDLE_HTTP_CONNECTION_MONITOR_THREAD_INTERVAL_MS =
+      TimeUnit.SECONDS.toMillis(5);
```
sfc-gh-tzhang, line 62: This is not used, please remove

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -42,9 +44,27 @@ public class HttpUtil {
...
+
+  // Only connections that are currently owned, not checked out, are subject to idle timeouts.
+  private static final int DEFAULT_IDLE_CONNECTION_TIMEOUT_SECONDS = 30;
```
sfc-gh-tzhang, line 65: same here

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -42,9 +44,27 @@ public class HttpUtil {
...
+  private static PoolingHttpClientConnectionManager connectionManager;
+
+  private static final int DEFAULT_CONNECTION_TIMEOUT_MINUTES = 1;
+  private static final int DEFAULT_HTTP_CLIENT_SOCKET_TIMEOUT_MINUTES = 5;
+
+  // Default is 2, but scaling it up to 100 to match with default_max_connections
+  private static final int DEFAULT_MAX_CONNECTIONS_PER_ROUTE = 100;
+
+  // 100 is close to max partition number we have seen for a kafka topic ingesting into snowflake.
+  private static final int DEFAULT_MAX_CONNECTIONS = 100;
```
sfc-gh-tzhang, line 58: This is depends on the number of tasks but not partitions, right?

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -42,9 +44,27 @@ public class HttpUtil {
...
   private static int MAX_RETRIES = 3;
 
-  private static HttpClient httpClient;
+  private static CloseableHttpClient httpClient;
 
-  public static HttpClient getHttpClient() {
+  private static PoolingHttpClientConnectionManager connectionManager;
+
+  private static final int DEFAULT_CONNECTION_TIMEOUT_MINUTES = 1;
+  private static final int DEFAULT_HTTP_CLIENT_SOCKET_TIMEOUT_MINUTES = 5;
```
sfc-gh-tzhang, line 52: In the log, we don't see any issue that is HTTP connection related, should we just use the old value?

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -42,9 +44,27 @@ public class HttpUtil {
...
+
+  // Default is 2, but scaling it up to 100 to match with default_max_connections
+  private static final int DEFAULT_MAX_CONNECTIONS_PER_ROUTE = 100;
```
sfc-gh-tzhang, line 55: DEFAULT_MAX_CONNECTIONS_PER_ROUTE = DEFAULT_MAX_CONNECTIONS means one partition could use all the connections, which seems like an issue? We have 3 APIs, so maybe 40? Or we could make DEFAULT_MAX_CONNECTIONS be match JDBC which is 300, and set DEFAULT_MAX_CONNECTIONS_PER_ROUTE to 100

pom.xml
sfc-gh-japatel: It is major.minor.patch so I guess 0.10.7 is correct. 
Also, 0.10.7 is already released, I thought we are not releasing a new version for them.
Should we do a prerelease?

src/main/java/net/snowflake/ingest/connection/ServiceResponseHandler.java
sfc-gh-japatel: Yess, we will.. But this was released in 0.10.6

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-japatel: This is per partition since we create SimpleINgestManager per partition.

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-japatel: It is per route so the API wont matter.. 

We can match it with JDBC

PR #182, opened 2022-06-02
SNOW-592855, SNOW-606515: fix channel.close time out issue due to high upload latency

sfc-gh-tzhang:
1. Update a few important log lines from DEBUG level to INFO level
2. It's possible that the get status request return an invalid channel status code, so we want to throw as well instead of returning the offset token that may not make sense
3. A bunch of telemetries were moved to JMX, add it back since it's easier to see it in the console
4. Increase the timeout for channel.close() due to high file upload tail latency, we will trust the server side more at this point instead of failing fast


sfc-gh-tzhang:
@sfc-gh-kbregula any idea about the GoogleJavaFormat failure?

sfc-gh-kbregula:
@sfc-gh-tzhang 
Screenshot 2022-06-02 at 20 51 07

sfc-gh-kbregula:
I pushed a fix to your branch.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -572,7 +572,7 @@ BlobMetadata upload(String filePath, byte[] blob, List<ChunkMetadata> metadata)
           metadata.stream().mapToLong(i -> i.getEpInfo().getRowCount()).sum());
     }
 
-    logger.logDebug("Finish uploading file={}", filePath);
+    logger.logInfo("Finish uploading file={}", filePath);
```
sfc-gh-japatel, line 575: if we are doing this logInfo, should we include timestamp diff between start and finish?

src/main/java/net/snowflake/ingest/streaming/internal/ChannelsStatusRequest.java
```diff
@@ -65,12 +59,6 @@ String getChannelName() {
     Long getClientSequencer() {
       return clientSequencer;
     }
-
-    @JsonInclude(Include.NON_EMPTY)
-    @JsonProperty("row_sequencer")
-    Long getRowSequencer() {
-      return rowSequencer;
-    }
```
sfc-gh-japatel, line 73: why remove this? Has the API changed?

sfc-gh-japatel:
> Increase the timeout for channel.close() due to high file upload tail latency, we will trust the server side more at this point instead of failing fast
From code, I see 5sec reduced to 3 seconds, is it the right to understand that we are increasing retries and decreasing timeout?

sfc-gh-tzhang:
> > Increase the timeout for channel.close() due to high file upload tail latency, we will trust the server side more at this point instead of failing fast
> > From code, I see 5sec reduced to 3 seconds, is it the right to understand that we are increasing retries and decreasing timeout?

Correct, but I changed it back and adding a timeout on the total time of waiting for uploading, please take a look at the big description in RegisterService.java and let me know if you have further question!

PR #183, opened 2022-06-03
NO-SNOW: fix test failure due to time zone difference

sfc-gh-tzhang:
Explicitly set the timezone to be UTC for NTZ so the test won't fail when running on non-UTC environment

### DIFF
--- a/src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
+++ b/src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
@@ -9,10 +9,12 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Calendar;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.TimeZone;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -87,7 +89,7 @@ public class StreamingIngestIT {
   @After
   public void afterAll() throws Exception {
     client.close();
-    jdbcConnection.createStatement().execute(String.format("drop database %s", TEST_DB));
+    // jdbcConnection.createStatement().execute(String.format("drop database %s", TEST_DB));
   }
 
   @Test
@@ -387,6 +389,8 @@ public class StreamingIngestIT {
     // Close the channel after insertion
     channel1.close().get();
 
+    Calendar cal = Calendar.getInstance();
+    cal.setTimeZone(TimeZone.getTimeZone("UTC"));
     for (int i = 1; i < 15; i++) {
       if (channel1.getLatestCommittedOffsetToken() != null
           && channel1.getLatestCommittedOffsetToken().equals("1")) {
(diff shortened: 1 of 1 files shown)


src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
```diff
@@ -462,7 +467,7 @@ public void testMultiColumnIngest() throws Exception {
 
     // Open a streaming ingest channel from the given client
     SnowflakeStreamingIngestChannel channel1 = client.openChannel(request1);
-    long timestamp = System.currentTimeMillis();
+    long timestamp = Instant.now().toEpochMilli();
```
sfc-gh-mnaides, line 470: Out of curiosity, why this change?

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
```diff
@@ -493,7 +500,7 @@ public void testMultiColumnIngest() throws Exception {
         Assert.assertEquals(3.14, result.getFloat("F"), 0.0001);
         Assert.assertEquals(1.1, result.getFloat("TINYFLOAT"), 0.001);
         Assert.assertEquals("{\n" + "  \"e\": 2.7\n" + "}", result.getString("VAR"));
-        Assert.assertEquals(timestamp * 1000, result.getTimestamp("T").getTime());
+        Assert.assertEquals(timestamp * 1000, result.getTimestamp("T", cal).getTime());
         Assert.assertEquals(-923, TimeUnit.MILLISECONDS.toDays(result.getDate("D").getTime()));
```
sfc-gh-tzhang, line 504: @sfc-gh-mnaides looks like we're still failing on the date comparison, any idea why?

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
sfc-gh-tzhang: I thought we want everything to be UTC, `System.currentTimeMillis` will be depended on the timezone, right?

sfc-gh-mnaides: https://docs.oracle.com/javase/7/docs/api/java/lang/System.html#currentTimeMillis() Will give you epoch millis, so not timezone dependent

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
sfc-gh-mnaides: I think you need to add cal to the getDate() call as well

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
sfc-gh-tzhang: I see, I think both of them are doing the same thing, revert it

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
sfc-gh-tzhang: ok, I think the issue is the Date returned has a timezone based on the session setting, so I need to set to UTC in the session

sfc-gh-mnaides (approved):

PR #184, opened 2022-06-07
SNOW-585421 Add PoolingHttpClientConnectionManager as the connection …

sfc-gh-japatel:
…manager with default maxConnections per route and maxTotalConns

- This commit has been approved for 0.10.7 here https://github.com/snowflakedb/snowflake-ingest-java/commit/3cd70979a4261af9048948f0d3dd96d8547908ad
- Confluent wants this commit on top of EOS feature for KC and Snowpipe. 
- Adding this on top of 0.10.4-beta.2 and will release next version of 1.7.x

### DIFF
--- a/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
+++ b/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
@@ -7,6 +7,7 @@ package net.snowflake.ingest.utils;
 import static net.snowflake.ingest.utils.StringsUtils.isNullOrEmpty;
 
 import java.security.Security;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import javax.net.ssl.SSLContext;
 import org.apache.http.HttpHost;
@@ -21,13 +22,15 @@ import org.apache.http.client.HttpRequestRetryHandler;
 import org.apache.http.client.ServiceUnavailableRetryStrategy;
 import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.protocol.HttpClientContext;
+import org.apache.http.conn.routing.HttpRoute;
 import org.apache.http.conn.ssl.DefaultHostnameVerifier;
 import org.apache.http.conn.ssl.SSLConnectionSocketFactory;
 import org.apache.http.impl.client.BasicCredentialsProvider;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClientBuilder;
-import org.apache.http.impl.client.HttpClients;
 import org.apache.http.impl.conn.DefaultProxyRoutePlanner;
+import org.apache.http.impl.conn.PoolingHttpClientConnectionManager;
+import org.apache.http.pool.PoolStats;
 import org.apache.http.protocol.HttpContext;
 import org.apache.http.ssl.SSLContexts;
 import org.slf4j.Logger;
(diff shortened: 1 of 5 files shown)


sfc-gh-japatel:
Hi @sfc-gh-tzhang I am planning to release this version because confluent wants to upgrade to 1.7.x and also use the commit for max connections from 0.10.7. They said 0.10.7 was working fine.. 

Let me know if you are okay, i wil release this and test out 1.7.3 (Already did on my machine)

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -125,6 +162,20 @@ private static void initHttpClient() {
     httpClient = clientBuilder.build();
   }
 
+  /** Starts a daemon thread to monitor idle connections in http connection manager. */
+  private static void initIdleConnectionMonitoringThread() {
```
sfc-gh-tzhang, line 166: IIRC this function was causing issue when we looked at the Fidelity case, right?

sfc-gh-japatel: Correct but there are no strong evidence apart from the stack trace, both confluent and I tested this commit on my local machine(with extensive testing) and didnt see any issue.. We will ofcourse wait for this version to be tested with KC before releasing KC. 

But IMO there is nothing in this function, the tests didnt catch this either..

Probably result of more threads? (Cant tell for sure)

sfc-gh-japatel: But, let me wrap this around try catch because  I vaguely remember this was the line
```idleConnectionMonitorThread.start();```

Comment (sfc-gh-japatel) [reply src/main/java/net/snowflake/ingest/utils/HttpUtil.java:166]:
```
040ac5fef93977329|pr102703|pr102703 {"@timestamp":"2022-05-11T18:33:32.931+00:00","message":"Wo rkerSinkTask{id=msm_fcm_snowflake_connector_prod-12} Task threw an uncaught and unrecoverable exception","logger_name":"org.apache.kafka.connect.runtime.WorkerTask","thread_name":"task-t hread-msm_fcm_snowflake_connector_prod-12","level":"ERROR","stack_trace":"com.snowflake.kafka.connector.internal.SnowflakeKafkaConnectorException: 
[SF_KAFKA_CONNECTOR] Exception: Inval id private key
[SF_KAFKA_CONNECTOR] Error Code: 0002
[SF_KAFKA_CONNECTOR] Detail: private key should be a valid PEM RSA private key
[SF_KAFKA_CONNECTOR] Message: null
[SF_KAFKA_CONNECTOR] java.lang.Thread.start(Thread.java:710)
[SF_KAFKA_CONNECTOR] net.snowflake.ingest.utils.HttpUtil.initIdleConnectionMonitoringThread(HttpUtil.java:173)
[SF_KAFKA_CONNECTOR] net.sn owflake.ingest.utils.HttpUtil.getHttpClient(HttpUtil.java:77)
[SF_KAFKA_CONNECTOR] net.snowflake.ingest.SimpleIngestManager.init(SimpleIngestManager.java:409)
[SF_KAFKA_CONNECTOR] net. snowflake.ingest.SimpleIngestManager.<init>(SimpleIngestManager.java:383)
[SF_KAFKA_CONNECTOR] com.snowflake.kafka.connector.internal.SnowflakeIngestionServiceV1.<init>(SnowflakeIngesti onServiceV1.java:46)
[SF_KAFKA_CONNECTOR] com.snowflake.kafka.connector.internal.SnowflakeIngestionServiceFactory$SnowflakeIngestionServiceBuilder.<init>(SnowflakeIngestionServiceFactor y.java:44)
[SF_KAFKA_CONNECTOR] com.snowflake.kafka.connector.internal.SnowflakeIngestionServiceFactory$SnowflakeIngestionServiceBuilder.<init>(SnowflakeIngestionServiceFactory.java:31) 
```

sfc-gh-tzhang: Yes, we want to return the actual exception so we know what happened next time

sfc-gh-tzhang (approved):

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-japatel: Almost certain that this is because of confluent version + threads and not our code. 

I tested this with 200 partition and continuously restarting the tasks to hit the close and open API.

sfc-gh-japatel: This is fixed in https://github.com/snowflakedb/snowflake-ingest-java/commit/de681c650f5ac221c5b13af34771c1fbb6eaec12

PR #187, opened 2022-06-14
SNOW-610062: use "/" as name separator instead of using Paths.get

sfc-gh-tzhang:
We have seen an issue that using the Client SDK under Windows would not work because we're using the wrong name separator with Paths.get, update to use "/" explicitly.


sfc-gh-mnaides (approved):

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -647,15 +645,7 @@ String getFilePath(Calendar calendar, String clientPrefix) {
             + this.counter.getAndIncrement()
             + "."
             + BLOB_EXTENSION_TYPE;
-    Path filePath =
-        Paths.get(
```
sfc-gh-tjones, line 651: Is there any way to pass in a separator here?

sfc-gh-tjones (approved):

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-tzhang: I had the same question but unfortunately I don't think we have it

PR #191, opened 2022-06-17
@snow SNOW-612516 Streaming ingest: add random file reader to scan single columns, format version 2

sfc-gh-azagrebin:
[SNOW-612516](https://snowflakecomputing.atlassian.net/browse/SNOW-612516)

# Problem

The high level problem is that even if the mixed table has 100 columns and a user needs to read only one from a BDEC file, the query scanner will always read and parse the whole file.

# New format

The BDEC file format is currently based on the Arrow Stream format (ArrowStreamWriter in Java Arrow client). This means that the arrow record batches are just written into the file with their own schema headers. The XP scan code has to use the same type of stream reader. The problem is that the stream read API does not have random access methods to read individual columns. This results in loading the whole file every time.

FlushService in client currently uses ArrowStreamWriter to write record batches. The Java Arrow library also provides an ArrowFileWriter that appends the footer at the end for subsequent random column seek.

# Compression

The Java Arrow client supports writing LZ4 and ZSTD compressed columns in VectorUnloader. Currently the Arrow vectors in client are serialised into ArrowRecordBatches and written with ArrowStreamWriter/ArrowWriter that also uses a private hidden VectorUnloader. Both ArrowStreamWriter and ArrowFileWriter extend ArrowWriter. The ArrowWriter does not forward the compression config API from the hidden VectorUnloader. Hence, FlushService cannot currently configure the compression that is set to no compression in ArrowWriter’s VectorUnloader.

The proposal here is to implement a custom ArrowFileWriter that configures its own VectorUnloader with compression. This also allows to fix some subtle Arrow buffer allocation issues with the custom compression implementation.

The compression rate comparison, based on `StreamingIngestIT#testDataTypes`, total size of produced files:
- No compression: 27Mb
- GZIP: 17Mb, compression time: ~2000 milliseconds
- New ZSTD: 16Mb - slightly better than GZIP, compression time: ~100-200 milliseconds

# Format version

The client SDK currently has a blob file version, hardcoded in constants to zero. It is not used anywhere except the BDEC header that is not written.

The client can send the version to GS in BlobMetadata / BlobDTO. The StreamingResource can send to the buffering service in chunk messages and then the buffering service can add the version to the queued BlobChunkDTOs.

See also design doc in JIRA.

### DIFF
--- a/scripts/check_content.sh
+++ b/scripts/check_content.sh
@@ -5,7 +5,7 @@
 set -o pipefail
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
-if jar tvf $DIR/../target/snowflake-ingest-sdk.jar  | awk '{print $8}' | grep -v -E "^(net|com)/snowflake" | grep -v -E "(com|net)/\$" | grep -v -E "^META-INF" | grep -v -E "^mozilla" | grep -v mime.types | grep -v project.properties | grep -v -E "javax" | grep -v -E "^org/" | grep -v -E "^com/google" | grep -v -E "^com/sun" | grep -v "log4j.properties" | grep -v "git.properties" | grep -v "io/" | grep -v "codegen/" | grep -v "com/codahale/" | grep -v "com/ibm/" | grep -v "LICENSE"; then
+if jar tvf $DIR/../target/snowflake-ingest-sdk.jar  | awk '{print $8}' | grep -v -E "^(net|com)/snowflake" | grep -v -E "(com|net)/\$" | grep -v -E "^META-INF" | grep -v -E "^mozilla" | grep -v mime.types | grep -v project.properties | grep -v -E "javax" | grep -v -E "^org/" | grep -v -E "^com/google" | grep -v -E "^com/sun" | grep -v "log4j.properties" | grep -v "git.properties" | grep -v "io/" | grep -v "codegen/" | grep -v "com/codahale/" | grep -v "com/ibm/" | grep -v "LICENSE" | grep -v "aix/" | grep -v "darwin/" | grep -v "win/" | grep -v "freebsd/" | grep -v "linux/" | grep -v "com/github/"; then
   echo "[ERROR] Ingest SDK jar includes class not under the snowflake namespace"
   exit 1
 fi
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ArrowFileWriterWithCompression.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ArrowFileWriterWithCompression.java
@@ -1,0 +1,65 @@
+package net.snowflake.ingest.streaming.internal;
+
+import java.io.IOException;
+import java.nio.channels.WritableByteChannel;
+import org.apache.arrow.vector.VectorSchemaRoot;
+import org.apache.arrow.vector.VectorUnloader;
+import org.apache.arrow.vector.compression.CompressionCodec;
+import org.apache.arrow.vector.compression.NoCompressionCodec;
+import org.apache.arrow.vector.ipc.ArrowFileWriter;
+import org.apache.arrow.vector.ipc.message.ArrowRecordBatch;
+
+/**
+ * {@link ArrowFileWriter} with configurable {@link CompressionCodec}
+ *
+ * <p>{@link ArrowFileWriter} cannot be configured to use a custom {@link CompressionCodec}. {@link
+ * ArrowFileWriter} internally has a {@link VectorUnloader} that can be configured with a custom
+ * {@link CompressionCodec} but {@link ArrowFileWriter} does not forward this configuration in its
+ * public constructors. This class extends {@link ArrowFileWriter} to override the underlying {@link
+ * VectorUnloader} and configure it with a custom {@link CompressionCodec}.
+ */
+class ArrowFileWriterWithCompression extends ArrowFileWriter {
+  private final VectorUnloader unloader;
+  private final boolean compressed;
+
+  ArrowFileWriterWithCompression(
+      VectorSchemaRoot root, WritableByteChannel out, CompressionCodec codec) {
+    // Note: the dictionary provider is null, this allows to simplify writeBatch() in this class
+    // and exclude ensureDictionariesWritten() call from the original ArrowWriter#writeBatch()
+    // implementation
+    super(root, null, out);
+
+    // we have to create an unloader, different to ArrowWriter#unloader
+    // because it is private and we need to configure it with CompressionCodec
+    // that is not available in ArrowWriter
+    this.unloader = new VectorUnloader(root, true, codec, true);
+    this.compressed = codec != NoCompressionCodec.INSTANCE;
+  }
+
+  // same as ArrowWriter#writeBatch but does not write dictionaries
+  // and uses an unloader configured with a custom CompressionCodec
+  @Override
+  public void writeBatch() throws IOException {
+    start(); // same as ensureStarted() in ArrowWriter#writeBatch()
+
+    // not called as in ArrowWriter#writeBatch
+    // because the dictionary provider is always null in ArrowFileWriterWithCompression()
+    // constructor
+    // ensureDictionariesWritten();
+
+    // ArrowWriter#unloader is private but we need to configure it with CompressionCodec
+    // hence we use here a different unloader
+    // the compression happenes in unloader.getRecordBatch() -> appendNodes() ->
+    // buffers.add(codec.compress(..)..)
+    try (ArrowRecordBatch batch = unloader.getRecordBatch()) {
+      writeRecordBatch(batch);
+      if (compressed) {
+        // compression creates new compressed buffers added to the ArrowRecordBatch
+        // their reference counter is incremented twice:
+        // once in CompressionCodec on create and once in ArrowRecordBatch to retain
+        // first the buffers are released here, second in the ArrowRecordBatch#close on try exit
+        batch.getBuffers().forEach(arrowBuf -> arrowBuf.getReferenceManager().release());
+      }
+    }
+  }
+}
--- a/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
@@ -8,7 +8,6 @@ import static net.snowflake.ingest.utils.Constants.BLOB_CHECKSUM_SIZE_IN_BYTES;
 import static net.snowflake.ingest.utils.Constants.BLOB_CHUNK_METADATA_LENGTH_SIZE_IN_BYTES;
 import static net.snowflake.ingest.utils.Constants.BLOB_EXTENSION_TYPE;
 import static net.snowflake.ingest.utils.Constants.BLOB_FILE_SIZE_SIZE_IN_BYTES;
-import static net.snowflake.ingest.utils.Constants.BLOB_FORMAT_VERSION;
 import static net.snowflake.ingest.utils.Constants.BLOB_NO_HEADER;
 import static net.snowflake.ingest.utils.Constants.BLOB_TAG_SIZE_IN_BYTES;
 import static net.snowflake.ingest.utils.Constants.BLOB_VERSION_SIZE_IN_BYTES;
@@ -24,6 +23,7 @@ import java.security.NoSuchAlgorithmException;
 import java.util.List;
 import java.util.zip.GZIPOutputStream;
 import javax.xml.bind.DatatypeConverter;
+import net.snowflake.ingest.utils.Constants;
 import net.snowflake.ingest.utils.Logging;
 import net.snowflake.ingest.utils.Pair;
 
@@ -96,12 +96,49 @@ class BlobBuilder {
   }
 
   /**
+   * Gzip compress the given chunk data if required by the given write mode and pads the compressed
+   * data for encryption.
+   *
+   * @param filePath blob file full path
+   * @param chunkData uncompressed chunk data
+   * @param blockSizeToAlignTo block size to align to for encryption
+   * @param arrowBatchWriteMode Arrow format write mode
+   * @return padded compressed chunk data, aligned to blockSizeToAlignTo, and actual length of
+   *     compressed data before padding at the end
+   * @throws IOException
+   */
+  static Pair<byte[], Integer> compressIfNeededAndPadChunk(
+      String filePath,
+      ByteArrayOutputStream chunkData,
+      int blockSizeToAlignTo,
+      Constants.ArrowBatchWriteMode arrowBatchWriteMode)
+      throws IOException {
+    // Encryption needs padding to the ENCRYPTION_ALGORITHM_BLOCK_SIZE_BYTES
+    // to align with decryption on the Snowflake query path starting from this chunk offset.
+    // The padding does not have arrow data and not compressed.
+    // Hence, the actual chunk size is smaller by the padding size.
+    // The compression on the Snowflake query path needs the correct size of the compressed
+    // data.
+    if (arrowBatchWriteMode == Constants.ArrowBatchWriteMode.STREAM) {
+      // Stream write mode does not support column level compression.
+      // Compress the chunk data and pad it for encryption.
+      return BlobBuilder.compress(filePath, chunkData, blockSizeToAlignTo);
+    } else {
+      int actualSize = chunkData.size();
+      int paddingSize = blockSizeToAlignTo - actualSize % blockSizeToAlignTo;
+      chunkData.write(new byte[paddingSize]);
+      return new Pair<>(chunkData.toByteArray(), actualSize);
+    }
+  }
+
+  /**
    * Build the blob file
    *
    * @param chunksMetadataList List of chunk metadata
    * @param chunksDataList List of chunk data
    * @param chunksChecksum Checksum for the chunk data portion
    * @param chunksDataSize Total size for the chunk data portion after compression
+   * @param bdecVersion BDEC file version
    * @return The blob file as a byte array
    * @throws JsonProcessingException
    */
@@ -109,7 +146,8 @@ class BlobBuilder {
       List<ChunkMetadata> chunksMetadataList,
       List<byte[]> chunksDataList,
       long chunksChecksum,
-      long chunksDataSize)
+      long chunksDataSize,
+      Constants.BdecVerion bdecVersion)
       throws IOException {
     byte[] chunkMetadataListInBytes = MAPPER.writeValueAsBytes(chunksMetadataList);
 
@@ -127,7 +165,7 @@ class BlobBuilder {
     ByteArrayOutputStream blob = new ByteArrayOutputStream();
     if (!BLOB_NO_HEADER) {
       blob.write(BLOB_EXTENSION_TYPE.getBytes());
-      blob.write(BLOB_FORMAT_VERSION);
+      blob.write(bdecVersion.toByte());
       blob.write(toByteArray(metadataSize + chunksDataSize));
       blob.write(toByteArray(chunksChecksum));
       blob.write(toByteArray(chunkMetadataListInBytes.length));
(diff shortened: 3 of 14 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/ArrowFileWriterWithCompression.java
```diff
@@ -0,0 +1,58 @@
...
+    try (ArrowRecordBatch batch = unloader.getRecordBatch()) {
+      writeRecordBatch(batch);
+      if (compressed) {
+        // compression creates new compressed buffers added to the ArrowRecordBatch
+        // their reference counter is incremented twice:
+        // once in CompressionCodec on create and once in ArrowRecordBatch to retain
+        // first the buffers are released here, second in the ArrowRecordBatch#close on try exit
+        for (ArrowBuf arrowBuf : batch.getBuffers()) {
+          arrowBuf.getReferenceManager().release();
+        }
```
sfc-gh-tjones, line 54: Nit, this can be consolidated with a `forEach`

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -109,7 +108,8 @@ static byte[] build(
       List<ChunkMetadata> chunksMetadataList,
       List<byte[]> chunksDataList,
       long chunksChecksum,
-      long chunksDataSize)
+      long chunksDataSize,
+      int bdecVersion)
```
sfc-gh-tjones, line 112: Let's add this to the docstring.

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -127,7 +127,7 @@ static byte[] build(
     ByteArrayOutputStream blob = new ByteArrayOutputStream();
     if (!BLOB_NO_HEADER) {
       blob.write(BLOB_EXTENSION_TYPE.getBytes());
-      blob.write(BLOB_FORMAT_VERSION);
+      blob.write((byte) bdecVersion);
```
sfc-gh-tjones, line 130: Are there any gotchas to doing this given that we're passing in an `int`?

src/main/java/net/snowflake/ingest/streaming/internal/CustomCompressionCodec.java
```diff
@@ -0,0 +1,118 @@
...
+ * <p>The custom code extraction is not optimal in a long term and there is a TODO: to follow up on
+ * this.
+ */
+public class CustomCompressionCodec implements CompressionCodec {
+  private final CompressionCodec actualCodec;
+
+  public CustomCompressionCodec(CompressionUtil.CodecType codecType) {
+    switch (codecType) {
+      case LZ4_FRAME:
+        // TODO: LZ4 compression currently breaks server side decompression
```
sfc-gh-tjones, line 32: Let's add a `SNOW-` JIRA tracking this.

src/test/java/net/snowflake/ingest/TestUtils.java
```diff
@@ -187,9 +189,16 @@ public static Properties getProperties() throws Exception {
     props.put(PRIVATE_KEY, privateKeyPem);
     props.put(ROLE, role);
     props.put(ACCOUNT_URL, getAccountURL());
+    props.put(BLOB_FORMAT_VERSION, getBlobFormatVersion());
     return props;
   }
 
+  private static int getBlobFormatVersion() {
+    return profile.has(BLOB_FORMAT_VERSION)
+        ? profile.get(BLOB_FORMAT_VERSION).asInt()
```
sfc-gh-tjones, line 198: Can we do a `getOrDefault` here?

sfc-gh-tzhang:
@sfc-gh-jfan any idea about the snyk failure in this PR? I need approval for access

sfc-gh-jfan:
> @sfc-gh-jfan any idea about the snyk failure in this PR? I need approval for access

Hey, @sfc-gh-tzhang  seems the new security vuln has been introduced via 

"Detailed paths
Introduced through: net.snowflake:snowflake-ingest-sdk@1.0.2-beta.2 › org.apache.arrow:arrow-compression@8.0.0 › org.apache.commons:commons-compress@1.20
"
seems that is included via your shell script. If you confirm you are not using this version, I will mark it as false positive in snyk backend

regarding access, I am working on access separation, before that is done, pls feel free to ping me if you see failed check.

sfc-gh-azagrebin:
@sfc-gh-jfan
this PR introduces the `org.apache.arrow:arrow-compression@8.0.0` dependency, we need it to compress ingested arrow data files
does it mean we cannot use this dependency?

sfc-gh-tzhang (commented):
Hi Andrey:

I left some comments in the design doc, please take a look and I feel like we want to conclude on what we want to do first before going through this PR in detail. My main concern is that the added metadata per batch and compression per column will make the file size significantly bigger, hence increases the upload latency. Right now, we're seeing that uploading is the main bottleneck of overall latency so we're trying to keep the file size small.

Thanks,
Toby

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -17,13 +17,19 @@ public class ParameterProvider {
...
   public static final long BUFFER_FLUSH_INTERVAL_IN_MILLIS_DEFAULT = 1000;
   public static final long BUFFER_FLUSH_CHECK_INTERVAL_IN_MILLIS_DEFAULT = 100;
   public static final long INSERT_THROTTLE_INTERVAL_IN_MILLIS_DEFAULT = 500;
   public static final long INSERT_THROTTLE_THRESHOLD_IN_PERCENTAGE_DEFAULT = 10;
   public static final boolean SNOWPIPE_STREAMING_METRICS_DEFAULT = false;
 
+  public static final int BLOB_FORMAT_VERSION_DEFAULT = 1;
+  public static final int BLOB_FORMAT_START_SUPPORTED_VERSION = 1;
+  public static final int BLOB_FORMAT_END_SUPPORTED_VERSION = 2;
+
```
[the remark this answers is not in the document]
sfc-gh-azagrebin: Yes, the idea was to support two formats for some time, in case of any noticed regressions for the newer format. Server side PR supports both versions anyway to handle the older format even now in PrPr.

scripts/check_content.sh
```diff
@@ -5,7 +5,7 @@
 set -o pipefail
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
-if jar tvf $DIR/../target/snowflake-ingest-sdk.jar  | awk '{print $8}' | grep -v -E "^(net|com)/snowflake" | grep -v -E "(com|net)/\$" | grep -v -E "^META-INF" | grep -v -E "^mozilla" | grep -v mime.types | grep -v project.properties | grep -v -E "javax" | grep -v -E "^org/" | grep -v -E "^com/google" | grep -v -E "^com/sun" | grep -v "log4j.properties" | grep -v "git.properties" | grep -v "io/" | grep -v "codegen/" | grep -v "com/codahale/" | grep -v "com/ibm/" | grep -v "LICENSE"; then
+if jar tvf $DIR/../target/snowflake-ingest-sdk.jar  | awk '{print $8}' | grep -v -E "^(net|com)/snowflake" | grep -v -E "(com|net)/\$" | grep -v -E "^META-INF" | grep -v -E "^mozilla" | grep -v mime.types | grep -v project.properties | grep -v -E "javax" | grep -v -E "^org/" | grep -v -E "^com/google" | grep -v -E "^com/sun" | grep -v "log4j.properties" | grep -v "git.properties" | grep -v "io/" | grep -v "codegen/" | grep -v "com/codahale/" | grep -v "com/ibm/" | grep -v "LICENSE" | grep -v "aix/" | grep -v "darwin/" | grep -v "win/" | grep -v "freebsd/" | grep -v "linux/" | grep -v "com/github/"; then
```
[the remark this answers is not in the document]
sfc-gh-azagrebin: Yes, they are all about the new Arrow compression algorithms

src/main/java/net/snowflake/ingest/streaming/internal/ArrowFileWriterWithCompression.java
```diff
@@ -0,0 +1,58 @@
...
+    start(); // same as ensureStarted() in ArrowWriter#writeBatch()
+
+    // not called as in ArrowWriter#writeBatch
+    // because the dictionary provider is always null in ArrowFileWriterWithCompression()
+    // constructor
+    // ensureDictionariesWritten();
+
+    // ArrowWriter#unloader is private but we need to configure it with CompressionCodec
+    // hence we use here a different unloader
+    try (ArrowRecordBatch batch = unloader.getRecordBatch()) {
```
sfc-gh-tzhang, line 45: I couldn't find where we put data into the unloader, could you elaborate?

src/main/java/net/snowflake/ingest/streaming/internal/BlobMetadata.java
```diff
@@ -33,4 +40,9 @@ String getMD5() {
   List<ChunkMetadata> getChunks() {
     return this.chunks;
   }
+
+  @JsonProperty("bdec_version")
+  Integer getVersion() {
+    return bdecVersion;
+  }
```
sfc-gh-tzhang, line 47: do we need to send this to server side?

src/main/java/net/snowflake/ingest/streaming/internal/CustomCompressionCodec.java
```diff
@@ -0,0 +1,118 @@
...
+ * into file results in double release when the original vector gets released as usual.
+ *
+ * <p>The custom code extraction is not optimal in a long term and there is a TODO: to follow up on
+ * this.
+ */
+public class CustomCompressionCodec implements CompressionCodec {
+  private final CompressionCodec actualCodec;
+
+  public CustomCompressionCodec(CompressionUtil.CodecType codecType) {
+    switch (codecType) {
```
sfc-gh-tzhang, line 30: I would suggest we remove all the other compression options if all we support now is ZSTD, we should add them whenever it's needed

src/main/java/net/snowflake/ingest/streaming/internal/CustomCompressionCodec.java
```diff
@@ -0,0 +1,118 @@
...
+ * This is custom wrapper on top of {@link CompressionCodec} implementations.
+ *
+ * <p>The main stream {@link CompressionCodec} implementations are based on {@link
+ * org.apache.arrow.vector.compression.AbstractCompressionCodec}. This implementation is similar to
+ * it with some tweaks.
+ *
+ * <p>The reason to have this custom implementation is to fix certain issues with mainstream codecs
+ * that lead to buffer reference counting problems due to creation of new compressed buffers and
+ * releasing uncompressed. The release of uncompressed buffer from the original vectors to write
+ * into file results in double release when the original vector gets released as usual.
```
sfc-gh-tzhang, line 21: Do you think we have a better way to resolve this issue? My concern is that we have a few assumptions on how the arrow library would work and that might change in the future

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -524,28 +531,69 @@ BlobMetadata buildAndUpload(String filePath, List<List<ChannelData>> blobData)
...
+        BlobBuilder.build(
+            chunksMetadataList, chunksDataList, crc.getValue(), curDataSize, bdecVersion);
     if (buildContext != null) {
       buildContext.stop();
     }
 
     return upload(filePath, blob, chunksMetadataList);
   }
 
+  private Pair<byte[], Integer> compressIfNeededAndPadChunk(
```
sfc-gh-tzhang, line 559: This function should be in `BlobBuilder`, and could you add comment for this function?

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -121,6 +125,9 @@ List<List<ChannelData>> getData() {
   // A map which stores the timer for each blob in order to capture the flush latency
   private final Map<String, Timer.Context> latencyTimerContextMap;
 
+  // blob file version
+  private final int bdecVersion;
```
sfc-gh-tzhang, line 129: we should be passing enums instead of numbers?

src/main/java/net/snowflake/ingest/utils/Constants.java
```diff
@@ -55,11 +54,16 @@ public class Constants {
...
-  public static enum WriteMode {
+  public enum WriteMode {
     CLOUD_STORAGE,
     REST_API,
   }
 
+  public enum ArrowBatchWriteMode {
+    STREAM,
+    FILE,
+  }
```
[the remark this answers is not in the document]
sfc-gh-tzhang: discussed offline and we could do this once we make sure the new writer type is working properly

src/test/java/net/snowflake/ingest/TestUtils.java
sfc-gh-azagrebin: `profile` is a json node, I do not see it having `getOrDefault`

src/main/java/net/snowflake/ingest/streaming/internal/ArrowFileWriterWithCompression.java
sfc-gh-azagrebin: I added into the comment:
unloader.getRecordBatch() -> appendNodes() -> buffers.add(codec.compress(..)..)

src/main/java/net/snowflake/ingest/streaming/internal/BlobMetadata.java
sfc-gh-azagrebin: yes, it is stored in the bdec file format metadata that is used by the scanner to parse the chunk.

sfc-gh-azagrebin:
@sfc-gh-tjones @sfc-gh-tzhang 
Thanks for the review!
I addressed the comments

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -127,7 +161,7 @@ static byte[] build(
     ByteArrayOutputStream blob = new ByteArrayOutputStream();
     if (!BLOB_NO_HEADER) {
       blob.write(BLOB_EXTENSION_TYPE.getBytes());
-      blob.write(BLOB_FORMAT_VERSION);
+      blob.write(bdecVersion.toByte());
```
sfc-gh-tzhang, line 164: Could we pass number to server side instead of enums? I'm imagine that in the future we would have a major version and a minor version, major version will be used for major changes and minor version is for small updates, so it would be 0.0 and 1.0 in this case

src/main/java/net/snowflake/ingest/streaming/internal/CustomCompressionCodec.java
sfc-gh-tzhang: I don't think we use NO_COMPRESSION neither, should we remove it? Or you believe that we will use it later?

src/main/java/net/snowflake/ingest/utils/Constants.java
```diff
@@ -55,11 +56,65 @@ public class Constants {
...
+
+  /** Thw write mode to generate Arrow BDEC file. */
+  public enum BdecVerion {
+    /** Uses Arrow to generate BDEC chunks with {@link ArrowBatchWriteMode#STREAM}. */
+    ONE(1),
+
+    /** Uses Arrow to generate BDEC chunks with {@link ArrowBatchWriteMode#FILE}. */
+    TWO(2);
```
sfc-gh-tzhang, line 86: nit: could we start from 0 instead of 1? I feel like most CS engineers like zero-based offset :)

src/test/java/net/snowflake/ingest/streaming/internal/FlushServiceTest.java
```diff
@@ -35,10 +34,7 @@
 import net.snowflake.client.jdbc.SnowflakeConnectionV1;
 import net.snowflake.client.jdbc.internal.org.bouncycastle.jce.provider.BouncyCastleProvider;
 import net.snowflake.ingest.streaming.OpenChannelRequest;
-import net.snowflake.ingest.utils.Cryptor;
-import net.snowflake.ingest.utils.ErrorCode;
-import net.snowflake.ingest.utils.ParameterProvider;
-import net.snowflake.ingest.utils.SFException;
+import net.snowflake.ingest.utils.*;
```
sfc-gh-tzhang, line 37: Please expand the imports instead of using * imports (according to SF's best coding practice)

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
```diff
@@ -973,25 +1011,164 @@ private void waitChannelFlushed(SnowflakeStreamingIngestChannel channel, int num
...
+    row.put("var", nextJson(r));
+    row.put("obj", nextJson(r));
+    row.put(
+        "arr",
+        Arrays.asList(r.nextInt(100), r.nextInt(100), r.nextInt(100)));
+    row.put("epochdays", Math.abs(r.nextInt()) % 18963); // DATE, 02.12.2021
+    // row.put("timesec", (5*60*60 + 12*60 + 43) * 10000 + 4536); // TIME(4), 05:12:43.4536 does not
+    // work for empty table, sf gives sb1 SNOW-514578
+    // row.put("timenano", (14*60*60 + 26*60 + 34) * 1000000000L + 437582643); // TIME(9),
+    // 14:26:34.437582643 does not work for empty table, sf gives sb1 SNOW-514578
```
sfc-gh-tzhang, line 1124: any issues with these rows?

sfc-gh-tzhang (approved):
Thanks Andrey! Left some comments, otherwise lgtm

sfc-gh-tzhang:
> [SNOW-612516](https://snowflakecomputing.atlassian.net/browse/SNOW-612516)
> 
> # Problem
> The high level problem is that even if the mixed table has 100 columns and a user needs to read only one from a BDEC file, the query scanner will always read and parse the whole file.
> 
> # New format
> The BDEC file format is currently based on the Arrow Stream format (ArrowStreamWriter in Java Arrow client). This means that the arrow record batches are just written into the file with their own schema headers. The XP scan code has to use the same type of stream reader. The problem is that the stream read API does not have random access methods to read individual columns. This results in loading the whole file every time.
> 
> FlushService in client currently uses ArrowStreamWriter to write record batches. The Java Arrow library also provides an ArrowFileWriter that appends the footer at the end for subsequent random column seek.
> 
> # Compression
> The Java Arrow client supports writing LZ4 and ZSTD compressed columns in VectorUnloader. Currently the Arrow vectors in client are serialised into ArrowRecordBatches and written with ArrowStreamWriter/ArrowWriter that also uses a private hidden VectorUnloader. Both ArrowStreamWriter and ArrowFileWriter extend ArrowWriter. The ArrowWriter does not forward the compression config API from the hidden VectorUnloader. Hence, FlushService cannot currently configure the compression that is set to no compression in ArrowWriter’s VectorUnloader.
> 
> The proposal here is to implement a custom ArrowFileWriter that configures its own VectorUnloader with compression. This also allows to fix some subtle Arrow buffer allocation issues with the custom compression implementation.
> 
> The compression rate comparison, based on `StreamingIngestIT#testDataTypes`, total size of produced files:
> 
> * No compression: 27Mb
> * GZIP: 17Mb, compression time: ~2000 milliseconds
> * New ZSTD: 16Mb - slightly better than GZIP, compression time: ~100-200 milliseconds
> 
> # Format version
> The client SDK currently has a blob file version, hardcoded in constants to zero. It is not used anywhere except the BDEC header that is not written.
> 
> The client can send the version to GS in BlobMetadata / BlobDTO. The StreamingResource can send to the buffering service in chunk messages and then the buffering service can add the version to the queued BlobChunkDTOs.
> 
> See also design doc in JIRA.

Could you please copy the paste the description in the JIRA instead? This is a public repo and I don't think anyone else other than the SF engineers could see the JIRA content.

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
sfc-gh-azagrebin: This lines passes a byte number to the server side `bdecVersion.toByte()` not enum, same as before.
Enum is now just a container for the byte number.
or did you mean something else?

PR #193, opened 2022-06-23
NO-SNOW http connection pooling manager

sfc-gh-japatel:
This is being approved earlier, just pushing it to master. 

https://github.com/snowflakedb/snowflake-ingest-java/pull/184

- Added a new commit which fixes the bug saw earlier in fidelity's case. 
- Will also fix the exception in KC once release 0.10.4-beta.3 is released

### DIFF
--- a/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
+++ b/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
@@ -50,8 +50,6 @@ public class HttpUtil {
 
   private static final String PROXY_SCHEME = "http";
   private static final int MAX_RETRIES = 3;
-  static final int DEFAULT_CONNECTION_TIMEOUT = 1; // minute
-  static final int DEFAULT_HTTP_CLIENT_SOCKET_TIMEOUT = 5; // minutes
   private static CloseableHttpClient httpClient;
 
   private static PoolingHttpClientConnectionManager connectionManager;
@@ -343,7 +341,7 @@ public class HttpUtil {
           }
         }
       } catch (InterruptedException ex) {
-        LOGGER.debug("Terminating Idle Connection Monitor Thread ");
+        LOGGER.warn("Terminating Idle Connection Monitor Thread ");
       }
     }
 
--- a/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
+++ b/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
@@ -10,6 +10,7 @@ import java.security.Security;
 import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.ReentrantLock;
 import javax.net.ssl.SSLContext;
 import net.snowflake.client.core.SFSessionProperty;
 import org.apache.http.HttpHost;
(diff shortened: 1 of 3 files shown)


sfc-gh-tjones (approved):

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -50,11 +54,39 @@ public class HttpUtil {
...
+  private static IdleConnectionMonitorThread idleConnectionMonitorThread;
+
+  /**
+   * This lock is to synchronize on idleConnectionMonitorThread to avoid setting starting a thread
+   * which was already started. (To avoid {@link IllegalThreadStateException})
+   */
+  private static ReentrantLock idleConnectionMonitorThreadLock = new ReentrantLock(true);
+
+  private static final int DEFAULT_CONNECTION_TIMEOUT_MINUTES = 1;
+  private static final int DEFAULT_HTTP_CLIENT_SOCKET_TIMEOUT_MINUTES = 5;
```
sfc-gh-tzhang, line 68: we already have these two above? line 53-54

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -126,6 +170,27 @@ private static void initHttpClient() {
...
+      if (connectionManager != null
+          && (idleConnectionMonitorThread == null || idleConnectionMonitorThread.isShutdown())) {
+        // Monitors in a separate thread where it closes any idle connections
+        // https://hc.apache.org/httpcomponents-client-4.5.x/current/tutorial/html/connmgmt.html
+        idleConnectionMonitorThread = new IdleConnectionMonitorThread(connectionManager);
+        idleConnectionMonitorThread.setDaemon(true);
+        idleConnectionMonitorThread.start();
+      }
+    } catch (Exception e) {
+      LOGGER.warn("Unable to start Daemon thread for Http Idle Connection Monitoring", e);
```
sfc-gh-tzhang, line 188: logger.error?

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -235,4 +300,86 @@ public static Properties generateProxyPropertiesForJDBC() {
...
+            // Close expired connections
+            connectionManager.closeExpiredConnections();
+            // Optionally, close connections
+            // that have been idle longer than 30 sec
+            connectionManager.closeIdleConnections(
+                DEFAULT_IDLE_CONNECTION_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+          }
+        }
+      } catch (InterruptedException ex) {
+        LOGGER.debug("Terminating Idle Connection Monitor Thread ");
```
sfc-gh-tzhang, line 346: logger.error?

sfc-gh-tzhang (approved):

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-japatel: Ouch, thanks for catching, let me fix it.

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-japatel: Didnt want to throw this as error, as this is more for debugging. warn should be fine.

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-japatel: Same as before. this is all BG task and is not essential to find the rootcause. will change this to warn.

PR #194, opened 2022-06-29
SNOW-619777, SNOW-606454: fix default min/max value in EP for real column

sfc-gh-tzhang:
When all the values in a REAL column are NULLs, the server side expects the min/max in EP to be 0.0, but the client SDK is sending NULL to the server side and causing exception. This PR fixes this issue and it also contains a few small related changes.


sfc-gh-tjones (approved):
Nice fix! Thank you!

src/main/java/net/snowflake/ingest/streaming/internal/RowBufferStats.java
```diff
@@ -252,11 +253,7 @@ byte[] performConversion(String input) {
 
       if (icuCollationString == null) {
         // No need for ICU key generation, convert (modified) input to UTF-8 bytes
-        try {
-          return input.getBytes("UTF-8");
-        } catch (java.io.UnsupportedEncodingException e) {
-          throw new SFException(ErrorCode.INTERNAL_ERROR, "Failed encoding string to UTF-8");
```
sfc-gh-mnaides, line 258: Why remove this error?

src/main/java/net/snowflake/ingest/streaming/internal/RowBufferStats.java
```diff
@@ -398,8 +395,8 @@ void addStrValue(String inputValue) {
     String value =
         inputValue.length() > MAX_LOB_LEN ? inputValue.substring(0, MAX_LOB_LEN) : inputValue;
 
-    byte[] valueBytes = value != null ? value.getBytes(StandardCharsets.UTF_8) : null;
-    byte[] collatedValueBytes = value != null ? getCollatedBytes(value) : null;
```
sfc-gh-mnaides, line 402: why remove the null check?

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
```diff
@@ -956,6 +957,7 @@ private float convertRowToArrow(
 
     // Insert nulls to the columns that doesn't show up in the input
     for (String columnName : Sets.difference(this.fields.keySet(), inputColumnNames)) {
+      rowBufferSize += 0.125; // 1/8 for null value bitmap
```
sfc-gh-mnaides, line 960: Interesting, how'd you come across this issue?

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
```diff
@@ -861,6 +861,38 @@ public void testChannelClose() throws Exception {
...
+    // Open a streaming ingest channel from the given client
+    SnowflakeStreamingIngestChannel channel1 = client.openChannel(request);
+    for (int val = 0; val < 1000; val++) {
+      Map<String, Object> row = new HashMap<>();
+      row.put("c0", val);
+      verifyInsertValidationResponse(channel1.insertRow(row, Integer.toString(val)));
+    }
+
+    // Close the channel to make sure everything is committed
+    channel1.close().get();
```
sfc-gh-mnaides, line 893: You should query the data in this test to make sure everything is correct and readable

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
sfc-gh-tzhang: I just read through the code and realize we didn't add this for null values

src/main/java/net/snowflake/ingest/streaming/internal/RowBufferStats.java
sfc-gh-tzhang: This change was recommended by Intellij

src/main/java/net/snowflake/ingest/streaming/internal/RowBufferStats.java
sfc-gh-tzhang: This is recommended by Intellj as well because we already do `inputValue.length` above, but I don't think this is correct since we could add a NULL string, will update

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
sfc-gh-tzhang: will do

src/main/java/net/snowflake/ingest/streaming/internal/RowBufferStats.java
sfc-gh-tzhang: Actually, `convertRowToArrow` already have a separate branch for NULL so we're good

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
```diff
@@ -889,8 +889,32 @@ public void testNullValuesOnMultiDataTypes() throws Exception {
       verifyInsertValidationResponse(channel1.insertRow(row, Integer.toString(val)));
     }
 
+    // Sleep for a few seconds to make sure a flush
+    Thread.sleep(5000);
```
sfc-gh-mnaides, line 893: Could you block on the flush finishing instead of sleeping?

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
```diff
@@ -889,8 +889,32 @@ public void testNullValuesOnMultiDataTypes() throws Exception {
...
+
+    // Query the table in the end to make sure everything is readable
+    ResultSet result =
+        jdbcConnection
+            .createStatement()
+            .executeQuery(
+                String.format(
+                    "select top 1 c1, c2, c3, c4, c5, c6, c7, c8, c9, c10, c11, c12, c13, c14,"
+                        + " c15, c16, c17 from %s.%s.%s",
+                    TEST_DB, TEST_SCHEMA, nullValuesOnMultiDataTypesTable));
```
sfc-gh-mnaides, line 913: Try adding an order by to make sure the result order is deterministic

sfc-gh-mnaides (approved):
Two comments, otherwise lgtm

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
sfc-gh-tzhang: i'm not sure how to do that, we don't have a flush API on the public interface anymore

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
sfc-gh-tzhang: Everything in these columns should always be null, so I don't think order by is not really necessary

PR #195, opened 2022-07-08
SNOW-356156: support client telemetry through client_telemetry_v

sfc-gh-tzhang:
This PR contains the following major changes:
- Add the ability in the ingest SDK to upload telemetries through the **sessionless** client telemetry framework
- Add a few client side metrics for Streaming Ingest and upload them to SF
- Add corresponding unit test, IT test should already covered by existing tests


src/main/java/net/snowflake/ingest/connection/SecurityManager.java
```diff
@@ -177,6 +190,11 @@ private void regenerateToken() {
     // atomically update the string
     LOGGER.info("Created new JWT");
     token.set(newToken);
+
+    // Refresh the token used in the telemetry service as well
+    if (telemetryService != null) {
+      telemetryService.refreshJWTToken(newToken);
```
sfc-gh-japatel, line 196: Will telemetryService ever be null?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -726,49 +747,71 @@ ParameterProvider getParameterProvider() {
...
-    jvmMemoryAndThreadMetrics = new MetricRegistry();
-    jvmMemoryAndThreadMetrics.register(
-        MetricRegistry.name("jvm", "memory"), new MemoryUsageGaugeSet());
-    jvmMemoryAndThreadMetrics.register(
-        MetricRegistry.name("jvm", "threads"), new ThreadStatesGaugeSet());
-
-    SharedMetricRegistries.add(
-        SNOWPIPE_STREAMING_JVM_MEMORY_AND_THREAD_METRICS_REGISTRY, jvmMemoryAndThreadMetrics);
+
+    if (ENABLE_TELEMETRY_TO_SF || this.parameterProvider.hasEnabledSnowpipeStreamingMetrics()) {
```
sfc-gh-japatel, line 770: May be we can remove `this.parameterProvider.hasEnabledSnowpipeStreamingMetrics()`? since ENABLE_TELEMETRY_TO_SF is always true!

pom.xml
```diff
@@ -228,33 +228,6 @@
             <version>2.13.2.1</version>
         </dependency>
 
-        <!-- Apache HTTP Components for actually sending requests over the network -->
```
sfc-gh-japatel, line 231: nit: may be do the refactoring of httpclient in separate PR to keep this only telemetry related. 

Also, thanks for doing it, no reason why we cant use it from jdbc!

src/test/java/net/snowflake/ingest/connection/TestKeyRenewal.java
```diff
@@ -30,7 +30,8 @@ public void doesRegenerateToken()
     KeyPair keyPair = keyGen.generateKeyPair();
 
     // create the security manager;
-    SecurityManager manager = new SecurityManager("account", "user", keyPair, 3, TimeUnit.SECONDS);
+    SecurityManager manager =
+        new SecurityManager("account", "user", keyPair, 3, TimeUnit.SECONDS, null);
```
sfc-gh-japatel, line 34: Strange that this test passing since we are passing null and it will not generate new token?

sfc-gh-japatel (approved):
Overall LGTM, minor suggestions, there is one snyk failure in merge gate.

src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
```diff
@@ -177,7 +204,7 @@ public RequestBuilder(
       String schemeName,
       String hostName,
       int portNum) {
-    this(accountName, userName, keyPair, schemeName, hostName, portNum, null);
+    this(accountName, userName, keyPair, schemeName, hostName, portNum, null, null, null);
```
sfc-gh-tjones, line 207: There's a lot of `null`s being given to this constructor, does it make sense to have it be its own specialized constructor? Might reduce the amount of code changed in this PR.

src/main/java/net/snowflake/ingest/connection/TelemetryService.java
```diff
@@ -0,0 +1,138 @@
...
+import net.snowflake.ingest.utils.Logging;
+
+/**
+ * Telemetry service to collect logs in the SDK and send them to Snowflake through the JDBC client
+ * telemetry API
+ */
+public class TelemetryService {
+  // Enum for different client telemetries
+  enum TelemetryType {
+    STREAMING_INGEST_LATENCY_IN_SEC("streaming_ingest_latency_in_sec"),
```
sfc-gh-tjones, line 25: Is this end to end latency? What latency is this measuring?

src/main/java/net/snowflake/ingest/connection/TelemetryService.java
```diff
@@ -0,0 +1,138 @@
...
+    STREAMING_INGEST_CLIENT_FAILURE("streaming_ingest_client_failure"),
```
sfc-gh-tjones, line 26: Should we add `_COUNT` or something here?

src/main/java/net/snowflake/ingest/streaming/internal/RegisterService.java
```diff
@@ -140,14 +140,28 @@ List<FlushService.BlobData> registerBlobs(Map<String, Timer.Context> latencyTime
...
-                  e,
-                  e.getMessage(),
-                  e.getCause(),
-                  e.getCause() == null ? null : e.getCause().getMessage());
+              StringBuilder stackTrace = new StringBuilder();
+              if (e.getCause() != null) {
+                for (StackTraceElement element : e.getCause().getStackTrace()) {
+                  stackTrace.append(System.lineSeparator()).append(element.toString());
+                }
+              }
```
sfc-gh-tjones, line 148: I think there might be a helper somewhere to do this in a common Java utility library but this is fine.

src/main/java/net/snowflake/ingest/utils/Constants.java
```diff
@@ -64,6 +65,7 @@ public static enum WriteMode {
   public static final boolean DISABLE_BACKGROUND_FLUSH = false;
   public static final boolean COMPRESS_BLOB_TWICE = false;
   public static final boolean BLOB_NO_HEADER = true;
+  public static final boolean ENABLE_TELEMETRY_TO_SF = true;
```
sfc-gh-tjones, line 68: This is always on? Should this be driven from a config file?

src/test/java/net/snowflake/ingest/connection/SecurityManagerTest.java
```diff
@@ -20,81 +20,24 @@ public class SecurityManagerTest {
...
   /** BASE64 ENCODED PUBLIC KEY */
   private String storedPublicKey =
-      "MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAtRPru42llC40VdmWnc8r\n"
-          + "7TI/AFemZw4Lh1HRnPIFRxwhOE/yxHHxFGuPLUouyHWM9rVT9N9eo6PTOB8TCnGw\n"
-          + "fwTW2jloSbjtycDdM3+UrBUpX7x/Ufhcwoeu0O3NR5pAhGJRVKCvSpmrD3k2l2vZ\n"
-          + "sRL0230IPGxeDB8m2Wia8QCKKou7AkSsmQ3/9kcKowLGf2axPHty2QSXx4NKvwe0\n"
-          + "B1NnLcQTBc6Z83Lym3gKn8YSINk1ZoO9G5oQKr64wnuQIOlXjcXD8BAEYKbv7VkG\n"
-          + "1vsikqixFpPfWrGUlzhoWWTcn4awzRFaX81ZzhAtfA/laqVvrqN/+O6Cc1k614kV\n"
-          + "GQIDAQAB\n";
+      "MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAtRPru42llC40VdmWnc8r7TI/AFemZw4Lh1HRnPIFRxwhOE/yxHHxFGuPLUouyHWM9rVT9N9eo6PTOB8TCnGwfwTW2jloSbjtycDdM3+UrBUpX7x/Ufhcwoeu0O3NR5pAhGJRVKCvSpmrD3k2l2vZsRL0230IPGxeDB8m2Wia8QCKKou7AkSsmQ3/9kcKowLGf2axPHty2QSXx4NKvwe0B1NnLcQTBc6Z83Lym3gKn8YSINk1ZoO9G5oQKr64wnuQIOlXjcXD8BAEYKbv7VkG1vsikqixFpPfWrGUlzhoWWTcn4awzRFaX81ZzhAtfA/laqVvrqN/+O6Cc1k614kVGQIDAQAB"; // pragma: allowlist secret
```
sfc-gh-tjones, line 23: I know these already exist but where did these credentials come from?

sfc-gh-tjones (approved):

src/main/java/net/snowflake/ingest/connection/TelemetryService.java
sfc-gh-tzhang: this is just client side latency, i didn't put client in the name because everything here will be put into the client_telemetry view

src/main/java/net/snowflake/ingest/connection/TelemetryService.java
sfc-gh-tzhang: this is not an aggregate counter, we will have one per failure and it will contain the actual failure exception

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
sfc-gh-tzhang: Having the ENABLE_TELEMETRY_TO_SF constant is to prevent the case when the telemetry change has some issues, so we can quickly create a patch release with ENABLE_TELEMETRY_TO_SF set to false

PR #197, opened 2022-07-13
NO-SNOW check the retry logic with millis instead of ms and seconds

sfc-gh-japatel:
Found out that it was not retrying while debugging streaming ingest with Client SDK. 

- Existing test which doesnt validate the retry mechanism though! 
- Added a test which checks the retries logic. (Ignored)

### DIFF
--- a/src/test/java/net/snowflake/ingest/streaming/internal/RegisterServiceTest.java
+++ b/src/test/java/net/snowflake/ingest/streaming/internal/RegisterServiceTest.java
@@ -5,6 +5,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import net.snowflake.ingest.utils.Pair;
 import org.junit.Assert;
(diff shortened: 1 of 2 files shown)


sfc-gh-tzhang (approved):
wow, thanks for the fix, I'm curious how did you identify this issue? With the old logic, looks like we will break the while loop for almost each blob?

sfc-gh-japatel:
> wow, thanks for the fix, I'm curious how did you identify this issue? With the old logic, looks like we will break the while loop for almost each blob?

Correct, I was debugging with testaccount and it wasnt retrying the upload logic after it got timeout exception. But if it works in first time, it should be fine.. So basically retries of upload wont work in  latest version.

sfc-gh-japatel:
This wasnt actually caught in the test since java wraps the completeExceptionally inside ExceutionException. I added a comment in the test which should have done the retries but it didnt.

src/test/java/net/snowflake/ingest/streaming/internal/RegisterServiceTest.java
```diff
@@ -53,6 +66,42 @@ public void testRegisterServiceTimeoutException() throws Exception {
     }
   }
 
+  // Ignore since it runs for BLOB_UPLOAD_TIMEOUT_IN_SEC * BLOB_UPLOAD_MAX_RETRY_COUNT
```
sfc-gh-japatel, line 69: @sfc-gh-tzhang added a test which will simulate retries, but ignored since it will take > 1 minute to complete.

PR #199, opened 2022-07-20
include blob upload duration in the registration requests

sfc-gh-gdoci:
This PR adds the file upload duration in ms and includes that information in the blob registration requests.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/BlobMetadata.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/BlobMetadata.java
@@ -16,17 +16,19 @@ class BlobMetadata {
   private final String md5;
   private final Constants.BdecVerion bdecVersion;
   private final List<ChunkMetadata> chunks;
+  private final long uploadDurationMs;
 
-  BlobMetadata(String path, String md5, List<ChunkMetadata> chunks) {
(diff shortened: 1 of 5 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/BlobMetadata.java
```diff
@@ -49,6 +51,11 @@ List<ChunkMetadata> getChunks() {
     return this.chunks;
   }
 
+  @JsonProperty("upload_duration")
```
sfc-gh-tjones, line 54: Let's change this to `upload_duration_ms`

sfc-gh-japatel (approved):
lgtm! can you run formatter. https://github.com/snowflakedb/snowflake-ingest-java#code-style before pushing? (Merge gate is failing)

sfc-gh-tzhang (changes requested):
I assume this is for calculating the end to end ingestion latency? If yes, we're missing two big things here:
1. The time it takes to build a blob
2. The time a blob in the registration queue, waiting be to registered

So the start timestamp we should use for each blob is when we kick off the flush job, please take a look at `flush(boolean isForce)` in `FlushService.java`

sfc-gh-gdoci:
Duration information is added in other PRs.

PR #200, opened 2022-07-22
@snow SNOW-633432 Parquet file generator

sfc-gh-azagrebin:
This PR is the first implementation of Parquet file generator for Streaming Ingest along with Arrow.
Parquet is a native format for Iceberg tables. This is the first step to support Iceberg for Streaming Ingest.

This Parquet implementation buffers the rows in `ParquetRowBuffer`. The `ParquetFlusher` uses classical simple Parquet file generation approach (as in official examples) that writes rows one-by-one. Nonetheless, small research shows that internally it relies on Parquet column buffers that can be used later for the sake of performance if needed for the price for the implementation complication.

@test run `StreamingIngestIT` (except timestamp tests).

The PR is based on a refactoring in #202 that prepares the code base to plugin other format implementations along the existing Arrow one.

@tests in #224

There are still some things that are not supported and planned as separate PRs:
- Scanner file caching (with encryption) is not supported by server side yet: [SNOW-656500](https://snowflakecomputing.atlassian.net/browse/SNOW-656500)
- Parquet V2 is not supported by server side yet: [SNOW-657238](https://snowflakecomputing.atlassian.net/browse/SNOW-657238)
- further optimisation:
    - use fixed len array for SF sb number types for non-Iceberg case
    - use lower level internal Parquet column buffers

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -389,6 +389,46 @@
                                     <pattern>org.apache</pattern>
                                     <shadedPattern>${shadeBase}.apache</shadedPattern>
                                 </relocation>
+                                <relocation>
+                                    <pattern>com.ctc.wstx</pattern>
+                                    <shadedPattern>${shadeBase}.com.ctc.wstx</shadedPattern>
+                                </relocation>
+                                <relocation>
+                                    <pattern>edu.umd.cs.findbugs.annotations</pattern>
+                                    <shadedPattern>${shadeBase}.edu.umd.cs.findbugs.annotations</shadedPattern>
+                                </relocation>
+                                <relocation>
+                                    <pattern>com.jcraft.jsch</pattern>
+                                    <shadedPattern>${shadeBase}.com.jcraft.jsch</shadedPattern>
+                                </relocation>
+                                <relocation>
+                                    <pattern>com.microsoft.sqlserver.jdbc</pattern>
+                                    <shadedPattern>${shadeBase}.com.microsoft.sqlserver.jdbc</shadedPattern>
+                                </relocation>
+                                <relocation>
+                                    <pattern>contribs.mx</pattern>
+                                    <shadedPattern>${shadeBase}.contribs.mx</shadedPattern>
+                                </relocation>
+                                <relocation>
+                                    <pattern>com.zaxxer.hikari</pattern>
+                                    <shadedPattern>${shadeBase}.com.zaxxer.hikari</shadedPattern>
+                                </relocation>
+                                <relocation>
+                                    <pattern>com.squareup.okhttp</pattern>
+                                    <shadedPattern>${shadeBase}.com.squareup.okhttp</shadedPattern>
+                                </relocation>
+                                <relocation>
+                                    <pattern>com.thoughtworks.paranamer</pattern>
+                                    <shadedPattern>${shadeBase}.com.thoughtworks.paranamer</shadedPattern>
+                                </relocation>
+                                <relocation>
+                                    <pattern>com.jamesmurty.utils</pattern>
+                                    <shadedPattern>${shadeBase}.com.jamesmurty.utils</shadedPattern>
+                                </relocation>
+                                <relocation>
+                                    <pattern>net.iharder.base64</pattern>
+                                    <shadedPattern>${shadeBase}.net.iharder.base64</shadedPattern>
+                                </relocation>
                             </relocations>
                             <filters>
                                 <filter>
--- a/pom.xml
+++ b/pom.xml
@@ -322,6 +322,12 @@
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-common</artifactId>
             <version>${hadoop.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>com.github.spotbugs</groupId>
+                    <artifactId>spotbugs-annotations</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
 
         <!-- https://mvnrepository.com/artifact/io.dropwizard.metrics/metrics-core -->
--- a/pom.xml
+++ b/pom.xml
@@ -307,16 +307,12 @@
             <groupId>org.apache.parquet</groupId>
             <artifactId>parquet-hadoop</artifactId>
             <version>1.11.2</version>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.hadoop</groupId>
-            <artifactId>hadoop-mapreduce-client-core</artifactId>
-            <version>${hadoop.version}</version>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.hadoop</groupId>
-            <artifactId>hadoop-client</artifactId>
-            <version>${hadoop.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>javax.annotation</groupId>
+                    <artifactId>javax.annotation-api</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>org.apache.hadoop</groupId>
@@ -326,6 +322,26 @@
                 <exclusion>
                     <groupId>com.github.spotbugs</groupId>
                     <artifactId>spotbugs-annotations</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>com.sun.jersey</groupId>
+                    <artifactId>jersey-core</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>com.sun.jersey</groupId>
+                    <artifactId>jersey-json</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>com.sun.jersey</groupId>
+                    <artifactId>jersey-server</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>javax.servlet</groupId>
+                    <artifactId>servlet-api</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>javax.servlet.jsp</groupId>
+                    <artifactId>jsp-api</artifactId>
                 </exclusion>
             </exclusions>
         </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -389,46 +389,6 @@
                                     <pattern>org.apache</pattern>
                                     <shadedPattern>${shadeBase}.apache</shadedPattern>
                                 </relocation>
-                                <relocation>
-                                    <pattern>com.ctc.wstx</pattern>
-                                    <shadedPattern>${shadeBase}.com.ctc.wstx</shadedPattern>
-                                </relocation>
-                                <relocation>
-                                    <pattern>edu.umd.cs.findbugs.annotations</pattern>
-                                    <shadedPattern>${shadeBase}.edu.umd.cs.findbugs.annotations</shadedPattern>
-                                </relocation>
-                                <relocation>
-                                    <pattern>com.jcraft.jsch</pattern>
-                                    <shadedPattern>${shadeBase}.com.jcraft.jsch</shadedPattern>
-                                </relocation>
-                                <relocation>
-                                    <pattern>com.microsoft.sqlserver.jdbc</pattern>
-                                    <shadedPattern>${shadeBase}.com.microsoft.sqlserver.jdbc</shadedPattern>
-                                </relocation>
-                                <relocation>
-                                    <pattern>contribs.mx</pattern>
-                                    <shadedPattern>${shadeBase}.contribs.mx</shadedPattern>
-                                </relocation>
-                                <relocation>
-                                    <pattern>com.zaxxer.hikari</pattern>
-                                    <shadedPattern>${shadeBase}.com.zaxxer.hikari</shadedPattern>
-                                </relocation>
-                                <relocation>
-                                    <pattern>com.squareup.okhttp</pattern>
-                                    <shadedPattern>${shadeBase}.com.squareup.okhttp</shadedPattern>
-                                </relocation>
-                                <relocation>
-                                    <pattern>com.thoughtworks.paranamer</pattern>
-                                    <shadedPattern>${shadeBase}.com.thoughtworks.paranamer</shadedPattern>
-                                </relocation>
-                                <relocation>
-                                    <pattern>com.jamesmurty.utils</pattern>
-                                    <shadedPattern>${shadeBase}.com.jamesmurty.utils</shadedPattern>
-                                </relocation>
-                                <relocation>
-                                    <pattern>net.iharder.base64</pattern>
-                                    <shadedPattern>${shadeBase}.net.iharder.base64</shadedPattern>
-                                </relocation>
                             </relocations>
                             <filters>
                                 <filter>
--- a/pom.xml
+++ b/pom.xml
@@ -343,6 +343,22 @@
                     <groupId>javax.servlet.jsp</groupId>
                     <artifactId>jsp-api</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>log4j</groupId>
+                    <artifactId>log4j</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.codehaus.jackson</groupId>
+                    <artifactId>jackson-mapper-asl</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>com.google.protobuf</groupId>
+                    <artifactId>protobuf-java</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.mortbay.jetty</groupId>
+                    <artifactId>jetty</artifactId>
+                </exclusion>
             </exclusions>
         </dependency>
 
--- a/pom.xml
+++ b/pom.xml
@@ -43,6 +43,7 @@
         <jacoco.skip.instrument>true</jacoco.skip.instrument>
         <jacoco.version>0.8.5</jacoco.version>
         <arrow.version>9.0.0</arrow.version>
+        <hadoop.version>2.10.1</hadoop.version>
     </properties>
 
     <build>
@@ -300,6 +301,27 @@
             <groupId>org.apache.arrow</groupId>
             <artifactId>arrow-compression</artifactId>
             <version>${arrow.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.parquet</groupId>
+            <artifactId>parquet-hadoop</artifactId>
+            <version>1.11.2</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-mapreduce-client-core</artifactId>
+            <version>${hadoop.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-client</artifactId>
+            <version>${hadoop.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-common</artifactId>
+            <version>${hadoop.version}</version>
         </dependency>
 
         <!-- https://mvnrepository.com/artifact/io.dropwizard.metrics/metrics-core -->
--- a/src/main/java/net/snowflake/ingest/streaming/internal/BlobMetadata.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/BlobMetadata.java
@@ -14,7 +14,7 @@ import net.snowflake.ingest.utils.ParameterProvider;
 class BlobMetadata {
   private final String path;
   private final String md5;
-  final Constants.BdecVersion bdecVersion;
+  private final Constants.BdecVersion bdecVersion;
   private final List<ChunkMetadata> chunks;
 
   BlobMetadata(String path, String md5, List<ChunkMetadata> chunks) {
@@ -55,8 +55,8 @@ class BlobMetadata {
    */
   static BlobMetadata createBlobMetadata(
       String path, String md5, Constants.BdecVersion bdecVersion, List<ChunkMetadata> chunks) {
-    // TODO: Unify BlobMetadata with BlobMetadataWithBdecVersion once server side BdecVersion in
-    // production
+    // TODO SNOW-659721: Unify BlobMetadata with BlobMetadataWithBdecVersion once server side
+    // BdecVersion in production
     return bdecVersion == Constants.BdecVersion.ONE
         ? new BlobMetadata(path, md5, bdecVersion, chunks)
         : new BlobMetadataWithBdecVersion(path, md5, bdecVersion, chunks);
--- a/src/main/java/net/snowflake/ingest/streaming/internal/BlobMetadata.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/BlobMetadata.java
@@ -14,7 +14,7 @@ import net.snowflake.ingest.utils.ParameterProvider;
 class BlobMetadata {
   private final String path;
   private final String md5;
-  private final Constants.BdecVersion bdecVersion;
+  final Constants.BdecVersion bdecVersion;
   private final List<ChunkMetadata> chunks;
 
   BlobMetadata(String path, String md5, List<ChunkMetadata> chunks) {
@@ -49,9 +49,16 @@ class BlobMetadata {
     return this.chunks;
   }
 
-  // TODO: send the bdec_version once server side supports this in production
-  //  @JsonProperty("bdec_version")
-  //  byte getVersionByte() {
-  //    return bdecVersion.toByte();
-  //  }
+  /**
+   * Create {@link BlobMetadata} in case of {@link Constants.BdecVersion#ONE} and {@link
+   * BlobMetadataWithBdecVersion} otherwise to send BDEC version to server side.
+   */
+  static BlobMetadata createBlobMetadata(
+      String path, String md5, Constants.BdecVersion bdecVersion, List<ChunkMetadata> chunks) {
+    // TODO: Unify BlobMetadata with BlobMetadataWithBdecVersion once server side BdecVersion in
+    // production
+    return bdecVersion == Constants.BdecVersion.ONE
+        ? new BlobMetadata(path, md5, bdecVersion, chunks)
+        : new BlobMetadataWithBdecVersion(path, md5, bdecVersion, chunks);
+  }
 }
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
@@ -14,7 +14,8 @@ import java.util.concurrent.ConcurrentHashMap;
  * during flush. The key is a fully qualified table name and the value is a set of channels that
  * belongs to this table
  *
- * @param <T> type of column data (Arrow {@link org.apache.arrow.vector.VectorSchemaRoot})
+ * @param <T> type of column data (Arrow {@link org.apache.arrow.vector.VectorSchemaRoot} or {@link
+ *     ParquetChunkData})
  */
 class ChannelCache<T> {
   // Cache to hold all the valid channels, the key for the outer map is FullyQualifiedTableName and
--- a/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
@@ -674,6 +674,17 @@ class DataValidationUtil {
         input.getClass(), "BOOLEAN", new String[] {"boolean", "Number", "String"});
   }
 
+  static void checkValueInRange(BigDecimal bigDecimalValue, int scale, int precision) {
+    if (bigDecimalValue.abs().compareTo(BigDecimal.TEN.pow(precision - scale)) >= 0) {
+      throw new SFException(
+          ErrorCode.INVALID_ROW,
+          bigDecimalValue,
+          String.format(
+              "Number out of representable exclusive range of (-1e%s..1e%s)",
+              precision - scale, precision - scale));
+    }
+  }
+
   static Set<String> allowedBooleanStringsLowerCased =
       Sets.newHashSet("1", "0", "yes", "no", "y", "n", "t", "f", "true", "false", "on", "off");
 
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
@@ -58,9 +58,11 @@ import org.apache.arrow.vector.VectorSchemaRoot;
  * <li>upload the blob to stage
  * <li>register the blob to the targeted Snowflake table
  *
- * @param <T> type of column data (Arrow {@link org.apache.arrow.vector.VectorSchemaRoot})
+ * @param <T> type of column data (Arrow {@link org.apache.arrow.vector.VectorSchemaRoot} or {@link
+ *     ParquetChunkData})
  */
 class FlushService<T> {
+  private static final long NO_ENCRYPTION_KEY_ID = 0L;
 
   // Static class to save the list of channels that are used to build a blob, which is mainly used
   // to invalidate all the channels when there is a failure
@@ -470,8 +472,10 @@ class FlushService<T> {
         // TODO: encryption is not yet supported by server side for Parquet
         long iv = curDataSize / Constants.ENCRYPTION_ALGORITHM_BLOCK_SIZE_BYTES;
         byte[] encryptedCompressedChunkData =
-            Cryptor.encrypt(
-                compressedAndPaddedChunkData, firstChannel.getEncryptionKey(), filePath, iv);
+            bdecVersion == Constants.BdecVersion.THREE
+                ? compressedAndPaddedChunkData
+                : Cryptor.encrypt(
+                    compressedAndPaddedChunkData, firstChannel.getEncryptionKey(), filePath, iv);
 
         // Compute the md5 of the chunk data
         String md5 = BlobBuilder.computeMD5(encryptedCompressedChunkData, compressedChunkLength);
@@ -490,7 +494,10 @@ class FlushService<T> {
                 .setChunkLength(compressedChunkLength)
                 .setChannelList(result.channelsMetadataList)
                 .setChunkMD5(md5)
-                .setEncryptionKeyId(firstChannel.getEncryptionKeyId())
+                .setEncryptionKeyId(
+                    bdecVersion == Constants.BdecVersion.THREE
+                        ? NO_ENCRYPTION_KEY_ID
+                        : firstChannel.getEncryptionKeyId())
                 .setEpInfo(
                     AbstractRowBuffer.buildEpInfoFromStats(
                         result.rowCount, result.columnEpStatsMapCombined))
@@ -559,7 +566,8 @@ class FlushService<T> {
         blob.length,
         System.currentTimeMillis() - startTime);
 
-    return new BlobMetadata(filePath, BlobBuilder.computeMD5(blob), bdecVersion, metadata);
+    return BlobMetadata.createBlobMetadata(
+        filePath, BlobBuilder.computeMD5(blob), bdecVersion, metadata);
   }
 
   /**
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
@@ -469,13 +469,10 @@ class FlushService<T> {
         // We need to maintain IV as a block counter for the whole file, even interleaved,
         // to align with decryption on the Snowflake query path.
         // TODO: address alignment for the header SNOW-557866
-        // TODO: encryption is not yet supported by server side for Parquet
         long iv = curDataSize / Constants.ENCRYPTION_ALGORITHM_BLOCK_SIZE_BYTES;
         byte[] encryptedCompressedChunkData =
-            bdecVersion == Constants.BdecVersion.THREE
-                ? compressedAndPaddedChunkData
-                : Cryptor.encrypt(
-                    compressedAndPaddedChunkData, firstChannel.getEncryptionKey(), filePath, iv);
+            Cryptor.encrypt(
+                compressedAndPaddedChunkData, firstChannel.getEncryptionKey(), filePath, iv);
 
         // Compute the md5 of the chunk data
         String md5 = BlobBuilder.computeMD5(encryptedCompressedChunkData, compressedChunkLength);
@@ -494,10 +491,7 @@ class FlushService<T> {
                 .setChunkLength(compressedChunkLength)
                 .setChannelList(result.channelsMetadataList)
                 .setChunkMD5(md5)
-                .setEncryptionKeyId(
-                    bdecVersion == Constants.BdecVersion.THREE
-                        ? NO_ENCRYPTION_KEY_ID
-                        : firstChannel.getEncryptionKeyId())
+                .setEncryptionKeyId(firstChannel.getEncryptionKeyId())
                 .setEpInfo(
                     AbstractRowBuffer.buildEpInfoFromStats(
                         result.rowCount, result.columnEpStatsMapCombined))
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
@@ -62,7 +62,6 @@ import org.apache.arrow.vector.VectorSchemaRoot;
  *     ParquetChunkData})
  */
 class FlushService<T> {
-  private static final long NO_ENCRYPTION_KEY_ID = 0L;
 
   // Static class to save the list of channels that are used to build a blob, which is mainly used
   // to invalidate all the channels when there is a failure
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
@@ -250,22 +250,22 @@ public class ParquetFlusher implements Flusher<ParquetChunkData> {
   private static class BdecWriteSupport extends WriteSupport<List<Object>> {
     MessageType schema;
     RecordConsumer recordConsumer;
-    Map<String, String> extraMetaData;
+    Map<String, String> extraMetadata;
     List<ChannelMetadata> channelsMetadataList;
 
     // TODO SNOW-672156: support specifying encodings and compression
     BdecWriteSupport(
         MessageType schema,
-        Map<String, String> extraMetaData,
+        Map<String, String> extraMetadata,
         List<ChannelMetadata> channelsMetadataList) {
       this.schema = schema;
-      this.extraMetaData = extraMetaData;
+      this.extraMetadata = extraMetadata;
       this.channelsMetadataList = channelsMetadataList;
     }
 
     @Override
     public WriteContext init(Configuration config) {
-      return new WriteContext(schema, extraMetaData);
+      return new WriteContext(schema, extraMetadata);
     }
 
     @Override
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
@@ -11,6 +11,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 import net.snowflake.ingest.utils.ErrorCode;
 import net.snowflake.ingest.utils.Logging;
 import net.snowflake.ingest.utils.SFException;
@@ -103,7 +104,7 @@ public class ParquetFlusher implements Flusher<ParquetChunkData> {
 
     Map<String, String> metadata = channelsDataPerTable.get(0).getVectors().metadata;
 
-    flushToParquetBdecChunk(chunkData, rows, metadata);
+    flushToParquetBdecChunk(chunkData, rows, metadata, channelsMetadataList);
     return new SerializationResult(channelsMetadataList, columnEpStatsMapCombined, rowCount);
   }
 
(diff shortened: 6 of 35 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -0,0 +1,356 @@
+package net.snowflake.ingest.streaming.internal;
+
+import java.util.*;
```
sfc-gh-tjones, line 3: Nit, avoid wildcard imports.

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -0,0 +1,356 @@
...
+    ARRAY(13, true),
+    OBJECT(12, true),
+    VARIANT(11, true),
+    ROW,
+    SEQUENCE,
+    FUNCTION,
+    USER_DEFINED_TYPE,
+    ;
+
+    private final int ordinal;
```
sfc-gh-tjones, line 53: Maybe add a comment explaining what the value of `-1` means

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -0,0 +1,356 @@
...
+    private final boolean object;
+
+    ColumnLogicalType() {
```
sfc-gh-tjones, line 56: Where is this used? Should we add a `ColumnLogicalType` of `INVALID` or `NOT_SET` corresponding to this value if we need it?

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -0,0 +1,356 @@
...
+   *
+   * @return the current buffer size
+   */
+  @Override
+  public float getSize() {
+    return this.bufferSize;
+  }
+
+  /**
+   * Verify that the input row columns are all valid
```
sfc-gh-tjones, line 151: Let's specify what `valid` means in this context. Should we also compare data / logical types here as well?

src/main/java/net/snowflake/ingest/streaming/internal/BlobMetadata.java
```diff
@@ -14,23 +14,23 @@
 class BlobMetadata {
   private final String path;
   private final String md5;
-  private final Constants.BdecVerion bdecVersion;
+  private final Constants.BdecVersion bdecVersion;
```
sfc-gh-tjones, line 17: Nit, for some of these changes it might make sense to pull them out into their own PR that we can quickly merge/review, it'll make reviewing this a bit easier.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -139,8 +139,21 @@ static String validateAndParseObject(Object input) {
    */
   static TimestampWrapper validateAndParseTimestampNtzSb16(
       Object input, Map<String, String> metadata) {
+    int scale = Integer.parseInt(metadata.get(ArrowRowBuffer.COLUMN_SCALE));
+    return validateAndParseTimestampNtzSb16(input, scale);
```
sfc-gh-tjones, line 143: Nit, inline `scale` here?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -207,10 +220,22 @@ private static int getFractionFromTimestamp(SFTimestamp input) {
    *     scale
    */
   static TimestampWrapper validateAndParseTimestampTz(Object input, Map<String, String> metadata) {
+    int scale = Integer.parseInt(metadata.get(ArrowRowBuffer.COLUMN_SCALE));
+    return validateAndParseTimestampTz(input, scale);
```
sfc-gh-tjones, line 224: Nit, inline `scale`?

sfc-gh-tjones (commented):
This is a really awesome PR and a great refactor. One thought I had is what if you open a PR WITHOUT the Parquet specific logic that makes the flush/buffer portions generic + has the existing implementation use them and then follow up with the Parquet implementation? That way we can separate out the addition of the base classes and the existing Arrow generation's use of them vs the newly introduced Parquet code. I don't think that'd be tough to do, basically just remove `ParquetFlusher`, `ParquetRowBuffer`, the `pom.xml` entires, and a few definitions from this PR.

sfc-gh-tzhang:
> This is a really awesome PR and a great refactor. One thought I had is what if you open a PR WITHOUT the Parquet specific logic that makes the flush/buffer portions generic + has the existing implementation use them and then follow up with the Parquet implementation? That way we can separate out the addition of the base classes and the existing Arrow generation's use of them vs the newly introduced Parquet code. I don't think that'd be tough to do, basically just remove `ParquetFlusher`, `ParquetRowBuffer`, the `pom.xml` entires, and a few definitions from this PR.

+1, there're 1855 lines of changes in the PR and most of them are not test related, so it's not very easy to navigate what's changing VS just some class/interface refactor, please consider divide this PR into multiple small PRs if possible, thanks!

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
```diff
@@ -14,24 +14,25 @@
  * during flush. The key is a fully qualified table name and the value is a set of channels that
  * belongs to this table
  */
-class ChannelCache {
+class ChannelCache<T> {
```
sfc-gh-japatel, line 17: This might just be me but I am not in favor of Single Letter generics. (I know this is the java norm and is standard but it would be much more readable if it is not T and something meaningful even though mouthful)

Or may be a single line description of what T is?

sfc-gh-azagrebin: sure, I forgot about the comment here!

sfc-gh-azagrebin:
> divide this PR into multiple small PRs if possible

true, I am on it

sfc-gh-azagrebin:
@sfc-gh-tjones @sfc-gh-tzhang 
I have split the PR into refactoring in #202 and Parquet implementation here.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -436,8 +437,10 @@ BlobMetadata buildAndUpload(String filePath, List<List<ChannelData<T>>> blobData
         // TODO: encryption is not yet supported by server side for Parquet
         long iv = curDataSize / Constants.ENCRYPTION_ALGORITHM_BLOCK_SIZE_BYTES;
         byte[] encryptedCompressedChunkData =
-            Cryptor.encrypt(
-                compressedAndPaddedChunkData, firstChannel.getEncryptionKey(), filePath, iv);
+            bdecVersion == Constants.BdecVersion.THREE
+                ? compressedAndPaddedChunkData
```
sfc-gh-tzhang, line 441: I don't think we can send unencrypted data, have you talk to the our security partners about this?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -0,0 +1,271 @@
+package net.snowflake.ingest.streaming.internal;
```
sfc-gh-tzhang, line 1: There're some comments I made in another PR which would apply here, so things like copyright, no star imports and package private classes, please go through all the newly added files and add them, thanks!

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -0,0 +1,271 @@
...
+          filePath);
+    }
+
+    Map<String, String> metadata = channelsDataPerTable.get(0).getVectors().metadata;
+
+    flushToParquet(chunkData, rows, metadata);
+    return new SerializationResult(channelsMetadataList, columnEpStatsMapCombined, rowCount);
+  }
+
+  private void flushToParquet(
```
sfc-gh-tzhang, line 102: Comments are missing for most of the functions in this file

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -0,0 +1,271 @@
...
+      ByteArrayOutputStream output, List<List<Object>> rows, Map<String, String> metadata)
+      throws IOException {
+    try {
+      ParquetWriter<List<Object>> writer =
+          new BdecParquetWriterBuilder(output, schema, metadata)
+              .withWriterVersion(ParquetProperties.WriterVersion.PARQUET_1_0)
```
sfc-gh-tzhang, line 108: should the WriterVersion and CompressionCodec be constants and put into Constants.java?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -0,0 +1,271 @@
...
+              .withWriteMode(ParquetFileWriter.Mode.CREATE)
+              .build();
+
+      for (List<Object> row : rows) {
+        writer.write(row);
+      }
+      writer.close();
+    } catch (Throwable t) {
+      logger.logError("Parquet Flusher: failed to write", t);
+      throw t;
```
sfc-gh-tzhang, line 121: Why do we need to catch it and log it? Can't we just let it throw?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
```diff
@@ -129,7 +131,10 @@ private RowBuffer<T> createRowBuffer(SnowflakeStreamingIngestClientInternal<T> c
     // can be probably reconsidered
     //noinspection unchecked
     return (RowBuffer<T>)
-        new ArrowRowBuffer((SnowflakeStreamingIngestChannelInternal<VectorSchemaRoot>) this);
+        (client.getParameterProvider().getBlobFormatVersion() == Constants.BdecVersion.THREE
+            ? new ParquetRowBuffer(
+                (SnowflakeStreamingIngestChannelInternal<ParquetRowBuffer.ParquetChunkData>) this)
+            : new ArrowRowBuffer((SnowflakeStreamingIngestChannelInternal<VectorSchemaRoot>) this));
```
sfc-gh-tzhang, line 137: nit: should be using `switch`?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -0,0 +1,271 @@
...
+    @Override
+    public void prepareForWrite(RecordConsumer recordConsumer) {
+      this.recordConsumer = recordConsumer;
+    }
+
+    @Override
+    public void write(List<Object> values) {
+      List<ColumnDescriptor> cols = schema.getColumns();
+      if (values.size() != cols.size()) {
+        throw new ParquetEncodingException(
```
sfc-gh-tzhang, line 218: is this during write to the buffer? If yes, we will need to wrap it with SFException, because that's the only thing we will catch when OnErrorOption=CONTINUE

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -0,0 +1,514 @@
...
+import org.apache.parquet.schema.MessageType;
+import org.apache.parquet.schema.PrimitiveType;
+import org.apache.parquet.schema.Type;
+import org.apache.parquet.schema.Types;
+
+/**
+ * The buffer in the Streaming Ingest channel that holds the un-flushed rows, these rows will be
+ * converted to Parquet format for faster processing
+ */
+public class ParquetRowBuffer extends AbstractRowBuffer<ParquetRowBuffer.ParquetChunkData> {
```
sfc-gh-tzhang, line 34: Same for this file, please add comments are the fields and functions

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -0,0 +1,514 @@
...
+    data = new ArrayList<>();
+    tempData = new ArrayList<>();
+  }
+
+  @Override
+  public void setupSchema(List<ColumnMetadata> columns) {
+    fieldIndex.clear();
+    metadata.clear();
+    List<Type> parquetTypes = new ArrayList<>();
+    int id = 1;
```
sfc-gh-tzhang, line 63: not sure why it starts with 1, comment please

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -0,0 +1,514 @@
...
+    switch (logicalType) {
+      case FIXED:
+        parquetType =
+            getFixedParquetType(
+                column, id, name, physicalType, decimalLogicalTypeAnnotation, repetition);
+        break;
+      case ARRAY:
+      case OBJECT:
+      case VARIANT:
+        this.metadata.put(id + ":obj_enc", "1");
```
sfc-gh-tzhang, line 117: no `parquertType` for this case?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -0,0 +1,514 @@
...
+                        .getTimeInScale();
+                break;
+              default:
+                break;
+            }
+            value = getSb16Bytes(parseSb16(value, stats, column.getScale()));
+            size += 16;
+            break;
+          default:
+            throw new ParquetEncodingException("Unsupported column type: " + typeName);
```
sfc-gh-tzhang, line 374: why we're throwing an encoding exception for non-encoding operations?

sfc-gh-tzhang (changes requested):
Thanks Andrey. I did a quick look through and 90% of code doesn't have any comments, so it's a bit hard to follow what you're trying to do. Please consider adding more comments and there is also no test.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
sfc-gh-gdoci, line 108: Why aren't we using the latest version `WriterVersion.PARQUET_2_0`? I would assume that `1_0` is used server-side and we have to match that.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-gdoci: +1, comments would really help to understand the function better.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -0,0 +1,514 @@
...
+
+    LogicalTypeAnnotation.DecimalLogicalTypeAnnotation decimalLogicalTypeAnnotation =
+        column.getScale() != null && column.getPrecision() != null
+            ? LogicalTypeAnnotation.DecimalLogicalTypeAnnotation.decimalType(
+                column.getScale(), column.getPrecision())
+            : null;
+
+    // Handle differently depends on the column logical and physical types
+    Type.Repetition repetition =
+        column.getNullable() ? Type.Repetition.OPTIONAL : Type.Repetition.REQUIRED;
```
sfc-gh-gdoci, line 107: For my understanding: for what do we need the repetion?

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
sfc-gh-azagrebin: I keep it mostly for readability to see what types cannot be actually packed into the chunk and read by server side.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-azagrebin: it is on the road map

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
sfc-gh-azagrebin: I think I wanted to explicitly log it in case of internal errors.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
sfc-gh-azagrebin: this is during flushing and not really expected.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-azagrebin: parquetType is same as the next one for ANY. CHAR etc
the object types needs this flag for the scanner

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-azagrebin: indeed changed:
```
throw new SFException(ErrorCode.UNKNOWN_DATA_TYPE, logicalType, physicalType);
```

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-tjones: Good catch Toby - we need to ensure that this unencrypted data cannot be queried in Snowflake as supplied from the Client SDK. Let's either follow up with a PR to support encryption (with a JIRA linked here we can track) or add encryption support now. If the latter would dramatically increase the size of the PR I'd be happy filing a JIRA, adding it here, and following up immediately afterward.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -456,7 +459,10 @@ BlobMetadata buildAndUpload(String filePath, List<List<ChannelData<T>>> blobData
                 .setChunkLength(compressedChunkLength)
                 .setChannelList(result.channelsMetadataList)
                 .setChunkMD5(md5)
-                .setEncryptionKeyId(firstChannel.getEncryptionKeyId())
+                .setEncryptionKeyId(
+                    bdecVersion == Constants.BdecVersion.THREE
+                        ? 0
```
sfc-gh-tjones, line 464: We might want to pull this out as a constant of `NO_ENCRYPTION_KEY_ID` so that it is obvious why `0` is being used here.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -0,0 +1,278 @@
...
+
+/**
+ * Converts {@link ChannelData} buffered in {@link RowBuffer} to the Parquet format for faster
+ * processing.
+ */
+public class ParquetFlusher implements Flusher<ParquetRowBuffer.ParquetChunkData> {
+  private static final Logging logger = new Logging(ParquetFlusher.class);
+  private final MessageType schema;
+
+  public ParquetFlusher(MessageType schema) {
```
sfc-gh-tjones, line 41: Nit, docstring for public constructor.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -0,0 +1,278 @@
...
+      String filePath)
+      throws IOException {
+    List<ChannelMetadata> channelsMetadataList = new ArrayList<>();
+    long rowCount = 0L;
+    SnowflakeStreamingIngestChannelInternal<ParquetRowBuffer.ParquetChunkData> firstChannel = null;
+    Map<String, RowBufferStats> columnEpStatsMapCombined = null;
+    List<List<Object>> rows = null;
+
+    for (ChannelData<ParquetRowBuffer.ParquetChunkData> data : channelsDataPerTable) {
+      // Create channel metadata
```
sfc-gh-tjones, line 58: Question about everything in the `for` loop here: a lot (all?) of it looks to be non-Parquet specific. Is this the same in the Arrow flusher? If so, would it be possible to refactor it to make the amount of duplication smaller?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -0,0 +1,278 @@
...
+  }
+
+  private void flushToParquet(
+      ByteArrayOutputStream output, List<List<Object>> rows, Map<String, String> metadata)
+      throws IOException {
+    try {
+      ParquetWriter<List<Object>> writer =
+          new BdecParquetWriterBuilder(output, schema, metadata)
+              .withWriterVersion(ParquetProperties.WriterVersion.PARQUET_1_0)
+              .withDictionaryEncoding(false)
```
sfc-gh-tjones, line 116: Why false here? Why does this need to be explicitly set?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -0,0 +1,278 @@
...
+              .enableValidation()
+              .withCompressionCodec(CompressionCodecName.UNCOMPRESSED)
+              .withWriteMode(ParquetFileWriter.Mode.CREATE)
+              .build();
+
+      for (List<Object> row : rows) {
```
sfc-gh-tjones, line 122: Nit, this can be consolidated to `rows.forEach(writer::write)`

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
sfc-gh-tjones: Is there something above this that catches/logs it?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -0,0 +1,278 @@
...
+        writer.write(row);
+      }
+      writer.close();
+    } catch (Throwable t) {
+      logger.logError("Parquet Flusher: failed to write", t);
+      throw t;
+    }
+  }
+
+  public static class BdecParquetWriterBuilder
```
sfc-gh-tjones, line 132: This might be better segmented in its own file with a proper docstring. Also, why is it public? Isn't it only used here?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -0,0 +1,595 @@
...
+      }
+      this.statsMap.put(column.getName(), new RowBufferStats(column.getCollation()));
+
+      if (this.owningChannel.getOnErrorOption() == OpenChannelRequest.OnErrorOption.ABORT) {
+        this.tempStatsMap.put(column.getName(), new RowBufferStats(column.getCollation()));
+      }
+
+      id++;
+    }
+    schema = new MessageType("bdec", parquetTypes);
```
sfc-gh-tjones, line 92: Should `bdec` be made a static final string somewhere? Also, is `MessageType` ours? Can we have a BDEC specific constructor?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -0,0 +1,595 @@
...
+  /**
+   * Adds a row to the parquet buffer.
+   *
+   * @param row row to add
+   * @param out internal buffer to add to
+   * @param statsMap column stats map
+   * @param inputColumnNames list of input column names after formatting
+   * @return row size
+   */
+  private float addRow(
```
sfc-gh-tjones, line 359: All this is and the data type specific logic below is tested where :) ?

src/main/java/net/snowflake/ingest/utils/Constants.java
```diff
@@ -84,7 +84,13 @@ public enum BdecVersion {
     ONE(1),
 
     /** Uses Arrow to generate BDEC chunks with {@link ArrowBatchWriteMode#FILE}. */
-    TWO(2);
+    TWO(2),
+
+    /**
+     * Uses Parquet to generate BDEC chunks with {@link
+     * net.snowflake.ingest.streaming.internal.ParquetRowBuffer}.
```
sfc-gh-tjones, line 91: Do we plan on extending this version to support encryption or is that going to be `FOUR`? If it is not finalized we should probably document that this is a work in progress, experimental version.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -0,0 +1,595 @@
...
+  }
+
+  public static class ParquetChunkData {
+    final List<List<Object>> rows;
+    final Map<String, String> metadata;
+
+    public ParquetChunkData(List<List<Object>> rows, Map<String, String> metadata) {
+      this.rows = rows;
+      this.metadata = metadata;
+    }
```
sfc-gh-tjones, line 588: Should this be its own class with a corresponding docstring given how frequently it gets referenced elsewhere?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
sfc-gh-azagrebin: I thought about it but this does not look very straight forward for me atm.
There are common pieces but scattered around, I would get back this if we see it growing.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
sfc-gh-azagrebin: I did not get to full understanding of all implications of this
it is switched off for simplicity now to avoid issues in the early version
we can revisit this later and reconsider
I will leave a comment

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
sfc-gh-azagrebin: `writer::write` throws an IO checked ex

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
sfc-gh-azagrebin: Indeed it can be a separate class. My point was that these classes are fairly small and used only here.
we can move them out if everything grows too much or/and we want them to be unit tested.
This one can indeed be private, good point.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-azagrebin: true, I pulled off a constant.
`MessageType` is a Parquet lib class.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-azagrebin: we are working on this in a separate PR to integrate with the test refactoring in #209.
@sfc-gh-gdoci looks into this.

src/main/java/net/snowflake/ingest/utils/Constants.java
sfc-gh-azagrebin: the encryption will be handled in a separate PR.
indeed version 3 is experimental for now, I will add comment.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
sfc-gh-azagrebin: My problem was that I did not see the error cause in the logs, maybe this has changed

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-azagrebin: The encryption is not supported by the server side scanner, there is an internal JIRA subtask for this.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
sfc-gh-azagrebin: PARQUET_2_0 uses `Encoding.DELTA_BYTE_ARRAY` for byte arrays (e.g. SF sb16)
Server side scanner does not support it yet, this may be an easy fix in future

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-azagrebin: Parquet `Type.Repetition` in general supports repeated values for the same row column, like a list of values.
This generator uses only
- either 0 or 1 value for nullable data type (`OPTIONAL`: 0 or none value if it is null)
- or exactly 1 value for non-nullable data type (`REQUIRED`)
Thanks for asking! I added all this to comments

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -0,0 +1,661 @@
...
+    PrimitiveType.PrimitiveTypeName type = getTimePrimitiveType(physicalType);
+    if (physicalType == ColumnPhysicalType.SB16) {
+      LogicalTypeAnnotation typeAnnotation = LogicalTypeAnnotation.decimalType(scale, 38);
+      return Types.primitive(type, repetition).as(typeAnnotation).length(16).id(id).named(name);
+    } else {
+      LogicalTypeAnnotation typeAnnotation =
+          logicalType == ColumnLogicalType.TIME
+              ? LogicalTypeAnnotation.timeType(false, timeUnit)
+              : LogicalTypeAnnotation.timestampType(false, timeUnit);
+      return Types.primitive(type, repetition).as(typeAnnotation).id(id).named(name);
```
sfc-gh-gdoci, line 288: are we missing `.length()`?

sfc-gh-gdoci: I see this is missing in some other places as well. Is the length not supposed to be defined for every `Types.primitive()`?

sfc-gh-azagrebin: This is my understanding, the bit/byte length of primitives is predefined (int64 etc). I think it even throws because length cannot be changed for primitives.

sfc-gh-gdoci: But we have in `L#230`.length(4) for `PrimitiveType.PrimitiveTypeName.INT32` and `L#241` .length(8). The reason I am asking is because I noticed that in tests for time columns the `type length` is 0.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -0,0 +1,661 @@
...
+      RowBufferStats stats = statsMap.get(columnName);
+      Utils.assertNotNull("Arrow column stats", stats);
+      ColumnMetadata column = fieldIndex.get(columnName).getFirst();
+      if (value != null) {
+        ColumnDescriptor columnDescriptor = schema.getColumns().get(colIndex);
+        PrimitiveType.PrimitiveTypeName typeName =
+            columnDescriptor.getPrimitiveType().getPrimitiveTypeName();
+        ColumnLogicalType logicalType = ColumnLogicalType.valueOf(column.getLogicalType());
+        ColumnPhysicalType physicalType = ColumnPhysicalType.valueOf(column.getPhysicalType());
+        switch (typeName) {
```
sfc-gh-gdoci, line 388: I think we are missing `case FLOAT`.

sfc-gh-azagrebin: Do we use FLOAT for something? So far I used only DOUBLE for SF REAL in `getColumnParquetType`.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetValueParser.java
```diff
@@ -0,0 +1,261 @@
...
+   * @param typeName Parquet primitive type name
+   * @param stats column stats to update
+   * @return parsed value and byte size of Parquet internal representation
+   */
+  static ParquetBufferValue parseColumnValueToParquet(
+      Object value,
+      ColumnMetadata columnMetadata,
+      PrimitiveType.PrimitiveTypeName typeName,
+      RowBufferStats stats) {
+    Utils.assertNotNull("Arrow column stats", stats);
```
sfc-gh-gdoci, line 51: "Arrow" -> "Parquet"

sfc-gh-gdoci: We also have a similiar problem in `ingest_eror_messages.properties` #L9 with `0004=The given row cannot be converted to Arrow format: {0}. {1}`

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-gdoci: update: @sfc-gh-azagrebin found out that we can use the encryption w/o file caching in XP (see changes in his last commit).

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-gdoci: I have added tests and merged here.

sfc-gh-tjones (approved):

src/main/java/net/snowflake/ingest/utils/Constants.java
```diff
@@ -83,7 +83,14 @@ public enum BdecVersion {
...
     /** Uses Arrow to generate BDEC chunks with {@link ArrowBatchWriteMode#FILE}. */
-    TWO(2);
+    TWO(2),
+
+    /**
+     * Uses Parquet to generate BDEC chunks with {@link
+     * net.snowflake.ingest.streaming.internal.ParquetRowBuffer}. This version is experimental and
+     * WIP at the moment.
+     */
+    THREE(3);
```
sfc-gh-tzhang, line 93: For Parquet, are you doing the compression at column level or chunk level?

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
```diff
@@ -31,16 +32,29 @@
...
 
 /** Example streaming ingest sdk integration test */
+@RunWith(Parameterized.class)
 public class StreamingIngestIT {
+  @Parameterized.Parameters(name = "{0}")
+  public static Collection<Object[]> bdecVersion() {
+    return Arrays.asList(
+        new Object[][] {
+          {"Arrow", Constants.BdecVersion.ONE},
+          // {"Parquet", Constants.BdecVersion.THREE}
```
sfc-gh-tzhang, line 54: uncomment or add a TODO?

sfc-gh-tzhang (commented):
Thanks @sfc-gh-gdoci and @sfc-gh-azagrebin, I left some comments, mainly about general recommendation, one question i have is that do you think we still need the change to support Arrow column level compression, if no, could we remove them (in another PR)?

sfc-gh-tzhang:
Also we need to check why the Snyk is failing with @sfc-gh-jfan and whether we could avoid of using that dependency, thanks!

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -0,0 +1,320 @@
...
+    try {
+      ParquetWriter<List<Object>> writer =
+          new BdecParquetWriterBuilder(bdecOutput, schema, metadata)
+              // PARQUET_2_0 uses Encoding.DELTA_BYTE_ARRAY for byte arrays (e.g. SF sb16)
+              // server side does not support it
+              // another fix could be to use lower level column writers and custom
+              // ValuesWriterFactory
+              // that uses plain byte array encoding used by PARQUET_1_0 and supported by server
+              // side
+              // TODO: SNOW-620624
```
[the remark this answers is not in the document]
sfc-gh-gdoci: yes, it is done by the auto formatter.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -0,0 +1,320 @@
...
+              // scanner yet
+              .withDictionaryEncoding(false)
+              .enableValidation()
+              .withCompressionCodec(CompressionCodecName.GZIP)
+              .withWriteMode(ParquetFileWriter.Mode.CREATE)
+              .build();
+
+      for (List<Object> row : chunkRows) {
+        writer.write(row);
+      }
```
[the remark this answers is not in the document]
sfc-gh-gdoci: Good point, created https://snowflakecomputing.atlassian.net/browse/SNOW-672142 and added its subtasks in the code todo comments.

src/main/java/net/snowflake/ingest/streaming/internal/BlobMetadata.java
```diff
@@ -14,7 +14,7 @@
 class BlobMetadata {
   private final String path;
   private final String md5;
-  private final Constants.BdecVersion bdecVersion;
+  final Constants.BdecVersion bdecVersion;
```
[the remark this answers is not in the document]
sfc-gh-gdoci: There is a getter function `getVersion()`. I changed the `bdecVersion` to private.

src/main/java/net/snowflake/ingest/streaming/internal/BlobMetadata.java
```diff
@@ -49,9 +49,16 @@ List<ChunkMetadata> getChunks() {
...
-  //  byte getVersionByte() {
-  //    return bdecVersion.toByte();
-  //  }
+  /**
+   * Create {@link BlobMetadata} in case of {@link Constants.BdecVersion#ONE} and {@link
+   * BlobMetadataWithBdecVersion} otherwise to send BDEC version to server side.
+   */
+  static BlobMetadata createBlobMetadata(
+      String path, String md5, Constants.BdecVersion bdecVersion, List<ChunkMetadata> chunks) {
+    // TODO: Unify BlobMetadata with BlobMetadataWithBdecVersion once server side BdecVersion in
```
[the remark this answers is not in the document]
sfc-gh-gdoci: sure, done.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -0,0 +1,221 @@
...
+import org.apache.parquet.schema.Type;
+
+/**
+ * The buffer in the Streaming Ingest channel that holds the un-flushed rows, these rows will be
+ * converted to Parquet format for faster processing
+ */
+public class ParquetRowBuffer extends AbstractRowBuffer<ParquetChunkData> {
+  private static final Logging logger = new Logging(ParquetRowBuffer.class);
+
+  private static final String PARQUET_MESSAGE_TYPE_NAME = "bdec";
```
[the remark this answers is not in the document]
sfc-gh-gdoci: Yes `BLOB_EXTENSION_TYPE`, but the constant  `PARQUET_MESSAGE_TYPE_NAME ` serves a different purpose (although it happens to have the same value as the first one). I would keep `PARQUET_MESSAGE_TYPE_NAME` as well.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -0,0 +1,221 @@
...
+  @Override
+  Object getVectorValueAt(String column, int index) {
+    int colIndex = fieldIndex.get(column).getSecond();
+    Object value = data.get(index).get(colIndex);
+    ColumnMetadata columnMetadata = fieldIndex.get(column).getFirst();
+    String physicalTypeStr = columnMetadata.getPhysicalType();
+    ColumnPhysicalType physicalType = ColumnPhysicalType.valueOf(physicalTypeStr);
+    String logicalTypeStr = columnMetadata.getLogicalType();
+    ColumnLogicalType logicalType = ColumnLogicalType.valueOf(logicalTypeStr);
+    if (logicalType == ColumnLogicalType.FIXED) {
```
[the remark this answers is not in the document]
sfc-gh-gdoci: SB4 and SB8 have direct mapping to parquet types int32 and int64, and don't need to be fixed.
(This method is used only for testing)

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
sfc-gh-gdoci: done

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
sfc-gh-gdoci: do you mean to add `channelName` in the exception message?

sfc-gh-gdoci:
> Thanks @sfc-gh-gdoci and @sfc-gh-azagrebin, I left some comments, mainly about general recommendation, one question i have is that do you think we still need the change to support Arrow column level compression, if no, could we remove them (in another PR)?

Thanks for the review @sfc-gh-tzhang. I addressed almost all of them in the last commit. I think we can remove it and created this [Jira](https://snowflakecomputing.atlassian.net/browse/SNOW-672231) to track that.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
sfc-gh-tzhang: Yes, otherwise we wouldn't know which channel is having issue, right?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -0,0 +1,221 @@
...
+  @Override
+  public void setupSchema(List<ColumnMetadata> columns) {
+    fieldIndex.clear();
+    metadata.clear();
+    List<Type> parquetTypes = new ArrayList<>();
+    // Snowflake column id that corresponds to the order in 'columns' received from server
+    // id is required to pack column metadata for the server scanner, e.g. decimal scale and
+    // precision
+    int id = 1;
+    for (ColumnMetadata column : columns) {
```
sfc-gh-tzhang, line 65: you could do something like `for (int id = 1; id < xx; id++) `instead of declaring id outside

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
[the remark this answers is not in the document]
sfc-gh-tzhang: thanks, would be good to mention it in the comment that this is only used for testing

src/main/java/net/snowflake/ingest/utils/Constants.java
sfc-gh-tzhang: Ah, so we're doing page level instead of column or chunk level? Would be good to mention it in the comment, thanks!

pom.xml
```diff
@@ -367,6 +389,46 @@
...
+                                    <shadedPattern>${shadeBase}.com.thoughtworks.paranamer</shadedPattern>
+                                </relocation>
+                                <relocation>
+                                    <pattern>com.jamesmurty.utils</pattern>
+                                    <shadedPattern>${shadeBase}.com.jamesmurty.utils</shadedPattern>
+                                </relocation>
+                                <relocation>
+                                    <pattern>net.iharder.base64</pattern>
+                                    <shadedPattern>${shadeBase}.net.iharder.base64</shadedPattern>
+                                </relocation>
```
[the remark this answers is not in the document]
sfc-gh-gdoci: As we discussed offline, I removed them from here and moved to `check_contents.sh`.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-gdoci: true, but I find the other `for(column: columns)` a bit more readable.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -0,0 +1,339 @@
...
+  /**
+   * A parquet specific write builder.
+   *
+   * <p>This class is implemented as parquet library API requires, mostly to provide {@link
+   * BdecWriteSupport} implementation.
+   */
+  private static class BdecParquetWriterBuilder
+      extends ParquetWriter.Builder<List<Object>, BdecParquetWriterBuilder> {
+    private final MessageType schema;
+    private final Map<String, String> extraMetaData;
```
sfc-gh-tzhang, line 162: nit 
```suggestion
    private final Map<String, String> extraMetadata;
```

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-tzhang: Yeah, this is probably personal preference, but I feel like having a separate index outside of the for loop seems redundant, and I believe most people will just declare it inline :)

src/main/java/net/snowflake/ingest/streaming/internal/ParquetValueParser.java
```diff
@@ -0,0 +1,334 @@
...
+        // explicitly match the BigDecimal input scale with the Snowflake data type scale
+        bigDecimalValue = bigDecimalValue.setScale(scale, RoundingMode.HALF_UP);
+        checkValueInRange(bigDecimalValue, scale, precision);
+        return bigDecimalValue.unscaledValue();
+      default:
+        throw new SFException(ErrorCode.UNKNOWN_DATA_TYPE, logicalType, physicalType);
+    }
+  }
+
+  private static void checkValueInRange(BigDecimal bigDecimalValue, int scale, int precision) {
```
sfc-gh-tzhang, line 262: I believe we have similar logic for Arrow as well, could you combine them?

sfc-gh-tzhang (approved):
Ship it! Left some minor comments, please make sure the Snyk checks are passing

src/main/java/net/snowflake/ingest/streaming/internal/ParquetValueParser.java
sfc-gh-gdoci: ack.

PR #201, opened 2022-07-26
@snow SNOW-634840 Disable sending blob_version to server side until supported in prod

sfc-gh-azagrebin:


src/main/java/net/snowflake/ingest/streaming/internal/BlobMetadata.java
```diff
@@ -49,8 +49,9 @@ List<ChunkMetadata> getChunks() {
     return this.chunks;
   }
 
-  @JsonProperty("bdec_version")
-  byte getVersionByte() {
-    return bdecVersion.toByte();
-  }
+  // TODO: send the bdec_version once server side supports this in production
```
sfc-gh-tzhang, line 52: could you add the JIRA number?

sfc-gh-tzhang (approved):
LGTM, thanks Andrey!

PR #206, opened 2022-07-31
SNOW-637927: Check max column length for Binary data type

sfc-gh-tzhang:
We need to check the max column length allowed for the binary data type, otherwise the data will be ingested successfully but not queryable.


sfc-gh-tzhang:
bump

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
```diff
@@ -945,7 +945,12 @@ private float convertRowToArrow(
               break;
             }
           case BINARY:
-            byte[] bytes = DataValidationUtil.validateAndParseBinary(value);
+            String maxLengthString = field.getMetadata().get(COLUMN_BYTE_LENGTH);
```
sfc-gh-mnaides, line 948: @sfc-gh-tzhang Can you confirm that the `COLUMN_BYTE_LENGTH` is what we expect here?

sfc-gh-tzhang: @sfc-gh-mnaides i'm not, based on my testing, byteLength and length always give me the same thing. I assume for char we should use length and for binary for we should use byte length?

sfc-gh-mnaides (approved):

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
sfc-gh-tzhang: verified at the server side that length and byte length are the same value so we should we be good

PR #207, opened 2022-08-02
Run streaming ingest integration tests on build

sfc-gh-mnaides:
Enable streaming ingest integration tests in the github build.  I had to split out the test connection used by the snowpipe and the snowpipe streaming tests since they were stepping on each with session state

### DIFF
--- a/src/test/java/net/snowflake/ingest/TestUtils.java
+++ b/src/test/java/net/snowflake/ingest/TestUtils.java
@@ -73,6 +73,8 @@ public class TestUtils {
 
   private static int port = 0;
 
+  // Keep separate test connections for snowpipe and snowpipe streaming so that session state is
+  // isolated
   private static Connection snowpipeConn = null;
 
   private static Connection streamingConn = null;
--- a/src/test/java/net/snowflake/ingest/TestUtils.java
+++ b/src/test/java/net/snowflake/ingest/TestUtils.java
@@ -100,7 +100,7 @@ public class TestUtils {
       warehouse = profile.get(WAREHOUSE).asText();
       host = profile.get(HOST).asText();
       scheme = profile.get(SCHEME).asText();
-      role = Optional.ofNullable(profile.get(ROLE)).map(r -> r.asText()).orElse("ACCOUNTADMIN");
+      role = Optional.ofNullable(profile.get(ROLE)).map(r -> r.asText()).orElse("DEFAULT_ROLE");
       privateKeyPem = profile.get(PRIVATE_KEY).asText();
 
       java.security.Security.addProvider(new BouncyCastleProvider());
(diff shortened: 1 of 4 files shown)


src/test/java/net/snowflake/ingest/TestUtils.java
```diff
@@ -73,7 +73,9 @@ public class TestUtils {
 
   private static int port = 0;
 
-  private static Connection conn = null;
+  private static Connection snowpipeConn = null;
+
+  private static Connection streamingConn = null;
```
sfc-gh-tzhang, line 78: Could you add comments why we need two connections?

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
```diff
@@ -642,7 +627,7 @@ public void testMultiThread() throws Exception {
         .execute(
             String.format("create or replace table %s (numcol NUMBER(10,2));", multiThreadTable));
     int numThreads = 20;
-    int numRows = 1000000;
+    int numRows = 10000;
```
sfc-gh-tzhang, line 630: any issue with the old value?

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
```diff
@@ -65,27 +66,16 @@ public void beforeAll() throws Exception {
...
-        .createStatement()
-        .execute(
-            String.format(
-                "alter database %s set ENABLE_PR_37692_MULTI_FORMAT_SCANSET=true;", TEST_DB));
-    jdbcConnection
-        .createStatement()
-        .execute(
-            String.format(
-                "alter table %s set ENABLE_PR_37692_MULTI_FORMAT_SCANSET=true;", TEST_TABLE));
-    jdbcConnection.createStatement().execute("alter session set ENABLE_UNIFIED_TABLE_SCAN=true;");
```
sfc-gh-tzhang, line 81: I assume the reason for removing these are because the default are already set to the correct value?

sfc-gh-tzhang (approved):
Thanks Matt! Left some comments/questions

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
sfc-gh-mnaides: It was excessive and caused the tests to run longer than necessary

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
sfc-gh-mnaides: The default is set correctly and we can't actually set these parameters from the github test account

sfc-gh-japatel (approved):
🚀 Thanks Matt!

PR #202, opened 2022-07-27
@snow SNOW-633432 Refactor RowBuffer and Flusher API

sfc-gh-azagrebin:
Now our client supports only flushing chunk in Arrow format.
This PR refactors buffering and flushing code to support plugins for other file formats.
It is a prerequisite for introducing support of Parquet file format in #200.

There are 2 major places that are made pluggable to support Parquet along with Arrow:
- `ArrowRowBuffer` is refactored to an abstract format-agnostic interface `RowBuffer`, common things are pulled into `AbstractRowBuffer`
- `FlushService` uses the `RowBuffer` to create an abstract format-agnostic interface `Flusher` to flush the rows into a chunk

`ChannelData` has been generified to hold either Arrow vectors or Parquet buffers in #200.
This triggered a lot of generification of other classes to avoid unsafe casts, better ideas are welcome.

@test tests are refactored in a separate PR #209.

PR #209, opened 2022-08-10
@snow SNOW-633432 Refactor tests to support various BDEC format versions

sfc-gh-azagrebin:
This is a follow up for #202 to refactor tests to run with various BDEC format versions.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
@@ -17,6 +17,7 @@ import java.util.stream.Collectors;
 import net.snowflake.ingest.streaming.InsertValidationResponse;
 import net.snowflake.ingest.streaming.OpenChannelRequest;
 import net.snowflake.ingest.streaming.SnowflakeStreamingIngestChannel;
+import net.snowflake.ingest.utils.Constants;
 import net.snowflake.ingest.utils.ErrorCode;
 import net.snowflake.ingest.utils.Logging;
 import net.snowflake.ingest.utils.SFException;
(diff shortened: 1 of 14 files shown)


sfc-gh-tjones (approved):

src/test/java/net/snowflake/ingest/streaming/internal/RowBufferTest.java
```diff
@@ -767,23 +407,22 @@ private void testFlushHelper(ArrowRowBuffer rowBuffer) {
     Assert.assertFalse(response.hasErrors());
     float bufferSize = rowBuffer.getSize();
 
-    ChannelData<VectorSchemaRoot> data = rowBuffer.flush();
+    ChannelData<?> data = rowBuffer.flush();
     Assert.assertEquals(2, data.getRowCount());
     Assert.assertEquals((Long) 1L, data.getRowSequencer());
-    Assert.assertEquals(7, data.getVectors().getFieldVectors().size());
+    // Assert.assertEquals(7, data.getVectors().getFieldVectors().size());
```
sfc-gh-tzhang, line 413: ?

src/test/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelTest.java
```diff
@@ -49,9 +49,9 @@ public void testChannelFactoryNullFields() {
     String dbName = "DATABASE";
     String schemaName = "SCHEMA";
     String tableName = "TABLE";
-    Long channelSequencer = 0L;
-    Long rowSequencer = 0L;
-    SnowflakeStreamingIngestClientInternal<VectorSchemaRoot> client =
+    long channelSequencer = 0L;
+    long rowSequencer = 0L;
```
sfc-gh-tzhang, line 53: why we're changing this?

src/test/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientTest.java
```diff
@@ -50,13 +50,8 @@
...
 import net.snowflake.ingest.streaming.SnowflakeStreamingIngestClient;
 import net.snowflake.ingest.streaming.SnowflakeStreamingIngestClientFactory;
-import net.snowflake.ingest.utils.ErrorCode;
-import net.snowflake.ingest.utils.Pair;
-import net.snowflake.ingest.utils.ParameterProvider;
-import net.snowflake.ingest.utils.SFException;
-import net.snowflake.ingest.utils.SnowflakeURL;
-import net.snowflake.ingest.utils.Utils;
-import org.apache.arrow.vector.VectorSchemaRoot;
+import net.snowflake.ingest.utils.*;
```
sfc-gh-tzhang, line 53: seems like you're still using *, do you want to update your intelliJ settings?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
```diff
@@ -97,6 +95,43 @@
...
+      String tableName,
+      String offsetToken,
+      Long channelSequencer,
+      Long rowSequencer,
+      SnowflakeStreamingIngestClientInternal<T> client,
+      String encryptionKey,
+      Long encryptionKeyId,
+      OpenChannelRequest.OnErrorOption onErrorOption,
+      Constants.BdecVersion bdecVersion,
+      BufferAllocator allocator) {
```
sfc-gh-tzhang, line 134: could you explain why we need a new constructor? I don't see we pass null allocator, do we?

sfc-gh-tzhang (approved):
Left some comments, otherwise LGTM! And I want to make sure this is just a refactor, we're not removing any coverage, are we?

src/test/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelTest.java
sfc-gh-azagrebin: Do you mean `Long -> long`? My motivation was that there is no need to box non-null primitives.
I can roll it back if needed

src/test/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientTest.java
sfc-gh-azagrebin: sorry leftover, I changed, also enabled warning on this.

src/test/java/net/snowflake/ingest/streaming/internal/ArrowBufferTest.java
```diff
@@ -0,0 +1,523 @@
...
+    Assert.assertEquals(
+        result.getChildren().get(0).getFieldType().getType(), Types.MinorType.BIGINT.getType());
+    Assert.assertEquals(
+        result.getChildren().get(1).getFieldType().getType(), Types.MinorType.INT.getType());
+    Assert.assertEquals(
+        result.getChildren().get(2).getFieldType().getType(), Types.MinorType.INT.getType());
+  }
+
+  @Test
+  public void buildFieldTimestampDate() {
```
sfc-gh-gdoci, line 347: nit: rename to `buildFieldDate`

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
sfc-gh-azagrebin: True we do not need three constructors. I also wanted to avoid a lot of logic in constructors and move it more into builder/factory. I removed the default constructor and made this default. In future we can consider a test builder for it.
This one is actually needed to configure the BDEC version directly in tests.
Less logic in constructors usually makes them easier to unit test and leads to less implicit dependencies that are moved to named factories.
I removed null allocator, it was used once.

src/test/java/net/snowflake/ingest/streaming/internal/RowBufferTest.java
sfc-gh-azagrebin: thanks, leftover, I moved it to `ArrowBufferTest#testFieldNumberAfterFlush()`

sfc-gh-tjones (approved):

PR #211, opened 2022-08-19
SNOW-644956: fix string data type  in ARRAY

sfc-gh-tzhang:
(Copied from JIRA) When an array (or list) of string is passed into the Ingest SDK, the string doesn’t have double quotes around it when the array is dumped into a string, leading to a failed query later on.

Desired behavior:

[“abc”] → “[“abc”]“

[[“abc”], “de”] → “[[“abc”],”de”]“

Current behavior:

[“abc”] → “[abc]“

[[“abc”], “de”] → “[[abc],de]“

The reason is that the toString method is called to make element of the array into string, and the method will not add quotes around the string as we intend. There could also be issues with other data types whose toString method is not implemented to match our interface.


src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -70,36 +69,22 @@ static String validateAndParseArray(Object input) {
...
     } catch (Exception e) {
       throw new SFException(
-          e, ErrorCode.INVALID_ROW, input, "Input column can't be convert to String.");
+          e,
+          ErrorCode.INVALID_ROW,
+          input.toString(),
+          "Input column can't be convert to a valid string");
+    }
+
+    if (output.length() > MAX_STRING_LENGTH) {
```
sfc-gh-japatel, line 82: Can you add a comment here describing we have this limitation?

sfc-gh-japatel (approved):
lgtm

PR #212, opened 2022-08-23
SNOW-649753: Update InsertError to include extra/missing column info for KC schema evolution

sfc-gh-tzhang:
Update InsertError to include extra/missing column information for KC schema evolution


sfc-gh-zefan (approved):
LGTM! Thanks for the prompt update!

PR #214, opened 2022-08-25
@snow SNOW-647377 Fix quoted columns scan and with inner space

sfc-gh-azagrebin:
This fixes the problem that ingested quoted columns are not queryable by server side.
The problem is that the server side expects column name in Arrow file metadata without quotes and space escape characters.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
@@ -19,7 +19,6 @@ import net.snowflake.ingest.streaming.OpenChannelRequest;
 import net.snowflake.ingest.utils.ErrorCode;
 import net.snowflake.ingest.utils.Logging;
 import net.snowflake.ingest.utils.SFException;
-import net.snowflake.ingest.utils.Utils;
 import org.apache.arrow.util.VisibleForTesting;
 
 /**
@@ -198,7 +197,7 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
       Map<String, Object> row, InsertValidationResponse.InsertError error) {
     Map<String, String> inputColNamesMap =
         row.keySet().stream()
-            .collect(Collectors.toMap(AbstractRowBuffer::formatColumnName, value -> value));
+            .collect(Collectors.toMap(LiteralQuoteUtils::formatColumnName, value -> value));
 
     // Check for extra columns in the row
     List<String> extraCols = new ArrayList<>();
@@ -436,14 +435,6 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
 
   @VisibleForTesting
   abstract int getTempRowCount();
-
-  /** Normalize the column name, given with the inserted row, to send to server side. */
-  static String formatColumnName(String columnName) {
-    Utils.assertStringNotNullOrEmpty("invalid column name", columnName);
-    return (columnName.charAt(0) == '"' && columnName.charAt(columnName.length() - 1) == '"')
-        ? columnName
-        : columnName.toUpperCase();
-  }
 
   /**
    * Given a set of col names to stats, build the right ep info
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
@@ -339,7 +339,8 @@ class ArrowRowBuffer extends AbstractRowBuffer<VectorSchemaRoot> {
 
     // Create the corresponding column field base on the column data type
     fieldType = new FieldType(column.getNullable(), arrowType, null, metadata);
-    return new Field(column.getName(), fieldType, children);
+    String columnName = LiteralQuoteUtils.unquoteColumnName(column.getName());
+    return new Field(columnName, fieldType, children);
   }
 
   @Override
@@ -443,7 +444,7 @@ class ArrowRowBuffer extends AbstractRowBuffer<VectorSchemaRoot> {
     float rowBufferSize = 0F;
     for (Map.Entry<String, Object> entry : row.entrySet()) {
       rowBufferSize += 0.125; // 1/8 for null value bitmap
-      String columnName = formatColumnName(entry.getKey());
+      String columnName = LiteralQuoteUtils.formatColumnName(entry.getKey());
       Object value = entry.getValue();
       Field field = this.fields.get(columnName);
       Utils.assertNotNull("Arrow column field", field);
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
@@ -106,7 +106,7 @@ class ArrowRowBuffer extends AbstractRowBuffer<VectorSchemaRoot> {
       Field field = buildField(column);
       FieldVector vector = field.createVector(this.allocator);
       if (!field.isNullable()) {
-        addNonNullableFieldName(field.getName());
+        addNonNullableFieldName(column.getName());
       }
       this.fields.put(column.getName(), field);
       vectors.add(vector);
--- a/src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
+++ b/src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
@@ -94,6 +94,88 @@ public class StreamingIngestIT {
   }
 
   @Test
+  public void testColumnNameQuotes() throws Exception {
+    final String tableName = "T_COLUMN_WITH_QUOTES";
+    final String unquotedColumn = "c1"; // user name <C1>
+    final String quotedColumn = "\"c 2\""; // user name <c 2>
+    final String doubleQuotedColumn = "\"c\"\"a\\ \"\"\\ 3\""; // user name <c"a\ "\ 3>
+    final String escapedSpacesColumn1 =
+        "c\\ 4"; // space is escaped with '\' in Snowflake SQL, user name <C 4>
+    final String escapedSpacesColumn2 =
+        "c\\ 5"; // space is escaped with '\' in Snowflake SQL, user name <C 5>
+    jdbcConnection
+        .createStatement()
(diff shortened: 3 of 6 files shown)


sfc-gh-tzhang (commented):
Instead of copying the `unquote` logic from server side, could we do it at server side to avoid code duplication and logic dependency?

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
```diff
@@ -87,6 +87,70 @@ public void afterAll() throws Exception {
     jdbcConnection.createStatement().execute(String.format("drop database %s", TEST_DB));
   }
 
+  @Test
+  public void testColumnNameQuotes() throws Exception {
+    final String tableName = "T_COLUMN_WITH_QUOTES";
+    final String quotedColumn = "\"c2\"";
+    final String doubleQuotedColumn = "\"c\"\"a\"\"3\"";
+    final String escapedSpacesColumn = "c\\ 4"; // space is escaped with '\' in Snowflake SQL
```
sfc-gh-tzhang, line 95: Not sure if I understand this case, I tried it in the worksheet, so something like `create or replace table t_streamingingest("c\\ 1" int);` , the column name is `c\\ 1`

sfc-gh-azagrebin: One slash here is to escape it in java, the actual string is 'c\ 1'. As I understood it is to escape space for SF.

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
```diff
@@ -618,14 +618,69 @@ Field buildField(ColumnMetadata column) {
...
+    String columnName = unquote(column.getName());
+    return new Field(columnName, fieldType, children);
   }
 
   private String formatColumnName(String columnName) {
     Utils.assertStringNotNullOrEmpty("invalid column name", columnName);
-    return (columnName.charAt(0) == '"' && columnName.charAt(columnName.length() - 1) == '"')
-        ? columnName
-        : columnName.toUpperCase();
+    if (columnName.charAt(0) == '"' && columnName.charAt(columnName.length() - 1) == '"') {
```
sfc-gh-tjones, line 627: What about `"""`?

sfc-gh-azagrebin: hmm I cannot find any special case for `"""` in server side code and [docs](https://docs.snowflake.com/en/sql-reference/identifiers-syntax.html). This will handled in `unquote()` depending on where it is as quoted or double-quote escape.
could it be that you are talking about const string expressions?

sfc-gh-azagrebin:
> Instead of copying the `unquote` logic from server side, could we do it at server side to avoid code duplication and logic dependency?

This PR fixes arrow file internals that are open in scanner, scanner does not have this code as far as I can see.

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
sfc-gh-tzhang: wow, learn something new today, thanks! Do you want to add a test case with double quotes and space?

sfc-gh-tzhang:
> > Instead of copying the `unquote` logic from server side, could we do it at server side to avoid code duplication and logic dependency?
> 
> This PR fixes arrow file internals that are open in scanner, scanner does not have this code as far as I can see.

Then we should add it in the scanner or is there anything you could reuse in XP? If no, then we need to store different thing in different places, because if you look at `validateAndProcessColumnEps`, we will get the column name and then call `unquote` on it.

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
```diff
@@ -618,14 +618,69 @@ Field buildField(ColumnMetadata column) {
...
+      return columnName;
+    } else if (columnName.contains(" ")) {
+      // Server side metadata gets appended by quotes in case of inner spaces
+      columnName = String.format("\"%s\"", columnName);
```
sfc-gh-tzhang, line 631: This won't work if you have a double quote + space, right?

sfc-gh-tzhang (dismissed):
LGTM if you could make sure it works for double quotes + space and adds a test for it, thanks!

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
sfc-gh-azagrebin: hmm I am not sure I understand why the space is a special case for double quotes?
do you mean if user already passed the column name in quotes and used double quotes to escape the internal user quotes in the name: `doubleQuotedColumn` but with spaces inside? like:
```
"\"c\"\"a \"\" 3\"" // space right after 'a' and right before '3'
```
I can change the test case be like this and the test passes.
It does not look special, the spaces will just stay and they do not need to be escaped.
If a space is escaped in the quoted string with '\', the '\' will just stay in the string.
The double quotes is a valid escape only in an originally already quoted string iiuc.

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -440,9 +440,63 @@ void reset() {
   /** Normalize the column name, given with the inserted row, to send to server side. */
   static String formatColumnName(String columnName) {
     Utils.assertStringNotNullOrEmpty("invalid column name", columnName);
-    return (columnName.charAt(0) == '"' && columnName.charAt(columnName.length() - 1) == '"')
-        ? columnName
-        : columnName.toUpperCase();
+    if (columnName.charAt(0) == '"' && columnName.charAt(columnName.length() - 1) == '"') {
```
sfc-gh-tzhang, line 443: Similar to `unquote`, Do you think we have any existing function at server side for `formatColumnName` that could be reused here?

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
```diff
@@ -93,6 +93,71 @@ public void afterAll() throws Exception {
...
+
+    for (int i = 1; i < 15; i++) {
+      if (channel1.getLatestCommittedOffsetToken() != null
+          && channel1.getLatestCommittedOffsetToken().equals("9")) {
+        ResultSet result =
+            jdbcConnection
+                .createStatement()
+                .executeQuery(
+                    String.format(
+                        "select * from %s.%s.%s order by c1", testDb, TEST_SCHEMA, tableName));
```
sfc-gh-tzhang, line 146: could we use column name instead of * to make sure we can select each individual column?

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
```diff
@@ -93,6 +93,71 @@ public void afterAll() throws Exception {
...
+            .build();
+
+    // Open a streaming ingest channel from the given client
+    SnowflakeStreamingIngestChannel channel1 = client.openChannel(request1);
+    for (int val = 0; val < 10; val++) {
+      Map<String, Object> row = new HashMap<>();
+      row.put("c1", val);
+      row.put(quotedColumn, val);
+      row.put(doubleQuotedColumn, val);
+      row.put("c 4", val);
```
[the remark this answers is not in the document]
sfc-gh-azagrebin: we can but the escape '\' is added because quoting is optional in UI
Java strings are always quoted, and this is what comes from server side in this case
I can add handling of escaping in this case but w/o escape it works out of the box

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
sfc-gh-azagrebin: TBH I think the problem is that server side does not return 'unquoted' literal that is stored internally.
Otherwise, the bellow `unquoteColumnName` already reflects the UI column identifier [doc](https://docs.snowflake.com/en/sql-reference/identifiers-syntax.html).
Except for the `QUOTED_IDENTIFIERS_IGNORE_CASE ` parameter that we can also add to channel builder if needed or send from server if it is not per session.

I think we should send the 'unquoted' literal that is stored internally long term in the open channel response
then we do not need this `formatColumnName` at all, just always apply `unquoteColumnName` to user input.

Now we can solve this by applying `unquoteColumnName` to column name in metadata, sent by server. The normalised literal, that is stored internally, should be the same at the end.

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
```diff
@@ -93,6 +93,88 @@ public void afterAll() throws Exception {
...
+    final String escapedSpacesColumn1 =
+        "c\\ 4"; // space is escaped with '\' in Snowflake SQL, user name <C 4>
+    final String escapedSpacesColumn2 =
+        "c\\ 5"; // space is escaped with '\' in Snowflake SQL, user name <C 5>
+    jdbcConnection
+        .createStatement()
+        .execute(
+            String.format(
+                "create or replace table %s.%s.%s (%s NUMBER(38,0), %s NUMBER(38,0), %s"
+                    + " NUMBER(38,0), %s NUMBER(38,0), AGGREGATE NUMBER(38,0));",
```
sfc-gh-tzhang, line 111: this needs to be reverted back :)

sfc-gh-tzhang (dismissed):
Thanks Andrey! In general I'm not a big fan of exposing more and more server side logic to the client because things will fail if they're not in sync anymore (like we added a new keyword at server side), but given that you promise to do it in a better way later, I won't block you from merging this. 

In fact, I'm not sure if you need to concern about backward compatible because column names with double quotes doesn't work anyway currently, and `.displayName` and `.name` will return the same thing for unquoted names so I would suggest to just go for the better way instead of doing this later :)

sfc-gh-tzhang (approved):

sfc-gh-tjones (approved):

sfc-gh-tjones (dismissed):

sfc-gh-azagrebin:
Closed because subsumed by #293.

PR #224, opened 2022-09-09
Add Parquet Tests for the parquet-file-generator

sfc-gh-gdoci:
This PR is based on Andrey's Parquet generator in #200. It
- refactors `ParquetRowBuffer.java`
- adds `ParquetTypeGeneratorTest.java`
- adds `ParquetValueParserTest.java`
- refactors `ArrowBufferTest.java`

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
@@ -10,7 +10,6 @@ import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -18,16 +17,12 @@ import java.util.Set;
 import net.snowflake.client.jdbc.internal.google.common.collect.Sets;
 import net.snowflake.ingest.streaming.OpenChannelRequest;
 import net.snowflake.ingest.utils.Constants;
-import net.snowflake.ingest.utils.ErrorCode;
 import net.snowflake.ingest.utils.Logging;
 import net.snowflake.ingest.utils.Pair;
-import net.snowflake.ingest.utils.SFException;
 import org.apache.parquet.column.ColumnDescriptor;
-import org.apache.parquet.schema.LogicalTypeAnnotation;
 import org.apache.parquet.schema.MessageType;
 import org.apache.parquet.schema.PrimitiveType;
 import org.apache.parquet.schema.Type;
-import org.apache.parquet.schema.Types;
 
 /**
  * The buffer in the Streaming Ingest channel that holds the un-flushed rows, these rows will be
@@ -36,10 +31,6 @@ import org.apache.parquet.schema.Types;
 public class ParquetRowBuffer extends AbstractRowBuffer<ParquetChunkData> {
   private static final Logging logger = new Logging(ParquetRowBuffer.class);
 
-  private static final Set<ColumnPhysicalType> TIME_SUPPORTED_PHYSICAL_TYPES =
-      new HashSet<>(Arrays.asList(ColumnPhysicalType.SB4, ColumnPhysicalType.SB8));
-  private static final Set<ColumnPhysicalType> TIMESTAMP_SUPPORTED_PHYSICAL_TYPES =
-      new HashSet<>(Arrays.asList(ColumnPhysicalType.SB8, ColumnPhysicalType.SB16));
   private static final String PARQUET_MESSAGE_TYPE_NAME = "bdec";
 
   private final Map<String, Pair<ColumnMetadata, Integer>> fieldIndex;
@@ -72,8 +63,10 @@ public class ParquetRowBuffer extends AbstractRowBuffer<ParquetChunkData> {
     // precision
     int id = 1;
     for (ColumnMetadata column : columns) {
-      Type type = getColumnParquetType(column, id);
-      parquetTypes.add(type);
+      ParquetTypeGenerator.ParquetTypeInfo typeInfo =
+          ParquetTypeGenerator.generateColumnParquetTypeInfo(column, id);
+      parquetTypes.add(typeInfo.getParquetType());
+      this.metadata.putAll(typeInfo.getMetadata());
       fieldIndex.put(column.getName(), new Pair<>(column, parquetTypes.size() - 1));
       if (!column.getNullable()) {
         addNonNullableFieldName(column.getName());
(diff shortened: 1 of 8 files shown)


src/test/java/net/snowflake/ingest/streaming/internal/ParquetBufferTest.java
```diff
@@ -0,0 +1,278 @@
...
+        result.getLogicalTypeAnnotation());
+    Assert.assertEquals(Type.Repetition.REQUIRED, result.getRepetition());
+  }
+
+  @Test
+  public void buildFieldFixedSB4() {
+    ColumnMetadata testCol = RowBufferTest.generateCol("SB4", "FIXED", true);
+    Type result = this.rowBufferOnErrorContinue.getColumnParquetType(testCol, 0);
+
+    Assert.assertEquals("testCol", result.getName());
```
sfc-gh-azagrebin, line 71: We also implicitly pack physical/logical types into `metadata` in `getColumnParquetType`.
This needs to be tested as well. If you consider a refactoring into a `ParquetTypeGenerator`,
It could return `type` + `metadata` something to test.

src/test/java/net/snowflake/ingest/streaming/internal/ParquetFlushServiceTest.java
```diff
@@ -0,0 +1,203 @@
...
+        .setSchemaName("schema1")
+        .setTableName("table1")
+        .setOffsetToken("offset2")
+        .setChannelSequencer(10L)
+        .setRowSequencer(100L)
+        .buildAndAdd();
+  }
+
+  @Test
+  public void testBuildAndUploadMoreDataTypes() throws Exception {
```
sfc-gh-azagrebin, line 52: What is Parquet specific in this test that is not covered by `FlushServiceTest`?
if there is something that is not covered in `FlushServiceTest` but it is not Parquet specific, should it rather be in `FlushServiceTest`?
If it just tests various data types at once then I am not sure it adds much to the coverage as we already have a similar test in IT that is fast enough.

src/test/java/net/snowflake/ingest/streaming/internal/ParquetBufferTest.java
```diff
@@ -0,0 +1,278 @@
...
+  public void setupRowBuffer() {
+    this.rowBufferOnErrorContinue = createTestBuffer(OpenChannelRequest.OnErrorOption.CONTINUE);
+    this.rowBufferOnErrorContinue.setupSchema(RowBufferTest.createSchema());
+  }
+
+  ParquetRowBuffer createTestBuffer(OpenChannelRequest.OnErrorOption onErrorOption) {
+    SnowflakeStreamingIngestClientInternal<ParquetChunkData> client =
+        new SnowflakeStreamingIngestClientInternal<>("client");
+    SnowflakeStreamingIngestChannelInternal<ParquetChunkData> channel =
+        new SnowflakeStreamingIngestChannelInternal<>(
```
sfc-gh-azagrebin, line 25: we should consider a test builder for `SnowflakeStreamingIngestChannelInternal` as well at some point :)

src/test/java/net/snowflake/ingest/streaming/internal/ParquetBufferTest.java
```diff
@@ -0,0 +1,278 @@
...
+    Assert.assertEquals(Type.Repetition.OPTIONAL, result.getRepetition());
+  }
+
+  @Ignore
+  @Test
+  public void buildFieldTimeSB8() {
+
+    ColumnMetadata testCol = RowBufferTest.generateCol("SB8", "TIME", true);
+    Type result = this.rowBufferOnErrorContinue.getColumnParquetType(testCol, 0);
+    // -> exception java.lang.IllegalStateException: TIME(MILLIS,false) can only annotate INT32
```
sfc-gh-azagrebin, line 234: We should check if this time annotation is necessary for correct scanning.
If it is not then we should not generate it in this case.
I think it is related to some Parquet limitations for data types, e.g. for Iceberg,
but that is not relevant to pure Snowflake tables.
I guess `MILLIS` (SF scale <=3) have to be always sb4 in Parquet but our scanner does not need this annotation. The unit is derived from scale, even if it is int64.

sfc-gh-azagrebin (commented):
Thanks Gloria, great! I left some comments.

src/test/java/net/snowflake/ingest/streaming/internal/ParquetFlushServiceTest.java
sfc-gh-gdoci: It tests various data types at once. You are right, I checked the coverage of both this test and `StreamingIngestIT`. Since this test doesn't add any coverage, I will remove it.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -100,7 +100,7 @@ public void setupSchema(List<ColumnMetadata> columns) {
    * @param column column metadata
    * @return column parquet type
    */
-  private Type getColumnParquetType(ColumnMetadata column, int id) {
+  protected Type getColumnParquetType(ColumnMetadata column, int id) {
```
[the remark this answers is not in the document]
sfc-gh-gdoci: Ok, I changed it to 'package-private' access.

src/test/java/net/snowflake/ingest/streaming/internal/ArrowBufferTest.java
```diff
@@ -58,15 +58,7 @@ public void testFieldNumberAfterFlush() {
...
-    ColumnMetadata testCol = new ColumnMetadata();
-    testCol.setName("testCol");
-    testCol.setPhysicalType("SB1");
-    testCol.setNullable(true);
-    testCol.setLogicalType("FIXED");
-    testCol.setByteLength(14);
-    testCol.setLength(11);
-    testCol.setScale(0);
-    testCol.setPrecision(4);
+    ColumnMetadata testCol = RowBufferTest.generateCol("SB1", "FIXED", true);
```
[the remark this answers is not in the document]
sfc-gh-gdoci: Makes sense, thanks!

src/test/java/net/snowflake/ingest/streaming/internal/ParquetBufferTest.java
sfc-gh-gdoci: done.

src/test/java/net/snowflake/ingest/streaming/internal/ParquetFlushServiceTest.java
sfc-gh-azagrebin: btw I just noticed we do not actually check the values returned from the select query in IT
it would be nice to add such a check at some point

src/test/java/net/snowflake/ingest/streaming/internal/ArrowBufferTest.java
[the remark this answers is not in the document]
sfc-gh-azagrebin: Looks good! I suggest to move to test code in a separate file if it is used only for tests

src/test/java/net/snowflake/ingest/streaming/internal/ParquetBufferTest.java
sfc-gh-azagrebin: Could you check this?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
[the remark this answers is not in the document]
sfc-gh-gdoci: Since it touches `metadata` member variable, I would leave the `getColumnParquetType` (renamed to `generateColumnparquetType`) in `ParquetRowBuffer.java`.

src/test/java/net/snowflake/ingest/streaming/internal/ParquetFlushServiceTest.java
sfc-gh-gdoci: I added a table count verification `select count(*)` and `select * where ` for the `StreamingIngestIT.testDataTypes` . But the select * fails with `net.snowflake.client.jdbc.SnowflakeSQLException: Time '19:40:55.-226005504' is not recognized`

src/test/java/net/snowflake/ingest/streaming/internal/ParquetBufferTest.java
sfc-gh-gdoci: Where in XP should I look for this information?

sfc-gh-gdoci: Is this what you meant https://github.com/snowflakedb/snowflake-ingest-java/blob/d3e91a01fe8e76a95269cd29bb73c3e51b73dab2/src/main/java/net/snowflake/ingest/streaming/internal/ParquetTypeGenerator.java#L239?

sfc-gh-azagrebin: Yes, looks so, you can check in debugger where it fails.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetTypeGenerator.java
```diff
@@ -0,0 +1,290 @@
...
+
+import java.util.*;
+import net.snowflake.ingest.utils.ErrorCode;
+import net.snowflake.ingest.utils.SFException;
+import org.apache.parquet.schema.LogicalTypeAnnotation;
+import org.apache.parquet.schema.PrimitiveType;
+import org.apache.parquet.schema.Type;
+import org.apache.parquet.schema.Types;
+
+public class ParquetTypeGenerator {
```
sfc-gh-azagrebin, line 11: it looks the team code style in the client is to leave java doc comments for all class methods and fields :)

src/test/java/net/snowflake/ingest/streaming/internal/ParquetBufferTest.java
sfc-gh-gdoci: The snippet above includes my change.
It was failing in `LogicalTypeAnnotation.timeType(false, timeUnit)`.
So I added ` (physicalType == AbstractRowBuffer.ColumnPhysicalType.SB8
                  ? LogicalTypeAnnotation.decimalType(scale, 18)`.

src/test/java/net/snowflake/ingest/streaming/internal/ParquetValueParserTest.java
```diff
@@ -0,0 +1,136 @@
...
+    Assert.assertEquals(BigInteger.valueOf(123), rowBufferStats.getCurrentMinIntValue());
+    Assert.assertEquals(BigInteger.valueOf(123), rowBufferStats.getCurrentMaxIntValue());
+  }
+
+  @Test
+  public void parseValueINT64() {
+    ColumnMetadata testCol =
+        ColumnMetadataBuilder.newBuilder()
+            .physicalType("SB8")
+            .logicalType("FIXED")
```
sfc-gh-azagrebin, line 36: Many parquet types are used to pack different SF logical types,
e.g. for INT64: FIXED, TIME, TIMESTAMP_*
BINARY is also used for OBJECT and VARIANT/ARRAY
some of those logical types are very different and each deserves its own test

sfc-gh-azagrebin (commented):
Thanks Gloria, added some more comments

src/test/java/net/snowflake/ingest/streaming/internal/ParquetFlushServiceTest.java
sfc-gh-azagrebin: Indeed, the test data is generated incorrectly, actually for all time data types in `getRandomRow`:
```
(14 * 60 * 60 + 26 * 60 + 34) * 1000000000 + 437582643
```
but should be:
```
(14 * 60 * 60 + 26 * 60 + 34) + 0.437582643
```
or the same but a string representation "123134.2343" for `BigDecimals`
maybe a good idea to not make it random and assert expected query result

src/test/java/net/snowflake/ingest/streaming/internal/ParquetBufferTest.java
sfc-gh-azagrebin: ah I see
I would just use always `LogicalTypeAnnotation.decimalType` for now and remove all `LogicalTypeAnnotation.time*` because they look faulty and not needed for scanning. The other time units could be applied for sb8 but this logic will complicate things for now.

sfc-gh-gdoci: ok. Should I use always `LogicalTypeAnnotation.decimalType(scale, 38)` 

 ```
      if (physicalType == AbstractRowBuffer.ColumnPhysicalType.SB16) {
      LogicalTypeAnnotation typeAnnotation = LogicalTypeAnnotation.decimalType(scale, 38);
      return Types.primitive(type, repetition).as(typeAnnotation).length(16).id(id).named(name);
    } else {
      LogicalTypeAnnotation typeAnnotation =
         LogicalTypeAnnotation.decimalType(scale, 18);
      return Types.primitive(type, repetition).as(typeAnnotation).id(id).named(name);
  }
```
or does the precision depend on the ColumnPhysicalType.SB8/4/2/1?

File: src/test/java/net/snowflake/ingest/streaming/internal/FlushServiceTest.java [hunk 9]
```diff
@@ -183,11 +183,6 @@ ChannelBuilder setEncryptionKeyId(Long encryptionKeyId) {
         return this;
       }
 
-      ChannelBuilder setOnErrorOption(OpenChannelRequest.OnErrorOption onErrorOption) {
```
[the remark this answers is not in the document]
sfc-gh-gdoci: correct.

src/test/java/net/snowflake/ingest/streaming/internal/ParquetBufferTest.java
sfc-gh-azagrebin: Yes, I think so
it is the total number of decimal digits allowed:
https://docs.snowflake.com/en/sql-reference/data-types-numeric.html#number
every SBx type has a limit that it can hold

src/test/java/net/snowflake/ingest/streaming/internal/ParquetFlushServiceTest.java
sfc-gh-gdoci: > maybe a good idea to not make it random and assert expected query result

done

src/test/java/net/snowflake/ingest/streaming/internal/ParquetTypeGeneratorTest.java
```diff
@@ -0,0 +1,471 @@
...
+        PrimitiveType.PrimitiveTypeName.INT32, type.asPrimitiveType().getPrimitiveTypeName());
+    Assert.assertEquals(
+        LogicalTypeAnnotation.decimalType(testCol.getScale(), testCol.getPrecision()),
+        type.getLogicalTypeAnnotation());
+    Assert.assertEquals(Type.Repetition.OPTIONAL, type.getRepetition());
+    Assert.assertEquals(
+        AbstractRowBuffer.ColumnLogicalType.FIXED.getOrdinal()
+            + ","
+            + AbstractRowBuffer.ColumnPhysicalType.SB1.getOrdinal(),
+        metadata.get(type.getId().toString()));
```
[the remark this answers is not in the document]
sfc-gh-gdoci: done

src/test/java/net/snowflake/ingest/streaming/internal/ParquetValueParserTest.java
sfc-gh-gdoci: @sfc-gh-azagrebin I added other tests. Please take a look and tell me if I missed any important cases.

src/test/java/net/snowflake/ingest/streaming/internal/ParquetValueParserTest.java
```diff
@@ -0,0 +1,552 @@
...
+        .parquetBufferValue(pv)
+        .rowBufferStats(rowBufferStats)
+        .expectedValueClass(String.class)
+        .expectedSize(7.0f)
+        .expectedMinMax("Length7")
+        .assertMatches();
+  }
+
+  @Test
+  public void parseValueVariantToBinary() {
```
sfc-gh-azagrebin, line 239: Do you think this and next one can be deduplicated code-wise?

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
```diff
@@ -1099,86 +1099,181 @@ public void testDataTypes() {
...
+    negRow.put("bl", true);
+    negRow.put("var", "{}");
+    negRow.put("obj", "{}");
+    negRow.put("arr", Arrays.asList());
+    negRow.put("epochdays", 0);
+    negRow.put("epochsec", 0);
+    negRow.put("epochnano", new BigDecimal("0"));
+    negRow.put("timesec", 0);
+    negRow.put("timenano", 0);
+    verifyInsertValidationResponse(channel.insertRow(negRow, Integer.toString(0)));
```
sfc-gh-azagrebin, line 1122: This is a lot of details in one method.
I suggest to move them into separate functions `getNegRow()` etc, also assertion code for each row.

sfc-gh-azagrebin (approved):
Thanks Gloria! I left some smaller comments to look into before merging.

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
sfc-gh-gdoci: ack.

src/test/java/net/snowflake/ingest/streaming/internal/ParquetValueParserTest.java
sfc-gh-gdoci: yes, extracted code in `testJsonWithLogicalType(String logicalType)`

sfc-gh-gdoci:
@sfc-gh-azagrebin  Thank you for the review and your suggestions :-)! 
@sfc-gh-tzhang can you please take a look before I merge it to the main Parquet generator PR? I think it makes the review of the other PR easier.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -72,8 +63,10 @@ public void setupSchema(List<ColumnMetadata> columns) {
     // precision
     int id = 1;
     for (ColumnMetadata column : columns) {
-      Type type = getColumnParquetType(column, id);
-      parquetTypes.add(type);
+      ParquetTypeGenerator.ParquetTypeInfo typeInfo =
+          ParquetTypeGenerator.generateColumnParquetTypeInfo(column, id);
+      parquetTypes.add(typeInfo.getParquetType());
+      this.metadata.putAll(typeInfo.getMetadata());
```
sfc-gh-tzhang, line 69: is there any concern about overwriting old entries with same key but different values

src/main/java/net/snowflake/ingest/streaming/internal/ParquetTypeGenerator.java
```diff
@@ -0,0 +1,302 @@
+package net.snowflake.ingest.streaming.internal;
+
+import java.util.*;
```
sfc-gh-tzhang, line 3: No * import please

src/main/java/net/snowflake/ingest/streaming/internal/ParquetTypeGenerator.java
```diff
@@ -0,0 +1,302 @@
+package net.snowflake.ingest.streaming.internal;
```
sfc-gh-tzhang, line 1: Please add copyright

src/main/java/net/snowflake/ingest/streaming/internal/ParquetTypeGenerator.java
```diff
@@ -0,0 +1,302 @@
...
+    } catch (IllegalArgumentException e) {
+      throw new SFException(
+          ErrorCode.UNKNOWN_DATA_TYPE, column.getLogicalType(), column.getPhysicalType());
+    }
+
+    metadata.put(Integer.toString(id), logicalType.getOrdinal() + "," + physicalType.getOrdinal());
+
+    // Parquet Type.Repetition in general supports repeated values for the same row column, like a
+    // list of values.
+    // This generator uses only
```
sfc-gh-tzhang, line 78: This two lines seems short, is this done by the auto formatter?

src/test/java/net/snowflake/ingest/streaming/internal/ColumnMetadataBuilder.java
```diff
@@ -0,0 +1,160 @@
+package net.snowflake.ingest.streaming.internal;
```
sfc-gh-tzhang, line 1: copy right please

src/test/java/net/snowflake/ingest/streaming/internal/ColumnMetadataBuilder.java
```diff
@@ -0,0 +1,89 @@
...
+
+public class ColumnMetadataBuilder {
```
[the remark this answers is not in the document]
sfc-gh-tzhang: Is it the concern of exposing server side information? If yes, I won't concern it too much since this is not some fancy algorithm :)

sfc-gh-tzhang (approved):
LGTM! Left some minor comments, will take another look when the entire change goes into master. Also @sfc-gh-lsembera is changing related areas, so it would be good to give him a FYI in case the person merging later might need a big refactor :)

sfc-gh-tzhang:
Btw, please make sure the merge gate is all passing before merging

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-gdoci: I don't think so, because the keys are column ids (and column_id+obj_enc for objects), which should be unique.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetTypeGenerator.java
sfc-gh-gdoci: no, but change to one line.

src/test/java/net/snowflake/ingest/streaming/internal/ColumnMetadataBuilder.java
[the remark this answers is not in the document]
sfc-gh-gdoci: There were missing comments of this public class and its methods, but I have added them in later commit.

sfc-gh-gdoci:
thanks for the review @sfc-gh-tzhang ! I am merging to Andrey's branch. The test failures don't seem related to my changes (e.g. `SQL compilation error:
invalid parameter 'DISABLE_PARQUET_CACHE'`).
 I will take care of those when, as agreed with @sfc-gh-azagrebin , I take over his PR and merge it to master when ready.

PR #229, opened 2022-09-19
SNOW-655614 Data type validation: Date and Time

sfc-gh-lsembera:
Type validation for Snowflake data types DATE, TIME, TIMESTAMP_*.
It tries to achieve consistency with the behavior of Snowflake
Worksheets.

* Only limited set of Java data types is accepted
    * No longer try to parse the result of Object#toString() call
* Never interpret any input in our code, always delegate the
  parsing logic to SnowflakeDateTimeFormat
* Numeric values are not allowed for any column (e.g. int 123)
* Decimal strings are not allowed (e.g. "1663580793.123")
* Numeric strings (e.g. "123") are still allowed because Snowflake
  Worksheets also allow them, but they are not documented on the
  website and the SDK also should not advertise them.
* Date and timestamp strings cannot be passed into TIME columns
* Time strings cannot be passed into DATE and TIMESTAMP_* columns
* Some objects from java.time.* package are accepted
    * LocalTime, OffsetTime for TIME column
    * LocalDate, LocalDateTime, OffsetDateTime, ZonedDateTime for
      DATE and TIMESTAMP_* columns
* Time zones conversions are handled correctly
    * For DATE, TIME and TIMESTAMP_NTZ the timestamp is truncated
    * For TIMESTAMP_LTZ, timezone is accounted for before converting
      into the timezone America/Los_Angeles, which is the default
      value of the Snowflake parameter TIMEZONE
    * For TIMESTAMP_TZ, the timezone is preserved
        * America/Los_Angeles is used when timezone is missing

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
@@ -553,11 +553,11 @@ class ArrowRowBuffer extends AbstractRowBuffer<VectorSchemaRoot> {
             }
           case TIMESTAMP_LTZ:
           case TIMESTAMP_NTZ:
+            boolean ignoreTimezone = logicalType == ColumnLogicalType.TIMESTAMP_NTZ;
+
             switch (physicalType) {
               case SB8:
                 {
-                  boolean ignoreTimezone = logicalType == ColumnLogicalType.TIMESTAMP_NTZ;
-
                   BigIntVector bigIntVector = (BigIntVector) vector;
                   TimestampWrapper timestampWrapper =
                       DataValidationUtil.validateAndParseTimestampNtzSb16(
@@ -569,8 +569,6 @@ class ArrowRowBuffer extends AbstractRowBuffer<VectorSchemaRoot> {
                 }
               case SB16:
                 {
-                  boolean ignoreTimezone = logicalType == ColumnLogicalType.TIMESTAMP_NTZ;
-
                   StructVector structVector = (StructVector) vector;
                   BigIntVector epochVector =
                       (BigIntVector) structVector.getChild(FIELD_EPOCH_IN_SECONDS);
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
@@ -556,16 +556,21 @@ class ArrowRowBuffer extends AbstractRowBuffer<VectorSchemaRoot> {
             switch (physicalType) {
               case SB8:
                 {
+                  boolean ignoreTimezone = logicalType == ColumnLogicalType.TIMESTAMP_NTZ;
+
                   BigIntVector bigIntVector = (BigIntVector) vector;
-                  BigInteger timeInScale =
-                      DataValidationUtil.validateAndParseTime(value, field.getMetadata());
-                  bigIntVector.setSafe(curRowIndex, timeInScale.longValue());
-                  stats.addIntValue(timeInScale);
+                  TimestampWrapper timestampWrapper =
+                      DataValidationUtil.validateAndParseTimestampNtzSb16(
+                          value, getColumnScale(field.getMetadata()), ignoreTimezone);
+                  bigIntVector.setSafe(curRowIndex, timestampWrapper.getTimeInScale().longValue());
+                  stats.addIntValue(timestampWrapper.getTimeInScale());
                   rowBufferSize += 8;
                   break;
                 }
               case SB16:
                 {
+                  boolean ignoreTimezone = logicalType == ColumnLogicalType.TIMESTAMP_NTZ;
+
                   StructVector structVector = (StructVector) vector;
                   BigIntVector epochVector =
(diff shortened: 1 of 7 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
```diff
@@ -556,16 +556,21 @@ private float convertRowToArrow(
             switch (physicalType) {
               case SB8:
                 {
+                  boolean ignoreTimezone = logicalType == ColumnLogicalType.TIMESTAMP_NTZ;
```
sfc-gh-tzhang, line 559: nit: seems like this could be move out of the switch?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -117,72 +138,51 @@ static String validateAndParseObject(Object input) {
...
+              ? ((ZonedDateTime) input).toLocalDateTime().toString()
+              : DateTimeFormatter.ISO_DATE_TIME.format(((ZonedDateTime) input).toOffsetDateTime());
+    else if (input instanceof OffsetDateTime)
+      valueString =
+          ignoreTimezone
+              ? ((OffsetDateTime) input).toLocalDateTime().toString()
+              : DateTimeFormatter.ISO_DATE_TIME.format((OffsetDateTime) input);
+    else
+      throw new SFException(
+          ErrorCode.INVALID_ROW, input.toString(), "Data type not supported for timestamp");
```
sfc-gh-tzhang, line 168: Could you add to the comment on the types that will be supported now?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -221,40 +208,36 @@ static TimestampWrapper validateAndParseTimestampTz(Object input, Map<String, St
...
+    else if (input instanceof LocalDate) stringInput = input.toString();
+    else if (input instanceof LocalDateTime) stringInput = input.toString();
+    else if (input instanceof ZonedDateTime)
+      stringInput =
+          DateTimeFormatter.ISO_DATE_TIME.format(((ZonedDateTime) input).toOffsetDateTime());
+    else if (input instanceof OffsetDateTime)
+      stringInput = DateTimeFormatter.ISO_DATE_TIME.format((OffsetDateTime) input);
+    else
+      throw new SFException(
+          ErrorCode.INVALID_ROW, input.toString(), "Data type not supported for timestamp");
```
sfc-gh-tzhang, line 223: looks like this logic could be moved to a helper to avoid duplication

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/DateTimeIT.java
```diff
@@ -0,0 +1,1715 @@
+package net.snowflake.ingest.streaming.internal.datatypes;
+
+import java.sql.SQLException;
+import java.time.*;
```
sfc-gh-tzhang, line 4: Please avoid using star import

sfc-gh-tzhang (approved):
Left some comments, otherwise LGTM, thanks!

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/DateTimeIT.java
sfc-gh-lsembera: Star import removed

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-lsembera: I added information about allowed data types both to the exception message, as well to the javadoc comment.

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
sfc-gh-lsembera: Done

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-lsembera: Done

sfc-gh-tzhang (approved):

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/DateTimeIT.java
```diff
@@ -0,0 +1,1721 @@
...
+ */
+public class DateTimeIT extends AbstractDataTypeTest {
+
+  @Test
+  public void testTimestampWithTimeZone() throws Exception {
+    useLosAngelesTimeZone();
+
+    // Test timestamp formats
+    testJdbcTypeCompatibility(
+        "TIMESTAMP_TZ",
```
sfc-gh-azagrebin, line 26: sorry for late comment but maybe for some future change
I was wondering if it makes sense to deduplicate the calls for tests with a lot of cases.
by declaring an array of input tuples, like:
```
cases = {
{"TIMESTAMP_TZ", "2/18/2008 02:36:48", "2008-02-18 02:36:48.000000000 -0800"},
...
};

for case in cases {
  testJdbcTypeCompatibility(case, ...);
}
```

PR #230, opened 2022-09-20
SNOW-655614 Data type validation: NUMBER and REAL

sfc-gh-lsembera:
* Only limited of Java types are allowed, documented in Javadoc and exceptions
* Allowed range for NUMBER is determined by the logical type * Previously, the max allowed value for a column was determined by the max value, which fit into the physical type, which is failing in Snowflake.
* Half up rounding used when the scale of the value is greater than the column scale

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
@@ -6,6 +6,7 @@ package net.snowflake.ingest.streaming.internal;
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.math.RoundingMode;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -14,7 +15,6 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import net.snowflake.client.jdbc.internal.google.common.collect.Sets;
-import net.snowflake.client.jdbc.internal.snowflake.common.util.Power10;
 import net.snowflake.ingest.streaming.OpenChannelRequest;
 import net.snowflake.ingest.utils.Constants;
 import net.snowflake.ingest.utils.ErrorCode;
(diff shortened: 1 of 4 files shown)


sfc-gh-tzhang (approved):
LGTM! One thing we're missing is IT tests on all the data types, I think what we could do is to insert rows using SDK and compare the result with using the INSERT command, WDYT?

sfc-gh-lsembera:
@sfc-gh-tzhang We have that, all those `testJdbcTypeCompatibility` calls ingest with SDK, insert with JDBC driver and then select with WHERE, which also tests pruning.

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
```diff
@@ -465,45 +465,47 @@ private float convertRowToArrow(
...
+              ((DecimalVector) vector).setSafe(curRowIndex, inputAsBigDecimal);
+              stats.addIntValue(inputAsBigDecimal.unscaledValue());
               rowBufferSize += 16;
             } else {
               switch (physicalType) {
                 case SB1:
-                  byte byteValue = DataValidationUtil.validateAndParseByte(value);
-                  ((TinyIntVector) vector).setSafe(curRowIndex, byteValue);
-                  stats.addIntValue(BigInteger.valueOf(byteValue));
+                  ((TinyIntVector) vector).setSafe(curRowIndex, inputAsBigDecimal.byteValueExact());
```
sfc-gh-azagrebin, line 491: will `inputAsBigDecimal.byteValueExact()` throw if `inputAsBigDecimal` is outside of byte range?
if so, do we need to wrap it into `SFException`?
cc @sfc-gh-tzhang

src/test/java/net/snowflake/ingest/streaming/internal/DataValidationUtilTest.java
```diff
@@ -277,74 +264,51 @@ public void testValidateAndPareTimestampTz() {
...
-    Assert.assertEquals(new BigDecimal("1"), DataValidationUtil.validateAndParseBigDecimal(1));
-    Assert.assertEquals(new BigDecimal("1.0"), DataValidationUtil.validateAndParseBigDecimal(1D));
-    Assert.assertEquals(new BigDecimal("1"), DataValidationUtil.validateAndParseBigDecimal(1L));
-    Assert.assertEquals(new BigDecimal("1.0"), DataValidationUtil.validateAndParseBigDecimal(1F));
-    Assert.assertEquals(
-        BigDecimal.valueOf(10).pow(37),
-        DataValidationUtil.validateAndParseBigDecimal(BigDecimal.valueOf(10).pow(37)));
-    Assert.assertEquals(
+    assertEquals(new BigDecimal("1"), validateAndParseBigDecimal("1"));
+    assertEquals(
```
sfc-gh-azagrebin, line 270: does it make sense to declare a list of cases for tests with many repetitions?
Like:
```
cases = {
  {"1000", "1e3"},
  ...
}

test() {
  for case in cases {
    input = case.1;
    expected = case.2;
    assertEquals(new BigDecimal(input).toBigInteger(), validateAndParseBigDecimal(expected).toBigInteger());
  }
}
```

sfc-gh-azagrebin (commented):
Thanks Lukas! My 2 cents

sfc-gh-tjones (approved):

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
sfc-gh-tzhang: I don't think so, I believe we have a big catch block in `insertRows` to catch everything and throw a SFException instead

sfc-gh-lsembera: We have a try-catch block in insertRows, which is catching not just SFException, but all exceptions. However, this will not happen because above in the `convertRowToArrow` method, there is now validation based on logical type, so if the value is outside of the physical byte range, for example, it would be also outside of logical range and the validation would catch it.

src/test/java/net/snowflake/ingest/streaming/internal/DataValidationUtilTest.java
sfc-gh-lsembera: I agree, it makes sense. Let's discuss it on our 1-1 and I can change it later for all test cases.

PR #232, opened 2022-09-27
NO-SNOW: add stacktrace during blob building failure

sfc-gh-tzhang:
This PR contains two changes:
1. As part of investigating an encryption related issue, we realize the stacktrace and error message are not captured in log for any encryption related exceptions, so this PR adds that support
2. Add a small message indicates the code path of channel invalidation

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
@@ -379,6 +379,12 @@ class FlushService<T> {
                           | IllegalBlockSizeException
                           | BadPaddingException
                           | InvalidKeyException e) {
+                        String errorMessage =
+                            String.format(
+                                "Building blob failed, client=%s, file=%s, exception=%s,"
+                                    + " detail=%s, all channels in the blob will be invalidated",
+                                this.owningClient.getName(), filePath, e, e.getMessage());
+                        logger.logError(errorMessage);
                         throw new SFException(e, ErrorCode.ENCRYPTION_FAILURE);
                       }
                     },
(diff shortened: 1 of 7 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -358,28 +358,38 @@ void distributeFlushTasks() {
                         logger.logDebug(
                             "buildUploadWorkers stats={}", this.buildUploadWorkers.toString());
                         return buildAndUpload(filePath, blobData);
-                      } catch (IOException e) {
+                      } catch (IOException
```
sfc-gh-tjones, line 361: Should we have a generic `Exception` here or below? If something unexpected happens where does that get caught?

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -358,28 +358,38 @@ void distributeFlushTasks() {
...
+                          | InvalidAlgorithmParameterException
+                          | NoSuchPaddingException
+                          | IllegalBlockSizeException
+                          | BadPaddingException
+                          | InvalidKeyException
+                          | NoSuchAlgorithmException e) {
+                        StringBuilder stackTrace = new StringBuilder();
+                        for (StackTraceElement element : e.getStackTrace()) {
```
sfc-gh-tjones, line 369: Nit, you could reduce to `e.getStackTrace().forEach(stackTrace.append(StackTraceElement::toString)` or something along those lines.

sfc-gh-tjones (approved):
Thanks Toby!

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-tzhang: getStackTrace() actually returns an array instead of a list

PR #233, opened 2022-09-28
SNOW-655614 Data type validation: Semi-structured types

sfc-gh-lsembera:
More restricted validation of VARIANT, ARRAY AND OBJECT has been implemented:

* Only selected types are allowed 
    * String **must** be a valid JSON, otherwise it is rejected 
        * i.e. `"foo"` is rejected, but `"\"foo\""` is allowed 
    * java primitive data types and their arrays 
    * `java.time.*` objects 
    * `T[]` if `T` is a valid variant type 
    * `List<T>` if `T` is a valid variant type 
    * `Map<String, T>` if `T` is a valid variant type
* Custom JSON serialization of `ZonedDateTime` and `byte[]`
* Collections (array, list, map) are recursively validated
* ARRAY: Non-array input is converted into a 1-element array
* OBJECT: Non-object input is rejected
* Due to SNOW-664249, the max allowed limit is reduced by 1KB
* Integration tests for ingesting `null` added

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
@@ -150,23 +150,37 @@ class DataValidationUtil {
    * @return If the passed object is allowed for ingestion into a VARIANT column
    */
   static boolean isAllowedVariantType(Object o) {
-    if (o == null) return true;
-    if (o instanceof String || o instanceof Boolean || o instanceof Character) return true;
+    if (o == null) {
+      return true;
+    }
+    if (o instanceof String || o instanceof Boolean || o instanceof Character) {
+      return true;
+    }
     if (o instanceof Long
         || o instanceof Integer
         || o instanceof Short
         || o instanceof Byte
-        || o instanceof BigInteger) return true;
-    if (o instanceof Float || o instanceof Double || o instanceof BigDecimal) return true;
-    if (o instanceof java.time.LocalTime || o instanceof OffsetTime) return true;
+        || o instanceof BigInteger) {
+      return true;
+    }
+    if (o instanceof Float || o instanceof Double || o instanceof BigDecimal) {
+      return true;
+    }
+    if (o instanceof java.time.LocalTime || o instanceof OffsetTime) {
+      return true;
+    }
     if (o instanceof LocalDate
         || o instanceof LocalDateTime
         || o instanceof ZonedDateTime
-        || o instanceof OffsetDateTime) return true;
+        || o instanceof OffsetDateTime) {
+      return true;
+    }
     if (o instanceof Map) {
       boolean allKeysAreStrings =
           ((Map<?, ?>) o).keySet().stream().allMatch(x -> x instanceof String);
-      if (!allKeysAreStrings) return false;
+      if (!allKeysAreStrings) {
+        return false;
+      }
       boolean allValuesAreAllowed =
           ((Map<?, ?>) o).values().stream().allMatch(DataValidationUtil::isAllowedVariantType);
       return allValuesAreAllowed;
@@ -178,11 +192,15 @@ class DataValidationUtil {
         || o instanceof float[]
         || o instanceof double[]
         || o instanceof boolean[]
-        || o instanceof char[]) return true;
-    if (o.getClass().isArray())
+        || o instanceof char[]) {
+      return true;
+    }
+    if (o.getClass().isArray()) {
       return Arrays.stream((Object[]) o).allMatch(DataValidationUtil::isAllowedVariantType);
-    if (o instanceof List)
+    }
+    if (o instanceof List) {
       return ((List<?>) o).stream().allMatch(DataValidationUtil::isAllowedVariantType);
+    }
     return false;
   }
 
--- a/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
@@ -45,7 +45,9 @@ class DataValidationUtil {
   public static final int BYTES_8_MB = 8 * 1024 * 1024;
   public static final int BYTES_16_MB = 2 * BYTES_8_MB;
 
-  static final int MAX_VARIANT_LENGTH = BYTES_16_MB - 1024; // SNOW-664249
+  // TODO SNOW-664249: There is a few-byte mismatch between the value sent by the user and its
+  // server-side representation. Validation leaves a small buffer for this difference.
+  static final int MAX_SEMI_STRUCTURED_LENGTH = BYTES_16_MB - 64;
 
   private static final TimeZone DEFAULT_TIMEZONE =
       TimeZone.getTimeZone("America/Los_Angeles"); // default value of TIMEZONE system parameter
@@ -53,6 +55,10 @@ class DataValidationUtil {
 
   private static final ObjectMapper objectMapper = new ObjectMapper();
 
+  // The version of Jackson we are using does not support serialization of date objects from the
+  // java.time package. Here we define a module with custom java.time serializers. Additionally, we
+  // define custom serializer for byte[] because the Jackson default is to serialize it as
+  // base64-encoded string, and we would like to serialize it as JSON array of numbers.
   static {
     SimpleModule module = new SimpleModule();
     module.addSerializer(byte[].class, new ByteArraySerializer());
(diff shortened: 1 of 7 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -52,80 +74,170 @@ private static SnowflakeDateTimeFormat createDateTimeFormatter() {
...
+   */
+  static String validateAndParseArray(Object input) {
+    JsonNode jsonNode = validateAndParseVariantAsJsonTree(input, "ARRAY");
+
+    // Non-array values are ingested as single-element arrays, mimicking the Worksheets behavior
+    if (!jsonNode.isArray()) {
+      jsonNode = objectMapper.createArrayNode().add(jsonNode);
     }
 
+    String output = jsonNode.toString();
```
sfc-gh-tjones, line 205: Same question here as below about checking length initially.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -52,80 +74,170 @@ private static SnowflakeDateTimeFormat createDateTimeFormatter() {
...
+   * </ul>
+   *
+   * @param o Object to validate
+   * @return If the passed object is allowed for ingestion into a VARIANT column
    */
-  static String validateAndParseArray(Object input) {
-    if (!input.getClass().isArray() && !(input instanceof List)) {
-      throw new SFException(
-          ErrorCode.INVALID_ROW, input, "Input column must be an Array or List object.");
+  static boolean isAllowedVariantType(Object o) {
```
sfc-gh-tjones, line 152: These checks are great but it might be more readable to have brackets for each of the conditions. I'm nervous that someone adding a check or changing this might inadvertently goof up a check otherwise.

sfc-gh-tjones (commented):
Nice tests!

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -52,80 +74,170 @@ private static SnowflakeDateTimeFormat createDateTimeFormatter() {
...
     }
 
-    if (output.length() > BYTES_16_MB) {
-      throw new SFException(
-          ErrorCode.INVALID_ROW,
-          input.toString(),
-          String.format("Object too large. length=%d maxLength=%d", output.length(), BYTES_16_MB));
+    String output = jsonNode.toString();
+    // Throw an exception if the size is too large
+    int stringLength = output.getBytes(StandardCharsets.UTF_8).length;
```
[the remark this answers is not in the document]
sfc-gh-lsembera: These checks should happen after parsing because if the user passes a string consisting of millions of whitespace characters and then `{}`, for example, the result would be just `{}`, which would also be ingested. Running the check before would be unnecessarily restrictive.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-lsembera: Braces added

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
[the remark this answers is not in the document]
sfc-gh-lsembera: We are essentially checking the length of the minimised JSON, not the pretty-printed JSON passed by the user.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -37,12 +45,26 @@ class DataValidationUtil {
   public static final int BYTES_8_MB = 8 * 1024 * 1024;
   public static final int BYTES_16_MB = 2 * BYTES_8_MB;
 
+  static final int MAX_VARIANT_LENGTH = BYTES_16_MB - 1024; // SNOW-664249
```
sfc-gh-tzhang, line 48: Please add TODO keyword, move the comment to the prev line and add some description about the JIRA
```suggestion
  // TODO SNOW-664249: [add your description]
  static final int MAX_VARIANT_LENGTH = BYTES_16_MB - 1024;
```

sfc-gh-tzhang: And I feel like 1024 is too much, have you try with 32 or 64?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -37,12 +45,26 @@ class DataValidationUtil {
...
+
   private static final TimeZone DEFAULT_TIMEZONE =
       TimeZone.getTimeZone("America/Los_Angeles"); // default value of TIMEZONE system parameter
   private static final TimeZone GMT = TimeZone.getTimeZone("GMT");
 
   private static final ObjectMapper objectMapper = new ObjectMapper();
 
+  static {
```
sfc-gh-tzhang, line 56: Could you add a comment explain what this is used for?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-tzhang: It still looks a bit messy, could you add comment for each `if` statement?

sfc-gh-tzhang: And this is used not only for VARIANT, should we call it `isAllowedSemiStructuredType`?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -52,80 +74,188 @@ private static SnowflakeDateTimeFormat createDateTimeFormatter() {
...
+   * @return JSON tree representing the input
    */
-  static String validateAndParseVariant(Object input) {
-    String output;
-    try {
-      output = input.toString();
-    } catch (Exception e) {
-      throw new SFException(
-          e, ErrorCode.INVALID_ROW, input, "Input column can't be convert to String.");
+  private static JsonNode validateAndParseVariantAsJsonTree(Object input, String snowflakeType) {
```
sfc-gh-tzhang, line 83:
```suggestion
  private static JsonNode validateAndParseSemiStructuredAsJsonTree(Object input, String snowflakeType) {
```

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -52,80 +74,188 @@ private static SnowflakeDateTimeFormat createDateTimeFormatter() {
...
     }
 
+    String output = jsonNode.toString();
     // Throw an exception if the size is too large
-    if (output.length() > BYTES_16_MB) {
-      throw new SFException(
-          ErrorCode.INVALID_ROW,
-          input.toString(),
-          String.format("Array too large. length=%d maxLength=%d", output.length(), BYTES_16_MB));
+    int stringLength = output.getBytes(StandardCharsets.UTF_8).length;
```
sfc-gh-tzhang, line 225: do you know if `StandardCharsets.UTF_8` is the only thing we support in SF?

src/main/java/net/snowflake/ingest/streaming/internal/serialization/ByteArraySerializer.java
```diff
@@ -0,0 +1,22 @@
+package net.snowflake.ingest.streaming.internal.serialization;
```
sfc-gh-tzhang, line 1: copyright for all newly added files

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-lsembera: It is, according to the documentation: https://docs.snowflake.com/en/sql-reference/collation.html#collation-not-supported-for-strings-in-variant-array-or-object

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-lsembera: Method renamed, I also restructured the if/else conditions to make the logic easier to follow + added comments.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-lsembera: Seems to be working fine with 64.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-lsembera: Added a comment

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-lsembera: Method renamed

sfc-gh-tzhang (approved):
LGTM, thanks!

PR #242, opened 2022-10-11
[SNOW-675611] Get ErrorCode enum from SFException

sfc-gh-rcheng:
Get the ErrorCode for Exception handling in Kafka Connector


src/main/java/net/snowflake/ingest/utils/SFException.java
```diff
@@ -58,4 +60,8 @@ public Object[] getParams() {
   public Throwable getCause() {
     return cause;
   }
+
+  public ErrorCode getErrorCode() {
```
sfc-gh-japatel, line 64: can you add a comment here saying this is error code returned from snowflake and nothing we created here?

sfc-gh-japatel (approved):
lgtm! Thanks

src/main/java/net/snowflake/ingest/utils/SFException.java
```diff
@@ -18,6 +18,7 @@ public class SFException extends RuntimeException {
   private Throwable cause;
   private String vendorCode;
   private Object[] params;
+  private ErrorCode errorCode;
```
sfc-gh-tzhang, line 21: Then we should be able to remove vendorCode?

sfc-gh-rcheng:
further investigation showed that this change is unnecessary

PR #272, opened 2022-11-01
SNOW-663621 SNOW-682477 Fix incorrect strlen of non-ascii strings

sfc-gh-lsembera:
Snowflake expects the length of non-ascii strings to be the number of bytes, not the number of characters.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
@@ -454,7 +454,7 @@ class DataValidationUtil {
     }
     int maxLength = maxLengthOptional.orElse(BYTES_16_MB);
 
-    if (output.length() > maxLength) {
+    if (output.getBytes(StandardCharsets.UTF_8).length > maxLength) {
       throw valueFormatNotAllowedException(
           input,
           "STRING",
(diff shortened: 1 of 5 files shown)


sfc-gh-tjones (approved):

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -454,7 +454,7 @@ static String validateAndParseString(Object input, Optional<Integer> maxLengthOp
     }
     int maxLength = maxLengthOptional.orElse(BYTES_16_MB);
 
-    if (output.length() > maxLength) {
+    if (output.getBytes(StandardCharsets.UTF_8).length > maxLength) {
```
sfc-gh-tzhang, line 457: could you add comment explain why we're using byte length?

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/AbstractDataTypeTest.java
```diff
@@ -310,4 +314,8 @@ <STREAMING_INGEST_WRITE> void assertVariant(
     Assert.assertEquals(objectMapper.readTree(expectedValue), objectMapper.readTree(value));
     Assert.assertEquals(expectedType, typeof);
   }
+
+  protected boolean debugMode() {
+    return false;
+  }
```
sfc-gh-tzhang, line 320: remove debug mode?

sfc-gh-tzhang (commented):
Looks like there're test failures, do you want to take a look?

sfc-gh-lsembera:
This issue will be fixed in https://github.com/snowflakedb/snowflake-ingest-java/pull/300.

PR #277, opened 2022-11-04
SNOW-637237 Thread-safe HTTP client initialisation

sfc-gh-lsembera:
* Use double checked locking for HTTP client initialisation. 
* Remove state from `ServiceUnavailableRetryStrategy`

### DIFF
--- a/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
+++ b/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
@@ -51,6 +51,7 @@ public class HttpUtil {
   public static final String HTTP_PROXY_PASSWORD = "http.proxyPassword";
 
   private static final String PROXY_SCHEME = "http";
+  private static final Duration RETRY_INTERVAL = Duration.of(3, ChronoUnit.SECONDS);
   private static final int MAX_RETRIES = 3;
   private static volatile CloseableHttpClient httpClient;
 
@@ -198,7 +199,6 @@ public class HttpUtil {
       final int REQUEST_TIMEOUT = 408;
       final int TOO_MANY_REQUESTS = 429;
       final int SERVER_ERRORS = 500;
-      final Duration RETRY_INTERVAL = Duration.of(3, ChronoUnit.SECONDS);
 
       @Override
       public boolean retryRequest(
--- a/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
+++ b/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
@@ -50,7 +50,7 @@ public class HttpUtil {
 
   private static final String PROXY_SCHEME = "http";
   private static final int MAX_RETRIES = 3;
-  private static CloseableHttpClient httpClient;
+  private static volatile CloseableHttpClient httpClient;
 
   private static PoolingHttpClientConnectionManager connectionManager;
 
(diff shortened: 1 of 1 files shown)


sfc-gh-tjones (approved):

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -220,11 +222,8 @@ public boolean retryRequest(
       }
 
       @Override
-      // The waiting time is backoff, and is
-      // the exponential of the executionCount.
       public long getRetryInterval() {
-        long interval = (1 << executionCount) * 1000; // milliseconds
-        return interval;
+        return 3000;
```
sfc-gh-japatel, line 226: Hi, why did we change from exponential backoff time to a constant time?

sfc-gh-lsembera: Hi Jay, this exponential backoff was not thread safe. `executionCount` is being stored as an instance variable in `ServiceUnavailableRetryStrategy`, which is shared among all clients in possibly multiple threads. Two threads failing their request could be overwriting executionCount in the shared single instance of ServiceUnavailableRetryStrategy.

sfc-gh-japatel: I see, thanks Lukas! 
Can you extract 3000 in a static constant? Also maybe rename it to `getRetryIntervalMs`?

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -80,7 +80,11 @@ public class HttpUtil {
 
   public static CloseableHttpClient getHttpClient() {
     if (httpClient == null) {
-      initHttpClient();
+      synchronized (HttpUtil.class) {
+        if (httpClient == null) {
+          initHttpClient();
+        }
+      }
```
sfc-gh-japatel, line 87: do we need outer check of (httpClient == null)

sfc-gh-japatel, line 87: do we need outer check of `(httpClient == null)`

sfc-gh-lsembera: Outer check is not required for correctness, but it reduces lock contention. Lock is only required during initialisation; once it is initialised, no lock is needed anymore. But if the outer `if` was not there, lock would have to be acquired every time even when the initialisation is long finished.

sfc-gh-japatel: True that, thanks Lukas!

sfc-gh-japatel (approved):
lgtm! Minor nit for extracting 3000 to use `TimeUnit.MILLIS..`

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-tzhang: Should we add a random jitter?

sfc-gh-tzhang (approved):
thanks for the fix!

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-lsembera: I extracted the interval into a static constant, the method cannot be renamed because it is an override. I don't think a random jitter is required here.

PR #278, opened 2022-11-04
SNOW-684044 Fix nullability mismatch incidents

sfc-gh-lsembera:
Nullability mismatch incidents were happening for columns with quoted names like "id". These fields were being stored in Arrow as "id", which XP did not understand, it expects just id.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ColumnMetadata.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ColumnMetadata.java
@@ -125,4 +125,12 @@ class ColumnMetadata {
     map.put("nullable", this.nullable);
     return map.toString();
   }
+
+  public String normalizedName() {
+    if (name.length() >= 2 && name.charAt(0) == '"' && name.charAt(name.length() - 1) == '"') {
+      return name.substring(1, name.length() - 1).replace("\"\"", "\"");
+    } else {
+      return name;
+    }
+  }
 }
--- a/src/test/java/net/snowflake/ingest/streaming/internal/datatypes/AbstractDataTypeTest.java
+++ b/src/test/java/net/snowflake/ingest/streaming/internal/datatypes/AbstractDataTypeTest.java
@@ -142,6 +142,10 @@ public abstract class AbstractDataTypeTest {
     }
   }
 
+  protected void migrateTable(String tableName) throws SQLException {
+    conn.createStatement().execute(String.format("alter table %s migrate;", tableName));
+  }
+
   protected <T> void expectNumberOutOfRangeError(
       String dataType, T value, int maxPowerOf10Exclusive) throws Exception {
     expectError(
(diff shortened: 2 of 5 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/ColumnMetadata.java
```diff
@@ -125,4 +125,12 @@ public String toString() {
     map.put("nullable", this.nullable);
     return map.toString();
   }
+
+  public String normalizedName() {
+    if (name.length() >= 2 && name.charAt(0) == '"' && name.charAt(name.length() - 1) == '"') {
+      return name.substring(1, name.length() - 1).replace("\"\"", "\"");
+    } else {
+      return name;
```
sfc-gh-tjones, line 133: Nit, you can reduce to:
```
return (name.length() >= 2 && name.charAt(0) == '"' && name.charAt(name.length() - 1) == '"') ?
   name.substring(1, name.length() - 1).replace("\"\"", "\"") : name;
```

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/AbstractDataTypeTest.java
```diff
@@ -142,6 +142,10 @@ private <T> void expectError(String dataType, T value, Predicate<Exception> erro
     }
   }
 
+  protected void migrateTable(String tableName) throws SQLException {
+    conn.createStatement().execute(String.format("alter table %s migrate;", tableName));
```
sfc-gh-tjones, line 146: I don't think this is executable outside of Snowflake?

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/ColumnNameIT.java
```diff
@@ -0,0 +1,63 @@
...
+    SnowflakeStreamingIngestChannel channel = openChannel(name);
+
+    Map<String, Object> row = new HashMap<>();
+    row.put(name, ingestedValue);
+    channel.insertRow(row, offset);
+
+    TestUtils.waitForOffset(channel, offset);
+    assertSingleRow(name, ingestedValue);
+
+    migrateTable(name);
```
sfc-gh-tjones, line 49: Can we trigger scanback without explicitly invoking `migrate`?

sfc-gh-tzhang (changes requested):
I believe @sfc-gh-azagrebin is planning to do a similar fix for this issue, where we handle names with both double quotes and space, please chat with him to see if that's something that need help, thanks!

The JIRA is https://snowflakecomputing.atlassian.net/browse/SNOW-647377

sfc-gh-lsembera:
I discussed this issue with @sfc-gh-azagrebin and I am going to implement passing normalised column name from the server side instead of doing column name normalisation in the client. The new PR will supersede both this PR and https://github.com/snowflakedb/snowflake-ingest-java/pull/214. Closing.

PR #281, opened 2022-11-07
SNOW-637237 Retry for 2 minutes since the first failure

sfc-gh-lsembera:
Retry HTTP requests for some fixed total time instead of fixed number of attempts. This provides better balance between retrying "long enough" and not running out of memory.

1. If the server fails quickly, it is desirable to retry more times.
2. If the server fails slowly (times out after 1 minute, for example), many retry attempts can be cause the SDK to run out of memory due to the large amount of buffered data.

Retrying for N minutes makes is easier to reason about the amount of data that is going to be buffered while waiting for the server to recover.


sfc-gh-tjones (approved):

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -203,23 +206,38 @@ private static ServiceUnavailableRetryStrategy getServiceUnavailableRetryStrateg
       @Override
       public boolean retryRequest(
           final HttpResponse response, final int executionCount, final HttpContext context) {
-        int statusCode = response.getStatusLine().getStatusCode();
-        if (executionCount == MAX_RETRIES + 1) {
```
sfc-gh-tzhang, line 207: Looks like MAX_RETRIES could be removed now?

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -203,23 +206,38 @@ private static ServiceUnavailableRetryStrategy getServiceUnavailableRetryStrateg
...
+        long totalRetryDurationSoFarInSeconds = 0;
+        if (firstFault == null) {
+          context.setAttribute(FIRST_FAULT_TIMESTAMP, Instant.now());
+        } else {
+          Instant firstFaultInstant = (Instant) firstFault;
+          Instant now = Instant.now();
+          totalRetryDurationSoFarInSeconds = Duration.between(firstFaultInstant, now).getSeconds();
+
+          if (totalRetryDurationSoFarInSeconds > TOTAL_RETRY_DURATION.getSeconds()) {
+            LOGGER.info("Reached the max retry time, not retrying anymore");
```
sfc-gh-tzhang, line 219: nit: could you add the max retry time to the message?

sfc-gh-tzhang (approved):
LGTM other than a few minor comments, I'm wondering whether we could do the retry with exponential backoff, because we will retry 40 times (120/3), and if we're hitting TOO_MANY_REQUERTS, should we backoff?

sfc-gh-lsembera:
> LGTM other than a few minor comments, I'm wondering whether we could do the retry with exponential backoff, because we will retry 40 times (120/3), and if we're hitting TOO_MANY_REQUERTS, should we backoff?

Ideally, we would like to use exponential backoff, but I didn't find a way how to do that in a thread-safe way with the current version of our apache http client. In HTTP client 5, the `getRetryInterval` gets number of retries as a parameter, so it can calculate the exponential backoff. In our version of the http client, the `getRetryInterval` method does not take any parameters, so it does not know how many attempts have been made.

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-lsembera: It is still used `HttpRequestRetryHandler`. It is used when no server response was produced (some connection timeouts, etc). There we don't have any control over retry interval, so I prefer to keep it as it is and not introduce this waiting for 2 minutes.

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-lsembera: Added

PR #282, opened 2022-11-08
Revert "@snow SNOW-612516 Streaming ingest: add random file reader to scan single columns, format version 2

sfc-gh-azagrebin:
this reverts #191
The PR reverts only code related to bdec version 2 that was an optimisation for Arrow and that we never released.
As we plan to move to parquet now, we do not need this code and arrow compression dependency that it needs.


sfc-gh-tjones (approved):

sfc-gh-azagrebin:
> the old PR has 580 lines of changes but this one only removes 235 lines, is there anything else that we could remove?

not really, as I mentioned in this PR's description, the original PR also introduced multi-version handling that we also need for Parquet and that has nothing to do with the Arrow optimisation and its dependencies.

PR #283, opened 2022-11-09
NO-SNOW: enable code coverage check in the SDK

sfc-gh-tzhang:
enable code coverage check in the SDK


sfc-gh-japatel:
I checked it here: https://app.codecov.io/gh/snowflakedb/snowflake-ingest-java
LGTM.

PR #284, opened 2022-11-09
@snow no-snow Streaming ingest: always send BDEC version as server side supports this

sfc-gh-azagrebin:
The server side supports BDEC version in blob metadata json.
Hence we can always send it now and remove the code that previously did not send for version 1.


sfc-gh-tjones (approved):

src/main/java/net/snowflake/ingest/streaming/internal/BlobMetadata.java
```diff
@@ -49,16 +49,14 @@ List<ChunkMetadata> getChunks() {
...
-   */
+  @JsonProperty("bdec_version")
+  byte getVersionByte() {
+    return bdecVersion.toByte();
+  }
+
+  /** Create {@link BlobMetadata}. */
   static BlobMetadata createBlobMetadata(
       String path, String md5, Constants.BdecVersion bdecVersion, List<ChunkMetadata> chunks) {
-    // TODO SNOW-659721: Unify BlobMetadata with BlobMetadataWithBdecVersion once server side
```
sfc-gh-tzhang, line 58: Looks like there're more places to be updated if you search for SNOW-659721?

sfc-gh-tzhang (approved):
LGTM if you could remove other places as well, thanks

src/main/java/net/snowflake/ingest/streaming/internal/BlobMetadata.java
sfc-gh-azagrebin: It looks that the JIRA was wrongly added in this context here.
SNOW-659721 is about parquet scanner on the server side,
the other comments, you can still find, are still relevant but not related to this PR.

PR #285, opened 2022-11-10
@snow SNOW-657667 Streaming ingest: factor out channel from the buffer

sfc-gh-azagrebin:
[SNOW-657667](https://snowflakecomputing.atlassian.net/browse/SNOW-657667)

clean up refactoring of the circular dependencies:
channel -> buffer -> channel
channel -> chunk data -> channel
this allows to use buffer independently from channel and client, e.g. in tests and test file generator

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
@@ -13,13 +13,16 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Consumer;
 import java.util.stream.Collectors;
 import net.snowflake.ingest.streaming.InsertValidationResponse;
 import net.snowflake.ingest.streaming.OpenChannelRequest;
+import net.snowflake.ingest.utils.Constants;
 import net.snowflake.ingest.utils.ErrorCode;
 import net.snowflake.ingest.utils.Logging;
 import net.snowflake.ingest.utils.SFException;
 import net.snowflake.ingest.utils.Utils;
+import org.apache.arrow.memory.BufferAllocator;
 import org.apache.arrow.util.VisibleForTesting;
 
 /**
@@ -147,14 +150,35 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
   // Current buffer size
   private volatile float bufferSize;
 
-  // Reference to the Streaming Ingest channel that owns this buffer
-  @VisibleForTesting final SnowflakeStreamingIngestChannelInternal<T> owningChannel;
-
   // Names of non-nullable columns
   private final Set<String> nonNullableFieldNames;
 
-  AbstractRowBuffer(SnowflakeStreamingIngestChannelInternal<T> channel) {
-    this.owningChannel = channel;
+  // buffer's channel fully qualified name with database, schema and table
+  final String channelFullyQualifiedName;
+
+  // metric callback to report size of inserted rows
+  private final Consumer<Float> rowSizeMetric;
+
+  // Allocator used to allocate the buffers
+  final BufferAllocator allocator;
+
+  // State of the owning channel
+  final ChannelRuntimeState channelState;
+
+  // ON_ERROR option for this channel
+  final OpenChannelRequest.OnErrorOption onErrorOption;
+
+  AbstractRowBuffer(
+      OpenChannelRequest.OnErrorOption onErrorOption,
+      BufferAllocator allocator,
+      String fullyQualifiedChannelName,
+      Consumer<Float> rowSizeMetric,
+      ChannelRuntimeState channelRuntimeState) {
+    this.onErrorOption = onErrorOption;
+    this.rowSizeMetric = rowSizeMetric;
+    this.channelState = channelRuntimeState;
+    this.channelFullyQualifiedName = fullyQualifiedChannelName;
+    this.allocator = allocator;
     this.nonNullableFieldNames = new HashSet<>();
     this.flushLock = new ReentrantLock();
     this.rowCount = 0;
@@ -256,7 +280,7 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
     InsertValidationResponse response = new InsertValidationResponse();
     this.flushLock.lock();
     try {
-      if (this.owningChannel.getOnErrorOption() == OpenChannelRequest.OnErrorOption.CONTINUE) {
+      if (onErrorOption == OpenChannelRequest.OnErrorOption.CONTINUE) {
         // Used to map incoming row(nth row) to InsertError(for nth row) in response
         long rowIndex = 0;
         for (Map<String, Object> row : rows) {
@@ -305,8 +329,8 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
                     RowBufferStats.getCombinedStats(stats, this.tempStatsMap.get(colName))));
       }
 
-      this.owningChannel.setOffsetToken(offsetToken);
-      this.owningChannel.collectRowSize(rowSize);
+      this.channelState.setOffsetToken(offsetToken);
+      this.rowSizeMetric.accept(rowSize);
     } finally {
       this.tempStatsMap.values().forEach(RowBufferStats::reset);
       clearTempRows();
@@ -324,7 +348,7 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
    */
   @Override
   public ChannelData<T> flush() {
-    logger.logDebug("Start get data for channel={}", this.owningChannel.getFullyQualifiedName());
+    logger.logDebug("Start get data for channel={}", channelFullyQualifiedName);
     if (this.rowCount > 0) {
       Optional<T> oldData = Optional.empty();
       int oldRowCount = 0;
@@ -334,8 +358,7 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
       Map<String, RowBufferStats> oldColumnEps = null;
 
       logger.logDebug(
-          "Arrow buffer flush about to take lock on channel={}",
-          this.owningChannel.getFullyQualifiedName());
+          "Arrow buffer flush about to take lock on channel={}", channelFullyQualifiedName);
 
       this.flushLock.lock();
       try {
@@ -344,8 +367,8 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
           oldData = getSnapshot();
           oldRowCount = this.rowCount;
           oldBufferSize = this.bufferSize;
-          oldRowSequencer = this.owningChannel.incrementAndGetRowSequencer();
-          oldOffsetToken = this.owningChannel.getOffsetToken();
+          oldRowSequencer = this.channelState.incrementAndGetRowSequencer();
+          oldOffsetToken = this.channelState.getOffsetToken();
           oldColumnEps = new HashMap<>(this.statsMap);
           // Reset everything in the buffer once we save all the info
           reset();
@@ -356,7 +379,7 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
 
       logger.logDebug(
           "Arrow buffer flush released lock on channel={}, rowCount={}, bufferSize={}",
-          this.owningChannel.getFullyQualifiedName(),
+          channelFullyQualifiedName,
           oldRowCount,
           oldBufferSize);
 
@@ -365,10 +388,10 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
         data.setVectors(oldData.get());
         data.setRowCount(oldRowCount);
         data.setBufferSize(oldBufferSize);
-        data.setChannel(owningChannel);
         data.setRowSequencer(oldRowSequencer);
         data.setOffsetToken(oldOffsetToken);
         data.setColumnEps(oldColumnEps);
+        data.setFlusherFactory(this::createFlusher);
         return data;
       }
     }
@@ -462,5 +485,38 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
       epInfo.getColumnEps().put(colName, dto);
     }
     return epInfo;
+  }
+
+  /** Row buffer factory. */
+  static <T> AbstractRowBuffer<T> createRowBuffer(
+      OpenChannelRequest.OnErrorOption onErrorOption,
+      BufferAllocator allocator,
+      Constants.BdecVersion bdecVersion,
+      String fullyQualifiedChannelName,
+      Consumer<Float> rowSizeMetric,
+      ChannelRuntimeState channelRuntimeState) {
+    switch (bdecVersion) {
+      case ONE:
+        //noinspection unchecked
+        return (AbstractRowBuffer<T>)
+            new ArrowRowBuffer(
+                onErrorOption,
+                allocator,
+                fullyQualifiedChannelName,
+                rowSizeMetric,
+                channelRuntimeState);
+      case THREE:
+        //noinspection unchecked
+        return (AbstractRowBuffer<T>)
+            new ParquetRowBuffer(
+                onErrorOption,
+                allocator,
+                fullyQualifiedChannelName,
+                rowSizeMetric,
+                channelRuntimeState);
+      default:
+        throw new SFException(
+            ErrorCode.INTERNAL_ERROR, "Unsupported BDEC format version: " + bdecVersion);
+    }
   }
 }
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
@@ -14,6 +14,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Consumer;
 import net.snowflake.client.jdbc.internal.google.common.collect.Sets;
 import net.snowflake.ingest.streaming.OpenChannelRequest;
 import net.snowflake.ingest.utils.ErrorCode;
@@ -77,17 +78,14 @@ class ArrowRowBuffer extends AbstractRowBuffer<VectorSchemaRoot> {
   // Map the column name to Arrow column field
   private final Map<String, Field> fields;
 
-  // Allocator used to allocate the buffers
-  private final BufferAllocator allocator;
-
-  /**
-   * Construct a ArrowRowBuffer object
-   *
-   * @param channel client channel
-   */
-  ArrowRowBuffer(SnowflakeStreamingIngestChannelInternal<VectorSchemaRoot> channel) {
-    super(channel);
-    this.allocator = channel.getAllocator();
+  /** Construct a ArrowRowBuffer object. */
+  ArrowRowBuffer(
+      OpenChannelRequest.OnErrorOption onErrorOption,
+      BufferAllocator allocator,
+      String fullyQualifiedChannelName,
+      Consumer<Float> rowSizeMetric,
+      ChannelRuntimeState channelState) {
+    super(onErrorOption, allocator, fullyQualifiedChannelName, rowSizeMetric, channelState);
     this.fields = new HashMap<>();
   }
 
@@ -111,7 +109,7 @@ class ArrowRowBuffer extends AbstractRowBuffer<VectorSchemaRoot> {
       vectors.add(vector);
       this.statsMap.put(column.getName(), new RowBufferStats(column.getCollation()));
 
-      if (this.owningChannel.getOnErrorOption() == OpenChannelRequest.OnErrorOption.ABORT) {
+      if (onErrorOption == OpenChannelRequest.OnErrorOption.ABORT) {
         FieldVector tempVector = field.createVector(this.allocator);
         tempVectors.add(tempVector);
         this.tempStatsMap.put(column.getName(), new RowBufferStats(column.getCollation()));
@@ -138,7 +136,7 @@ class ArrowRowBuffer extends AbstractRowBuffer<VectorSchemaRoot> {
     logger.logInfo(
         "Trying to close arrow buffer for channel={} from function={}, allocatedBeforeRelease={},"
             + " allocatedAfterRelease={}",
-        this.owningChannel.getName(),
+        channelFullyQualifiedName,
         name,
         allocatedBeforeRelease,
         allocatedAfterRelease);
@@ -146,12 +144,12 @@ class ArrowRowBuffer extends AbstractRowBuffer<VectorSchemaRoot> {
 
     // If the channel is valid but still has leftover data, throw an exception because it should be
     // cleaned up already before calling close
-    if (allocatedBeforeRelease > 0 && this.owningChannel.isValid()) {
+    if (allocatedBeforeRelease > 0 && channelState.isValid()) {
       throw new SFException(
           ErrorCode.INTERNAL_ERROR,
           String.format(
               "Memory leaked=%d by allocator=%s, channel=%s",
-              allocatedBeforeRelease, this.allocator, this.owningChannel.getFullyQualifiedName()));
+              allocatedBeforeRelease, this.allocator, channelFullyQualifiedName));
     }
   }
 
--- a/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
@@ -18,12 +18,20 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.security.InvalidAlgorithmParameterException;
+import java.security.InvalidKeyException;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
+import java.util.ArrayList;
 import java.util.List;
+import java.util.zip.CRC32;
 import java.util.zip.GZIPOutputStream;
+import javax.crypto.BadPaddingException;
+import javax.crypto.IllegalBlockSizeException;
+import javax.crypto.NoSuchPaddingException;
 import javax.xml.bind.DatatypeConverter;
 import net.snowflake.ingest.utils.Constants;
+import net.snowflake.ingest.utils.Cryptor;
 import net.snowflake.ingest.utils.Logging;
 import net.snowflake.ingest.utils.Pair;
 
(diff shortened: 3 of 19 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -147,14 +151,30 @@ public int getOrdinal() {
...
   private final Set<String> nonNullableFieldNames;
 
-  AbstractRowBuffer(SnowflakeStreamingIngestChannelInternal<T> channel) {
-    this.owningChannel = channel;
+  // ON_ERROR option for this channel
+  final OpenChannelRequest.OnErrorOption onErrorOption;
+
+  final String channelFullyQualifiedName;
+
+  // Indicates whether the channel is still valid
```
sfc-gh-gdoci, line 162: nit: `the channel owing this buffer` or `the buffer`.

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -48,6 +56,106 @@ class BlobBuilder {
...
+
+        // Compute the md5 of the chunk data
+        String md5 = computeMD5(encryptedCompressedChunkData, compressedChunkLength);
+        int encryptedCompressedChunkDataSize = encryptedCompressedChunkData.length;
+
+        // Create chunk metadata
+        long startOffset = curDataSize;
+        ChunkMetadata chunkMetadata =
+            ChunkMetadata.builder()
+                .setOwningTable(firstChannelContext)
```
sfc-gh-gdoci, line 116: This line sounds weird. Either `.setOwningChannel(firstChannelContext)` or `.setOwingTable(firstChannelContext.getTable)` would make sense.

sfc-gh-gdoci, line 116: This line sounds weird. Maybe `.setOwningTableFromContext(firstChannelContext)`?

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
sfc-gh-gdoci, line 162: the channel owning this buffer or the buffer?

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -437,95 +434,12 @@ BlobMetadata buildAndUpload(String filePath, List<List<ChannelData<T>>> blobData
...
-            encryptedCompressedChunkDataSize,
-            bdecVersion);
-      }
-    }
-
-    // Build blob file, and then upload to streaming ingest dedicated stage
-    byte[] blob =
-        BlobBuilder.build(
-            chunksMetadataList, chunksDataList, crc.getValue(), curDataSize, bdecVersion);
+    BlobBuilder.Blob blob = BlobBuilder.build(filePath, blobData, bdecVersion);
```
sfc-gh-gdoci, line 438: nice!!

sfc-gh-gdoci (commented):
Thanks Andrey for the changes! Left a few comments.

sfc-gh-azagrebin:
@sfc-gh-gdoci 
Thanks for the review Gloria! Addressed!

sfc-gh-tzhang:
Any idea why we don't have the code coverage report for this PR?

sfc-gh-azagrebin:
> Any idea why we don't have the code coverage report for this PR?

No idea, this is something new for me that I saw only couple of times last week.
Could it be some timing issue with when the PR was open?
Do you know who is introduced the  code coverage report?

src/test/java/net/snowflake/ingest/streaming/internal/FlushServiceTest.java
```diff
@@ -474,7 +478,21 @@ public void testBuildAndUpload() throws Exception {
     channel1.insertRows(rows1, "offset1");
     channel2.insertRows(rows2, "offset2");
 
-    ChannelData<?> channel1Data = testContext.flushChannel(channel1.getName());
+    String name1 = channel1.getName();
+    ChannelData<?> channel1Data;
+    try {
+      channel1Data = testContext.flushChannel(name1);
```
sfc-gh-azagrebin, line 484: This line is very weird.
If I do not add this try/catch block, this line throws NullPointerException but only in PR CI and it is not reproducible locally. Same with the next test. It also fails sporadically in CI for this or next test.
I already tried everything, CI even passes notNull assertion for `testContext` and then throws immediately.
There is also nothing to be null in `flushChannel()` (I thought maybe the stack trace is incorrect).

@sfc-gh-tzhang 
Have you encountered such problem?

sfc-gh-azagrebin:
Suddenly there is a coverage report today.

src/test/java/net/snowflake/ingest/streaming/internal/FlushServiceTest.java
sfc-gh-azagrebin: Ok, now passed after rebase

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -463,4 +506,50 @@ static EpInfo buildEpInfoFromStats(long rowCount, Map<String, RowBufferStats> co
...
+        //noinspection unchecked
+        return (AbstractRowBuffer<T>) new ParquetRowBuffer(bufferConfig);
+      default:
+        throw new SFException(
+            ErrorCode.INTERNAL_ERROR,
+            "Unsupported BDEC format version: " + bufferConfig.bdecVersion);
+    }
+  }
+
+  static class BufferConfig {
```
sfc-gh-tzhang, line 529: comment?

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -463,4 +506,50 @@ static EpInfo buildEpInfoFromStats(long rowCount, Map<String, RowBufferStats> co
...
+    final BufferAllocator allocator;
+    final Constants.BdecVersion bdecVersion;
+    final OpenChannelRequest.OnErrorOption onErrorOption;
+    final String fullyQualifiedName;
```
sfc-gh-tzhang, line 533: fullyQualifiedChannelName?

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
sfc-gh-tzhang: And could we add getter function for all the fields?

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
```diff
@@ -80,14 +80,10 @@ class ArrowRowBuffer extends AbstractRowBuffer<VectorSchemaRoot> {
...
-   *
-   * @param channel client channel
-   */
-  ArrowRowBuffer(SnowflakeStreamingIngestChannelInternal<VectorSchemaRoot> channel) {
-    super(channel);
-    this.allocator = channel.getAllocator();
+  /** Construct a ArrowRowBuffer object. */
+  ArrowRowBuffer(BufferConfig bufferConfig) {
+    super(bufferConfig);
+    this.allocator = bufferConfig.allocator;
```
sfc-gh-tzhang, line 86: Why we're saving the allocator in both the super class and here?

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -48,6 +56,106 @@ class BlobBuilder {
...
+            result.rowCount,
+            startOffset,
+            chunkData.size(),
+            compressedChunkLength,
+            encryptedCompressedChunkDataSize,
+            bdecVersion);
+      }
+    }
+
+    // Build blob file, and then upload to streaming ingest dedicated stage
```
sfc-gh-tzhang, line 152: Please update the comment

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
```diff
@@ -38,23 +36,11 @@
...
-  private final String dbName;
-  private final String schemaName;
-  private final String tableName;
-  private volatile String offsetToken;
-  private final AtomicLong rowSequencer;
-
-  // Sequencer for this channel, corresponding to client sequencer at server side because each
-  // connection to a channel at server side will be seen as a connection from a new client
-  private final Long channelSequencer;
+  private final ChannelContext channelContext;
```
sfc-gh-tzhang, line 39: comment please, it's not clear what does channel context mean

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
```diff
@@ -128,41 +105,24 @@
...
-        return (RowBuffer<T>)
-            new ParquetRowBuffer((SnowflakeStreamingIngestChannelInternal<ParquetChunkData>) this);
-      default:
-        throw new SFException(
-            ErrorCode.INTERNAL_ERROR, "Unsupported BDEC format version: " + bdecVersion);
-    }
+    this.channelContext =
+        new ChannelContext(
+            name, dbName, schemaName, tableName, channelSequencer, encryptionKey, encryptionKeyId);
+    AbstractRowBuffer.BufferConfig bufferConfig =
```
sfc-gh-tzhang, line 114: I feel like the channelContext and bufferConfig are confusing, for example, the onErrorOption, offsetToken or rowSequencer should be associated with a channel instead of a configuration on a buffer, should we just create a channelContext and passing it around?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
```diff
@@ -246,24 +201,28 @@ public String getFullyQualifiedTableName() {
...
   @Override
   public boolean isValid() {
-    return this.isValid;
+    return this.rowBuffer.isValid();
   }
 
   /** Mark the channel as invalid, and release resources */
   void invalidate(String message) {
-    this.isValid = false;
+    this.rowBuffer.invalidate();
```
sfc-gh-tzhang, line 219: Similar concern here, the isValid should be on a channel instead of a rowBuffer

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
```diff
@@ -500,12 +459,84 @@ private void checkValidation() {
...
     return rowBuffer;
   }
+
+  /** Returns underlying channel's attributes. */
+  public ChannelContext getChannelContext() {
+    return channelContext;
+  }
+
+  /** Channel attributes. */
+  static class ChannelContext {
```
sfc-gh-tzhang, line 473: Could we add getter functions as well?

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -48,6 +56,106 @@ class BlobBuilder {
...
+  /**
+   * Builds blob.
+   *
+   * @param filePath Path of the destination file in cloud storage
+   * @param blobData All the data for one blob. Assumes that all ChannelData in the inner List
+   *     belongs to the same table. Will error if this is not the case
+   * @param bdecVersion version of blob
+   * @return {@link Blob} data
+   */
+  static <T> Blob build(
```
sfc-gh-tzhang, line 68: nit: `build` and `buildBlobBytes` might be confusing, should we say something like `constructBlobAndMetadata` and `buildBlob`?

sfc-gh-tzhang (changes requested):
Thanks Andrey, left some comments about my concerns, PTAL and see if it makes sense!

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
sfc-gh-azagrebin: This was one of the key points of refactoring.
The `isValid` flag is used in the `ArrowRowBuffer#close` method.
As I understand there is 1-1 relationship between channel and its buffer.
If channel is invalid then the buffer as well.
I would not touch it if the `isValid` flag was not used in `ArrowRowBuffer#close`.

Another option is that `ArrowRowBuffer#close` takes `isValid` as argument
but it is a volatile field and I was not sure about all consequences of capturing its value in a local function argument
before calling `ArrowRowBuffer#close`. If you say it is safe, it can be done like this.

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -147,14 +151,30 @@ public int getOrdinal() {
...
+
+  private final Consumer<Float> rowSizeMetric;
+
+  AbstractRowBuffer(BufferConfig bufferConfig) {
+    this.rowSizeMetric = bufferConfig.rowSizeMetric;
+    this.isValid = true;
+    this.offsetToken = bufferConfig.offsetToken;
+    this.rowSequencer = new AtomicLong(bufferConfig.rowSequencer);
+    this.onErrorOption = bufferConfig.onErrorOption;
+    this.channelFullyQualifiedName = bufferConfig.fullyQualifiedName;
```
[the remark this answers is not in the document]
sfc-gh-azagrebin: pass where?

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -165,6 +185,30 @@ public int getOrdinal() {
...
+  public String getOffsetToken() {
+    return this.offsetToken;
+  }
+
+  @Override
+  public AtomicLong getRowSequencer() {
+    return this.rowSequencer;
+  }
+
+  OpenChannelRequest.OnErrorOption getOnErrorOption() {
```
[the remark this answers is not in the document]
sfc-gh-azagrebin: Actually this method is not needed, removed it, `onErrorOption` can be used directly in subclasses

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
sfc-gh-azagrebin: I can do this but for such internal classes, it feels like just cluttering the code.

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
sfc-gh-azagrebin: I do not see it in super class here but actually we will need it there eventually for Parquet as well so I will move it there.

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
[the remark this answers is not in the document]
sfc-gh-azagrebin: oh maybe you mean to keep directly in a class field.
this config (I decided to rename it to `BufferingContext`) is not fully directly used in the buffer.
Hence I am not sure about coupling it with the buffer, e.g. row sequencer is just an initial sequencer (renamed it in the config btw) and bdec version is not used.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
sfc-gh-azagrebin: From what I see, onErrorOption, offsetToken or rowSequencer are basically used and managed mostly in the buffer.

> should we just create a channelContext and passing it around?

they have pretty different contents, I do not like mixing concerns into one object that basically results in couplings that this PR tries to address. I think each class should contain only what it needs to serve its purpose.

`BufferingContext` (renamed `BufferConfig`) is about setting up the buffer.
`ChannelContext` (I renamed it to ChannelAttributes) is to describe the channel (identification, encryption), it does not have much to do with buffering directly.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
sfc-gh-azagrebin: I understand getters as a part of public API to get a chance of adding later some hidden details but I do not understand the value of getters for internal classes. I would not clutter the code now with getters. If this code grows and e.g. we refactor this into different packages then I would add getters. Now it is just a simple immutable struct.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
sfc-gh-tzhang: Thanks for the reply. I do understand that the fields like `onErrorOption`, `offsetToken` or `rowSequencer` is just used/updated in the buffer, but these fields should still be associated with a channel and I find it confusing if we associate it with the buffer instead of channel. It's not clear to the readers that channel and buffer are 1:1 mapped, what if we update the logic in the future that all the channels belong to the same table are inserting into the same buffer? If there is no better way, I would prefer to keep this logic as is since it's easier to understand and live with the circular dependency. WDYT?

sfc-gh-tzhang: Shared object sounds good to me, thanks!

sfc-gh-gdoci: As agreed with @sfc-gh-azagrebin, I pushed a commit that does that. PTAL.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
sfc-gh-azagrebin: now part of the `ChannelState`

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
```diff
@@ -128,41 +111,19 @@
...
-        throw new SFException(
-            ErrorCode.INTERNAL_ERROR, "Unsupported BDEC format version: " + bdecVersion);
-    }
+    this.channelContext =
+        new ChannelContext(
+            name, dbName, schemaName, tableName, channelSequencer, encryptionKey, encryptionKeyId);
+    AbstractRowBuffer.InitialisationContext bufferConfig =
+        new AbstractRowBuffer.InitialisationContext(
+            allocator, bdecVersion, getFullyQualifiedName(), this::collectRowSize);
+    this.channelState = new ChannelState(onErrorOption, offsetToken, rowSequencer, true);
```
sfc-gh-tzhang, line 123: Could we try to minimum the number of static classes used? Maybe just create one ChannelProperties class and put everything there? I find it confusing by these fields as a person who reads the code and it's not clear the exact purpose of each one, and if I need to add a new field i'm not sure which one to use

sfc-gh-azagrebin: The ChannelState is something that is used and mutated by both channel and buffer. I do not want to mix anything else into this because otherwise it is hard to just about what state channel and buffer are coupled with. Btw, just realised that `onErrorOption` actually does not belong to it, I was just trying to somehow keep with the channel but it is not used by channel class.

InitialisationContext is something that is needed only for the buffer.
It is mostly to avoid constructors and methods with a lot of arguments. I can remove it and pass again everything to the constructor.

sfc-gh-azagrebin: The ChannelState is something that is used and mutated by both channel and buffer. I do not want to mix anything else into this because otherwise it is hard to just about what state channel and buffer are coupled with. Btw, just realised that `onErrorOption` actually does not belong to it, I was just trying to somehow keep with the channel but it is not used by channel class.

InitialisationContext is something that is needed only for the buffer.
It is mostly to avoid constructors and methods with a lot of arguments. I can remove it and pass again everything to the constructor.

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -147,14 +151,26 @@ public int getOrdinal() {
...
+  final BufferAllocator allocator;
+
+  // State of the owning channel
+  final ChannelState channelState;
+
+  AbstractRowBuffer(InitialisationContext initialisationContext, ChannelState channelState) {
+    this.rowSizeMetric = initialisationContext.rowSizeMetric;
+    this.channelState = channelState;
+    this.channelFullyQualifiedName = initialisationContext.fullyQualifiedChannelName;
+    this.allocator = initialisationContext.allocator;
```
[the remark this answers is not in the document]
sfc-gh-azagrebin: If you want to share more mutable state between channel and buffer, you put it into the channelState.
If you want to add constructor argument to the buffer, you add it to the initialisationContext, I can just put everything again as constructor arguments.
if you want to share immutable channel attribute between channel and its data then you add to ChannelContext.

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -48,6 +56,104 @@ class BlobBuilder {
...
 
+  /**
+   * Builds blob.
+   *
+   * @param filePath Path of the destination file in cloud storage
+   * @param blobData All the data for one blob. Assumes that all ChannelData in the inner List
+   *     belongs to the same table. Will error if this is not the case
+   * @param bdecVersion version of blob
+   * @return {@link Blob} data
+   */
```
sfc-gh-kkloudas, line 67: This is a nice change because I think that now the remaining methods (or at least the big ones like the `compressIfNeededAndPadChunk` and the `compress` and maybe more) can become `private` so that this class does not "bleed" implementation details to its neighbours.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
```diff
@@ -500,12 +434,13 @@ private void checkValidation() {
...
-  OpenChannelRequest.OnErrorOption getOnErrorOption() {
-    return this.onErrorOption;
-  }
-
   /** Returns underlying channel's row buffer implementation. */
   RowBuffer<T> getRowBuffer() {
     return rowBuffer;
   }
+
+  /** Returns underlying channel's attributes. */
```
sfc-gh-kkloudas, line 442: This is only visible for tests right? So we could annotate it as `@VisibleForTests` to make it clear that it is not used in "production". WDYT?

src/main/java/net/snowflake/ingest/streaming/internal/ChannelContext.java
```diff
@@ -0,0 +1,108 @@
+package net.snowflake.ingest.streaming.internal;
+
```
sfc-gh-kkloudas, line 2: Personally I like this class as it is immutable and it summarises the "id" of the channel. This makes is a nice candidate for "sharing" information with other classes like the `channelData`. 

Actually why not passing the context as is to the `flush()` method as argument so that we can pass it to the constructor of the `ChannelData` directly, instead of having the `channelData#setChannelContext()`? I think this makes it cleaner. WDYT @sfc-gh-azagrebin ?

src/main/java/net/snowflake/ingest/streaming/internal/ChannelState.java
```diff
@@ -0,0 +1,59 @@
+package net.snowflake.ingest.streaming.internal;
+
+import java.util.concurrent.atomic.AtomicLong;
+
+/** Internal state of channel that is used and mutated by both channel and its buffer. */
+class ChannelState {
```
sfc-gh-kkloudas, line 6: Maybe rename to `ChannelRuntimeState`? To show that this is an object holding the state of the channel that changes at runtime, contrary to the `ChannelData` which contains the "identity" data about the channel. WDYT?

src/main/java/net/snowflake/ingest/streaming/internal/ChannelContext.java
sfc-gh-azagrebin: True, I think the builder pattern also makes sense in a separate PR to make the channelData immutable

src/main/java/net/snowflake/ingest/streaming/internal/ChannelState.java
sfc-gh-azagrebin: Sure if it is more readable :)

sfc-gh-kkloudas (commented):
In general, I like a lot the changes here as it breaks the circular dependencies that the `SFStreamingIngestChannelInternal` was involved in. In addition I personally prefer small classes and whenever possible splitting the "data structures" which are only to keep data (eg. the `ChannelData`) from the "business logic" (e.g. the `SFStreamingIngestChannelInternal`).

I left some minor comments but nothing  that would stop it from being merged but I will not give a +1 given that @sfc-gh-tzhang has some comments and he is more involved in the Client development.

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -463,4 +481,43 @@ static EpInfo buildEpInfoFromStats(long rowCount, Map<String, RowBufferStats> co
...
+
+    BufferInitialisationContext(
+        OpenChannelRequest.OnErrorOption onErrorOption,
+        BufferAllocator allocator,
+        Constants.BdecVersion bdecVersion,
+        String fullyQualifiedChannelName,
+        Consumer<Float> rowSizeMetric) {
+      this.onErrorOption = onErrorOption;
+      this.allocator = allocator;
+      this.bdecVersion = bdecVersion;
```
sfc-gh-tzhang, line 518: Looks like this is not used?

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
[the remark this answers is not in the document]
sfc-gh-tzhang: Sounds good, could you put what you mentioned here in the comment?

src/main/java/net/snowflake/ingest/streaming/internal/ChannelContext.java
```diff
@@ -0,0 +1,108 @@
+package net.snowflake.ingest.streaming.internal;
```
sfc-gh-tzhang, line 1: Please add copyright for all the newly added files

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -147,14 +150,30 @@ public int getOrdinal() {
...
+  // Allocator used to allocate the buffers
+  final BufferAllocator allocator;
+
+  // State of the owning channel
+  final ChannelState channelState;
+
+  // ON_ERROR option for this channel
+  final OpenChannelRequest.OnErrorOption onErrorOption;
+
+  AbstractRowBuffer(BufferInitialisationContext initialisationContext, ChannelState channelState) {
```
sfc-gh-tzhang, line 171: What about combine the `BufferInitialisationContext` and `ChannelState`, and rename `ChannelState` to `ChannelProperties`? This way all we left are `ChannelProperties` and `ChannelFlushContext`, where `ChannelProperties` is used to store the properties associated with a channel and for passing whatever is needed for the buffer, and `ChannelFlushContext` is an immutable object that is used during flush.

sfc-gh-tzhang (approved):
Discussed offline with @sfc-gh-azagrebin and left some comments, my major comment is about whether we really need `BufferInitialisationContext`, `ChannelState ` and `ChannelState`, could we combine the first two to be `ChannelProperties` and then `ChannelFlushContext`? I feel like that makes the code a bit easier to understand, please think about it and I'm happy with whatever decision you make. Thanks for removing the circular dependency and all the other smaller changes!

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
sfc-gh-azagrebin: It is used to decide what buffer to create, I can pull it out to the method

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
sfc-gh-azagrebin: `BufferInitialisationContext` is not needed to be in the channel. It is used only in the buffer.
Eventually we should create the buffer outside of the channel to simplify testing. All stuff in the `BufferInitialisationContext` is not needed in the channel at all. If we keep them in `ChannelState/ChannelProperties `, the `ChannelState` will have more than one responsibility that is now to share channel/buffer state (See [Single responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle)). More over readers will think that everything in the `BufferInitialisationContext` is needed in the `ChannelProperties` in the channel and the buffer. This will complicate understanding and moving/refactoring things, like it did when we introduced Parquet, me being the reader :)

I can remove `BufferInitialisationContext` and just pass all its fields to the buffer constructor if the number of classes is the concern.

sfc-gh-lsembera:
These changes are great, I prefer to split the functionality into more classes, immutable when possible. The ability to construct a buffer without a corresponding channel is really nice. Thanks a lot for for the work, @sfc-gh-azagrebin!

PR #286, opened 2022-11-10
V1.0.2-beta.6 release

sfc-gh-tzhang:
This release contains a few bug fixes and improvements for Snowpipe Streaming:

- [Improvement] Add parquet file support, this will be our default file format in the future
- [Improvement] Add support to specify a memory limit per client instead of using the system max memory
- [Improvement] Remove a bunch of unused code and setup code coverage support

This release contains a few bug fixes and improvements for both Snowpipe and Snowpipe Streaming:

- [Improvement] Improve the HTTP retry logic

PR #280, opened 2022-11-07
@snow SNOW-633432 Parquet file generator (fixed dep)

sfc-gh-azagrebin:
reopen #200 again after bumping Hadoop to the latest version 3.3.4.

PR #287, opened 2022-11-14
@snow no-snow Fix more parquet dependencies causing vulnerabilities

sfc-gh-azagrebin:
follow-up for #280 to fix more parquet dependencies causing vulnerabilities in #286


sfc-gh-gdoci (approved):
LGTM, thanks! DEW tests also are passing.

sfc-gh-tjones (approved):

PR #291, opened 2022-11-21
SNOW-647377 Quoted columns support

sfc-gh-lsembera:
This PR is a resurrection of @sfc-gh-azagrebin 's PR https://github.com/snowflakedb/snowflake-ingest-java/pull/214 with adaptations for Parquet and added some integration tests. 

The new code is using column display name everywhere except in Arrow/Parquet field names.

Originally, the idea was to pass the unquoted name from the server side, but it is not enough because we still need to translate between the name entered by the user and the display name of the column. For example, client-side quoting is required when ingesting into the column named `"CREATE"`.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/LiteralQuoteUtils.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/LiteralQuoteUtils.java
@@ -1,0 +1,170 @@
+package net.snowflake.ingest.streaming.internal;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.regex.Pattern;
+import net.snowflake.ingest.utils.Utils;
+
+/**
+ * Util class to normalise literals to match server side metadata.
+ *
+ * <p>Note: The methods in this class have to be kept in sync with the respective methods on server
+ * side.
+ */
+class LiteralQuoteUtils {
+  private static final String[] reservedKeywords = {
+    // ANSI reserved KW
+    "ALL",
+    "ALTER",
+    "AND",
+    "ANY",
+    "AS",
+    "BETWEEN",
+    "BY",
+    "CHECK",
+    "COLUMN",
+    "CONNECT",
+    "CREATE",
+    "CURRENT",
+    "DELETE",
+    "DISTINCT",
+    "DROP",
+    "ELSE",
+    "EXISTS",
+    "FOLLOWING",
+    "FOR",
+    "FROM",
+    "GRANT",
+    "GROUP",
+    "HAVING",
+    "IN",
+    "INSERT",
+    "INTERSECT",
+    "INTO",
+    "IS",
+    "LIKE",
+    "NOT",
+    "NULL",
+    "OF",
+    "ON",
+    "OR",
+    "ORDER",
+    "REVOKE",
+    "ROW",
+    "ROWS",
+    "SAMPLE",
+    "SELECT",
+    "SET",
+    "START",
+    "TABLE",
+    "TABLESAMPLE",
+    "THEN",
+    "TO",
+    "TRIGGER",
+    "UNION",
+    "UNIQUE",
+    "UPDATE",
+    "VALUES",
+    "WHENEVER",
+    "WHERE",
+    "WITH",
+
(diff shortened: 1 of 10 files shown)


sfc-gh-lsembera:
Later, we could move the quoting-related code to GSCommon to avoid duplication.

src/main/java/net/snowflake/ingest/streaming/internal/LiteralQuoteUtils.java
```diff
@@ -0,0 +1,170 @@
...
+import net.snowflake.ingest.utils.Utils;
+
+/**
+ * Util class to normalise literals to match server side metadata.
+ *
+ * <p>Note: The methods in this class have to be kept in sync with the respective methods on server
+ * side.
+ */
+class LiteralQuoteUtils {
+  private static final String[] reservedKeywords = {
```
sfc-gh-tjones, line 16: Question: should we even maintain this list? Let's say that we don't do any checks w/ respect to these. The operation should fail server side at registration time, right?

src/main/java/net/snowflake/ingest/streaming/internal/LiteralQuoteUtils.java
```diff
@@ -0,0 +1,170 @@
...
+    String normalisedInternalName = unquoteColumnName(columnName);
+    // display name sent by server side to client in open channel column metadata response
+    return quoteColumnNameIfNeeded(normalisedInternalName);
+  }
+
+  /**
+   * Quote user column name to match server side metadata.
+   *
+   * <p>This needs to be in sync with server side to code, until we decide to send the normalised
+   * unquoted column name from server.
```
sfc-gh-tjones, line 109: Would it be better to do the unquoted column name instead rather than these changes?

src/main/java/net/snowflake/ingest/streaming/internal/LiteralQuoteUtils.java
sfc-gh-lsembera: If a user creates table with column "CREATE", for example, which is a valid name, and then ingests a row with `row.put("CREATE", 1)`, we need to quote the input to match it with the column display name.

src/main/java/net/snowflake/ingest/streaming/internal/LiteralQuoteUtils.java
sfc-gh-lsembera: We need both quoted and unquoted - everything on REST API expects quoted and everything in BDEC expects unquoted (XP expects unquoted).

sfc-gh-lsembera: And I need to update this comment - even if we send both quoted and unquoted from the server side, we still need quote user input to match the column display name.

sfc-gh-tzhang:
> Originally, the idea was to pass the unquoted name from the server side, but it is not enough because we still need to translate between the name entered by the user and the display name of the column. For example, client-side quoting is required when ingesting into the column named `"CREATE"`.

Thanks Lukas, I might miss something, but I think if we pass both the display name and unquoted name we should be ok? Like you said, the unquoted name is for creating the Field, and display name is for anything else. I'm not sure if I understand the example you mentioned above, if the column name is "CREATE", then the input should be "\"CREATE\"", anything else shouldn't be matched, could you elaborate more? 

In general, I don't like the idea of exposing more and more server side logic to the client, especially the `reservedKeywords` in this case, so this option is really the last thing we should do if nothing else works.

sfc-gh-lsembera:
@sfc-gh-tzhang Yeah, we can do what you suggest. The user would have to provide a valid column `displayName`, so that the SDK can match it without any additional quoting logic to the display name from the server. There are some cases, though, when the display name is not obvious. For example, a column created as `ab\ c` (unquoted) has display name `"AB C"`, so the user would have to pass it like that: `row.put("\"AB C\"", X)`. Are we ok with that? I guess it is fine as long as we have really good error messages, which would point out what the column display names are and what should the input formats be. WDYT?

sfc-gh-lsembera:
Closing in favor of https://github.com/snowflakedb/snowflake-ingest-java/pull/293

PR #292, opened 2022-11-23
SNOW-696155 fix binary values for Parquet

sfc-gh-gdoci:
This PR implements a fix for an issue with binary column values in the Parquet flusher. We were creating a utf-8 String for values with type_name=binary  and for byte arrays (hexadecimal strings) that is wrong.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
@@ -190,7 +190,7 @@ public class ParquetRowBuffer extends AbstractRowBuffer<ParquetChunkData> {
       }
     }
     if (logicalType == ColumnLogicalType.BINARY && value != null) {
-      value = ((String) value).getBytes(StandardCharsets.UTF_8);
+      value = value instanceof String ? ((String) value).getBytes(StandardCharsets.UTF_8) : value;
     }
     return value;
   }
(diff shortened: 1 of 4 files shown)


sfc-gh-tjones (approved):

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -190,7 +190,7 @@ Object getVectorValueAt(String column, int index) {
       }
     }
     if (logicalType == ColumnLogicalType.BINARY && value != null) {
-      value = ((String) value).getBytes(StandardCharsets.UTF_8);
+      value = value instanceof String ? ((String) value).getBytes(StandardCharsets.UTF_8) : value;
```
sfc-gh-tzhang, line 193: I believe we should always use hex format for BINARY data type, see https://docs.snowflake.com/en/user-guide/binary-input-output.html. CC: @sfc-gh-lsembera for confirmation

sfc-gh-lsembera: Internally in the SDK we should use `byte[]` for BINARY, but in EpInfo we should use hex. It is in scope of my work on SNOW-663704. Server-side change was required, which was recently deployed, and later I am going to change it also in the SDK. I talked with @sfc-gh-gdoci about it, it is fine for now, I am going to fix it later.

sfc-gh-tzhang (approved):
lgtm!

sfc-gh-tjones (approved):

PR #293, opened 2022-11-24
SNOW-647377 Support quoted column names

sfc-gh-lsembera:
This PR implements support for quoted column names. In the code, internal unquoted name is used everywhere with the exception of statistics because REST API expects column display name in EpInfo.

Testing
Integration tests have been added with coverage for various corner cases.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
@@ -107,14 +107,16 @@ class ArrowRowBuffer extends AbstractRowBuffer<VectorSchemaRoot> {
       if (!field.isNullable()) {
         addNonNullableFieldName(field.getName());
       }
-      this.fields.put(column.getName(), field);
+      this.fields.put(column.getInternalName(), field);
       vectors.add(vector);
-      this.statsMap.put(column.getName(), new RowBufferStats(column.getCollation()));
+      this.statsMap.put(
+          column.getInternalName(), new RowBufferStats(column.getName(), column.getCollation()));
 
       if (this.owningChannel.getOnErrorOption() == OpenChannelRequest.OnErrorOption.ABORT) {
         FieldVector tempVector = field.createVector(this.allocator);
         tempVectors.add(tempVector);
-        this.tempStatsMap.put(column.getName(), new RowBufferStats(column.getCollation()));
+        this.tempStatsMap.put(
+            column.getInternalName(), new RowBufferStats(column.getName(), column.getCollation()));
       }
     }
 
@@ -338,7 +340,7 @@ class ArrowRowBuffer extends AbstractRowBuffer<VectorSchemaRoot> {
 
     // Create the corresponding column field base on the column data type
     fieldType = new FieldType(column.getNullable(), arrowType, null, metadata);
-    return new Field(column.getName(), fieldType, children);
+    return new Field(column.getInternalName(), fieldType, children);
   }
 
   @Override
@@ -442,7 +444,7 @@ class ArrowRowBuffer extends AbstractRowBuffer<VectorSchemaRoot> {
     float rowBufferSize = 0F;
     for (Map.Entry<String, Object> entry : row.entrySet()) {
       rowBufferSize += 0.125; // 1/8 for null value bitmap
-      String columnName = formatColumnName(entry.getKey());
+      String columnName = LiteralQuoteUtils.unquoteColumnName(entry.getKey());
       Object value = entry.getValue();
       Field field = this.fields.get(columnName);
       Utils.assertNotNull("Arrow column field", field);
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ColumnMetadata.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ColumnMetadata.java
@@ -111,6 +111,10 @@ class ColumnMetadata {
     return this.nullable;
   }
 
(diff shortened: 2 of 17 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
```diff
@@ -442,7 +444,7 @@ private float convertRowToArrow(
     float rowBufferSize = 0F;
     for (Map.Entry<String, Object> entry : row.entrySet()) {
       rowBufferSize += 0.125; // 1/8 for null value bitmap
-      String columnName = formatColumnName(entry.getKey());
+      String columnName = LiteralQuoteUtils.unquoteColumnName(entry.getKey());
```
sfc-gh-tzhang, line 447: I don't think this is correct since we need the display name for KC to do schema evolution, for example if the input has a column called "\"c1\"" and the table doesn't have it, then we will return this extra column name in InsertValidationResponse.InsertError and KC will create the column "c1" in the table, so the wrong column will be created if you pass the unquoted name. I believe the unquotedName is only used in the Arrow scanner and all other places should be using the display name, free feel to schedule a chat with me if needed

sfc-gh-tzhang (changes requested):

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
sfc-gh-azagrebin: Maybe it is easier to do this the other way around: to use display name only for `InsertValidationResponse.InsertError`
if we send both display and unquoted name from server side?

The problem with using display name to match the user input is that we have to unquote it and remove quote and space escapes to match the server side logic and then quote it back to match against display name. and we still need it for Arrow atm.
We could write custom logic to resolve user input name against display name but then this is yet another way to do things that we do not have on server side but have to keep align.
Now we can just move `unquoteColumnName` to SF common later and reuse it in the SDK.

sfc-gh-lsembera: @sfc-gh-tzhang We can discuss it in the evening, but I don't know how support this KC feature without introducing client-side quoting again. For example, if it tries to ingest into column `CREATE` and such column does not exists, KC would have to create it, but it needs to know what it is keyword and therefore quote it, right? Or is the user required to pass it quoted already?

sfc-gh-tjones (approved):

sfc-gh-tzhang:
> * Extra columns in user input => Original (unquoted) user input

To make sure, we will return the original input name? For example, if the input is "CREATE", then we will return "CREATE", and if the input is "\"create\"", and we will return "\"create\""?

src/main/java/net/snowflake/ingest/streaming/internal/ColumnMetadata.java
```diff
@@ -111,6 +111,10 @@ boolean getNullable() {
     return this.nullable;
   }
 
+  String getInternalName() {
```
sfc-gh-tzhang, line 114: Should we just store the internal name as well? Looks like there're a bunch of places are using it

src/main/java/net/snowflake/ingest/streaming/internal/LiteralQuoteUtils.java
```diff
@@ -0,0 +1,62 @@
+package net.snowflake.ingest.streaming.internal;
```
sfc-gh-tzhang, line 1: Please add copyright

src/main/java/net/snowflake/ingest/streaming/internal/RowBufferStats.java
```diff
@@ -306,20 +306,22 @@ byte[] performConversion(String input) {
   private long currentMaxLength;
   private CollationDefinition collationDefinition;
   private final String collationDefinitionString;
+  private final String columnDisplayName;
```
sfc-gh-tzhang, line 309: could you add a comment explain why we need it here?

sfc-gh-tzhang (approved):
Left some comments, otherwise LGTM, thanks!

src/main/java/net/snowflake/ingest/streaming/internal/ColumnMetadata.java
sfc-gh-lsembera: Good point, I changed it to calculate and store the internal name from name when `setName` is called. In the future, we will read it from the server.

src/main/java/net/snowflake/ingest/streaming/internal/LiteralQuoteUtils.java
sfc-gh-lsembera: Added

src/main/java/net/snowflake/ingest/streaming/internal/RowBufferStats.java
sfc-gh-lsembera: Comment added

sfc-gh-lsembera:
> To make sure, we will return the original input name? For example, if the input is "CREATE", then we will return "CREATE", and if the input is ""create"", and we will return ""create""?

Yes, this is correct. In case of an extra column in the map, the validation response contains exactly what the used entered, unmodified. There is an [integration test for that](https://github.com/snowflakedb/snowflake-ingest-java/pull/293/files#diff-6e2db18062631e042dc4f6ea7a3a52d3ef5aa99f1d1465e5e476a8c4c27bcb35R118).

PR #294, opened 2022-11-28
SNOW-637237 Test opening several thousand channels in parallel

sfc-gh-lsembera:


sfc-gh-kkloudas (commented):
Thanks a lot for the work @sfc-gh-lsembera ! This looks great! 

Do you think it makes sense to have also a test like the `reopenSameChannel` where we have multiple threads trying to concurrently open the same channel, but at the end, we try to push a chunk through each one of them and we check that only the one with the max channel sequencer succeeds while the others fail with the appropriate error code?

sfc-gh-lsembera:
@sfc-gh-kkloudas Good idea, I added an additional verification to `reopenSameChannel` test that checks that only the channel with the highest client sequencer can be used to ingest data.

sfc-gh-tjones (approved):
This is awesome - great test! Very easy to follow along as well.

sfc-gh-tzhang (approved):
lgtm!

PR #295, opened 2022-11-30
SNOW-701710: [Client side] Add telemetry to track the end2end latency for every BDEC

sfc-gh-tzhang:
In order to track the exact end2end latency for Snowpipe Streaming, we're adding two more fields in the BDEC registration request (first_insert_time_in_ms and last_insert_time_in_ms) which corresponds to the first row insert timestamp in the BDEC and the last row insert timestamp in the BDEC, then we could use it to calculate the end2end latency for each BDEC at server side and output it in the telemetry


src/main/java/net/snowflake/ingest/streaming/internal/ChannelData.java
```diff
@@ -56,6 +58,18 @@ public static Map<String, RowBufferStats> getCombinedColumnStatsMap(
     return result;
   }
 
+  /**
+   * Combines the two paris of min/max insert timestamp together
```
sfc-gh-tjones, line 62: Why is this needed? Why do we need to combine two pairs?

sfc-gh-tzhang: because this is at chunk level, not channel level

sfc-gh-tjones: Got it, makes sense!

sfc-gh-tjones (approved):

PR #296, opened 2022-12-02
SNOW-703928 add two system properties for IT tests

sfc-gh-gdoci:
Since a couple of IT tests are not run with Parquet format as part of merge gates, we want to have the ability to run them internally (in precommits) before merging changes. This can be done be setting the `-DenableParquetTests=true` and `-DtestProfile_path=profile.json` properties.

### DIFF
--- a/src/test/java/net/snowflake/ingest/TestUtils.java
+++ b/src/test/java/net/snowflake/ingest/TestUtils.java
@@ -91,7 +91,11 @@ public class TestUtils {
    * @throws IOException if can't read profile
    */
   private static void init() throws Exception {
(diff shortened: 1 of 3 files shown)


sfc-gh-azagrebin (approved):
Thanks a lot Gloria!
My only recommendation is to factor out methods in `TestUtils` for these changes:
```
- static String getTestProfilePath()
- static Collection<Object[]> getBdecVersionItCases()
```

sfc-gh-tjones (approved):
Thanks Gloria!

sfc-gh-tzhang (approved):
lgtm

PR #297, opened 2022-12-02
NO-SNOW: fix client telemetry logic

sfc-gh-tzhang:
This PR contains two changes on the client telemetry:
- Only send the telemetry when the count is NOT 0
- Update the latency unit to milliseconds


sfc-gh-japatel (approved):
lgtm

sfc-gh-tjones (approved):

PR #298, opened 2022-12-05
SNOW-704566 Add filename in Parquet extra MD

sfc-gh-kkloudas:
This is the client SDK part of the fix for streams on replicated mixed tables, as described [here](https://docs.google.com/document/d/1MZWhyQznTSinT9ZhKYnOO3UIof2n_qAUWVu1S3qZdaY/edit#heading=h.w6d46ulzf6d2). It will only work for Parquet, as we are not planning to go PuPr with Arrow.

The XP and the GS parts are implemented in these two PRs: https://github.com/snowflakedb/snowflake/pull/80212 (for GS) and https://github.com/snowflakedb/snowflake/pull/80209 (for XP).

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
@@ -341,6 +341,8 @@ class FlushService<T> {
       List<List<ChannelData<T>>> blobData = new ArrayList<>();
       float totalBufferSize = 0;
 
+      final String filePath = getFilePath(this.targetStage.getClientPrefix());
+
       // Distribute work at table level, create a new blob if reaching the blob size limit
       while (itr.hasNext() && totalBufferSize <= MAX_BLOB_SIZE_IN_BYTES) {
         ConcurrentHashMap<String, SnowflakeStreamingIngestChannelInternal<T>> table =
@@ -349,7 +351,7 @@ class FlushService<T> {
         // TODO: we could do parallel stream to get the channelData if needed
         for (SnowflakeStreamingIngestChannelInternal<T> channel : table.values()) {
           if (channel.isValid()) {
-            ChannelData<T> data = channel.getData();
+            ChannelData<T> data = channel.getData(filePath);
             if (data != null) {
               channelsDataPerTable.add(data);
               totalBufferSize += data.getBufferSize();
@@ -362,8 +364,11 @@ class FlushService<T> {
       }
 
       // Kick off a build job
-      if (!blobData.isEmpty()) {
-        String filePath = getFilePath(this.targetStage.getClientPrefix());
+      if (blobData.isEmpty()) {
+        // we decrement the counter so that we do not have gaps in the filenames created by this
+        // client. See method getFilePath() below.
+        this.counter.decrementAndGet();
+      } else {
         if (this.owningClient.flushLatency != null) {
           latencyTimerContextMap.putIfAbsent(filePath, this.owningClient.flushLatency.time());
         }
--- a/src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
@@ -226,7 +226,7 @@ class StreamingIngestStage {
     JsonNode responseNode = this.parseClientConfigureResponse(response);
     // Do not change the prefix everytime we have to refresh credentials
     if (Utils.isNullOrEmpty(this.clientPrefix)) {
-      this.clientPrefix = responseNode.get("prefix").textValue();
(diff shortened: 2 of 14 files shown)


sfc-gh-tzhang:
> It will only work for Parquet, as we are not planning to go PuPr with Arrow.

It doesn't look hard to support Arrow, why not added in ArrowRowBuffer as well?

sfc-gh-tzhang:
> > It will only work for Parquet, as we are not planning to go PuPr with Arrow.
> 
> It doesn't look hard to support Arrow, why not add the line in ArrowRowBuffer as well?

Nvm, looks like we don't expose the metadata for the ArrowRowBuffer

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -341,6 +341,7 @@ void distributeFlushTasks() {
       List<List<ChannelData<T>>> blobData = new ArrayList<>();
       float totalBufferSize = 0;
 
+      final String filePath = getFilePath(this.targetStage.getClientPrefix());
```
sfc-gh-tzhang, line 344: The issue with moving this here is that the blob name sequencer would have gaps now, because it's possible that blobData is empty, and we kind of replying on the continuous sequencer in the blob name to figure out the name of the next blob during debugging, could you take a look at how it was done in https://github.com/snowflakedb/snowflake-ingest-java/pull/136/files and update it accordingly? Basically we want to save the previous filePath if the blobData is empty

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
```diff
@@ -226,7 +226,8 @@ synchronized SnowflakeFileTransferMetadataWithAge refreshSnowflakeMetadata(boole
     JsonNode responseNode = this.parseClientConfigureResponse(response);
     // Do not change the prefix everytime we have to refresh credentials
     if (Utils.isNullOrEmpty(this.clientPrefix)) {
-      this.clientPrefix = responseNode.get("prefix").textValue();
+      this.clientPrefix = responseNode.get("prefix").textValue() +
+              (responseNode.has("deployment_id") ? ("_" + responseNode.get("deployment_id").longValue()) : "");
```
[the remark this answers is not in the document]
sfc-gh-tzhang: Ideally we should have an object on the response, but I guess we just didn't do it before :(

src/main/java/net/snowflake/ingest/utils/Constants.java
```diff
@@ -25,6 +25,8 @@ public class Constants {
   public static final String PRIVATE_KEY = "private_key";
   public static final String PRIVATE_KEY_PASSPHRASE = "private_key_passphrase";
   public static final String JDBC_PRIVATE_KEY = "privateKey";
+
```
sfc-gh-tzhang, line 28: nit: remove extra line

sfc-gh-tzhang (approved):
LGTM if you could update the logic to reuse filePath when the blobData is empty, thanks!

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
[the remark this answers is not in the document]
sfc-gh-kkloudas: Hehe if we had an object, this change would have been even trickier because I think that old clients would fail if they found a new field and a jackson exception. I think in that case, we would have to use some jackson annotation magic to tell the client to ignore the field if it exists and we do not expect it (new SF old client) and to put a default value if we expect it and it is not there (old SF new client).

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-kkloudas: I see. Thanks for the explanation @sfc-gh-tzhang . I will integrate this.

sfc-gh-kkloudas: I checked the code in the PR you mentioned and I think that it does not prevent gaps totally. Imagine that we have 5 channels, we create a filename for the 5th (so we bump the counter) and that channel is empty so we do nothing and we exit the method. Now the `filePath` variable you have in the PR that keeps your previous filename is deleted because we exited the method. At the next iteration, we will re-enter this method and bump again the counter, so now we will have gaps, right?

sfc-gh-kkloudas:
@sfc-gh-tzhang I updated the PR, please let me know what you think and I will merge :)

sfc-gh-kkloudas:
Yes, I will test this today actually.

sfc-gh-kkloudas:
I also tried this change on a snowflake version without the `deployment_id` and it is working without problems.

PR #300, opened 2022-12-06
SNOW-663621 Fix problems with unicode strings

sfc-gh-lsembera:
This PR implements support for unicode strings, aligns truncation and last byte incrementation of EPs with the server-side. String and semi-structured validation ensures that input values are valid UTF-8. 

Min and max metadata values are passed as hex-encoded, the server is ready to interpret them as such, which is why this PR must be released as SDK version `beta.8`.

Additionally the PR fixes these closely related issues:

* SNOW-693446 Disable ingestion into collated strings
* SNOW-686944 Fix string comparison
* SNOW-682477 Use UTF-8 bytes count for 16MB limit check and UTF-8 characters count for optional character limit check - `VARCHAR(N)`


src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -152,6 +152,8 @@ public int getOrdinal() {
   // Names of non-nullable columns
   private final Set<String> nonNullableFieldNames;
 
+  private final Set<String> collatedColumnsInternalNames;
```
sfc-gh-tjones, line 155: Nit, add docstring here. Might be good to clarify what an "internal name" is here.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -412,15 +419,34 @@ static String validateAndParseString(
...
+          String.format(
+              "String too long: length=%d bytes maxLength=%d bytes",
+              utf8Bytes.length, BYTES_16_MB));
     }
+
+    // If max allowed length is specified (e.g. VARCHAR(10)), the number of unicode characters must
+    // not exceed this value
+    maxLengthOptional.ifPresent(
+        maxAllowedCharacters -> {
+          int actualCharacters = output.codePointCount(0, output.length());
```
sfc-gh-tjones, line 439: What does `codePointCount` do?

sfc-gh-tjones (approved):
Looks good to me - see my minor comments. Awesome job with the tests!

sfc-gh-tzhang:
> * SNOW-693446 Disable ingestion into collated strings (only `NULL` values are allowed)

I don't see why we want to allow NULLs here, should we just return an error as part of the open channel if there is a COLLATE clause?

pom.xml
```diff
@@ -7,7 +7,7 @@
     <!-- Arifact name and version information -->
     <groupId>net.snowflake</groupId>
     <artifactId>snowflake-ingest-sdk</artifactId>
-    <version>1.0.2-beta.7</version>
+    <version>1.0.2-beta.8</version>
```
sfc-gh-tzhang, line 10: let's revert this in case there're other changes that we want to be included in beta.8, and we want this change to be the last change of the beta.8

src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
```diff
@@ -127,7 +127,7 @@ public class RequestBuilder {
   // Don't change!
   public static final String CLIENT_NAME = "SnowpipeJavaSDK";
 
-  public static final String DEFAULT_VERSION = "1.0.2-beta.7";
+  public static final String DEFAULT_VERSION = "1.0.2-beta.8";
```
sfc-gh-tzhang, line 130: same here

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-tzhang: "The codePointCount() method is used to count the number of Unicode code points in the specified text range of a given String.

The text range begins at the specified beginIndex and extends to the char at index endIndex - 1. Thus the length (in chars) of the text range is endIndex-beginIndex. Unpaired surrogates within the text range count as one code point each."

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
```diff
@@ -238,4 +251,28 @@ public int hashCode() {
         nullCount,
         maxLength);
   }
+
+  /** XP-compatible string truncation (FdnColDataContainer::Metrics::updateStrMinMax) */
+  static String truncateBytesAsHex(byte[] bytes, boolean truncateUp) {
```
sfc-gh-tzhang, line 256: Should we move this to some util class?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -401,6 +407,7 @@ static String validateAndParseString(
     String output;
     if (input instanceof String) {
       output = (String) input;
+      verifyValidUtf8(output, columnName, "STRING");
```
sfc-gh-tzhang, line 410: Any performance implication on this?

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
```diff
@@ -61,12 +65,21 @@ class FileColumnProperties {
...
+    if (stats.getCurrentMinStrValue() != null) {
+      String truncatedAsHex = truncateBytesAsHex(stats.getCurrentMinStrValue(), false);
+      this.setMinStrValue(truncatedAsHex);
+    }
+
+    // current hex-encoded max value, truncated up to 32 bytes
+    if (stats.getCurrentMaxStrValue() != null) {
+      String truncatedAsHex = truncateBytesAsHex(stats.getCurrentMaxStrValue(), true);
+      this.setMaxStrValue(truncatedAsHex);
+    }
```
sfc-gh-tzhang, line 81: Is this what we do for FDN as well?

sfc-gh-tzhang (commented):
Thanks @sfc-gh-lsembera ! Left some comments, my major concern is about the behavior for collated columns, is this what PM has suggested?

pom.xml
sfc-gh-lsembera: Revert of this would test failures. I suggest we keep it there, keep the PR open and only merge it just before `beta.8` release.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-lsembera: Yeah, there is a performance impact. I ran some micro benchmarks on my machine and validation of 1GB of string data (1 million of 1KB strings) with Java 11 takes around 2.5 s. With Java 17 there is an improvement to 1.4 s.

sfc-gh-lsembera:
> > * SNOW-693446 Disable ingestion into collated strings (only `NULL` values are allowed)
> 
> I don't see why we want to allow NULLs here, should we just return an error as part of the open channel if there is a COLLATE clause?

@sfc-gh-tzhang Previous versions of the SDK were allowing ingestion of collated strings, rejecting it now altogether would be unnecessarily strict and force users to recreate the DB schema, which may not always be easy. If we allow at least NULLs, they would only have to change the code ingesting data.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-lsembera: Yeah, this calculates the number of unicode characters in the string, which we need for `VARCHAR(N)`. For example:

```java
String s = "🍞";
Assert.assertEquals(2, s.length());
Assert.assertEquals(1, s.codePointCount(0, s.length()));
```

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
sfc-gh-lsembera: Yes, we do this for FDN, too.

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
sfc-gh-lsembera: Comment added

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
```diff
@@ -238,4 +251,28 @@ public int hashCode() {
         nullCount,
         maxLength);
   }
+
+  /** XP-compatible string truncation (FdnColDataContainer::Metrics::updateStrMinMax) */
```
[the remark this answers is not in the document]
sfc-gh-lsembera: Changed the comment

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
sfc-gh-lsembera: Move to `BinaryStringUtils`.

sfc-gh-lsembera:
@sfc-gh-tzhang I updated the PR to throw an exception while creating the channel if one of the columns has a collation defined.

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -191,14 +191,27 @@ public int getOrdinal() {
...
+   * provided.
    *
    * @param nonNullableFieldName non-nullable filed name
    */
   void addNonNullableFieldName(String nonNullableFieldName) {
     nonNullableFieldNames.add(nonNullableFieldName);
   }
 
+  /** Throws an exception if the column has a collation defined. */
+  void validateColumnCollation(ColumnMetadata column) {
```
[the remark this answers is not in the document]
sfc-gh-lsembera: I will add it to the server-side, too.

PR #302, opened 2022-12-07
SNOW-706682 Log retry-related messages as INFO

sfc-gh-lsembera:
429 responses are going to occasionally happen, the SDK will retry and there is no need to scare the users by logging these events on WARN level.


sfc-gh-kkloudas (approved):

sfc-gh-tjones:
Do we know how many requests will trigger a 429 from GS?

sfc-gh-lsembera:
@sfc-gh-tzhang When customers see WARN logs they are naturally interested what can they do to fix that warning. I would argue that if there is nothing that customer can do to fix it, it should not be a WARN and adding a message to a `WARN` log that it is actually DEBUG would just further confuse things. The log line is still there as INFO, retry logic is in place to make sure the request is retried. I also created SNOW-707518 to review our rate limiting rules.

PR #303, opened 2022-12-07
SNOW-705594: fix ConcurrentModificationException on the telemetry hashmap

sfc-gh-tzhang:
This PR contains two changes:
- Fix the ConcurrentModificationException on the telemetry hashmap, use ConcurrentHashMap instead
- Update the logging for SDK version to INFO from DEBUG


sfc-gh-tjones (approved):
Nice catch!

PR #304, opened 2022-12-08
SNOW-707520 Fix incorrect fractional part calculation of TIMESTAMP

sfc-gh-lsembera:
When TIMESTAMP columns scale is less than 9, fractional part in Arrow is stored with less than 9 digits, which is causing mismatch during scanback. Arrow  expects 9 digits in the fractional part.


sfc-gh-tzhang (approved):
lgtm

sfc-gh-lsembera:
@sfc-gh-tjones This issue only applies to Arrow - timestamps are stored differently in Parquet.
@sfc-gh-tzhang Could you please include this PR into the beta.7 release? Thanks!

PR #306, opened 2022-12-13
SNOW-710291 Allow inf, -inf and nan strings into FLOAT columns

sfc-gh-lsembera:


sfc-gh-tjones (approved):
Nice!

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -638,10 +638,21 @@ static double validateAndParseReal(Object input) {
...
       return ((Number) input).doubleValue();
     } else if (input instanceof String) {
+      String stringInput = (String) input;
       try {
-        return Double.parseDouble((String) input);
+        return Double.parseDouble(stringInput);
       } catch (NumberFormatException err) {
-        throw valueFormatNotAllowedException(input, "REAL", "Not a valid decimal number");
+        stringInput = stringInput.toLowerCase();
+        switch (stringInput.toLowerCase()) {
```
sfc-gh-tzhang, line 646: According to https://docs.snowflake.com/en/sql-reference/data-types-numeric.html#precision, the server side behavior is that everything needs to be in a single quote and are case sensitive, I think we should be following the same thing

sfc-gh-tzhang: My bad, i thought it says case-sensitive

sfc-gh-tzhang (approved):
lgtm

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -638,10 +638,21 @@ static double validateAndParseReal(Object input) {
     } else if (input instanceof Number) {
       return ((Number) input).doubleValue();
     } else if (input instanceof String) {
+      String stringInput = (String) input;
       try {
-        return Double.parseDouble((String) input);
+        return Double.parseDouble(stringInput);
       } catch (NumberFormatException err) {
-        throw valueFormatNotAllowedException(input, "REAL", "Not a valid decimal number");
+        stringInput = stringInput.toLowerCase();
```
sfc-gh-tzhang, line 645: extra?

sfc-gh-lsembera: Fixed, good catch, thanks.

PR #310, opened 2023-01-03
SNOW-711338 Convert empty variant strings into Variant column to NULL

sfc-gh-asen:
Snowflake parse_json('') is returning NULL so we should use the same behavior.

Tested with unit tests

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
@@ -444,6 +444,7 @@ class ArrowRowBuffer extends AbstractRowBuffer<VectorSchemaRoot> {
       rowBufferSize += 0.125; // 1/8 for null value bitmap
       String columnName = LiteralQuoteUtils.unquoteColumnName(entry.getKey());
       Object value = entry.getValue();
+      Object parsedValue = value;
       Field field = this.fields.get(columnName);
       Utils.assertNotNull("Arrow column field", field);
       FieldVector vector = sourceVectors.getVector(field);
@@ -454,14 +455,7 @@ class ArrowRowBuffer extends AbstractRowBuffer<VectorSchemaRoot> {
           ColumnLogicalType.valueOf(field.getMetadata().get(COLUMN_LOGICAL_TYPE));
       ColumnPhysicalType physicalType =
           ColumnPhysicalType.valueOf(field.getMetadata().get(COLUMN_PHYSICAL_TYPE));
-
-      if (value == null) {
-        if (!field.getFieldType().isNullable()) {
-          throw new SFException(
-              ErrorCode.INVALID_ROW, columnName, "Passed null to non nullable field");
-        }
-        insertNull(vector, stats, curRowIndex);
-      } else {
+      if (value != null) {
         switch (logicalType) {
           case FIXED:
             int columnPrecision = Integer.parseInt(field.getMetadata().get(COLUMN_PRECISION));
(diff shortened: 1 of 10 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -120,8 +120,14 @@ private static JsonNode validateAndParseSemiStructuredAsJsonTree(
...
    */
   static String validateAndParseVariant(String columnName, Object input) {
-    String output =
-        validateAndParseSemiStructuredAsJsonTree(columnName, input, "VARIANT").toString();
+    JsonNode node = validateAndParseSemiStructuredAsJsonTree(columnName, input, "VARIANT");
+
+    // Missing nodes are not valid json, ingest them as NULL instead
+    if (node.isMissingNode()) {
+      return null;
+    }
```
sfc-gh-lsembera, line 128: This will not work because null is not an ordinary value, it needs some special treatment in both arrow and parquet row buffers. We need to increase null count, check if the column is nullable, etc. 

For Arrow, take a look at `net.snowflake.ingest.streaming.internal.ArrowRowBuffer#insertNull`. For Parquet it is the branch at `net/snowflake/ingest/streaming/internal/ParquetValueParser.java:127`. This code needs to be triggered when we ingest null.

sfc-gh-lsembera (changes requested):

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-lsembera: Until now, `NULL` was stored in the data file only if user directly ingested java `null`. Now, the user can insert non-null value (an empty string in this case) and we would like to ingest null. The code for will need to be restructured a bit to make sure we still trigger the same code path as if the user ingested null directly.

sfc-gh-asen: Thanks for the context and detailed explanation, this is very helpful, let me investigate further and update the PR

sfc-gh-asen: I changed code structure to handle returning null from the validation by doing the null check after getting a parsedValue object. 

Another approach that I could think about was changing the return type of validateAndParseXXXX() methods to return Optional<String> to force handle the null cases in the downstream calls in a more defensive manner, i am not sure about the convention here so I'll leave that up to you to decide if returning null is okay here

src/test/java/net/snowflake/ingest/streaming/internal/RowBufferTest.java
```diff
@@ -1201,4 +1220,136 @@ public void testOnErrorAbortFailures() {
...
+
+    // Check stats generation
+    ChannelData<?> result = innerBuffer.flush("my_snowpipe_streaming.bdec");
+    Assert.assertEquals(5, result.getRowCount());
+    Assert.assertEquals(2, result.getColumnEps().get("COLVARIANT").getCurrentNullCount());
+  }
+
+
+  @Test
+  public void testE2EObject() {
```
sfc-gh-asen, line 1275: Added E2E object and array type after I realized they are missing

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
```diff
@@ -474,28 +468,33 @@ private float convertRowToArrow(
...
 
             if (columnScale != 0 || physicalType == ColumnPhysicalType.SB16) {
+              parsedValue = inputAsBigDecimal;
               ((DecimalVector) vector).setSafe(curRowIndex, inputAsBigDecimal);
               stats.addIntValue(inputAsBigDecimal.unscaledValue());
               rowBufferSize += 16;
             } else {
               switch (physicalType) {
                 case SB1:
+                  parsedValue = inputAsBigDecimal.byteValueExact();
```
sfc-gh-lsembera, line 478: This wouldn't detect `null` because a `NullPointerException` would be thrown if `validateAndParseBigDecimal` returned null.

Do we need to carry over the `parsedValue` for all the types? There are many types and this can only occur for VARIANTs. Wouldn't it be easier to only have a boolean like `nullDetected` initially as false and set it to true only in the VARIANT case for empty string? WDYT?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-lsembera: I think returning null from the validate() methods is fine.

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
sfc-gh-asen: I think nullDetected makes a lot of sense and would be easier. The reason I added parsedValue to every single value is to have same kind of parallelism between different logical types and also parquetParser but it turns for it complicated the logic.

IIUC, it is the same call as the vector.setSafe() which should make it null safe but regardless i ll go with the nulldetected approach

src/main/java/net/snowflake/ingest/streaming/internal/ParquetValueParser.java
```diff
@@ -61,7 +62,7 @@ static ParquetBufferValue parseColumnValueToParquet(
         case BOOLEAN:
           int intValue =
               DataValidationUtil.validateAndParseBoolean(columnMetadata.getName(), value);
-          value = intValue > 0;
+          parsedValue = intValue > 0;
```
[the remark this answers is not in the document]
sfc-gh-asen: I just kept the logic same, i think it is to convert to a boolean 1 for true and 0 for false

sfc-gh-lsembera:
Thank you for the changes, @sfc-gh-asen! Could you please add an integration test to `SemiStructuredIT.java`, so we verify that the ingestion works end to end?

sfc-gh-lsembera (approved):

PR #311, opened 2023-01-09
SNOW-710401 Avoid datetime parsing wherever possible

sfc-gh-lsembera:
This patch implements more efficient calculation of `SFTimestamp` from supported `java.time.*` objects, without calling `toString` on them and parsing the result.

Users should be advised to pass date/time values either as unix timestamp strings or as `java.time.*` objects. Passing dates as formatted strings requires parsing, which is less efficient.

A simple microbenchmark on my local machine shows ≈19x improvement. Sequentially parsing and validating 10M ZonedDateTime objects took ≈36s with the old approach compared to 1.9s with the new approach.

Additionally, this commit unifies processing of timestamps with and without timezones.


sfc-gh-tzhang (approved):
LGTM! Thanks!

sfc-gh-tzhang:
> Users should be advised to pass date/time values either as unix timestamp strings or as `java.time.*` objects. Passing dates as formatted strings requires parsing, which is less efficient.

Could you add a performance section and add this to our online doc?

PR #316, opened 2023-01-16

sfc-gh-wfateem:
This change checks if the Snowflake account matches any of the hosts provided by the -Dhttp.nonProxyHosts JVM argument. If so, we connect to the host directly without going through the proxy server.


sfc-gh-wfateem:
I need to investigate this a bit further because somehow this functionality is working in the main branch without this code change. Closing this for now.

PR #320, opened 2023-01-24
SNOW-711299 New date/time processing

sfc-gh-lsembera:
`SnowflakeDateTimeFormat` has some issues. This PR implements the following changes to our date/time processing:

* We no longer depend on `SnowflakeDateTimeFormat`, custom validation and serialization of date/time values has been implemented. 
    * `java.time.OffsetDateTime` is the canonical representation of timestamps in the SDK. 
    * The following date/time values are going to be accepted: 
        * Types from `java.time` package (`LocalDate`, `LocalDateTime`, `OffsetDateTime`, `ZonedDateTime`, `Instant`, `LocalTime`, `OffsetTime`) 
        * Strings in ISO-8601 format such as _2011-12-03T10:15:30+01:00_. We attempt to parse one of the above-mentioned types using the `java.time.format.DateTimeFormatter`. 
        * Integer-stored strings
* Input values for `TIMESTAMP_TZ` and `TIMESTAMP_LTZ`, which do not carry timezone information, are by default interpreted in the timezone `America/Los_Angeles`. This can now be changed by calling `OpenChannelRequestBuilder#setDefaultTimezone`. 
    * Just like in Snowflake, integer-stored strings, as well as `java.time.Instant` in our case, are always interpreted in UTC.
* Detailed Javadoc for `insertRow()` has been added. It describes accepted values for all Snowflake types, so that users don't have to search the documentation.
* Comprehensive suite of integration tests has been added.

This PR fixes SNOW-663646, as well.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
+++ b/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
@@ -184,7 +184,7 @@ public interface SnowflakeStreamingIngestChannel {
    *
    *             </ul>
    *
-   *             For TIMESTAMP_LTZ and TIMESTAMP_TZ, all input without timezone will be by default interpreted in UTC. This can be changed by calling {@link net.snowflake.ingest.streaming.OpenChannelRequest.OpenChannelRequestBuilder#setDefaultTimezone(ZoneId)}.
+   *             For TIMESTAMP_LTZ and TIMESTAMP_TZ, all input without timezone will be by default interpreted in the timezone "America/Los_Angeles". This can be changed by calling {@link net.snowflake.ingest.streaming.OpenChannelRequest.OpenChannelRequestBuilder#setDefaultTimezone(ZoneId)}.
    *         </td>
    *         <tr>
    *             <td>VARIANT, ARRAY</td>
--- a/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
+++ b/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
@@ -4,6 +4,7 @@
 
 package net.snowflake.ingest.streaming;
 
+import java.time.ZoneId;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import javax.annotation.Nullable;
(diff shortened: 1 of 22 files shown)


src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
```diff
@@ -75,7 +76,149 @@ public interface SnowflakeStreamingIngestChannel {
...
+   *                 </ul>
+   *             </td>
+   *         </tr>
+   *         <tr>
+   *            <td>GEOGRAPHY, GEOMETRY</td>
+   *            <td>Not supported</td>
+   *         </tr>
+   *     </tr>
+   *
+   * </table>
```
sfc-gh-tjones, line 221: Nice table/explanation!

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -303,6 +300,93 @@ static String validateAndParseObject(String columnName, Object input) {
...
+        if (instant != null) return instant.atOffset(ZoneOffset.UTC);
+      }
+
+      // Couldn't parse anything, throw an exception
+      throw valueFormatNotAllowedException(
+          columnName,
+          input.toString(),
+          typeName,
+          "Not a valid value, see"
+              + " https://docs.snowflake.com/en/LIMITEDACCESS/snowpipe-streaming.html"
```
sfc-gh-tjones, line 370: FYI have we updated the docs yet with what is supported here?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -303,6 +300,93 @@ static String validateAndParseObject(String columnName, Object input) {
...
+        input.getClass(),
+        typeName,
+        new String[] {"String", "LocalDate", "LocalDateTime", "ZonedDateTime", "OffsetDateTime"});
+  }
+
+  private static <T> T catchParsingError(Supplier<T> op) {
+    try {
+      return op.get();
+    } catch (DateTimeParseException | NumberFormatException e) {
+      return null;
```
sfc-gh-tjones, line 386: Is this right? Should we throw an exception here?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-lsembera: I will update the docs, but a little bit later. This change will go to `beta.9` earliest, so I will wait until this release is about to come out.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-lsembera: This is right, it is just helper to translate an exception to null. The reason is that we are trying to parse input in 5 different formats, any of them can fail and I wanted to avoid 5 nested try/catch blocks.

sfc-gh-tjones: Ack, makes sense.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-tjones: SGTM!

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -552,44 +566,83 @@ static byte[] validateAndParseBinary(
...
+      }
 
-    if (timestamp == null) {
       throw valueFormatNotAllowedException(
           columnName,
           input,
           "TIME",
           "Not a valid time, see"
-              + " https://docs.snowflake.com/en/user-guide/date-time-input-output.html#time-formats"
+              + " https://docs.snowflake.com/en/LIMITEDACCESS/snowpipe-streaming.html"
```
sfc-gh-tzhang, line 613: could you add a TODO comment to update this in the future?

src/main/java/net/snowflake/ingest/streaming/internal/TimestampWrapper.java
```diff
@@ -1,40 +1,48 @@
...
-      long epoch, int fraction, BigInteger timeInScale, SFTimestamp sfTimestamp) {
-    this.epoch = epoch;
-    this.fraction = fraction;
-    this.timeInScale = timeInScale;
-    this.sfTimestamp = Optional.ofNullable(sfTimestamp);
+import java.math.RoundingMode;
+import java.time.OffsetDateTime;
+import net.snowflake.client.jdbc.internal.snowflake.common.util.Power10;
+
+public class TimestampWrapper {
```
sfc-gh-tzhang, line 9: comment for fields and functions in this file?

sfc-gh-tzhang:
> Input values for `TIMESTAMP_TZ` and `TIMESTAMP_LTZ`, which do not carry timezone information, are by default interpreted in UTC.
> * This can now be changed by calling `OpenChannelRequestBuilder#setDefaultTimezone`.
> * Just like in Snowflake, integer-stored strings, as well as `java.time.Instant` in our case, are always interpreted in UTC.

Could you run this with @sfc-gh-xhuang to see if this is something we want to support now? I believe there're a few session parameters that we can't support and this is one of it, if we decide this is the way we want, we would add them all; otherwise we would just support the default value and see what we want to do later.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -43,17 +39,18 @@
 
 /** Utility class for parsing and validating inputs based on Snowflake types */
 class DataValidationUtil {
+
+  private static final long SECONDS_LIMIT_FOR_EPOCH = 31536000000L;
```
sfc-gh-japatel, line 43: nit: can we add a comment here describing how we came up with this number?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -552,44 +566,83 @@ static byte[] validateAndParseBinary(
...
+  /**
+   * Attempts to parse integer-stored date from string input. Tries to guess the scale according to
+   * the rules documented at
+   * https://docs.snowflake.com/en/user-guide/date-time-input-output.html#auto-detection-of-integer-stored-date-time-and-timestamp-values.
+   *
+   * @param input String to parse, must represent a valid long
+   * @return Instant representing the input
+   * @throws NumberFormatException If the input in not a valid long
+   */
+  private static Instant parseInstantGuessScale(String input) {
```
sfc-gh-japatel, line 631: wondering if we are the first one in snowflakedb repos to guess this? there should be some commons package which handles this. WDYT?

sfc-gh-lsembera: We are not the first to guess this scale, this logic is taken from `SnowflakeDateTimeFormat#parse`, but it is working with `SFTimestamp`, which we are trying to move away from. It would be nice to extract this logic to GSCommon, but it may take some time, so I will create a JIRA to collect all the duplicated logic we have in the SDK and move it to GSCommon.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -303,6 +300,93 @@ static String validateAndParseObject(String columnName, Object input) {
...
+      // Just like integer-stored timestamps, instants are always interpreted in UTC
+      return ((Instant) input).atZone(ZoneOffset.UTC).toOffsetDateTime();
+    }
+
+    if (input instanceof String) {
+      String stringInput = (String) input;
+      {
+        // First, try to parse ZonedDateTime
+        ZonedDateTime zoned = catchParsingError(() -> ZonedDateTime.parse(stringInput));
+        if (zoned != null) return zoned.toOffsetDateTime();
```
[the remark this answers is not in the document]
sfc-gh-lsembera: Braces added

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-lsembera: TODO added

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-lsembera: Comments added, this refers to constants from `SFTimestamp`

sfc-gh-lsembera:
> > Input values for `TIMESTAMP_TZ` and `TIMESTAMP_LTZ`, which do not carry timezone information, are by default interpreted in UTC.
> > 
> > * This can now be changed by calling `OpenChannelRequestBuilder#setDefaultTimezone`.
> > * Just like in Snowflake, integer-stored strings, as well as `java.time.Instant` in our case, are always interpreted in UTC.
> 
> Could you run this with @sfc-gh-xhuang to see if this is something we want to support now? I believe there're a few session parameters that we can't support and this is one of it, if we decide this is the way we want, we would add them all; otherwise we would just support the default value and see what we want to do later.

We discussed this privately with @sfc-gh-xhuang and @sfc-gh-tzhang and decided to change the default timezone to `America/Los_Angeles` (to be aligned with the default value of the `TIMEZONE`), but keep the `setDefaultTimezone` method. Support for additional session parameters may be added later, depending on the demand. Example parameters that may be added: [BINARY_INPUT_FORMAT](https://docs.snowflake.com/en/user-guide/binary-input-output.html#session-parameters-for-binary-values)

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-lsembera: JIRA created: SNOW-740636

PR #321, opened 2023-01-25
SNOW-715618: Improve build latency for high throughput case with many channels

sfc-gh-tzhang:
We're seeing some issue where the build time is high during high throughput case (20 million+ rows) with 10+ parallel channels going into the same table, and the root cause is that the buffer takes longer to transfer ownership and the next buffer already builds up with enough rows, so the file size becomes very big which is causing the compression time to be very high. In longer term, instead of checking the channel size, I think checking the allocator size per table would help this case. CC: @sfc-gh-azagrebin @sfc-gh-gdoci 

In this PR, it contains a few changes to improve the build time:
- Fix a wrong buffer size logic which causes the flush to be scheduled unexpectedly
- Use parallel stream to get the buffer data
- Reduce the channel size limit to schedule the flush faster

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
@@ -274,7 +274,7 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
   @Override
   public InsertValidationResponse insertRows(
       Iterable<Map<String, Object>> rows, String offsetToken) {
-    float rowsSize = 0F;
+    float rowsSizeInBytes = 0F;
     if (!hasColumns()) {
       throw new SFException(ErrorCode.INTERNAL_ERROR, "Empty column fields");
     }
@@ -290,7 +290,7 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
               new InsertValidationResponse.InsertError(row, rowIndex);
           try {
             Set<String> inputColumnNames = verifyInputColumns(row, error);
-            rowsSize += addRow(row, this.rowCount, this.statsMap, inputColumnNames);
+            rowsSizeInBytes += addRow(row, this.rowCount, this.statsMap, inputColumnNames);
             this.rowCount++;
           } catch (SFException e) {
             error.setException(e);
@@ -307,17 +307,17 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
         }
       } else {
         // If the on_error option is ABORT, simply throw the first exception
-        float tempRowsSize = 0F;
+        float tempRowsSizeInBytes = 0F;
         int tempRowCount = 0;
         for (Map<String, Object> row : rows) {
           Set<String> inputColumnNames = verifyInputColumns(row, null);
-          tempRowsSize += addTempRow(row, tempRowCount, this.tempStatsMap, inputColumnNames);
+          tempRowsSizeInBytes += addTempRow(row, tempRowCount, this.tempStatsMap, inputColumnNames);
           tempRowCount++;
         }
 
         moveTempRowsToActualBuffer(tempRowCount);
 
-        rowsSize = tempRowsSize;
+        rowsSizeInBytes = tempRowsSizeInBytes;
         if ((long) this.rowCount + tempRowCount >= Integer.MAX_VALUE) {
           throw new SFException(ErrorCode.INTERNAL_ERROR, "Row count reaches MAX value");
         }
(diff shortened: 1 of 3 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -274,7 +274,7 @@ Set<String> verifyInputColumns(
   @Override
   public InsertValidationResponse insertRows(
       Iterable<Map<String, Object>> rows, String offsetToken) {
-    float rowSize = 0F;
+    float rowsSize = 0F;
```
sfc-gh-tjones, line 277: Nit, maybe rename to `rowsSizeBytes` so we know what the unit is that we're tracking.

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -308,31 +307,31 @@ public InsertValidationResponse insertRows(
         }
       } else {
         // If the on_error option is ABORT, simply throw the first exception
-        float tempRowSize = 0F;
+        float tempRowsSize = 0F;
```
sfc-gh-tjones, line 310: Same here regarding unit in the name.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -339,25 +341,28 @@ void distributeFlushTasks() {
 
     while (itr.hasNext()) {
       List<List<ChannelData<T>>> blobData = new ArrayList<>();
-      float totalBufferSize = 0;
+      AtomicReference<Float> totalBufferSize = new AtomicReference<>((float) 0);
```
sfc-gh-tjones, line 344: Nit, add unit here as well.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -339,25 +341,28 @@ void distributeFlushTasks() {
...
-            ChannelData<T> data = channel.getData(filePath);
-            if (data != null) {
-              channelsDataPerTable.add(data);
-              totalBufferSize += data.getBufferSize();
-            }
-          }
-        }
+        List<ChannelData<T>> channelsDataPerTable = Collections.synchronizedList(new ArrayList<>());
+        // Use parallel stream since getData could be the performance bottleneck when the number of
+        // channels are big
```
sfc-gh-tjones, line 354: Will this cause performance when the number of channels is small to degrade?

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -339,25 +341,28 @@ void distributeFlushTasks() {
...
-          if (channel.isValid()) {
-            ChannelData<T> data = channel.getData(filePath);
-            if (data != null) {
-              channelsDataPerTable.add(data);
-              totalBufferSize += data.getBufferSize();
-            }
-          }
-        }
+        List<ChannelData<T>> channelsDataPerTable = Collections.synchronizedList(new ArrayList<>());
+        // Use parallel stream since getData could be the performance bottleneck when the number of
```
sfc-gh-tjones, line 353: Nit, `when we have a high number of channels`

src/main/java/net/snowflake/ingest/utils/Constants.java
```diff
@@ -34,8 +34,8 @@ public class Constants {
       7L; // Don't change, should match server side
   public static final int BLOB_UPLOAD_TIMEOUT_IN_SEC = 5;
   public static final int INSERT_THROTTLE_MAX_RETRY_COUNT = 60;
-  public static final long MAX_BLOB_SIZE_IN_BYTES = 512000000L;
-  public static final long MAX_CHUNK_SIZE_IN_BYTES = 32000000L;
+  public static final long MAX_BLOB_SIZE_IN_BYTES = 256000000L;
+  public static final long MAX_CHUNK_SIZE_IN_BYTES = 16000000L;
```
sfc-gh-tjones, line 38: Nit, might be good to add a comment here explaining why these constants were choosen.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-tzhang: Not really based on my quick testing, the overhead of using a parallel stream is negligible compared with the time to get the buffer data

src/main/java/net/snowflake/ingest/utils/Constants.java
sfc-gh-tzhang: do we want to mention SF's internal file setting here in public comment?

sfc-gh-tjones: Ehhhhh probably not, let's leave this as-is then.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-tjones: Ack

sfc-gh-tjones (approved):

PR #322, opened 2023-01-27
SNOW-733699 make Parquet the default blob format version instead of Arrow

sfc-gh-gdoci:
This PR switches the default blob format version from Arrow to Parquet.

### DIFF
--- a/src/test/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelTest.java
+++ b/src/test/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelTest.java
@@ -36,7 +36,6 @@ import net.snowflake.ingest.utils.ParameterProvider;
 import net.snowflake.ingest.utils.SFException;
 import net.snowflake.ingest.utils.SnowflakeURL;
 import net.snowflake.ingest.utils.Utils;
-import org.apache.arrow.vector.VectorSchemaRoot;
 import org.junit.Assert;
 import org.junit.Test;
 import org.mockito.Mockito;
@@ -463,9 +462,9 @@ public class SnowflakeStreamingIngestChannelTest {
 
   @Test
   public void testInsertRow() {
-    SnowflakeStreamingIngestClientInternal<VectorSchemaRoot> client =
+    SnowflakeStreamingIngestClientInternal<ParquetChunkData> client =
         new SnowflakeStreamingIngestClientInternal<>("client");
-    SnowflakeStreamingIngestChannelInternal<VectorSchemaRoot> channel =
+    SnowflakeStreamingIngestChannelInternal<ParquetChunkData> channel =
         new SnowflakeStreamingIngestChannelInternal<>(
             "channel",
             "db",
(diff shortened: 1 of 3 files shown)


sfc-gh-azagrebin (commented):
Thanks Gloria!
I think `System.getProperty("enableParquetTests")` should be also removed to run both arrow and parquet.

sfc-gh-azagrebin (approved):

src/test/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelTest.java
```diff
@@ -463,9 +462,9 @@ public void testOpenChannelSuccessResponse() throws Exception {
 
   @Test
   public void testInsertRow() {
-    SnowflakeStreamingIngestClientInternal<VectorSchemaRoot> client =
+    SnowflakeStreamingIngestClientInternal<ParquetChunkData> client =
```
sfc-gh-tzhang, line 465: would this still work with Arrow?

sfc-gh-tzhang (approved):
Thanks @sfc-gh-gdoci ! Could you make sure everything still work with Arrow if we set the version to one in the config file? In case there is an issue, customer could mitigate it by switching back.

src/test/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelTest.java
sfc-gh-gdoci: It wouldn't. So I changed this test. Other tests should not need any modification and work with both Arrow/Parquet as default versions.

src/test/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelTest.java
```diff
@@ -463,9 +463,16 @@ public void testOpenChannelSuccessResponse() throws Exception {
...
-        new SnowflakeStreamingIngestClientInternal<>("client");
-    SnowflakeStreamingIngestChannelInternal<VectorSchemaRoot> channel =
+    SnowflakeStreamingIngestClientInternal<?> client;
+    boolean isArrowDefault =
+        ParameterProvider.BLOB_FORMAT_VERSION_DEFAULT == Constants.BdecVersion.ONE;
+    if (isArrowDefault) {
+      client = new SnowflakeStreamingIngestClientInternal<VectorSchemaRoot>("client_ARROW");
+    } else {
+      client = new SnowflakeStreamingIngestClientInternal<ParquetChunkData>("client_PARQUET");
+    }
```
sfc-gh-tjones, line 473: Nit, can consolidate to:
```
SnowflakeStreamingIngestClientInternal<?> client = ParameterProvider.BLOB_FORMAT_VERSION_DEFAULT == Constants.BdecVersion.ONE ?
    new SnowflakeStreamingIngestClientInternal<VectorSchemaRoot>("client_ARROW") :
    new SnowflakeStreamingIngestClientInternal<ParquetChunkData>("client_PARQUET");
```

sfc-gh-tjones (approved):

sfc-gh-tjones (approved):

PR #323, opened 2023-01-29

sfc-gh-wfateem:
The JDBC driver was upgraded to 3.13.18 and the HttpUtil code was changed to take into account that a user may need to bypass the proxy by providing the -Dhttp.nonProxyHosts JVM argument

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
@@ -101,6 +101,9 @@ public class SnowflakeStreamingIngestClientInternal<T> implements SnowflakeStrea
   // Name of the client
   private final String name;
 
+  // Account name
+  private String account;
+
   // Snowflake role for the client to use
   private String role;
 
@@ -165,8 +168,9 @@ public class SnowflakeStreamingIngestClientInternal<T> implements SnowflakeStrea
     this.parameterProvider = new ParameterProvider(parameterOverrides, prop);
 
     this.name = name;
+    this.account = accountURL.getAccount();
     this.isTestMode = isTestMode;
-    this.httpClient = httpClient == null ? HttpUtil.getHttpClient() : httpClient;
+    this.httpClient = httpClient == null ? HttpUtil.getHttpClient(account) : httpClient;
     this.channelCache = new ChannelCache<>();
     this.allocator = new RootAllocator();
     this.isClosed = false;
--- a/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
+++ b/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
@@ -14,6 +14,8 @@ import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReentrantLock;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import javax.net.ssl.SSLContext;
 import net.snowflake.client.core.SFSessionProperty;
 import net.snowflake.client.jdbc.internal.apache.http.HttpHost;
@@ -47,9 +49,10 @@ public class HttpUtil {
   public static final String USE_PROXY = "http.useProxy";
   public static final String PROXY_HOST = "http.proxyHost";
   public static final String PROXY_PORT = "http.proxyPort";
-
+  public static final String NON_PROXY_HOSTS = "http.nonProxyHosts";
   public static final String HTTP_PROXY_USER = "http.proxyUser";
   public static final String HTTP_PROXY_PASSWORD = "http.proxyPassword";
+  private static final String SNOWFLAKE_DOMAIN_NAME = ".snowflakecomputing.com";
 
   private static final String PROXY_SCHEME = "http";
   private static final String FIRST_FAULT_TIMESTAMP = "FIRST_FAULT_TIMESTAMP";
@@ -84,11 +87,11 @@ public class HttpUtil {
   // Only connections that are currently owned, not checked out, are subject to idle timeouts.
   private static final int DEFAULT_IDLE_CONNECTION_TIMEOUT_SECONDS = 30;
 
-  public static CloseableHttpClient getHttpClient() {
+  public static CloseableHttpClient getHttpClient(String account) {
     if (httpClient == null) {
       synchronized (HttpUtil.class) {
         if (httpClient == null) {
-          initHttpClient();
+          initHttpClient(account);
         }
       }
     }
@@ -100,7 +103,7 @@ public class HttpUtil {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(HttpUtil.class);
 
-  private static void initHttpClient() {
+  private static void initHttpClient(String account) {
 
     Security.setProperty("ocsp.enable", "true");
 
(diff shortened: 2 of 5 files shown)


src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -401,4 +411,29 @@ private static String createPoolStatsInfo(String title, PoolStats poolStats) {
     }
     return title;
   }
+
+  private static Boolean bypassProxy(String account){
+    String targetHost = account + SNOWFLAKE_DOMAIN_NAME;
+    if(System.getProperty(NON_PROXY_HOSTS) != null){
+      if(isInNonProxyHosts(targetHost))
+        return true;
```
sfc-gh-tjones, line 419: Same here regarding brackets.

sfc-gh-tjones, line 419: You can just do return `return System.getProperty(NON_PROXY_HOSTS) != null && isInNonProxyHosts(targetHost)`

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -401,4 +411,29 @@ private static String createPoolStatsInfo(String title, PoolStats poolStats) {
...
+  }
+
+  private static Boolean isInNonProxyHosts(String targetHost){
+    String nonProxyHosts = System.getProperty(NON_PROXY_HOSTS)
+            .replace(".", "\\.")
+            .replace("*", ".*");
+    String[] nonProxyHostsArray = nonProxyHosts.split("\\|");
+    Pattern p;
+    Matcher m;
+    for(String i : nonProxyHostsArray){
```
sfc-gh-tjones, line 431: Nit, I think we require a space after the `for`

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -401,4 +411,29 @@ private static String createPoolStatsInfo(String title, PoolStats poolStats) {
...
+    return false;
+  }
+
+  private static Boolean isInNonProxyHosts(String targetHost){
+    String nonProxyHosts = System.getProperty(NON_PROXY_HOSTS)
+            .replace(".", "\\.")
+            .replace("*", ".*");
+    String[] nonProxyHostsArray = nonProxyHosts.split("\\|");
+    Pattern p;
+    Matcher m;
```
sfc-gh-tjones, line 430: Why not move these below the `for`

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -401,4 +411,29 @@ private static String createPoolStatsInfo(String title, PoolStats poolStats) {
...
+    for(String i : nonProxyHostsArray){
+      p = Pattern.compile(i);
+      m = p.matcher(targetHost);
+      if(m.matches())
```
sfc-gh-tjones, line 434: Nit, you could always consolidate: `if Pattern.compile(i).matcher(targetHost).matches() ...`

sfc-gh-tjones (commented):
I had some nits but otherwise LGTM. Let's have Toby review as well.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -101,6 +101,9 @@
   // Name of the client
   private final String name;
 
+  // Account name
+  private String account;
```
sfc-gh-tzhang, line 105: could we use accountName to be more clear?

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -147,29 +150,31 @@ private static void initHttpClient() {
...
-      // Check if proxy username and password are set
-      final String proxyUser = System.getProperty(HTTP_PROXY_USER);
-      final String proxyPassword = System.getProperty(HTTP_PROXY_PASSWORD);
-      if (!isNullOrEmpty(proxyUser) && !isNullOrEmpty(proxyPassword)) {
-        Credentials credentials = new UsernamePasswordCredentials(proxyUser, proxyPassword);
-        AuthScope authScope = new AuthScope(proxyHost, proxyPort);
-        CredentialsProvider credentialsProvider = new BasicCredentialsProvider();
-        credentialsProvider.setCredentials(authScope, credentials);
-        clientBuilder = clientBuilder.setDefaultCredentialsProvider(credentialsProvider);
+      if(!bypassProxy(account)) {
```
sfc-gh-tzhang, line 153: nit: The two if checks could be combined into one
```suggestion
      if("true".equalsIgnoreCase(System.getProperty(USE_PROXY)) && !bypassProxy(account)) {
```

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -316,6 +321,12 @@ public static Properties generateProxyPropertiesForJDBC() {
         proxyProperties.put(SFSessionProperty.PROXY_USER.getPropertyKey(), proxyUser);
         proxyProperties.put(SFSessionProperty.PROXY_PASSWORD.getPropertyKey(), proxyPassword);
       }
+
+      // Check if http.nonProxyHosts was set
+      final String nonProxyHosts = System.getProperty(NON_PROXY_HOSTS);
+      if (!isNullOrEmpty(nonProxyHosts)) {
+        proxyProperties.put(SFSessionProperty.NON_PROXY_HOSTS.getPropertyKey(), nonProxyHosts);
+      }
```
sfc-gh-tzhang, line 329: Should we still need to check and put everything in `proxyProperties` if NON_PROXY_HOSTS = true?

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -401,4 +412,22 @@ private static String createPoolStatsInfo(String title, PoolStats poolStats) {
...
+
+  private static Boolean bypassProxy(String account) {
+    String targetHost = account + SNOWFLAKE_DOMAIN_NAME;
+    return System.getProperty(NON_PROXY_HOSTS) != null && isInNonProxyHosts(targetHost);
+  }
+
+  private static Boolean isInNonProxyHosts(String targetHost) {
+    String nonProxyHosts = System.getProperty(NON_PROXY_HOSTS)
+            .replace(".", "\\.")
+            .replace("*", ".*");
```
sfc-gh-tzhang, line 424: Could you add comments on why the replacement is needed?

sfc-gh-tzhang (approved):
Thanks Waleed! Left some minor comments, otherwise LGTM!

sfc-gh-tzhang:
@sfc-gh-wfateem Looks like the Java format and tests are failing, please fix them before merging, thanks!

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-wfateem: I'm assuming you meant if `USE_PROXY = true`, correct? 
Yes, someone may need to connect to host A and can only reach it by going through a proxy server, and at the same time, they need to connect to host B but can only reach that by going around the proxy server. `NON_PROXY_HOSTS `will contain the list of hosts that someone wants to connect to directly, so that has to be taken into account if it is set.

sfc-gh-japatel:
Hi @sfc-gh-wfateem can you open this PR in the same repo? instead of opening from a fork. Github forked PRs cannot access secrets.

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -84,11 +87,11 @@ public class HttpUtil {
   // Only connections that are currently owned, not checked out, are subject to idle timeouts.
   private static final int DEFAULT_IDLE_CONNECTION_TIMEOUT_SECONDS = 30;
 
-  public static CloseableHttpClient getHttpClient() {
+  public static CloseableHttpClient getHttpClient(String account) {
```
sfc-gh-japatel, line 90: nit: `accountName`
Also, Javadocs for public method and param

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -147,29 +150,31 @@ private static void initHttpClient() {
...
-      // Check if proxy username and password are set
-      final String proxyUser = System.getProperty(HTTP_PROXY_USER);
-      final String proxyPassword = System.getProperty(HTTP_PROXY_PASSWORD);
-      if (!isNullOrEmpty(proxyUser) && !isNullOrEmpty(proxyPassword)) {
-        Credentials credentials = new UsernamePasswordCredentials(proxyUser, proxyPassword);
-        AuthScope authScope = new AuthScope(proxyHost, proxyPort);
-        CredentialsProvider credentialsProvider = new BasicCredentialsProvider();
-        credentialsProvider.setCredentials(authScope, credentials);
-        clientBuilder = clientBuilder.setDefaultCredentialsProvider(credentialsProvider);
+      if("true".equalsIgnoreCase(System.getProperty(USE_PROXY)) && !bypassProxy(account)) {
```
sfc-gh-japatel, line 153: this can be simplified to just `if (!bypassProxy(account))`

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-japatel: Is it possible to add tests for this function too?

sfc-gh-japatel:
You can use this to run formatter to pass the merge gate: https://github.com/snowflakedb/snowflake-ingest-java#code-style

Also, thanks for the contribution and making changes to this repo.

sfc-gh-wfateem:
I'll close the PR and create a new branch in the original repo instead.

PR #324, opened 2023-02-01
SNOW-620624 enforce single rowcount Parquet files

sfc-gh-gdoci:
It looks like we are flushing buffered rows a bit late and this leads to large BDEC files. A side effect of 'late flushing' is that we are buffering too many rows and spilling to multiple rowgroups for Parquet. We had already increased the page size and row count limit for Parquet writers. 

But looking deeper into Parquet lib, it seems that  there is another [check](https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/InternalParquetRecordWriter.java#L151-L156) that compares the size of the buffer where rows are buffered to the block size limit and flushes to rowgroup (and then pages) if limit is reached. The check occurs not for every write, but after a configurable amount of record writes. 

So we have to either increase block size or increase the amount of records after which the check happens. 
In this PR, I increased the second one (`minRowCountForPageSizeCheck`) to `Integer.MAX_VALUE`, so basically de-activating the check. Optionally I can also set the block size to the estimated uncompressed chunk size that we have, but not sure if that is a good estimate for the block size.

Another important point: I added this `estimatedUncompressedChunkSizeInBytes` that we can use to set the page limit for Parquet, but for cases with few data, like `StreamingIngestIT.nullableColumns()` where the estimate is 8 bytes, Parquet actually compresses it to a 64 bytes large page.  Setting the limit to `estimatedUncompressedChunkSizeInBytes * 8` seemed too high to me for other cases, that's why I left it as it was = `MAX_CHUNK_SIZE_IN_BYTES * 2`.


src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
```diff
@@ -119,7 +121,10 @@ public void close() throws IOException {
     }
   }
 
-  private static ParquetProperties createParquetProperties() {
+  private static ParquetProperties createParquetProperties(
+      float estimatedUncompressedChunkSizeInBytes) {
+    int pageSizeLimit =
+        (int) MAX_CHUNK_SIZE_IN_BYTES * 2; // (int) (estimatedUncompressedChunkSizeInBytes * 2);
```
sfc-gh-tjones, line 127: Why is this commented out? And what's the rationale for the `* 2`?

sfc-gh-tjones (approved):
LGTM but seem my comment.

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
```diff
@@ -134,8 +139,9 @@ private static ParquetProperties createParquetProperties() {
         // pages is the same.
         // The quick fix is to effectively disable the page size/row limit
         // to always have one page per chunk until server side is generalised.
-        .withPageSize((int) Constants.MAX_CHUNK_SIZE_IN_BYTES * 2)
+        .withPageSize(pageSizeLimit)
         .withPageRowCountLimit(Integer.MAX_VALUE)
+        .withMinRowCountForPageSizeCheck(Integer.MAX_VALUE)
```
sfc-gh-tzhang, line 144: I assume this is all we need? If yes, do we still need the estimatedUncompressedChunkSizeInBytes?

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
```diff
@@ -134,8 +139,9 @@ private static ParquetProperties createParquetProperties() {
         // pages is the same.
         // The quick fix is to effectively disable the page size/row limit
         // to always have one page per chunk until server side is generalised.
-        .withPageSize((int) Constants.MAX_CHUNK_SIZE_IN_BYTES * 2)
+        .withPageSize(pageSizeLimit)
```
sfc-gh-tzhang, line 142: do you know the page size limit is before or after compression?

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
sfc-gh-tzhang: Also let's add comment on why we need to do it and a TODO to update it later

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
sfc-gh-gdoci: Page size limit is for compressed data, i.e. after compression.

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
sfc-gh-gdoci: > I assume this is all we need? If yes, do we still need the estimatedUncompressedChunkSizeInBytes?

Yes. The idea was to estimate size after compression (pageSizeLimit for Parquet)) based on size of data before compression (estimatedUncompressedChunkSizeInBytes). Since we don't have a good way to estimate it atm, I can remove the estimatedUncompressedChunkSizeInBytes for now.

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
sfc-gh-gdoci: The idea was to estimate size after compression (pageSizeLimit for Parquet)) based on size of data before compression (estimatedUncompressedChunkSizeInBytes). 

> And what's the rationale for the * 2?
It is simply an assumption that the size after compression will be at most twice the size before compression (e.g. for small rows the size after compression will be larger). But I am not sure that this is a valid assumption that's why it is commented out.

sfc-gh-gdoci: We may be able to remove this, see https://snowflake.slack.com/archives/C041W9578S1/p1675359412548999?thread_ts=1673361185.095149&cid=C041W9578S1

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
sfc-gh-tzhang: thanks, what if they see the size is bigger than the size limit after compression? Will they uncompressed it and split?

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
sfc-gh-tzhang: No strong preference here since the `MAX_CHUNK_SIZE_IN_BYTES * 2` doesn't make much sense already :) I think the safest option is set the limit to be MAX_BLOB_SIZE_IN_BYTES

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
```diff
@@ -119,7 +121,36 @@ public void close() throws IOException {
...
+     * SNOW-737331 *
+     */
+
+    // an estimate of the final page size after compression and encoding.
+    // For every column/value, Parquet stores other metadata like: nullability, repetition which
+    // require more space. To be on the safe side, we are multiplying the estimated uncompressed
+    // size with a factor.
+    int pageSizeLimit =
+        (int)
+            Math.max((int) MAX_CHUNK_SIZE_IN_BYTES * 2, estimatedUncompressedChunkSizeInBytes * 2);
```
sfc-gh-tzhang, line 153: could you add a log to print both values so we know what is the value we use?

sfc-gh-tzhang (approved):
lgtm, thanks!

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
sfc-gh-azagrebin: I would agree with `MAX_BLOB_SIZE_IN_BYTES`, but the problem is that we saw cases when this limits are used to allocate buffers in mem taking this size as initial or for estimates. We risk to overallocate too much memory with such conservative limits imo.

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestBigFilesIT.java
```diff
@@ -0,0 +1,199 @@
...
+        .createStatement()
+        .execute(String.format("use warehouse %s", TestUtils.getWarehouse()));
+
+    prop = TestUtils.getProperties(bdecVersion);
+    if (prop.getProperty(ROLE).equals("DEFAULT_ROLE")) {
+      prop.setProperty(ROLE, "ACCOUNTADMIN");
+    }
+    client =
+        (SnowflakeStreamingIngestClientInternal<?>)
+            SnowflakeStreamingIngestClientFactory.builder("client1").setProperties(prop).build();
```
sfc-gh-azagrebin, line 71: can this code be de-duped with `StreamingIngestIT`?

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestBigFilesIT.java
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    ExecutorService testThreadPool = Executors.newFixedThreadPool(numChannels);
+    CompletableFuture[] futures = new CompletableFuture[numChannels];
+    List<SnowflakeStreamingIngestChannel> channelList = new ArrayList<>();
+    for (int i = 0; i < numChannels; i++) {
+      final String channelName = "CHANNEL" + i;
+      int finalI = i;
+      futures[i] =
+          CompletableFuture.runAsync(
```
sfc-gh-azagrebin, line 147: just curious does it have to be concurrent ingest to produce big files?

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestBigFilesIT.java
```diff
@@ -0,0 +1,199 @@
...
+    client.close();
+    jdbcConnection.createStatement().execute(String.format("drop database %s", testDb));
+  }
+
+  @Test
+  public void testManyRowsMultipleChannelsToMultipleTable()
+      throws SQLException, ExecutionException, InterruptedException {
+    String tableNamePrefix = "t_big_table_";
+
+    int numTables = 2;
```
sfc-gh-azagrebin, line 85: does it have to be 2 tables to produce big pages / row groups?

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestBigFilesIT.java
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+  private void createTableForTest(String tableName) {
+    try {
+      jdbcConnection
+          .createStatement()
+          .execute(
+              String.format(
+                  "create or replace table %s (\n"
+                      + "                                    num_2_1 NUMBER(2, 1),\n"
```
sfc-gh-azagrebin, line 173: I would test it for each data type separately.

sfc-gh-azagrebin (commented):
Thanks Gloria!

I also think we need to create a follow-up JIRA to review our size estimates for Parquet.
Eventually we might want to abstract it: parquet value, type -> size
because it will depend on page version and encodings.

Anyways we indeed do not need the estimated size here because `.withMinRowCountForPageSizeCheck(Integer.MAX_VALUE)` disables multiple row groups and multiple pages.

Once Parquet scanner fix for multiple page support is in prod and we tested it, we can remove `.withMinRowCountForPageSizeCheck(Integer.MAX_VALUE)` but we still have to force one row group. We will have to do by byte size then. Hence we can already now set row group size limit to something big like 2x BLOB size (better to revisit parquet size estimates by then) instead of `ParquetWriter.DEFAULT_BLOCK_SIZE` in `ParquetFileWriter ` and `InternalParquetRecordWriter `. This looks safe because I do not see any memory allocation for this config.

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
sfc-gh-gdoci: I removed the estimate because we don't even need to set `withPageSize(limit)`. The line I added `.withMinRowCountForPageSizeCheck(Integer.MAX_VALUE)` will disable both checks (1) when to flush to rowgroup and then (2) when to flush column data to pages. The page size limit we set doesn't have any effect on the decision when to flush, it just allocates memory with that initial size.

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
sfc-gh-gdoci: > thanks, what if they see the size is bigger than the size limit after compression?

IIUC there is no size check after compression, only before to decide when to start a page.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -117,7 +117,8 @@ private void createFileWriter() {
     fileOutput = new ByteArrayOutputStream();
     try {
       if (enableParquetInternalBuffering) {
-        bdecParquetWriter = new BdecParquetWriter(fileOutput, schema, metadata, channelName);
+        bdecParquetWriter =
+            new BdecParquetWriter(fileOutput, schema, metadata, channelName, getSize());
```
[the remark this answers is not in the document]
sfc-gh-gdoci: I removed the size estimate completely.

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestBigFilesIT.java
sfc-gh-gdoci: no, also one table with many channels and rows can produce multiple rowgroups. I just increased it to 2 to cover multi tables case as well.

sfc-gh-gdoci:
Thanks for the reviews! I added a TODO and linked the Jira for size estimates. Since we don't need any size estimates at the moment, I think we can go with the defaults for now and change them later when the scanner fixes are deployed and tested.

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestBigFilesIT.java
sfc-gh-gdoci: sure, done.

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestBigFilesIT.java
sfc-gh-gdoci: yes, I tried the same configuration but not concurrent and it produced many small BDEC files, < 20 MB.

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestBigFilesIT.java
sfc-gh-azagrebin: This is probably not urgent but more relevant if we start to rely on any byte size again. Especially after revisiting Parquet size calculation.

src/test/java/net/snowflake/ingest/TestUtils.java
```diff
@@ -24,17 +25,13 @@
...
-import java.sql.Connection;
-import java.sql.DriverManager;
-import java.sql.ResultSet;
-import java.sql.Statement;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Optional;
-import java.util.Properties;
+import java.sql.*;
+import java.util.*;
```
sfc-gh-azagrebin, line 29: We do not use * imports (IDE setting). I already got a a lot of comments to revert * imports :)

sfc-gh-azagrebin (approved):

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestBigFilesIT.java
sfc-gh-gdoci: Agree.

src/test/java/net/snowflake/ingest/TestUtils.java
sfc-gh-gdoci: argh, thanks! I changed my IDE settings related to that and reverted the * imports.

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestBigFilesIT.java
sfc-gh-gdoci: `StreamingIngestIT` has another create database that this one doesn't have. Also not everything in the `beforeAll` can be extracted in one method, just four statements (create database, schema, set session tz and warehouse) could be de-duped, I don't think that will help much.

PR #325, opened 2023-02-01

sfc-gh-wfateem:
The JDBC driver was upgraded to 3.13.18 and the HttpUtil code was changed to take into account that a user may need to bypass the proxy by providing the -Dhttp.nonProxyHosts JVM argument


sfc-gh-tjones (approved):
LGTM since this is the same PR that was opened before. Thanks!

sfc-gh-tzhang (approved):
LGTM, Thanks!

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestStageTest.java
```diff
@@ -426,6 +433,44 @@ public void testGenerateProxyPropertiesForJDBC() {
...
         System.setProperty(HTTP_PROXY_PASSWORD, oldPassword);
       }
+      if (oldNonProxyHosts != null) {
+        System.setProperty(NON_PROXY_HOSTS, oldNonProxyHosts);
+      }
+    }
+  }
+
+  @Test
+  public void testShouldBypassProxy() {
```
sfc-gh-japatel, line 443: thanks for adding tests!

sfc-gh-japatel (approved):

PR #329, opened 2023-02-07
SNOW-663704 Send hex-encoded min/max for BINARY

sfc-gh-lsembera:
When populating min/max BINARY values, we should send them hex-encoded to the server, just like in case of strings. The server-side is ready to accept hex-encoded min/max values for BINARY since the release 6.36.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -7,7 +7,7 @@
     <!-- Arifact name and version information -->
     <groupId>net.snowflake</groupId>
     <artifactId>snowflake-ingest-sdk</artifactId>
-    <version>1.0.2-beta.8</version>
+    <version>1.0.3-beta</version>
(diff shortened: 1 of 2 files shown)


.github/workflows/End2EndTest.yml
```diff
@@ -8,7 +8,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
```
sfc-gh-lsembera, line 11: Without this, the builds are failing.

sfc-gh-tjones: April fool's day I guess?

sfc-gh-tjones (approved):

PR #330, opened 2023-02-07
NO-SNOW: various fixes/improvements in the SDK

sfc-gh-tzhang:
This PR contains multiple fixes/improvements:
- Include client version in the failure event
- Add some basic verifications in the generated EP
- Add logging to debug a stats mismatch issue
- Add server side failure status code to client telemetry as well

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/EpInfo.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/EpInfo.java
@@ -1,7 +1,11 @@
 package net.snowflake.ingest.streaming.internal;
 
+import static net.snowflake.ingest.utils.Constants.EP_NDV_UNKNOWN;
+
 import com.fasterxml.jackson.annotation.JsonProperty;
 import java.util.Map;
+import net.snowflake.ingest.utils.ErrorCode;
+import net.snowflake.ingest.utils.SFException;
 
 /** Class used to serialize/deserialize EP information */
 class EpInfo {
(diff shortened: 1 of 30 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/EpInfo.java
```diff
@@ -17,6 +21,30 @@ class EpInfo {
     this.columnEps = columnEps;
   }
 
+  /** Some basic verification logic to make sure the EP info is correct */
+  public void verifyEpInfo() {
```
sfc-gh-tjones, line 25: FYI we have this check server side now as well. We could always remove it here if you want?

sfc-gh-tzhang: This is before the file is built and customer can see it right away so they can save the log and report it, it doesn't hurt to verify it twice but i feel like adding the check at server side is too late.

sfc-gh-lsembera, line 25: We could have more here: min is always smaller than max (both int and string), neither min nor max strings are longer than 32 bytes (in hex),

sfc-gh-lsembera (approved):

src/main/java/net/snowflake/ingest/streaming/internal/EpInfo.java
sfc-gh-tzhang: We could add these later as we go, will merge this first to have a new release

PR #332, opened 2023-02-10
Remove JAXB

sfc-gh-lsembera:
This PR remove the dependency on `javax.xml.bind` that we used for hex/encoding decoding. We are going to use commons-codec instead.

PR is based on https://github.com/snowflakedb/snowflake-ingest-java/pull/328.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
@@ -29,11 +29,11 @@ import java.util.zip.GZIPOutputStream;
 import javax.crypto.BadPaddingException;
 import javax.crypto.IllegalBlockSizeException;
 import javax.crypto.NoSuchPaddingException;
-import javax.xml.bind.DatatypeConverter;
 import net.snowflake.ingest.utils.Constants;
 import net.snowflake.ingest.utils.Cryptor;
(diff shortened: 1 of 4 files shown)


sfc-gh-tjones (approved):

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -319,7 +319,7 @@ static String computeMD5(byte[] data, int length) throws NoSuchAlgorithmExceptio
         MessageDigest.getInstance("MD5");
     md.update(data, 0, length);
     byte[] digest = md.digest();
-    return DatatypeConverter.printHexBinary(digest).toLowerCase();
+    return Hex.encodeHexString(digest);
```
sfc-gh-tzhang, line 322: Is the result always the same by these two functions?

sfc-gh-tzhang:
> PR based on #328.

Let's have a meaningful description as well, thanks!

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
sfc-gh-lsembera: Yes, I verified that they are the same, `Hex.encodeHexString(String)` returns lower-case hexadecimal string.

PR #333, opened 2023-02-13
SNOW-738614 improve uncompressed size estimate for Parquet row buffers

sfc-gh-gdoci:
In this PR we try to make better estimates of uncompressed, but encoded data that will be written in Parquet pages. It means taking into consideration extra data/metadata added by Parquet when encoding the values. This can lead to a more timely-accurate decision of when to flush the rows to BDEC files.

Parquet encoding details: https://parquet.apache.org/docs/file-format/data-pages/encodings/

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ParquetValueParser.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ParquetValueParser.java
@@ -52,7 +52,10 @@ class ParquetValueParser {
       RowBufferStats stats,
       ZoneId defaultTimezone) {
     Utils.assertNotNull("Parquet column stats", stats);
-    float size = 0F;
+    float estimatedParquetSize = 0F;
+    estimatedParquetSize +=
+        (1 + 0.125f); // 1 byte + 1 bit for the definition level value in the RLE+bitpack encoded
+    // Parquet column
     if (value != null) {
       AbstractRowBuffer.ColumnLogicalType logicalType =
           AbstractRowBuffer.ColumnLogicalType.valueOf(columnMetadata.getLogicalType());
@@ -64,7 +67,8 @@ class ParquetValueParser {
               DataValidationUtil.validateAndParseBoolean(columnMetadata.getName(), value);
           value = intValue > 0;
           stats.addIntValue(BigInteger.valueOf(intValue));
-          size = 1;
+          estimatedParquetSize =
+              0.125f; // Parquet uses BitPacking for encoding boolean, hence 1 bit
           break;
         case INT32:
           int intVal =
@@ -77,7 +81,7 @@ class ParquetValueParser {
                   physicalType);
           value = intVal;
           stats.addIntValue(BigInteger.valueOf(intVal));
-          size = 4;
+          estimatedParquetSize = 4;
           break;
         case INT64:
           long longValue =
@@ -91,25 +95,27 @@ class ParquetValueParser {
                   defaultTimezone);
           value = longValue;
           stats.addIntValue(BigInteger.valueOf(longValue));
-          size = 8;
+          estimatedParquetSize = 8;
           break;
         case DOUBLE:
           double doubleValue =
               DataValidationUtil.validateAndParseReal(columnMetadata.getName(), value);
           value = doubleValue;
           stats.addRealValue(doubleValue);
-          size = 8;
+          estimatedParquetSize = 8;
           break;
         case BINARY:
           if (logicalType == AbstractRowBuffer.ColumnLogicalType.BINARY) {
             value = getBinaryValueForLogicalBinary(value, stats, columnMetadata);
-            size = ((byte[]) value).length;
+            estimatedParquetSize = ((byte[]) value).length;
           } else {
             String str = getBinaryValue(value, stats, columnMetadata);
             value = str;
             if (str != null) {
-              size = str.getBytes().length;
+              estimatedParquetSize = str.getBytes().length;
             }
+            estimatedParquetSize +=
+                4; // Parquet stores length in 4 bytes before the actual data bytes
           }
           break;
         case FIXED_LEN_BYTE_ARRAY:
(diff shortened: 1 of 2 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/ParquetValueParser.java
```diff
@@ -77,7 +81,7 @@ static ParquetBufferValue parseColumnValueToParquet(
                   physicalType);
           value = intVal;
           stats.addIntValue(BigInteger.valueOf(intVal));
-          size = 4;
+          estimatedParquetSize = 4;
```
sfc-gh-tzhang, line 84: what's the reason of using = here instead of +=?

sfc-gh-tzhang (approved):
Left one question, otherwise LGTM!

src/main/java/net/snowflake/ingest/streaming/internal/ParquetValueParser.java
sfc-gh-gdoci: You are right, it should be `+=` everywhere. I had the `1 + 0.125` at the end and forgot to change all `+` to `+=` when moving that line to the top. Fixed it.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetValueParser.java
```diff
@@ -91,25 +95,27 @@ static ParquetBufferValue parseColumnValueToParquet(
...
           } else {
             String str = getBinaryValue(value, stats, columnMetadata);
             value = str;
             if (str != null) {
-              size = str.getBytes().length;
+              estimatedParquetSize += str.getBytes().length;
             }
+            estimatedParquetSize +=
+                4; // Parquet stores length in 4 bytes before the actual data bytes
           }
```
sfc-gh-azagrebin, line 119:
```
        int length = 0;
        if (logicalType == AbstractRowBuffer.ColumnLogicalType.BINARY) {
          value = getBinaryValueForLogicalBinary(value, stats, columnMetadata);
          length = ((byte[]) value).length;
        } else {
          String str = getBinaryValue(value, stats, columnMetadata);
          value = str;
          if (str != null) {
            length = str.getBytes().length;
          }
        }
        if (value != null) {
          // Parquet stores length in 4 bytes before the actual data bytes
          estimatedParquetSize += 4 + length;
        }
```
Otherwise 4 is not counted for `ColumnLogicalType.BINARY` but counted for str that is null.

src/test/java/net/snowflake/ingest/streaming/internal/ParquetValueParserTest.java
```diff
@@ -14,6 +14,9 @@
 
 public class ParquetValueParserTest {
 
+  public static final float DEFINITION_LEVEL_ENCODING_BYTES = 1 + 1.0f / 8;
+  private static final int BYTE_ARRAY_LENGTH_ENCODING_BYTES = 4;
```
sfc-gh-azagrebin, line 18: I would put these statics and also use in the `ParquetValueParser`:
```
public static final float BIT_ENCODING_BYTE_LEN = 1.0f / 8;
public static final float DEFINITION_LEVEL_ENCODING_BYTE_LEN = 2.0f / 8;
public static final int BYTE_ARRAY_LENGTH_ENCODING_BYTE_LEN = 4;
```

sfc-gh-azagrebin (commented):
Thanks Gloria!
btw, do we test TEXT logical type somewhere?
I could not spot it

src/main/java/net/snowflake/ingest/streaming/internal/ParquetValueParser.java
```diff
@@ -52,7 +52,10 @@ static ParquetBufferValue parseColumnValueToParquet(
       RowBufferStats stats,
       ZoneId defaultTimezone) {
     Utils.assertNotNull("Parquet column stats", stats);
-    float size = 0F;
+    float estimatedParquetSize = 0F;
+    estimatedParquetSize +=
+        (1 + 0.125f); // 1 byte + 1 bit for the definition level value in the RLE+bitpack encoded
```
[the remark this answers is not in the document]
sfc-gh-gdoci: We assume the worst case, right? `null (0), not null (1), null (0), not null (1)...` that will be encoded to `10 11 10  1 ...`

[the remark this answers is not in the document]
sfc-gh-azagrebin: I think this will be encoded as 2 bytes:
| length byte (e.g. 8) | data byte |
where data byte is 8 bits: 01010101

src/main/java/net/snowflake/ingest/streaming/internal/ParquetValueParser.java
sfc-gh-gdoci: done.

src/test/java/net/snowflake/ingest/streaming/internal/ParquetValueParserTest.java
sfc-gh-gdoci: done.

sfc-gh-gdoci:
> Thanks Gloria! btw, do we test TEXT logical type somewhere? I could not spot it

Thanks for the review :)

src/main/java/net/snowflake/ingest/streaming/internal/ParquetValueParser.java
```diff
@@ -17,6 +17,17 @@
 
 /** Parses a user column value into Parquet internal representation for buffering. */
 class ParquetValueParser {
+
+  // Parquet uses BitPacking to encode boolean, hence 1 bit per value
+  public static final float BIT_ENCODING_BYTE_LEN = 1.0f / 8;
+
+  /* (on average) 2 bytes / 8 values for the RLE+bitpack encoded definition level.
+  Since we don't have nested types, repetition level is always 0 and is not stored at all by Parquet. */
```
sfc-gh-azagrebin, line 25: Could you add details about justification that we discussed? that we have 2 cases:
- repeat count + repeated value bits
- value count + bits of values
with links to parquet class and methods
add byte/bit examples

sfc-gh-azagrebin, line 25: Could you add details about justification that we discussed? that we have 2 cases:
- repeat count + repeated value bits
- value count + bits of values
with links to parquet class and methods
add byte/bit examples

sfc-gh-azagrebin (approved):
Thanks Gloria!

PR #334, opened 2023-02-16
NO-SNOW: revert https://github.com/snowflakedb/snowflake-kafka-connector/pull/546/files as it's causing failure on GCS

sfc-gh-tzhang:
original commit: a12afc32e5624e6b7377d871d1e7dc8b201e6ad8

PR #335, opened 2023-02-17
V1.0.3-beta.1 release

sfc-gh-tzhang:
The 1.0.3-beta release has an issue with GCS, so we have to revert https://github.com/snowflakedb/snowflake-ingest-java/pull/334 and then create a new release on top

### DIFF
--- a/src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
+++ b/src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
@@ -4,9 +4,22 @@
 
 package net.snowflake.ingest.connection;
 
+import static net.snowflake.ingest.utils.Constants.ENABLE_TELEMETRY_TO_SF;
+import static net.snowflake.ingest.utils.Utils.isNullOrEmpty;
+
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import java.io.FileNotFoundException;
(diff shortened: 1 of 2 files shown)


src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
```diff
@@ -36,6 +23,16 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.FileNotFoundException;
+import java.io.InputStream;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.security.KeyPair;
+import java.util.*;
```
sfc-gh-rcheng, line 31: nit: unsure why these imports were changed?

sfc-gh-rcheng (approved):
lgtm, one nit, but probably caught by the formatting gate

src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
sfc-gh-japatel: +1 no wildcard import. Probably some intellij magic.

sfc-gh-japatel (approved):
lgtm

sfc-gh-tjones (approved):

src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
sfc-gh-tzhang: will update, i update one intellJ so all the settings seems get lost

PR #336, opened 2023-02-21
SNOW-745946 SNOW-735517 Create defensive copy of ingested rows

sfc-gh-lsembera:
As a protection against concurrent modifications of the input map,  this PR creates a defensive shallow copy of the map, as well as deep copy of its mutable values (currently only `byte[]` for `BINARY` columns). Concurrent modifications could cause issues like data/metadata mismatch or double counting of `null` values.

For semi-structured types, we accept more mutable values (`List` and `Map`), but for these values we do not calculate min/max EP values, so there is no risk of a mismatch between data and metadata.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
@@ -91,14 +91,6 @@ class DataValidationUtil {
   }
 
   /**
-   * Creates a new SnowflakeDateTimeFormat. In order to avoid SnowflakeDateTimeFormat's
-   * synchronization blocks, we create a new instance when needed instead of sharing one instance.
-   */
-  private static SnowflakeDateTimeFormat createDateTimeFormatter() {
-    return SnowflakeDateTimeFormat.fromSqlFormat("auto");
-  }
-
-  /**
    * Validates and parses input as JSON. All types in the object tree must be valid variant types,
    * see {@link DataValidationUtil#isAllowedSemiStructuredType}.
    *
@@ -590,7 +582,12 @@ class DataValidationUtil {
       String columnName, Object input, Optional<Integer> maxLengthOptional) {
     byte[] output;
     if (input instanceof byte[]) {
-      output = (byte[]) input;
+      // byte[] is a mutable object, we need to create a defensive copy to protect against
+      // concurrent
+      // modifications of the array, which could lead to mismatch between data and metadata
+      byte[] originalInputArray = (byte[]) input;
+      output = new byte[originalInputArray.length];
+      System.arraycopy(originalInputArray, 0, output, 0, originalInputArray.length);
     } else if (input instanceof String) {
       try {
         output = Hex.decodeHex((String) input);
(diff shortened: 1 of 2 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -90,14 +90,6 @@ class DataValidationUtil {
     objectMapper.registerModule(module);
   }
 
-  /**
-   * Creates a new SnowflakeDateTimeFormat. In order to avoid SnowflakeDateTimeFormat's
-   * synchronization blocks, we create a new instance when needed instead of sharing one instance.
-   */
-  private static SnowflakeDateTimeFormat createDateTimeFormatter() {
-    return SnowflakeDateTimeFormat.fromSqlFormat("auto");
```
sfc-gh-tjones, line 98: unused?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
```diff
@@ -353,7 +355,16 @@ public InsertValidationResponse insertRows(
...
     }
 
-    InsertValidationResponse response = this.rowBuffer.insertRows(rows, offsetToken);
+    // We create a shallow copy to protect against concurrent addition/removal of columns, which can
+    // lead to double counting of null values, for example. Individual mutable values may still be
+    // concurrently modified (e.g. byte[]). Before validation and EP calculation, we must make sure
+    // that defensive copies of all mutable objects are created.
+    final List<Map<String, Object>> rowsCopy = new LinkedList<>();
+    for (Map<String, Object> row : rows) {
+      rowsCopy.add(new HashMap<>(row));
```
sfc-gh-tjones, line 364: Nit, can reduce to `rows.forEach(r -> rowsCopy.add(new HashMap<>(r)))`;

sfc-gh-tjones (approved):

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-lsembera: Yeah, this was an unused method from the previous release.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
sfc-gh-lsembera: Will fix in a separate PR, I want to merge this one quickly, so it makes it to the next release.

sfc-gh-tjones: SGTM!

sfc-gh-lsembera:
@sfc-gh-tzhang This PR has just been merge, would you include it into `1.0.3-beta.1`, please?

sfc-gh-tzhang:
> @sfc-gh-tzhang This PR has just been merge, would you include it into `1.0.3-beta.1`, please?

Unfortunately this is merged after the version upgrade change, but it will included in the PuPr release. If you feel like this is a risky change, please revert it and add it back later as we really need a stable release as the first PuPr release.

PR #337, opened 2023-02-22
SNOW-744974: avoid combining channel data with different encryption key to the same chunk

sfc-gh-tzhang:
We need this behavior in two cases:
- when there is a key rotation and it's possible that old channels and new channels are using different encryption key
- when the table is dropped and recreated with the same name, we would want to separate the data from old channels and new channels because the old channels data will be invalidated


src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -338,33 +338,72 @@ void distributeFlushTasks() {
                 String, ConcurrentHashMap<String, SnowflakeStreamingIngestChannelInternal<T>>>>
         itr = this.channelCache.iterator();
     List<Pair<BlobData<T>, CompletableFuture<BlobMetadata>>> blobs = new ArrayList<>();
+    List<ChannelData<T>> leftoverChannelsDataPerTable = new ArrayList<>();
```
sfc-gh-tjones, line 341: Is it possible to forward channel data such that it causes us to go over the size threshold in the next batch?

sfc-gh-tzhang: good question, if you see line 353, we will add everything in `leftoverChannelsDataPerTable` to `channelsDataPerTable`, then the same split logic will apply again on `channelsDataPerTable`

sfc-gh-tjones: Got it, so is it possible to continually move data around if there's a high ingest rate? Probably an edge case but something I wanted to call out.

sfc-gh-tzhang: I guess the word "split" is not accurate, we basically have an array of chunks and we will see how many blobs we need so the size of each blob stay below the limit, so there is no "data moving", but if the ingest rate is high and we don't have enough threads to build the actual blobs, that will be an issue and we will see a high build latency.

sfc-gh-tjones: Got it - makes sense.

sfc-gh-tjones (approved):

PR #338, opened 2023-02-24
V1.1.0 release

sfc-gh-tzhang:
This will be the first non-beta release after v0.10.8 and it contains the support to do Snowpipe Streaming

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -7,7 +7,7 @@
     <!-- Arifact name and version information -->
     <groupId>net.snowflake</groupId>
     <artifactId>snowflake-ingest-sdk</artifactId>
-    <version>1.0.3-beta.1</version>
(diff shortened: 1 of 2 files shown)


sfc-gh-japatel (approved):
YAYYY! SHIP IT

sfc-gh-tjones (approved):

sfc-gh-xhuang:
Do we want https://github.com/snowflakedb/snowflake-ingest-java/pull/337 in this too?

We should rebase at x time as the final cutoff for 1.1.0 today

sfc-gh-tzhang:
> Do we want #337 in this too?
> 
> We should rebase at x time as the final cutoff for 1.1.0 today

No, that only fixes a very rare case and I feel it's safer to include it after the PuPr release

PR #339, opened 2023-02-24
SNOW-749216 Allow leading and trailing whitespace

sfc-gh-lsembera:
Allow surrounding whitespace around string input values for various data types, just like Snowflake Worksheets do.


sfc-gh-tjones (approved):
Looks great, thanks for the changes Lukas!

PR #345, opened 2023-03-06
Reduce info logging around JWT token regeneration

sfc-gh-rcheng:
Combined INFO logging for JWT properties and moved to where the properties are set. Reduced the logging level to DEBUG.

Per this customer issue: https://github.com/snowflakedb/snowflake-kafka-connector/issues/526

### DIFF
--- a/src/main/java/net/snowflake/ingest/connection/SecurityManager.java
+++ b/src/main/java/net/snowflake/ingest/connection/SecurityManager.java
@@ -154,14 +154,12 @@ final class SecurityManager implements AutoCloseable {
     // create our JWT claim builder object
     JWTClaimsSet.Builder builder = new JWTClaimsSet.Builder();
 
-    // set the subject to the fully qualified username
+    // get the subject to the fully qualified username
     String subject = String.format("%s.%s", account, user);
-    LOGGER.info("Creating Token with subject {}", subject);
 
-    // set the issuer
+    // get the issuer
     String publicKeyFPInJwt = calculatePublicKeyFp(keyPair);
     String issuer = String.format("%s.%s.%s", account, user, publicKeyFPInJwt);
-    LOGGER.info("Creating Token with issuer {}", issuer);
 
     // iat set to now
     Date iat = new Date(System.currentTimeMillis());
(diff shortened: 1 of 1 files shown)


src/main/java/net/snowflake/ingest/connection/SecurityManager.java
```diff
@@ -154,14 +154,12 @@ private void regenerateToken() {
...
     JWTClaimsSet.Builder builder = new JWTClaimsSet.Builder();
 
-    // set the subject to the fully qualified username
+    // get the subject to the fully qualified username
     String subject = String.format("%s.%s", account, user);
-    LOGGER.info("Creating Token with subject {}", subject);
 
-    // set the issuer
+    // get the issuer
     String publicKeyFPInJwt = calculatePublicKeyFp(keyPair);
```
sfc-gh-rcheng, line 161: could 'calculatePublicKeyFp' ever fail? If so would we need the subject property in the log for debugging? If so, I can revert the changes and just set the log level to debug instead of info

sfc-gh-tjones (approved):

src/main/java/net/snowflake/ingest/connection/SecurityManager.java
sfc-gh-japatel: It can throw RTE, I would rather keep them two separate to have more information for debugging if we ever need that.

sfc-gh-japatel (approved):
lgtm!

sfc-gh-tzhang (commented):
The change itself looks ok, but should we first understand if this is a regression, maybe asking for a full log or see if we have a local repro? We haven't change this code for a very long time so I don't think this is the issue.

sfc-gh-rcheng:
They called out this specifically in the issue and I confirmed that we create the JWT token multiple times, not just during setup (JWT regeneration/refreshing i think). 

I'll follow up with them for the full log, in case they're having JWT issues or if this was just an example of too many info logs.

sfc-gh-tzhang:
> They called out this specifically in the issue and I confirmed that we create the JWT token multiple times, not just during setup (JWT regeneration/refreshing i think).
> 
> I'll follow up with them for the full log, in case they're having JWT issues or if this was just an example of too many info logs.

But I think we only do refresh once 54 minutes, see RENEWAL_INTERVAL, seems like an issue if you see many log entires

PR #349, opened 2023-03-15
SNOW-735517 Remove nullCount <= rowCount sanity check

sfc-gh-lsembera:
In order to be notified when SNOW-735517 occurs, remove the check from SDK. It is going to be caught server-side during blob registration request.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/EpInfo.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/EpInfo.java
@@ -26,14 +26,6 @@ class EpInfo {
     for (Map.Entry<String, FileColumnProperties> entry : columnEps.entrySet()) {
       String colName = entry.getKey();
       FileColumnProperties colEp = entry.getValue();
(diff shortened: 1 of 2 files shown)


sfc-gh-tzhang:
I don't think we want to do this, as it's generating noise for the metadata team, also we need to disable compactor and contact the customer to drop the table. If the purpose is to be notified when this case happens, we could send the exception through the client telemetry to server side and setup an alert on this

sfc-gh-tjones:
> I don't think we want to do this, as it's generating noise for the metadata team, also we need to disable compactor and contact the customer to drop the table. If the purpose is to be notified when this case happens, we could send the exception through the client telemetry to server side and setup an alert on this

They don't get notified, we do because the component is set to `DP-Streaming Ingest`. Customers can disable client telemetry and this allows us to know immediately when we hit an occurence.

sfc-gh-tzhang:
closing since we're identified the root cause, thanks @sfc-gh-lsembera !

PR #352, opened 2023-03-16
SNOW-761399: exclude nimbus-jose-jwt from hadoop-common dependency

sfc-gh-dwang:
This is to exclude nimbus-jose-jwt from hadoop-common dependency as we already introduces a newer version of nimbus-jose-jwt in direct dependency


sfc-gh-tjones (approved):
Looks good but let's have Toby review.

sfc-gh-tzhang (approved):
LGTM, thanks

PR #353, opened 2023-03-17
@snow SNOW-759764 Streaming Ingest Parquet: use Snowflake column ordinal

sfc-gh-azagrebin:
Currently, Parquet file generator does not use Snowflake internal column ordinal.
It uses the column index from the column list that comes from server side.
It works when the table just created because the original ordinals are initially the column indexes.
Nonetheless, this breaks table column evolution (add/drop) because the column ordinal is not the index anymore.

@test
`StreamingIngestIT.testTableColumnEvolution`
the test is disabled until the server side fix is released in 7.10.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ParquetTypeGenerator.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ParquetTypeGenerator.java
@@ -60,10 +60,11 @@ public class ParquetTypeGenerator {
    * side.
    *
    * @param column column metadata as received from server side
-   * @param id column id
+   * @param id column id if column.getOrdinal() is not available
    * @return column parquet type
    */
   static ParquetTypeInfo generateColumnParquetTypeInfo(ColumnMetadata column, int id) {
+    id = column.getOrdinal() == null ? id : column.getOrdinal();
     ParquetTypeInfo res = new ParquetTypeInfo();
     Type parquetType;
     Map<String, String> metadata = new HashMap<>();
--- a/src/test/java/net/snowflake/ingest/streaming/internal/ColumnMetadataBuilder.java
+++ b/src/test/java/net/snowflake/ingest/streaming/internal/ColumnMetadataBuilder.java
@@ -17,6 +17,8 @@ public class ColumnMetadataBuilder {
   private boolean nullable;
   private String collation;
 
+  private Integer ordinal;
+
   /**
    * Returns a new ColumnMetadata builder
    *
@@ -143,6 +145,17 @@ public class ColumnMetadataBuilder {
   }
 
   /**
+   * Set column ordinal
+   *
+   * @param ordinal ordinal
+   * @return columnMetadataBuilder object
+   */
+  public ColumnMetadataBuilder ordinal(int ordinal) {
+    this.ordinal = ordinal;
+    return this;
+  }
+
+  /**
    * Build a columnMetadata object from the set values
    *
    * @return columnMetadata object
@@ -159,6 +172,7 @@ public class ColumnMetadataBuilder {
     colMetadata.setScale(scale);
     colMetadata.setPrecision(precision);
     colMetadata.setCollation(collation);
+    colMetadata.setOrdinal(ordinal);
     return colMetadata;
   }
 }
--- a/src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
+++ b/src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
@@ -3,6 +3,8 @@ package net.snowflake.ingest.streaming.internal;
 import static net.snowflake.ingest.utils.Constants.BLOB_NO_HEADER;
 import static net.snowflake.ingest.utils.Constants.COMPRESS_BLOB_TWICE;
 import static net.snowflake.ingest.utils.Constants.ROLE;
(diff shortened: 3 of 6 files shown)


src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
```diff
@@ -36,10 +38,7 @@
 import net.snowflake.ingest.utils.ErrorCode;
 import net.snowflake.ingest.utils.ParameterProvider;
 import net.snowflake.ingest.utils.SFException;
-import org.junit.After;
-import org.junit.Assert;
-import org.junit.Before;
-import org.junit.Test;
+import org.junit.*;
```
sfc-gh-lsembera, line 41: nit, wildcard import

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
```diff
@@ -994,6 +993,119 @@ public void testColumnNameQuotes() throws Exception {
...
+
+    // remove last column
+    executeQuery(String.format("alter table %s drop column c4;", tableFullName));
+    ingestByColIndexAndVerifyTable(tableName, channelNum, Arrays.asList(2, 3), rowNum);
+    channelNum++;
+
+    // add last column
+    executeQuery(String.format("alter table %s add column c5 varchar;", tableFullName));
+    ingestByColIndexAndVerifyTable(tableName, channelNum, Arrays.asList(2, 3, 5), rowNum);
+  }
```
sfc-gh-lsembera, line 1028: Can we also test with changetracking columns?

sfc-gh-lsembera (approved):
Thank you!

src/main/java/net/snowflake/ingest/streaming/internal/ParquetTypeGenerator.java
```diff
@@ -60,10 +60,11 @@ public void setMetadata(Map<String, String> metadata) {
    * side.
    *
    * @param column column metadata as received from server side
-   * @param id column id
+   * @param id column id if column.getOrdinal() is not available
    * @return column parquet type
    */
   static ParquetTypeInfo generateColumnParquetTypeInfo(ColumnMetadata column, int id) {
+    id = column.getOrdinal() == null ? id : column.getOrdinal();
```
sfc-gh-gdoci, line 67: We will remove the `int id` when the server side fix is deployed, right?

sfc-gh-azagrebin: yes, I think so

sfc-gh-gdoci (approved):
Thanks Andrey!

src/main/java/net/snowflake/ingest/streaming/internal/ParquetTypeGenerator.java
sfc-gh-azagrebin: we need this for the backwards server side compatibility for the release rollback period

sfc-gh-tjones (approved):

src/test/java/net/snowflake/ingest/streaming/internal/ColumnMetadataBuilder.java
```diff
@@ -17,6 +17,8 @@ public class ColumnMetadataBuilder {
   private boolean nullable;
   private String collation;
```
sfc-gh-tzhang, line 19: nit: remove space?

sfc-gh-tzhang (commented):
LGTM! We should wait for the server side change to be deployed before merging this, otherwise any new releases we do in between will need to exclude this one manually

sfc-gh-azagrebin:
[Precommit](https://buwa.ci1.us-west-2.aws-dev.app.snowflake.com/builds/precommit/ci-dev-131.int.snowflakecomputing.com/8133)

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
```diff
@@ -1019,7 +1147,7 @@ private SnowflakeStreamingIngestChannel openChannel(String tableName, String cha
             .setDBName(testDb)
             .setSchemaName(TEST_SCHEMA)
             .setTableName(tableName)
-            .setOnErrorOption(OpenChannelRequest.OnErrorOption.CONTINUE)
+            .setOnErrorOption(OpenChannelRequest.OnErrorOption.ABORT)
```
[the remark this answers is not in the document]
sfc-gh-azagrebin: I think it is better to fail fast

sfc-gh-azagrebin:
> We should wait for the server side change to be deployed before merging this, otherwise any new releases we do in between will need to exclude this one manually

Sorry, merged too fast. I checked it works with the server side w/o fix.

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
```diff
@@ -994,6 +997,131 @@ public void testColumnNameQuotes() throws Exception {
     Assert.fail("Row sequencer not updated before timeout");
   }
 
+  @Ignore
```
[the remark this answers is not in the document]
sfc-gh-azagrebin: https://snowflakecomputing.atlassian.net/browse/SNOW-763199

PR #355, opened 2023-03-19
Add maven-dependency-plugin

Preston4tw:
Adding maven-dependency plugin surfaces these dependency issues, also fixed by this PR:

- Declare used undeclared dependencies
  - com.google.code.findbugs:jsr305:3.0.2
  - org.apache.parquet:parquet-common:1.12.3
  - com.fasterxml.jackson.core:jackson-core:2.14.0
  - io.netty:netty-common:4.1.82.Final
  - com.fasterxml.jackson.core:jackson-annotations:2.14.0
  - commons-codec:commons-codec:1.15
  - org.apache.arrow:arrow-memory-core:10.0.0
  - org.apache.parquet:parquet-column:1.12.3
  - org.apache.commons:commons-lang3:3.12.0
  - com.google.guava:guava:27.0-jre
- Remove unused declared dependency
  - org.apache.hadoop:hadoop-mapreduce-client-core:3.3.4:compile

A dependency is considered used when it is imported by source java files present in this project, and unused when not. It is good hygiene to declare a dependency when used to avoid relying on transitive dependencies.

Related Java Library Best Practice: https://jlbp.dev/JLBP-22

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -44,6 +44,7 @@
         <jacoco.version>0.8.5</jacoco.version>
         <arrow.version>10.0.0</arrow.version>
         <hadoop.version>3.3.4</hadoop.version>
+        <version.plugin.dependency>3.3.0</version.plugin.dependency>
     </properties>
 
     <build>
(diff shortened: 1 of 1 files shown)


sfc-gh-tzhang:
@Preston4tw Thanks for contributing to the repo and help us improve our dependency logic, I don't think our team has any expert on this area :)

I have given you access to the repo, do you want to open a PR against the original repo?

sfc-gh-pbennes:
Fixed in #359

PR #358, opened 2023-03-21
Formats pom.xml

sfc-gh-tjones:
This correctly formats pom.xml according to best practices. Nothing is changed other than formatting.

Verified via `mvn clean install -DskipTests=true`


sfc-gh-japatel (approved):
lgtm

sfc-gh-tzhang:
We plan to release 1.1.1 by end of this week which contains a few important fixes for customers, please wait for 1.1.1 to be released before merging, thanks!

PR #362, opened 2023-03-22
SNOW-766522 Cache unquoted column names

sfc-gh-lsembera:
The SDK is unnecessarily unquoting column names from user input again and again, even though they are (likely) always the same in each row. This PR introduces a loading cache of already computed unquoted column names to avoid unnecessary re-computations.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/LiteralQuoteUtils.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/LiteralQuoteUtils.java
@@ -46,7 +46,9 @@ class LiteralQuoteUtils {
       return unquotedColumnNamesCache.get(columnName);
     } catch (ExecutionException e) {
       throw new SFException(
-          e, ErrorCode.INTERNAL_ERROR, String.format("Exception thrown while unquoting column name %s", columnName));
+          e,
+          ErrorCode.INTERNAL_ERROR,
+          String.format("Exception thrown while unquoting column name %s", columnName));
     }
   }
 


sfc-gh-japatel (approved):
lgtm!

sfc-gh-tjones (approved):

src/main/java/net/snowflake/ingest/streaming/internal/LiteralQuoteUtils.java
```diff
@@ -11,6 +20,38 @@
...
+            .build(
+                new CacheLoader<String, String>() {
+                  @Override
+                  public String load(String key) {
+                    return unquoteColumnNameInternal(key);
+                  }
+                });
+  }
+
+  static String unquoteColumnName(String columnName) {
```
sfc-gh-tzhang, line 46: add comment?

sfc-gh-tzhang (approved):
LGTM, have we seen any performance improvement with this? We will release 1.1.1 and we don't want Informatica comes back to us complaining that this does not help :)

src/main/java/net/snowflake/ingest/streaming/internal/LiteralQuoteUtils.java
sfc-gh-lsembera: @sfc-gh-tzhang 

Comment added.

I locally micro-benchmarked unquoting of 500M quoted column names yesterday and for quoted columns there is ≈4.5x (≈40sec vs 1380sec) speedup. However, today I found that for simple, not quoted column names, there is ≈50% slowdown (25s vs 38s) - for simple columns it is cheaper to call `unquote` than go through the cache. Since the majority of column names is not quoted, I added `isUnquotingExpensive` heuristics, which just check first and last character (it is very fast) and if they are both quotes, the unquoting is considered expensive and it will cache the result.

sfc-gh-lsembera: I ran jmh micro-benchmarks that test unquote performance for various column names. Every iteration cleared the cache and unquoted 10M pre-created column names, here are the results:

```
Column format: COLUMN_1_NAME_EXAMPLE[0-1000]
Benchmark                          Mode  Cnt  Score   Error  Units
JmhBench.benchmarkAlwaysRecompute  avgt    2  0.315           s/op
JmhBench.benchmarkFromCache        avgt    2  0.377           s/op

Column format: column_1_name_example[0-1000]
Benchmark                          Mode  Cnt  Score   Error  Units
JmhBench.benchmarkAlwaysRecompute  avgt    2  0.545           s/op
JmhBench.benchmarkFromCache        avgt    2  0.377           s/op

Column format: "column_1_name_example[0-1000]"
Benchmark                          Mode  Cnt  Score   Error  Units
JmhBench.benchmarkAlwaysRecompute  avgt    2  1.257           s/op
JmhBench.benchmarkFromCache        avgt    2  0.402           s/op

Column format: "column_""1_name""_example[0-1000]"
Benchmark                          Mode  Cnt  Score   Error  Units
JmhBench.benchmarkAlwaysRecompute  avgt    2  3.401           s/op
JmhBench.benchmarkFromCache        avgt    2  0.387           s/op
```


Depending on the input column name, the improvement is between 0.4x - 10x. Only for the optimal case when no transformation is needed (column is not quoted and user passes upper-cased input), there is a small regression (16%), which will be compensated later when we move to from guava cache to scaffeine. These are ballpark numbers for some example column names, the actual numbers will of course depend on the actual column names, number of quotes and their length, but it gives us some idea. 

In a followup PR, I would like to switch to scaffeine that, according to my preliminary benchmarking, will give us additional performance boost.

src/main/java/net/snowflake/ingest/streaming/internal/LiteralQuoteUtils.java
```diff
@@ -11,6 +18,38 @@
...
+  /**
+   * Unquote column name expected to be used from the outside. It decides is unquoting would be
+   * expensive. If not, it unquotes directly, otherwise it return a value from a loading cache.
+   */
+  static String unquoteColumnName(String columnName) {
+    try {
+      return unquotedColumnNamesCache.get(columnName);
+    } catch (ExecutionException e) {
+      throw new SFException(
+          e, ErrorCode.INTERNAL_ERROR, "Exception thrown while unquoting column name");
```
sfc-gh-tzhang, line 49: could we add the column name?

src/main/java/net/snowflake/ingest/streaming/internal/LiteralQuoteUtils.java
sfc-gh-tzhang: sounds good, let's make sure the security team is ok with the new library, @sfc-gh-japatel should know the process

sfc-gh-japatel: As long as it has a valid allowed license, i think we are good to go! https://lift.snowflake.com/$viewer.do?sysparm_stack=no&sysparm_sys_id=b1287f0e1b9c691404ad33fe034bcb84

sfc-gh-lsembera: Great, thank you @sfc-gh-japatel! If that is the case I will squeeze caffeine cache into 1.1.1.

sfc-gh-lsembera: Updated benchmarks for caffeine cache:

```
Column format: COLUMN_1_NAME_EXAMPLE[0-1000]
Benchmark                          Mode  Cnt  Score   Error  Units
JmhBench.benchmarkAlwaysRecompute  avgt    2  0.321           s/op
JmhBench.benchmarkFromCache        avgt    2  0.115           s/op

Column format: column_1_name_example[0-1000]
Benchmark                          Mode  Cnt  Score   Error  Units
JmhBench.benchmarkAlwaysRecompute  avgt    2  0.538           s/op
JmhBench.benchmarkFromCache        avgt    2  0.129           s/op


Column format: "column_1_name_example[0-1000]"
Benchmark                          Mode  Cnt  Score   Error  Units
JmhBench.benchmarkAlwaysRecompute  avgt    2  1.197           s/op
JmhBench.benchmarkFromCache        avgt    2  0.143           s/op

Column format: "column_""1_name""_example[0-1000]"
Benchmark                          Mode  Cnt  Score   Error  Units
JmhBench.benchmarkAlwaysRecompute  avgt    2  3.415           s/op
JmhBench.benchmarkFromCache        avgt    2  0.138           s/op
```

The numbers look much better with caffeine. Now we get ~3x speedup for the optimal case, ~8x speedup for the basic quoted case and 25x speedup for the worst case.

src/main/java/net/snowflake/ingest/streaming/internal/LiteralQuoteUtils.java
sfc-gh-lsembera: This line is not needed for caffeine because its `get` method doesn't throw checked exceptions.

sfc-gh-lsembera: Fixed

PR #366, opened 2023-03-24
SNOW-766522 Avoid duplicate calls to unquoteColumnName

sfc-gh-lsembera:
During the lifetime of an input row, we are calling `unquote` for each column name twice. Even though the values are cached now, it is unnecessary because in the input row, we can in-place replace the input column names with their unquoted variants.


sfc-gh-lsembera:
There is no need for this anymore, caffeine cache lookups are very fast.

PR #369, opened 2023-03-27
SNOW-766066 Run tests on Windows

sfc-gh-lsembera:
This PR adds a new GitHub Actions job `build-windows`, which runs our unit and integration tests on Windows. The job runs in parallel with our existing Linux job, so it shouldn't slow down the overall PR build time. The Windows job only executes tests, it does not check code coverage nor does it execute `check_content.sh`.

### DIFF
--- a/.github/workflows/End2EndTest.yml
+++ b/.github/workflows/End2EndTest.yml
@@ -36,3 +36,22 @@ jobs:
 
       - name: Code Coverage
         uses: codecov/codecov-action@v1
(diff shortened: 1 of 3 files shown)


sfc-gh-tjones (approved):

sfc-gh-tzhang (approved):
LGTM, thanks! I think a more important support we need to add is to run the tests against all cloud providers, I believe there is a JIRA to track it but I couldn't find it

PR #426, opened 2023-03-31
SNOW-773778 make the absolute size threshold limit for throttling configurable, default 200 MB

sfc-gh-gdoci:
This PR makes the size threshold limit for throttling configurable and increases the default value from `100MB` to `200MB`.


src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -172,6 +182,16 @@ public int getInsertThrottleThresholdInPercentage() {
     }
     return (int) val;
   }
+  /** @return Absolute size in bytes of free total memory at which we throttle row inserts */
+  public int getInsertThrottleThresholdInBytes() {
+    Object val =
+        this.parameterMap.getOrDefault(
+            INSERT_THROTTLE_THRESHOLD_IN_BYTES, INSERT_THROTTLE_THRESHOLD_IN_BYTES_DEFAULT);
+    if (val instanceof String) {
+      return Integer.parseInt(val.toString());
```
sfc-gh-tzhang, line 191: Any chance to add a test for this?

sfc-gh-tzhang (approved):
LGTM, thanks!

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-tjones, line 191: FYI Codecov is complaining about this.

sfc-gh-tjones (approved):
LGTM barring Toby's comments! Thank you for the quick fix!

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-gdoci: I extended the `ParameterProviderTest`. 
> Consider uploading reports for the commit https://github.com/snowflakedb/snowflake-ingest-java/commit/9f186da9e3036d941bf28abb53c785d8e8a03051 to get more accurate results

Any idea how can I make the CodeCov to give a new report?

PR #427, opened 2023-03-31
[SNOW-766867] Add blob level latency telemetry

sfc-gh-rcheng:


src/main/java/net/snowflake/ingest/connection/TelemetryService.java
```diff
@@ -73,7 +81,9 @@ public void close() {
     this.telemetry.close();
   }
 
-  /** Report the Streaming Ingest latency metrics */
+  /** Reports aggregated the Streaming Ingest latency metrics */
+  // TODO @rcheng question - can I rename the telemetry method and type to
```
sfc-gh-rcheng, line 85: I think it would be better to have the following, however I'm not sure if renaming telemetry types/names would break anything downstream

- reportAggregatedLatencyInSec
- reportBlobLatencyInSec

src/main/java/net/snowflake/ingest/connection/TelemetryService.java
```diff
@@ -86,6 +96,14 @@ public void reportLatencyInSec(
     }
   }
 
+  // TODO @rcheng question - is it useful to report all the blob metadata here, or just latencies?
```
sfc-gh-rcheng, line 99: Is blob metadata already reported elsewhere? If so do we need to report this telemetry? I'm worried about overloading the telemetry service with too many messages. Alternatively we could buffer or aggregate messages, but then the message size might be too large.

src/main/java/net/snowflake/ingest/streaming/internal/BlobMetadata.java
```diff
@@ -11,12 +11,20 @@
 import net.snowflake.ingest.utils.ParameterProvider;
 
 /** Metadata for a blob that sends to Snowflake as part of the register blob request */
-class BlobMetadata {
+public class BlobMetadata {
```
sfc-gh-rcheng, line 14: Is there a reason this class was not public before?

src/main/java/net/snowflake/ingest/streaming/internal/BlobMetadata.java
```diff
@@ -54,6 +62,39 @@ byte getVersionByte() {
     return bdecVersion.toByte();
   }
 
+  // TODO: @rcheng question - do these need @jsonproperties?
```
sfc-gh-rcheng, line 65: The above methods aren't called in a Json context, so why do they need @JsonProperty translation? Would the properties that I add need the tag as well?

src/main/java/net/snowflake/ingest/streaming/internal/RegisterService.java
```diff
@@ -171,10 +174,9 @@ List<FlushService.BlobData<T>> registerBlobs(Map<String, Timer.Context> latencyT
                       e.getCause() == null ? null : e.getCause().getMessage(),
                       getStackTrace(e.getCause()));
               logger.logError(errorMessage);
-              if (this.owningClient.getTelemetryService() != null) {
-                this.owningClient
-                    .getTelemetryService()
-                    .reportClientFailure(this.getClass().getSimpleName(), errorMessage);
+              // TODO: @rcheng question - should i try another telemetryservice get here?
```
sfc-gh-rcheng, line 177: I didn't see any additional telemetry client setup above this point in the method, so I'm assuming it is safe to move the getTelemetryClient() call to the beginning of the method?

src/main/java/net/snowflake/ingest/streaming/internal/RegisterService.java
```diff
@@ -200,20 +202,25 @@ List<FlushService.BlobData<T>> registerBlobs(Map<String, Timer.Context> latencyT
...
                             return null;
                           }));
             }
           }
+
+          telemetryService.reportAllBlobLatenciesInSec(blobs, true);
         }
       }
     }
+    // TODO @rcheng question - do we want whatever latencies we can get from error blobs too?
```
sfc-gh-rcheng, line 223: Unfortunately the errorBlobs don't have the BlobMetadata object, however are the errorBlob latencies (if any) useful for debugging?

sfc-gh-rcheng (commented):
PR still in draft, just had a couple questions. Will add tests and comments if the general idea of this PR looks ok

PR #429, opened 2023-04-01
[SNOW-766867] Add blob latency and start timestamps to registerBlobs call

sfc-gh-rcheng:
Add the following latencies to the registerBlobs request. These latencies will be caught and logged on the server side in this PR: https://github.com/snowflakedb/snowflake/pull/95716

buildLatency
- how long it took to build the blob
- added to the blobMetadata 

uploadLatency
- how long it took to upload the blob
- added to the blobMetadata

registerStartTimestamp
- when the register request was sent from the client
- added to the request payload
- the register latency will be calculated when the request hits the server

flushStartTimestamp
- when the client began flushing
- added to the request payload
- the flush latency will be calculated when the request hits the server
- flushLatency SHOULD be >= buildLatency + uploadLatency + registerLatency


src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
```diff
@@ -117,6 +119,11 @@ public void afterAll() throws Exception {
 
   @Test
   public void testSimpleIngest() throws Exception {
+    // TODO @rcheng - dont want to change factory, so inject
```
sfc-gh-rcheng, line 122: This test suite uses a factory to create the client. I decided it was better to inject the requestBuilder after it was created by the factory, instead of changing the factory for the purposes of this test. Open to alternatives though!

sfc-gh-rcheng:
Confirmed this works by looking at GS logs, namely:
```
"clientBlobBuildDurationMs":410
"clientBlobUploadDurationMs":4
"clientBlobFlushDurationMs":497
"clientBlobRegisterDurationMs":65
```

`
snowflake.log:INFO  [2023-04-05 17:31:59,980 - 13250073] [CommitServiceScheduler-56] [sid:0 aid:4 oid:270345 tbl:270345 rid:8af3ed2d-c09f-4c0f-9820-dc6eef2040c0 job:01ab6f3f-0100-0001-0000-000000045ba5 uid:0] EVENT_LOGGER: INGEST_BDEC_COMMIT_SUCCESS [time=1680741119980,type=INGEST_BDEC_COMMIT_SUCCESS,event={"accountId":4,"tableId":270345,"channelIds":[270341],"bdecFileName":"2023/4/6/0/31/rso45b_PlvltBctxiJXPM9eUGkp951pmJBBt9OT3GShDNlh8z3ICC_3_17_0.bdec","bdecNumRows":1000,"chunkSizeBytes":2560,"registrationDurationMs":47,"queueingDurationMs":8,"commitValidationDurationMs":1,"totalProcessingDurationMs":149,"lifetimeStats":{"registrationReceivedTimestampMs":1680741119831,"registrationDurationMs":47,"enqueuedTimestampMs":1680741119878,"queueingDurationMs":8,"commitBatchCreationTimestamp":1680741119886,"commitValidationDurationMs":1,"commitDurationMs":92,"totalProcessingDurationMs":149,"end2endMinDurationMs":753,"end2endMaxDurationMs":878,"clientBlobBuildDurationMs":410,"clientBlobUploadDurationMs":4,"clientBlobFlushDurationMs":497,"clientBlobRegisterDurationMs":65},"sdkVersion":"1.1.1","commitDurationMs":92,"end2endMinDurationMs":753,"end2endMaxDurationMs":878}]
`

sfc-gh-tjones (approved):

src/main/java/net/snowflake/ingest/streaming/internal/BlobMetadata.java
```diff
@@ -12,21 +12,42 @@
 
 /** Metadata for a blob that sends to Snowflake as part of the register blob request */
 class BlobMetadata {
+  public static final long DEFAULT_BLOB_LATENCY = -1;
```
sfc-gh-tzhang, line 15: is there a particular reason of using -1, could we just use Long and pass NULL? That's what we did for most of the values, for example `first_insert_time_in_ms`

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -258,6 +260,8 @@ private CompletableFuture<Void> registerFuture() {
    */
   CompletableFuture<Void> flush(boolean isForce) {
     long timeDiffMillis = System.currentTimeMillis() - this.lastFlushTime;
+    long flushJobStartTime = System.currentTimeMillis();
```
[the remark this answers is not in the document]
sfc-gh-tzhang: I thought we have a timer for flush, could we put it in the same place? Looks like we're passing this down multiple levels in the call stack which doesn't look pretty, or maybe use some global variables/classes?

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -487,10 +492,15 @@ BlobMetadata buildAndUpload(String filePath, List<List<ChannelData<T>>> blobData
...
-    return upload(filePath, blob.blobBytes, blob.chunksMetadataList);
+
+    return buildContext != null
+        ? upload(
+            filePath,
+            blob.blobBytes,
+            blob.chunksMetadataList,
+            TimeUnit.NANOSECONDS.toMillis(buildContext.stop()))
+        : upload(
+            filePath, blob.blobBytes, blob.chunksMetadataList, BlobMetadata.DEFAULT_BLOB_LATENCY);
```
sfc-gh-tzhang, line 503: nit: get the time first and then call upload

buildlatency = !=null? a: b
return upload (xx,xx,xx,a)

sfc-gh-tzhang (commented):
Thanks Revi, left some comments, PTAL! Also I think you need to merge the server side change first, otherwise the new client won't work for old server side code?

src/main/java/net/snowflake/ingest/streaming/internal/BlobMetadata.java
```diff
@@ -12,21 +12,42 @@
...
   private final String path;
   private final String md5;
   private final Constants.BdecVersion bdecVersion;
   private final List<ChunkMetadata> chunks;
 
+  // latencies
+  private final long buildLatencyMs;
+  private final long uploadLatencyMs;
+
+  // used for testing only
```
sfc-gh-japatel, line 26: nit: lets use annotation `visiblefortesting` from guava

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestUtils.java
```diff
@@ -122,4 +122,12 @@ public static String getShortname(final String fullname) {
     final String[] parts = fullname.split("/");
     return parts[parts.length - 1];
   }
+
+  public static String getPayloadAsStr(Map<Object, Object> payload, String message) {
```
sfc-gh-japatel, line 126: nit: javadocs on public method.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -429,6 +431,8 @@ void registerBlobs(List<BlobMetadata> blobs, final int executionCount) {
           "request_id", this.flushService.getClientPrefix() + "_" + counter.getAndIncrement());
       payload.put("blobs", blobs);
       payload.put("role", this.role);
+      payload.put("flush_start_timestamp", flushJobStartTime);
+      payload.put("register_start_timestamp", System.currentTimeMillis());
```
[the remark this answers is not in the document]
sfc-gh-rcheng: Added a BlobLatencies object that will be added into BlobMetadata

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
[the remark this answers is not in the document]
sfc-gh-rcheng: Cleaned up the call stack, passing just a blobLatency object that will contain all the latencies

src/main/java/net/snowflake/ingest/streaming/internal/BlobMetadata.java
sfc-gh-rcheng: Using NULL Long instead, thanks for the suggestion

src/main/java/net/snowflake/ingest/streaming/internal/BlobLatencies.java
```diff
@@ -0,0 +1,68 @@
...
+import com.codahale.metrics.Timer;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.concurrent.TimeUnit;
+
+/** Latency information for a blob */
+class BlobLatencies {
+  public static final long DEFAULT_BLOB_LATENCY = -1;
+
+  private long buildLatencyMs;
+  private long uploadLatencyMs;
```
sfc-gh-tjones, line 16: Should these be renamed to use `duration` instead of `latency`?

src/main/java/net/snowflake/ingest/streaming/internal/BlobLatencies.java
```diff
@@ -0,0 +1,68 @@
...
+
+  private long flushStartTimestamp;
+  private long registerStartTimestamp;
```
sfc-gh-tjones, line 19: Nit, let's include units so something like:
```
private long flushStartTimeMs;
private long registerStartTimeMS;
```

src/main/java/net/snowflake/ingest/streaming/internal/BlobLatencies.java
```diff
@@ -0,0 +1,68 @@
...
+
+  public BlobLatencies() {
+    this.buildLatencyMs = DEFAULT_BLOB_LATENCY;
+    this.uploadLatencyMs = DEFAULT_BLOB_LATENCY;
+
+    this.flushStartTimestamp = DEFAULT_BLOB_LATENCY;
+    this.registerStartTimestamp = DEFAULT_BLOB_LATENCY;
```
sfc-gh-tjones, line 26: Do these need to be initialized? Why not use the default value of `0`?

src/main/java/net/snowflake/ingest/streaming/internal/RegisterService.java
```diff
@@ -198,8 +196,13 @@ List<FlushService.BlobData<T>> registerBlobs(
             Timer.Context registerContext =
                 Utils.createTimerContext(this.owningClient.registerLatency);
 
+            for (BlobMetadata blobMetadata : blobs) {
+              blobMetadata.getBlobLatencies().setRegisterStartTimestamp(System.currentTimeMillis());
+              blobMetadata.getBlobLatencies().setFlushStartTimestamp(System.currentTimeMillis());
```
sfc-gh-tjones, line 201: Nit, let's set `now` as a variable and pass that down:
```
long timeStamp = System.currentTimeMillis();
blobMetadata.getBlobLatencies().setRegisterStartTimestamp(System.currentTimeMillis());
blobMetadata.getBlobLatencies().setFlushStartTimestamp(System.currentTimeMillis());
```

src/main/java/net/snowflake/ingest/streaming/internal/BlobLatencies.java
sfc-gh-rcheng: They do not neeeed to be, but I want a value that indicates error. It will reduce confusion - so we know if we are missing the blob latency or if it was actually 0.

src/main/java/net/snowflake/ingest/streaming/internal/RegisterService.java
sfc-gh-rcheng: pulled out currentTimeMillis to a different variable

src/main/java/net/snowflake/ingest/streaming/internal/BlobLatencies.java
sfc-gh-rcheng: That would make more sense, renamed to duration

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestUtils.java
sfc-gh-rcheng: method removed, so will not add javadocs

src/main/java/net/snowflake/ingest/streaming/internal/BlobLatencies.java
```diff
@@ -0,0 +1,70 @@
...
+
+package net.snowflake.ingest.streaming.internal;
+
+import com.codahale.metrics.Timer;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.concurrent.TimeUnit;
+
+/** Latency information for a blob */
+class BlobLatencies {
+  public static final Long DEFAULT_BLOB_LATENCY = null;
```
sfc-gh-tjones, line 13: FYI any boxed type gets initialized to `null`

src/main/java/net/snowflake/ingest/streaming/internal/BlobLatencies.java
```diff
@@ -0,0 +1,70 @@
...
+  // because the end time is when the request hits the server
+  private Long flushStartMs;
+  private Long registerStartMs;
+
+  public BlobLatencies() {
+    this.buildDurationMs = DEFAULT_BLOB_LATENCY;
+    this.uploadDurationMs = DEFAULT_BLOB_LATENCY;
+
+    this.flushStartMs = DEFAULT_BLOB_LATENCY;
+    this.registerStartMs = DEFAULT_BLOB_LATENCY;
```
sfc-gh-tjones, line 28: Are these useful? I think we want `flush duration` which I believe is covered by `uploadDurationMs`. But `registerStartMs` would be used how?

src/main/java/net/snowflake/ingest/streaming/internal/BlobLatencies.java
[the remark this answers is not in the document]
sfc-gh-rcheng: I discussed it in these two threads, but it doesn't seem like something that is needed so I'll remove it

https://github.com/snowflakedb/snowflake-ingest-java/pull/429#discussion_r1161032069
https://github.com/snowflakedb/snowflake-ingest-java/pull/429#discussion_r1163261989

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
```diff
@@ -136,6 +154,18 @@ public void testSimpleIngest() throws Exception {
     // Close the channel after insertion
     channel1.close().get();
 
+    // verify expected request sent to server
+    String[] expectedPayloadParams = {
+      "request_id", "blobs", "role", "flush_start_timestamp", "register_start_timestamp"
```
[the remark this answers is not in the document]
sfc-gh-rcheng: thats true, updated the payload

src/main/java/net/snowflake/ingest/streaming/internal/RegisterService.java
```diff
@@ -196,6 +196,12 @@ List<FlushService.BlobData<T>> registerBlobs(Map<String, Timer.Context> latencyT
             Timer.Context registerContext =
                 Utils.createTimerContext(this.owningClient.registerLatency);
 
+            for (BlobMetadata blobMetadata : blobs) {
+              long currentTimeMs = System.currentTimeMillis();
+              blobMetadata.getBlobLatencies().setRegisterStartMs(currentTimeMs);
+              blobMetadata.getBlobLatencies().setFlushStartMs(currentTimeMs);
```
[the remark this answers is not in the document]
sfc-gh-rcheng: great catch, moved to where we build the latencyTimerContextMap, in distributeFlushTasks()

src/main/java/net/snowflake/ingest/streaming/internal/BlobLatencies.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+package net.snowflake.ingest.streaming.internal;
+
+import com.codahale.metrics.Timer;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.concurrent.TimeUnit;
+
+/** Latency information for a blob */
+class BlobLatencies {
```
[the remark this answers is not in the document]
sfc-gh-rcheng: changed to blobStats

src/main/java/net/snowflake/ingest/streaming/internal/BlobStats.java
```diff
@@ -0,0 +1,62 @@
...
+    this.registerStartMs = registerStartMs;
+  }
+
+  void setBuildDurationMs(Timer.Context buildLatencyContext) {
+    if (buildLatencyContext != null) {
+      this.buildDurationMs = TimeUnit.NANOSECONDS.toMillis(buildLatencyContext.stop());
+    }
+  }
+
+  @JsonIgnore
```
sfc-gh-japatel, line 56: Why is this only setter which is marked  @ json ignore? AFAIK, you also need to have ignore on field if you want to skip serializing this property. (Which I dont think you want to)

sfc-gh-japatel (approved):
lgtm! Barring jackson annotation question. 
[Please revert the unrelated changes before pushing(which were used for testing)] 

Also Toby mentioned this already but we would want server side changes first?

src/main/java/net/snowflake/ingest/streaming/internal/BlobStats.java
sfc-gh-tzhang: +1

sfc-gh-tzhang (approved):
LGTM, left one question. Also please make sure the server side change should be merged first. Thanks!

src/main/java/net/snowflake/ingest/streaming/internal/BlobStats.java
sfc-gh-rcheng: I was running into some strange JSON issues, mostly around dev vm stuff so i added that as a sanity check. Updated to have the correct json

sfc-gh-rcheng:
Waiting on server side change, then will remove testing changes for final review

sfc-gh-rcheng:
server side change went in, need to wait a couple releases to push this

sfc-gh-tjones:
> server side change went in, need to wait a couple releases to push this

I think we can merge, I don't think we'll break server side on additional fields being sent. We just won't use them.

sfc-gh-rcheng:
> 

The integration tests are not passing because the register call doesn't have a valid payload. The server changes will allow it to accept requests without the additional properties, but I don't think it works the other way around

```
detail=Register blob request failed: HTTP Status: 400 ErrorBody: {
Message: Incoming request does not contain a valid payload.,
Data: null
}
```

sfc-gh-rcheng:
Server side changes should be coming out in 7.17, which is targeted for May 15/16. Will check in on this again then

sfc-gh-tzhang:
@sfc-gh-rcheng Please use Squash and Merge in the future so the commit history could be a bit cleaner, thanks!

sfc-gh-rcheng:
Ah my bad, will do that in the future, thanks!

PR #433, opened 2023-04-05
@snow SNOW-779794 Client SDK Parquet buffer: fix perf degradation due to calling suboptimal lib method per row

sfc-gh-azagrebin:
Calling this Parquet lib method in addRow() turned to be very expensive as its implementation is subotimal, `MessageType#getColumns`:
```
ColumnDescriptor columnDescriptor = schema.getColumns().get(colIndex);
      PrimitiveType.PrimitiveTypeName typeName =
          columnDescriptor.getPrimitiveType().getPrimitiveTypeName();
```
Instead we can just cache `PrimitiveTypeName` that we derived in setup method in column metadata index.

@test run DEW locally + precommit

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
@@ -22,11 +22,8 @@ import net.snowflake.client.jdbc.internal.google.common.collect.Sets;
 import net.snowflake.ingest.streaming.OpenChannelRequest;
 import net.snowflake.ingest.utils.Constants;
 import net.snowflake.ingest.utils.ErrorCode;
-import net.snowflake.ingest.utils.Logging;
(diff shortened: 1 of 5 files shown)


sfc-gh-gdoci (approved):

sfc-gh-tjones (approved):
What a subtle issue - thank you for digging into this and fixing so quickly!

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -313,4 +306,17 @@ void closeInternal() {
   public Flusher<ParquetChunkData> createFlusher() {
     return new ParquetFlusher(schema, enableParquetInternalBuffering);
   }
+
+  private static class ParquetColumn {
```
sfc-gh-tzhang, line 310: comment?

sfc-gh-tzhang (approved):
LGTM, thanks!

PR #367, opened 2023-03-24
SNOW-766522 Use caffeine cache

sfc-gh-lsembera:
This PR switches from guava LoadingCache to caffeine because it gives us better performance.

PR #450, opened 2023-04-06
NO-SNOW: InsertRow perf improvements

sfc-gh-tzhang:
Over the past few months, we keep adding data validation logic in order to make sure everything ingested is valid, some of the changes cause perf regressions, in this PR, we try to improve a few logic:
- Remove un-necessary Map that stores columnName - unquotedColName mapping
- Use a different way to valid utf-8 strings instead of using CharsetEncoder
- Use a different caching library that improves performance (from https://github.com/snowflakedb/snowflake-ingest-java/pull/367)


src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -846,14 +842,9 @@ private static String sanitizeValueForExceptionMessage(Object value) {
...
   private static void verifyValidUtf8(String input, String columnName, String dataType) {
-    CharsetEncoder charsetEncoder =
-        StandardCharsets.UTF_8
-            .newEncoder()
-            .onMalformedInput(CodingErrorAction.REPORT)
-            .onUnmappableCharacter(CodingErrorAction.REPORT);
-    try {
-      charsetEncoder.encode(CharBuffer.wrap(input));
-    } catch (CharacterCodingException e) {
+    String roundTripStr =
```
sfc-gh-azagrebin, line 845: Interesting idea :)
I guess the only difference is how informative `CharsetEncoder`'s exception could have been.
This new approach just detects that something is wrong.
Should we parameterise this change for debugging purposes?

sfc-gh-tzhang: We output the data so it should sufficient for us to debug?

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
```diff
@@ -737,12 +732,13 @@ private float convertRowToArrow(
 
     // All input values passed validation, iterate over the columns again and combine their existing
     // statistics with the forked statistics for the current row.
-    for (String userInputColumnName : row.keySet()) {
-      String columnName = userInputToUnquotedColumnNameMap.get(userInputColumnName);
-      RowBufferStats stats = statsMap.get(columnName);
-      RowBufferStats forkedStats = forkedStatsMap.get(columnName);
-      statsMap.put(columnName, RowBufferStats.getCombinedStats(stats, forkedStats));
+    for (Map.Entry<String, RowBufferStats> forkedColStats : forkedStatsMap.entrySet()) {
```
sfc-gh-tjones, line 735:
```
forkedStatsMap.entrySet().forEach(fcs -> statsMap.put(fcs.getKey(), RowBufferStats.getCombinedStats(statsMap.get(fcs.getKey()), fcs.getValue()));
```

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -846,14 +842,9 @@ private static String sanitizeValueForExceptionMessage(Object value) {
...
+        new String(input.getBytes(StandardCharsets.UTF_8), StandardCharsets.UTF_8);
+    if (!input.equals(roundTripStr)) {
       throw valueFormatNotAllowedException(columnName, input, dataType, "Invalid Unicode string");
```
sfc-gh-tjones, line 848: Should we throw an exception with `roundTripStr` as well?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -209,11 +204,11 @@ private float addRow(
...
-      String columnName = userInputToUnquotedColumnNameMap.get(userInputColumnName);
-      RowBufferStats stats = statsMap.get(columnName);
-      RowBufferStats forkedStats = forkedStatsMap.get(columnName);
-      statsMap.put(columnName, RowBufferStats.getCombinedStats(stats, forkedStats));
+    for (Map.Entry<String, RowBufferStats> forkedColStats : forkedStatsMap.entrySet()) {
+      String columnName = forkedColStats.getKey();
+      statsMap.put(
+          columnName,
+          RowBufferStats.getCombinedStats(statsMap.get(columnName), forkedColStats.getValue()));
     }
```
sfc-gh-tjones, line 212: Same logic as above right? Should this be pulled into a helper?

sfc-gh-tjones: Got it, that's fair. Thanks!

src/main/java/net/snowflake/ingest/streaming/internal/ArrowRowBuffer.java
sfc-gh-tjones: That's fine, I think it'll reduce to the same Java bytecode anyway most likely.

sfc-gh-tjones (approved):

PR #452, opened 2023-04-06
@no-snow fix bug in OnErrorOption.Abort for Parquet

sfc-gh-gdoci:
This PR fixes a bug for OnErrorOption.Abort for Parquet. There the rows were being added directly into the data buffer and not in the temporary buffer.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
@@ -156,10 +156,6 @@ public class ParquetRowBuffer extends AbstractRowBuffer<ParquetChunkData> {
     }
   }
 
-  void writeTempRow(List<Object> row) {
-    tempData.add(row);
-  }
-
   @Override
   float addTempRow(
       Map<String, Object> row,
@@ -168,7 +164,7 @@ public class ParquetRowBuffer extends AbstractRowBuffer<ParquetChunkData> {
       Set<String> formattedInputColumnNames,
       long insertRowIndex) {
     return addRow(
-        row, curRowIndex, this::writeTempRow, statsMap, formattedInputColumnNames, insertRowIndex);
+        row, curRowIndex, tempData::add, statsMap, formattedInputColumnNames, insertRowIndex);
   }
 
   /**
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
@@ -156,6 +156,10 @@ public class ParquetRowBuffer extends AbstractRowBuffer<ParquetChunkData> {
     }
   }
 
+  void writeTempRow(List<Object> row) {
+    tempData.add(row);
+  }
+
   @Override
   float addTempRow(
       Map<String, Object> row,
@@ -164,7 +168,7 @@ public class ParquetRowBuffer extends AbstractRowBuffer<ParquetChunkData> {
       Set<String> formattedInputColumnNames,
       long insertRowIndex) {
     return addRow(
-        row, curRowIndex, this::writeRow, statsMap, formattedInputColumnNames, insertRowIndex);
+        row, curRowIndex, this::writeTempRow, statsMap, formattedInputColumnNames, insertRowIndex);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -154,13 +154,17 @@ void writeRow(List<Object> row) {
...
+  }
+
   @Override
   float addTempRow(
       Map<String, Object> row,
       int curRowIndex,
       Map<String, RowBufferStats> statsMap,
       Set<String> formattedInputColumnNames) {
-    return addRow(row, this::writeRow, statsMap, formattedInputColumnNames);
+    return addRow(row, this::writeTempRow, statsMap, formattedInputColumnNames);
```
sfc-gh-gdoci, line 167: @sfc-gh-azagrebin no easy way to test this because `AbstractRowBuffer.insertRows` has a finally block, which cleans the temporary buffer.
```
} finally {
      this.tempStatsMap.values().forEach(RowBufferStats::reset);
      clearTempRows();
      this.flushLock.unlock();
    }

**Review by sfc-gh-tjones — APPROVED** [review]

**Review by sfc-gh-tzhang — APPROVED** [review]
any way we could add a test to avoid this from happening again?

**Comment by sfc-gh-gdoci** [comment]
> any way we could add a test to avoid this from happening again?

ack, I added a simple test.

File: src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java [hunk 2]
```diff
@@ -164,7 +168,7 @@ float addTempRow(
       Set<String> formattedInputColumnNames,
       long insertRowIndex) {
     return addRow(
-        row, curRowIndex, this::writeRow, statsMap, formattedInputColumnNames, insertRowIndex);
+        row, curRowIndex, this::writeTempRow, statsMap, formattedInputColumnNames, insertRowIndex);
```
sfc-gh-azagrebin, line 171:
```suggestion
        row, curRowIndex, tempData::add, statsMap, formattedInputColumnNames, insertRowIndex);
```

sfc-gh-azagrebin: Then you do not need `writeTempRow()`

src/test/java/net/snowflake/ingest/streaming/internal/RowBufferTest.java
```diff
@@ -1498,4 +1498,56 @@ private void testE2EArrayHelper(OpenChannelRequest.OnErrorOption onErrorOption)
...
+    Assert.assertEquals(
+        "a",
+        innerBufferOnErrorContinue.getVectorValueAt(
+            "COLCHAR", 0)); // only first row is in the buffer
+    Assert.assertThrows(
+        IndexOutOfBoundsException.class,
+        () -> {
+          innerBufferOnErrorAbort.getVectorValueAt(
+              "COLCHAR", 0); // neither of the rows were copied from the temp buffer
+        });
```
sfc-gh-azagrebin, line 1551: I would rather call getSnapshot and check the actual contents because this is what matters eventually. Using `getVectorValueAt` is an invasive check that does not check the public API contract of the buffer class.

sfc-gh-gdoci: `getSnapshot` doesn't test this scenario because we always copy the rows from temp buffer to actual channel buffer before the `getSnapshot call.

sfc-gh-gdoci: agree, I replaced the test.

src/test/java/net/snowflake/ingest/streaming/internal/RowBufferTest.java
```diff
@@ -1498,4 +1499,71 @@ private void testE2EArrayHelper(OpenChannelRequest.OnErrorOption onErrorOption)
...
+    mixedRows.add(Collections.singletonMap("colChar", "b"));
+    mixedRows.add(Collections.singletonMap("colChar", "1111111111111111111111")); // too big
+
+    response = innerBufferOnErrorContinue.insertRows(mixedRows, "3");
+    Assert.assertTrue(response.hasErrors());
+
+    Assert.assertThrows(
+        SFException.class, () -> innerBufferOnErrorAbort.insertRows(mixedRows, "3"));
+
+    if (bdecVersion == Constants.BdecVersion.THREE) {
```
sfc-gh-azagrebin, line 1542: What about the other possible versions in future?

I would at least use a switch/case and throw on unimplemented versions here to not loose coverage for new versions.

Long term if we have to maintain more versions, we might want to implement API for snapshot deserialiazation into List<List<Object>> for assertions like this or create and compare good/bad snapshots.

sfc-gh-azagrebin (approved):

src/test/java/net/snowflake/ingest/streaming/internal/RowBufferTest.java
sfc-gh-gdoci: ack, I changed it to use a switch/case with throw exception in `default:`.

PR #474, opened 2023-04-12
SNOW-726924 Re-add jvmNonProxy host

sfc-gh-japatel:
This reverts commit ca63c27621f2f143638bde113fcfae4a400d145f.

- This is required for Kafka Connector, corresponding PR: https://github.com/snowflakedb/snowflake-kafka-connector/pull/614
- But also uses JDBC 3.13.25

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -60,7 +60,7 @@
     <shadeBase>net.snowflake.ingest.internal</shadeBase>
     <slf4j.version>1.7.36</slf4j.version>
     <snappy.version>1.1.8.3</snappy.version>
-    <snowjdbc.version>3.13.15</snowjdbc.version>
+    <snowjdbc.version>3.13.25</snowjdbc.version>
     <yetus.version>0.13.0</yetus.version>
   </properties>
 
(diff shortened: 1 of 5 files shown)


sfc-gh-tzhang (approved):
LGTM!

sfc-gh-tzhang:
@sfc-gh-tjones @sfc-gh-dwang @sfc-gh-pbennes Any idea about the failure in the merge gate? We just upgrade the JDBC and it starts to complain about shading JAR, and for some reason it only complains in Windows

sfc-gh-pbennes:
Unsure. I think @sfc-gh-japatel has some leads. It's worth noting that the latest version of the maven shade plugin is 3.4.1 and we're on 3.1.1. It's possibel the issue has already been fixed upstream, whatever it is, and all that may be required is to update the shade plugin. That'd probably be the first thing I'd suggest.

sfc-gh-japatel:
> Unsure. I think @sfc-gh-japatel has some leads. It's worth noting that the latest version of the maven shade plugin is 3.4.1 and we're on 3.1.1. It's possibel the issue has already been fixed upstream, whatever it is, and all that may be required is to update the shade plugin. That'd probably be the first thing I'd suggest.

Thanks @sfc-gh-tzhang and @sfc-gh-pbennes for looking into this, upgrading it at least fixed the issue locally so I am guessing it should work on Github actions too!

PR #477, opened 2023-04-18
NO-SNOW Fix bug for ParquetRowBuffer regarding row Index

sfc-gh-japatel:
- Andrey pointed out this. Thanks @sfc-gh-azagrebin 
  - There are two indices used - bufferedRowIndex and insertRowIndex. BufferedRow index was not used in parquetBuffer but was used in arrow. insertRowIndex was introduced earlier 1310e0591491647696aeb755d5b4ecbf0d1b1c67 but was incorrectly used in parquet buffer. The index used there was bufferedRowIndex. Since we didnt use the bufferedRowIndex anywhere in parquetBuffer, I replaced it with `insertRowsBuffer`

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ParquetValueParser.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ParquetValueParser.java
@@ -75,7 +75,8 @@ class ParquetValueParser {
    * @param columnMetadata column metadata
    * @param typeName Parquet primitive type name
    * @param stats column stats to update
-   * @param curRowIndex Row index corresponding the row to parse
+   * @param insertRowsCurrIndex Row index corresponding the row to parse (w.r.t input rows in
+   *     insertRows API, and not buffered row)
    * @return parsed value and byte size of Parquet internal representation
    */
   static ParquetBufferValue parseColumnValueToParquet(
@@ -84,7 +85,7 @@ class ParquetValueParser {
       PrimitiveType.PrimitiveTypeName typeName,
       RowBufferStats stats,
       ZoneId defaultTimezone,
-      int curRowIndex) {
+      long insertRowsCurrIndex) {
     Utils.assertNotNull("Parquet column stats", stats);
     float estimatedParquetSize = 0F;
     estimatedParquetSize += DEFINITION_LEVEL_ENCODING_BYTE_LEN;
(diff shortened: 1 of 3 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/ParquetValueParser.java
```diff
@@ -196,6 +199,7 @@ static ParquetBufferValue parseColumnValueToParquet(
    * @param precision data type precision
    * @param logicalType Snowflake logical type
    * @param physicalType Snowflake physical type
+   * @param insertRowsCurrIndex
```
sfc-gh-gdoci, line 202: nit: comment

src/main/java/net/snowflake/ingest/streaming/internal/ParquetValueParser.java
```diff
@@ -354,7 +361,10 @@ static byte[] getSb16Bytes(BigInteger intRep) {
    * @return string representation
```
sfc-gh-gdoci, line 361: nit: parameter comment about `insertRowsCurrIndex`

sfc-gh-gdoci (commented):
Thanks @sfc-gh-japatel. Is there a way to add a test that would have caught this bug or in general about `insertRowIndex`?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetValueParser.java
sfc-gh-japatel: Ack!

src/main/java/net/snowflake/ingest/streaming/internal/ParquetValueParser.java
sfc-gh-japatel: Ack

sfc-gh-japatel:
> Thanks @sfc-gh-japatel. Is there a way to add a test that would have caught this bug or in general about `insertRowIndex`?

The reason it wasnt caught was because parquet and arrow rows has two separate use case for row index. previous index existing in code was used in arrow but not in parquet.

sfc-gh-tjones (approved):

sfc-gh-tzhang:
> * Andrey pointed out this. Thanks @sfc-gh-azagrebin

Could you describe the issue in the description?

sfc-gh-tzhang (approved):

sfc-gh-japatel:
> > Thanks @sfc-gh-japatel. Is there a way to add a test that would have caught this bug or in general about `insertRowIndex`?
> 
> The reason it wasnt caught was because parquet and arrow rows has two separate use case for row index. previous index existing in code was used in arrow but not in parquet.

Thanks @sfc-gh-gdoci, added the test - Good to add the test always!

PR #481, opened 2023-04-20
SNOW-796657 Avoid unnecessary String#getBytes allocations

sfc-gh-lsembera:
Some calls to `String#getBytes` we are making are not necessary. This PR restructures the code, so that we can String#getBytes once early after receiving the string from the user. 

Affected data types are `VARCHAR`, `VARIANT`, `OBJECT` and `ARRAY`.


sfc-gh-lsembera:
This change is not needed.

PR #484, opened 2023-04-21
NO-SNOW Fix flaky Windows test

sfc-gh-lsembera:
The test `SnowflakeStreamingIngestChannelTest#testInsertRowThrottling` is flaky on Windows. The reason is that we are using mockito to mock methods from java.lang.Runtime in one thread and read them in another. These changes are not always guaranteed to be visible in the other thread.

This PR introduces a new interface with two implementations to provide memory info, so we don't need to mock anything anymore.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/MemoryInfoProvider.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/MemoryInfoProvider.java
@@ -1,0 +1,14 @@
+/*
+ * Copyright (c) 2023 Snowflake Computing Inc. All rights reserved.
+ */
+
+package net.snowflake.ingest.streaming.internal;
+
+/** Provider information about available system memory */
+public interface MemoryInfoProvider {
+  long getMaxMemory();
+
+  long getTotalMemory();
+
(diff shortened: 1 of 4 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/MemoryInfoProvider.java
```diff
@@ -0,0 +1,14 @@
...
+
+package net.snowflake.ingest.streaming.internal;
+
+/** Provider information about available system memory */
+public interface MemoryInfoProvider {
+  long getMaxMemory();
+
+  long getTotalMemory();
+
+  long getFreeMemory();
```
sfc-gh-japatel, line 13: nit: javadocs on public interface methods!

sfc-gh-japatel (approved):
thanks for fixing flakiness!

src/main/java/net/snowflake/ingest/streaming/internal/MemoryInfoProvider.java
sfc-gh-lsembera: Javadocs added

PR #489, opened 2023-05-03
SNOW-806047 fix logged uncompressedSize for Parquet

sfc-gh-gdoci:
This PR fixes the uncompressed size that we log. Previously it had an incorrect value for Parquet.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ArrowFlusher.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ArrowFlusher.java
@@ -34,6 +34,7 @@ public class ArrowFlusher implements Flusher<VectorSchemaRoot> {
     ByteArrayOutputStream chunkData = new ByteArrayOutputStream();
     List<ChannelMetadata> channelsMetadataList = new ArrayList<>();
     long rowCount = 0L;
+    float chunkUncompressedSize = 0f;
     VectorSchemaRoot root = null;
     ArrowWriter arrowWriter = null;
     VectorLoader loader = null;
@@ -94,6 +95,7 @@ public class ArrowFlusher implements Flusher<VectorSchemaRoot> {
         // Write channel data using the stream writer
         arrowWriter.writeBatch();
         rowCount += data.getRowCount();
+        chunkUncompressedSize += data.getBufferSize();
 
         logger.logDebug(
             "Finish building channel={}, rowCount={}, bufferSize={} in blob={}",
@@ -112,6 +114,7 @@ public class ArrowFlusher implements Flusher<VectorSchemaRoot> {
         channelsMetadataList,
         columnEpStatsMapCombined,
         rowCount,
+        chunkUncompressedSize,
         chunkData,
         chunkMinMaxInsertTimeInMs);
   }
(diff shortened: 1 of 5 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -50,6 +50,7 @@ private SerializationResult serializeFromParquetWriteBuffers(
       throws IOException {
     List<ChannelMetadata> channelsMetadataList = new ArrayList<>();
     long rowCount = 0L;
+    float chunkUncompressedSize = 0f;
```
sfc-gh-tjones, line 53: Codecov is saying that this line is not covered by tests.

sfc-gh-tjones (approved):
LGTM barring the Codecov comment.

src/main/java/net/snowflake/ingest/streaming/internal/ArrowFlusher.java
```diff
@@ -94,6 +95,7 @@ public Flusher.SerializationResult serialize(
         // Write channel data using the stream writer
         arrowWriter.writeBatch();
         rowCount += data.getRowCount();
+        chunkUncompressedSize += data.getBufferSize();
```
sfc-gh-tzhang, line 98: chunkData.size and the accumulation of buffer size could be different, right? I think the buffer size is the estimation  and we should use the actual uncompressed data size?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -176,14 +180,16 @@ private SerializationResult serializeFromJavaObjects(
       rows.addAll(data.getVectors().rows);
 
       rowCount += data.getRowCount();
+      chunkUncompressedSize += data.getBufferSize();
```
sfc-gh-tzhang, line 183: Same here, is it possible we get the actual uncompressed buffer size instead of using the estimated buffer size?

sfc-gh-tzhang (commented):
Left some comments, PTAL!

src/main/java/net/snowflake/ingest/streaming/internal/ArrowFlusher.java
sfc-gh-gdoci: I think it is fine to use the estimation. We use here in other places as well, eg this [log line](https://github.com/snowflakedb/snowflake-ingest-java/pull/489/files#diff-0b5d892d276fc68717bae6286a9fc43ae7ff3d033b0e260c66e781179c7143ddL60) and this [other](https://github.com/snowflakedb/snowflake-ingest-java/pull/489/files#diff-0b5d892d276fc68717bae6286a9fc43ae7ff3d033b0e260c66e781179c7143ddL102).

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
sfc-gh-gdoci: Yes, `bufferSize()` is an estimation. For Parquet we have no access to the 'real' uncompressed data size because data is compressed internally. Our estimation can be slightly larger than the real value.

sfc-gh-tzhang (approved):

src/main/java/net/snowflake/ingest/streaming/internal/ArrowFlusher.java
sfc-gh-azagrebin: We can actually get back to the original behaviour for Arrow if `SerializationResult` had getUncompressedSize() method overriden for Arrow and Parquet accordingly or only for parquet.

sfc-gh-gdoci: Arrow will be deprecated soon anyway.

Issue #356, opened 2023-03-19
Dependency shading is not complete, or is not configured correctly


jdcaperon:
When adding the shaded version of the snowflake dependency to our closure, we noticed that it appears that shading is not configured correctly as there are a large number of conflicts in our dependency closure. It appears that shading is configured correctly for some dependencies and not for others, particularly some transitive dependencies are not covered it seems.

I couldn't find why shading was introduced in the first place, there's very little information other than a Relocate all dependencies to internal to solve dependency conflict problem comment: https://github.com/snowflakedb/snowflake-ingest-java/pull/19

In the current 1.1.0 shaded jar there are quite a few packages that conflict with the ones in our dependency closure, including com.google.common, com.codahale.metrics, and org.eclipse.jetty

For now we are going to switch to the unshaded dependency to resolve these issues but it would be good to reconfigure the shading to be correct for the shaded version of the dependency.

leszekgruchala:
We have faced the issue too. Our solution was to change the order of loaded jars, to make snowflake-sdk the last one.

sfc-gh-tzhang:
@sfc-gh-tjones @sfc-gh-dwang Any idea about the latest question? I think the protobuf library is part of the hadoop-common? But I'm not sure why they didn't hit this issue before the refactoring.

sfc-gh-dwang:
@sfc-gh-tzhang com.google.protobuf:protobuf-java 2.5 is brought by hadoop-common 3.3.5, likely 3.3.4 uses some old version of protobuf

leszekgruchala:
@sfc-gh-tzhang @sfc-gh-tjones @sfc-gh-dwang Can you please explain why you keep the same packages of the shaded libraries? They should be different, because now the shaded jars are still available on the classpath with the original packages and are conflicting with others.

As I understand, you bring an uber jar with all your dependencies still available in the client's classpaths even though you try to hide them. It's worse than if you provide them as regular dependencies defined in the pom.xml, because now the client's can't properly evict the incorrect library versions.

I don't know about Maven, but in SBT there is a possibility to rename the packages
https://stackoverflow.com/a/40404371/804812

```
assemblyShadeRules in assembly ++= Seq(
  ShadeRule.rename("com.typesafe.config.**" -> "my_conf.@1")
    .inLibrary("com.typesafe" % "config" % "1.3.0")
    .inProject
)
```

sfc-gh-pbennes:
[...]

PR #490, opened 2023-05-03
SNOW-801398 Tidy up shaded jar

sfc-gh-lsembera:
This PR reorganizes our shaded JAR, so that the only top level packages are snowflake ones. All dependencies are shaded into the package `net.snowflake.ingest.internal`.

The script `check_content.sh` has been adapted accordingly, it now only ignores a limited set of required top-level files and directories.

Additionally, this PR removes some unnecessary transitive dependencies, particularly those pulled in by `hadoop-common`.

Tests executed:
* dew tests (both Arrow and Parquet) have passed.
* SDK integration tests have passed. The tests were extracted into a separate maven project and the SDK was added as a dependency, to make sure it tests the shaded JAR.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -62,7 +62,6 @@
     <slf4j.version>1.7.36</slf4j.version>
     <snappy.version>1.1.8.3</snappy.version>
     <snowjdbc.version>3.13.29</snowjdbc.version>
-
     <yetus.version>0.13.0</yetus.version>
   </properties>
 
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
   <!-- Arifact name and version information -->
   <groupId>net.snowflake</groupId>
   <artifactId>snowflake-ingest-sdk</artifactId>
-  <version>1.1.3-shaded20</version>
+  <version>1.1.3</version>
   <packaging>jar</packaging>
   <name>Snowflake Ingest SDK</name>
   <description>Snowflake Ingest SDK</description>
@@ -62,6 +62,7 @@
     <slf4j.version>1.7.36</slf4j.version>
     <snappy.version>1.1.8.3</snappy.version>
     <snowjdbc.version>3.13.29</snowjdbc.version>
+
     <yetus.version>0.13.0</yetus.version>
   </properties>
 
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
   <!-- Arifact name and version information -->
   <groupId>net.snowflake</groupId>
   <artifactId>snowflake-ingest-sdk</artifactId>
-  <version>1.1.3</version>
+  <version>1.1.3-shaded20</version>
   <packaging>jar</packaging>
   <name>Snowflake Ingest SDK</name>
   <description>Snowflake Ingest SDK</description>
@@ -788,6 +788,11 @@
             <artifactId>maven-shade-plugin</artifactId>
             <version>3.4.1</version>
             <configuration>
+              <artifactSet>
+                <excludes>
+                  <exclude>net.snowflake:snowflake-jdbc</exclude>
+                </excludes>
+              </artifactSet>
               <relocations>
                 <relocation>
                   <pattern>com.nimbusds</pattern>
--- a/pom.xml
+++ b/pom.xml
@@ -154,16 +154,68 @@
             <artifactId>reload4j</artifactId>
           </exclusion>
           <exclusion>
+            <groupId>com.github.pjfanning</groupId>
+            <artifactId>jersey-json</artifactId>
+          </exclusion>
+          <exclusion>
+            <groupId>com.sun.jersey</groupId>
+            <artifactId>jersey-core</artifactId>
+          </exclusion>
+          <exclusion>
+            <groupId>com.sun.jersey</groupId>
+            <artifactId>jersey-server</artifactId>
+          </exclusion>
+          <exclusion>
+            <groupId>com.sun.jersey</groupId>
+            <artifactId>jersey-servlet</artifactId>
+          </exclusion>
+          <exclusion>
+            <groupId>dnsjava</groupId>
+            <artifactId>dnsjava</artifactId>
+          </exclusion>
+          <exclusion>
             <groupId>javax.activation</groupId>
             <artifactId>activation</artifactId>
+          </exclusion>
+          <exclusion>
+            <groupId>javax.servlet</groupId>
+            <artifactId>javax.servlet-api</artifactId>
+          </exclusion>
+          <exclusion>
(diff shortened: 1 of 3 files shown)


sfc-gh-pbennes:
Ah nice, didn't know someone else was looking at #356. I opened a draft: #492 but given the discussion in #356 I'm more inclined to suggest removal of shading altogether.

leszekgruchala:
@sfc-gh-pbennes That's fine, but the problem is that the shaded jars are not shaded correctly and are conflicting with same libraries in the class path https://github.com/snowflakedb/snowflake-ingest-java/issues/356

sfc-gh-pbennes:
> @sfc-gh-pbennes I agree that unshaded JAR is a better option and maybe it should be our primary artifact (i.e. instead of having `1.1.4` and `1.1.4-unshaded`, we could have `1.1.4` and `1.1.4-shaded`). But I would argue against the removal of shading because it can help in situations when the client requires a specific version of a dependency, but we are shipping with a different, incompatible version. Some potential suspects would be Hadoop or Jackson.

Yeah if we need to shade one or two things that's fine but we probably don't need to shade everything.

sfc-gh-tjones (approved):
I'm good with this but let's have Preston have the final say as I am not a Java wizard.

sfc-gh-pbennes (changes requested):
I ran the version of check_content.sh from https://github.com/snowflakedb/snowflake-ingest-java/blob/f0137e6ef51a52a7ab90029fe8102043728a5a60/scripts/check_content.sh and noticed that unrelocated shaded snowflake-jdbc classes are present. snowflake-jdbc should either be relocated or excluded from shading:

```
[pbennes@SDP_DevVM-pbennes snowflake-ingest-java]$ scripts/check_content.sh | grep -v META-INF | head
[ERROR] Ingest SDK jar includes classes outside net.snowflake.ingest
Found classes in jar:
com/snowflake/client/jdbc/SnowflakeDriver.class
net/snowflake/client/core/AssertUtil.class
net/snowflake/client/core/BasicEvent$QueryState.class
```

sfc-gh-japatel:
> Ah nice, didn't know someone else was looking at #356. I opened a draft: #492 but given the discussion in #356 I'm more inclined to suggest removal of shading altogether.

Wouldnt removing shading altogether considered a behavior change? IIUC, this will force users of this jar to add more dependencies?

sfc-gh-pbennes:
> > Ah nice, didn't know someone else was looking at #356. I opened a draft: #492 but given the discussion in #356 I'm more inclined to suggest removal of shading altogether.
> 
> Wouldnt removing shading altogether considered a behavior change? IIUC, this will force users of this jar to add more dependencies?

For relocated dependencies yes because there would now be an interaction for dependencies with multiple uses in a project consuming the ingest sdk. If any our first-order dependencies (ex. hadoop) require us to pull in ancient versions of commonly used dependencies (ex. protobuf-java) continuing to shade those may make sense, if we can't just outright exclude them, to avoid irreconcilable differences for people consuming snowflake-ingest-sdk.

For unrelocated shaded dependencies _basically_ no because it's the same as if the dependency weren't shaded at all. The caveat here is that unrelocated shaded dependencies have the potential to be problematic due to the class conflict issue described in #492 where it's not going to be immediately obvious to someone consuming snowflake-ingest-sdk that they're going to get a version of those shaded unrelocated classes they may not expect depending on where snowflake-ingest-sdk appears in their pom.xml (assuming maven, IDK offhand how gradle etc. handles this problem).

Basically:
- Anything shaded MUST be relocated
- We should be highly selective about what gets shaded/relocated

scripts/check_content.sh
```diff
@@ -1,33 +1,40 @@
...
-    | grep -v "lookup.class" | grep -v "update.class" | grep -v "dig.class" | grep -v "jnamed"  \
-    ; then
-  echo "[ERROR] Ingest SDK jar includes class not under the snowflake namespace"
+
+if jar tvf $DIR/../target/snowflake-ingest-sdk.jar  | awk '{print $8}' | \
+    # Ignore directories
+    grep -v -E '/$' | \
+
+    # Snowflake top-level packages are allowed
+    grep -v -E "^(net|com)/snowflake" | \
```
sfc-gh-pbennes, line 16: This line should be adjusted to only cover the classes in this project. Currently it misses the shaded unrelocated snowflake-jdbc classes.

sfc-gh-pbennes (approved):
This LGTM with snowflake-jdbc excluded and the update to check_content.sh

sfc-gh-tzhang:
> @sfc-gh-pbennes I agree that unshaded JAR is a better option and maybe it should be our primary artifact (i.e. instead of having `1.1.4` and `1.1.4-unshaded`, we could have `1.1.4` and `1.1.4-shaded`). But I would argue against the removal of shading because it can help in situations when the client requires a specific version of a dependency, but we are shipping with a different, incompatible version. Some potential suspects would be Hadoop or Jackson.

I don't think now it's a good time to update the default behavior, this would be a big behavior change that needs to be communicated with customers carefully and we haven't heard many complains so far. I suggest we keep releasing `1.1.4` and `1.1.4-unshaded,` customers that want the unshaded version could use that and then we could see the usage from internal telemetry to conclude on the direction we want to go.

pom.xml
```diff
@@ -62,6 +62,7 @@
     <slf4j.version>1.7.36</slf4j.version>
     <snappy.version>1.1.8.3</snappy.version>
     <snowjdbc.version>3.13.29</snowjdbc.version>
+
```
sfc-gh-tzhang, line 65: nit: remove extra space?

sfc-gh-tzhang (approved):
LGTM, thanks @sfc-gh-lsembera !

sfc-gh-pbennes:
> > @sfc-gh-pbennes I agree that unshaded JAR is a better option and maybe it should be our primary artifact (i.e. instead of having `1.1.4` and `1.1.4-unshaded`, we could have `1.1.4` and `1.1.4-shaded`). But I would argue against the removal of shading because it can help in situations when the client requires a specific version of a dependency, but we are shipping with a different, incompatible version. Some potential suspects would be Hadoop or Jackson.
> 
> I don't think now it's a good time to update the default behavior, this would be a big behavior change that needs to be communicated with customers carefully and we haven't heard many complains so far. I suggest we keep releasing `1.1.4` and `1.1.4-unshaded,` customers that want the unshaded version could use that and then we could see the usage from internal telemetry to conclude on the direction we want to go.

Agreed. My comments are more on the direction and long term goals and less on the immediate PR which is a step in the right direction.

pom.xml
sfc-gh-lsembera: Fixed

PR #491, opened 2023-05-03
SNOW-807102 Disable OpenManyChannelsIT

sfc-gh-lsembera:
The `OpenManyChannelsIT` integration test takes a very long time and generates a large log output. It is not a good fit for SDK integration tests running with each PR and will be re-implemented in dew.


sfc-gh-gdoci (approved):

sfc-gh-tzhang (approved):
the issue for DEW is that it can only be run on released version, do we have similar test here with reduced workload?

sfc-gh-tjones (approved):

sfc-gh-rcheng (approved):

sfc-gh-lsembera:
@sfc-gh-tzhang reduced workload would reduce the value of this test as it is trying to detect race conditions in multi-threaded code. I think dew is the ideal place for this kind of test.

PR #492, opened 2023-05-03
Fix shading class relocation: net.snowflake.ingest.shade

### LINKED ISSUE #356: Dependency shading is not complete, or is not configured correctly

sfc-gh-pbennes:
- Relocate all dependency classes under net.snowflake.ingest.shade

As of release 1.1.3 the shaded jar contains a large number of unrelocated classes. This is problematic for library consumers that include any of the same dependencies in their projects that snowflake-ingest-sdk uses in addition to snowflake-ingest-sdk. This can result in duplicate classes in a project and confusing errors for users depending on dependency ordering of their project. When duplicate classes are present on the classpath the first class found wins.

This means that for consumers of snowflake-ingest-sdk if it's a dependency listed high in a pom.xml's dependency section, its duplicate classes take precedence, overriding any other dependency definitions in their project.

Here's a short example of the problem. If I have a project using the ingest SDK with a pom.xml as follows:

```
      <dependency>
        <version>1.1.3</version>
      </dependency>
      <dependency>
        <groupId>com.google.protobuf</groupId>
        <artifactId>protobuf-java</artifactId>
        <version>3.22.3</version>
      </dependency>
```

snowflake-ingest-sdk includes class files from
com.google.protobuf:protobuf-java:2.5.0 which is pulled in transitively from hadoop-common. For any class files that are duplicates of protobuf-java 3.22.3 the class files of 2.5.0 will take precedence at runtime due to snowflake-ingest-sdk being first in the pom.xml and therefore first on the classpath.

Conversely if snowflake-ingest-sdk is last then protobuf-java 3.22.3 will be active which may cause runtime issues with snowflake-ingest-sdk if it depends on APIs only present in 2.5.0.

Fixes #356

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -465,6 +465,11 @@
           <version>3.0.0-M5</version>
         </plugin>
         <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-site-plugin</artifactId>
+          <version>4.0.0-M8</version>
+        </plugin>
+        <plugin>
           <groupId>org.jacoco</groupId>
           <artifactId>jacoco-maven-plugin</artifactId>
           <version>${jacoco.version}</version>
@@ -493,11 +498,6 @@
               </configuration>
             </execution>
           </executions>
-        </plugin>
-        <plugin>
-          <groupId>org.apache.maven.plugins</groupId>
-          <artifactId>maven-site-plugin</artifactId>
-          <version>4.0.0-M8</version>
         </plugin>
       </plugins>
     </pluginManagement>
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
   <!-- Arifact name and version information -->
   <groupId>net.snowflake</groupId>
   <artifactId>snowflake-ingest-sdk</artifactId>
-  <version>1.1.3</version>
+  <version>1.1.4-SNAPSHOT</version>
   <packaging>jar</packaging>
   <name>Snowflake Ingest SDK</name>
   <description>Snowflake Ingest SDK</description>
@@ -57,7 +57,7 @@
     <objenesis.version>3.1</objenesis.version>
     <parquet.version>1.12.3</parquet.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <shadeBase>net.snowflake.ingest.internal</shadeBase>
+    <shadeBase>net.snowflake.ingest.shade</shadeBase>
     <slf4j.version>1.7.36</slf4j.version>
     <snappy.version>1.1.8.3</snappy.version>
     <snowjdbc.version>3.13.29</snowjdbc.version>
@@ -384,7 +384,6 @@
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
-      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>org.apache.arrow</groupId>
@@ -495,6 +494,11 @@
             </execution>
           </executions>
         </plugin>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-site-plugin</artifactId>
+          <version>4.0.0-M8</version>
+        </plugin>
       </plugins>
     </pluginManagement>
     <plugins>
@@ -510,7 +514,7 @@
           <sortDependencyExclusions>groupId,artifactId</sortDependencyExclusions>
           <sortExecutions>true</sortExecutions>
           <sortModules>true</sortModules>
-          <sortPlugins>groupId,artifactId</sortPlugins>
+          <sortPlugins>groupId</sortPlugins>
           <sortProperties>true</sortProperties>
           <verifyFail>stop</verifyFail>
           <verifyFailOn>strict</verifyFailOn>
@@ -676,7 +680,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
-        <version>2.19.1</version>
+        <version>2.22.2</version>
         <configuration>
           <excludes>
             <exclude>**/TestSimpleIngestLocal.java</exclude>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -6,7 +6,7 @@
   <!-- Arifact name and version information -->
   <groupId>net.snowflake</groupId>
   <artifactId>snowflake-ingest-sdk</artifactId>
-  <version>1.1.3</version>
+  <version>1.1.4-SNAPSHOT</version>
```
sfc-gh-pbennes, line 9: I think once a version has been published this should be bumped to the next version + `-SNAPSHOT`

pom.xml
```diff
@@ -57,7 +57,7 @@
     <objenesis.version>3.1</objenesis.version>
     <parquet.version>1.12.3</parquet.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <shadeBase>net.snowflake.ingest.internal</shadeBase>
+    <shadeBase>net.snowflake.ingest.shade</shadeBase>
```
sfc-gh-pbennes, line 60: I think `shade` makes the use more clear vs internal could be something else, ex. https://github.com/snowflakedb/snowflake-ingest-java/tree/master/src/main/java/net/snowflake/ingest/streaming/internal

pom.xml
```diff
@@ -384,7 +384,6 @@
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
-      <scope>provided</scope>
```
sfc-gh-pbennes, line 387: Linkage errors on the shade jar show provided is probably incorrect

pom.xml
```diff
@@ -495,6 +494,11 @@
             </execution>
           </executions>
         </plugin>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-site-plugin</artifactId>
+          <version>4.0.0-M8</version>
+        </plugin>
```
sfc-gh-pbennes, line 501: Unrelated and can be removed, but adding this enables `mvn site` to work

pom.xml
```diff
@@ -510,7 +514,7 @@
           <sortDependencyExclusions>groupId,artifactId</sortDependencyExclusions>
           <sortExecutions>true</sortExecutions>
           <sortModules>true</sortModules>
-          <sortPlugins>groupId,artifactId</sortPlugins>
+          <sortPlugins>groupId</sortPlugins>
```
sfc-gh-pbennes, line 517: When two plugins have an execution in the same phase, the plugins run in order they appear in pom.xml. This was problematic for the antrun plugin and had to be changed

pom.xml
```diff
@@ -724,66 +728,48 @@
...
       <build>
         <plugins>
-          <!-- Relocate all dependencies to internal to solve dependency conflict problem -->
           <plugin>
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-shade-plugin</artifactId>
             <version>3.4.1</version>
             <configuration>
               <relocations>
                 <relocation>
```
sfc-gh-pbennes, line 737: All relocations were adjusted to cover the TLDs

pom.xml
```diff
@@ -724,66 +728,48 @@
...
-                    <exclude>org/apache/commons/logging/impl/AvalonLogger.class</exclude>
-                  </excludes>
-                </filter>
-                <filter>
-                  <artifact>org.slf4j:slf4j-simple</artifact>
-                  <excludes>
-                    <exclude>org/slf4j/**</exclude>
-                  </excludes>
-                </filter>
-              </filters>
```
sfc-gh-pbennes, line 786: It's not obvious if any of this needs to stay but it seemed like it might be able to go.

pom.xml
```diff
@@ -798,6 +784,33 @@
...
+                    <move file="${project.build.directory}/tmp/META-INF/versions/9/net/snowflake/client" todir="${project.build.directory}/tmp/META-INF/versions/9/net/snowflake/ingest/shade/net/snowflake/"/>
+                    <move file="${project.build.directory}/tmp/META-INF/versions/11/net/snowflake/client" todir="${project.build.directory}/tmp/META-INF/versions/9/net/snowflake/ingest/shade/net/snowflake/"/>
+                    <move file="${project.build.directory}/tmp/META-INF/versions/15/net/snowflake/client" todir="${project.build.directory}/tmp/META-INF/versions/9/net/snowflake/ingest/shade/net/snowflake/"/>
+                    <zip basedir="${project.build.directory}/tmp" destfile="${project.build.directory}/${project.build.finalName}.jar"/>
+                    <delete dir="${project.build.directory}/tmp"/>
+                  </target>
+                </configuration>
+              </execution>
+            </executions>
+          </plugin>
```
sfc-gh-pbennes, line 813: The relocation plugin doesn't handle multi-release jars, but https://stackoverflow.com/questions/74119618/maven-shade-plugin-relocate-classes-under-meta-inf-versions-in-multi-release-jar provides a hacky workaround

scripts/check_content.sh
sfc-gh-pbennes, line 1: It looks like the original intent of this script was to catch the problems fixed by this PR, however I get the impression that people didn't fully understand the errors this script would report and kept adding exclusions. The modifications attempt to return the script to its apparent intent: report on any classes outside package net.snowflake.ingest after shading.

sfc-gh-pbennes:
#490 addresses this.

PR #494, opened 2023-05-09
SNOW-811371 Shade and Relocate Snowflake JDBC Classes

sfc-gh-wfateem:
Users may run more than one application in a single JVM for different purposes. If one application is built with the ingest SDK and another one with the JDBC driver, then class conflicts will occur. Excluding the JDBC driver is a breaking change as well that would impact these applications and potentially the Snowflake Kafka Connector as well.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -788,15 +788,18 @@
             <artifactId>maven-shade-plugin</artifactId>
             <version>3.4.1</version>
             <configuration>
-              <artifactSet>
-                <excludes>
-                  <exclude>net.snowflake:snowflake-jdbc</exclude>
-                </excludes>
-              </artifactSet>
               <relocations>
                 <relocation>
                   <pattern>com.nimbusds</pattern>
                   <shadedPattern>${shadeBase}.com.nimbusds</shadedPattern>
+                </relocation>
+                <relocation>
+                  <pattern>net.snowflake.client</pattern>
+                  <shadedPattern>${shadeBase}.net.snowflake.client</shadedPattern>
+                </relocation>
+                <relocation>
+                  <pattern>com.snowflake.client</pattern>
+                  <shadedPattern>${shadeBase}.com.snowflake.client</shadedPattern>
                 </relocation>
                 <relocation>
                   <pattern>net.jcip</pattern>


sfc-gh-tzhang (approved):

sfc-gh-pbennes:
> Users may run more than one application in a single JVM for different purposes. If one application is built with the ingest SDK and another one with the JDBC driver, then class conflicts will occur. Excluding the JDBC driver is a breaking change as well that would impact these applications and potentially the Snowflake Kafka Connector as well.

I'm not sure I understand what this would look like. Can you give an example?

sfc-gh-wfateem:
@sfc-gh-pbennes I'm not sure how we would be able to test this because it's a runtime problem. I've given an example in SNOW-811371 of just a recent outage for a partner of ours. 
The Snowflake JDBC driver and the Ingest SDK are gaining popularity and more vendors are including those as part of their offerings. NiFi is maybe a good example to consider. Multiple users can use no-code development to move data and transform that from multiple types of sources to multiple targets. You could create a generic JDBC node to write data into Snowflake and select the JDBC driver you want NiFi to load for you, let's say that's v3.13.18. Another user logged in can create another node that loads data into Snowflake using Java code that leverages the ingest SDK, let's say that's v1.1.3. You're going to end up with those two libraries on the classpath and both of those are going to include the exact same classes of the Snowflake JDBC driver (but from versions 3.13.18 and 3.13.29). In return, those two are shading two different versions of Bouncy Castle.
This isn't a problem on the main branch of Ingest SDK because it's been excluded as part of the changes in #490. Following the same example, that's great, but then poses a nightmare from a supportability standpoint because we don't know what JDBC driver version is going to be added manually on the classpath by a user. A slightly older JDBC driver version will break all proxy connection-related functionality for example. 
The best short-term solution I can think of is to include the JDBC driver with the Ingest SDK and relocate the classes for the JDBC driver.

sfc-gh-akazantsev:
FWIW: shading is generally risky pattern and the only acceptable way is to shade complete dependency graph with all transitives.

sfc-gh-lsembera:
> A Java application built against this library and runtime environments that only have the Ingest SDK library on the classpath will fail as soon as they upgrade (since we now removed the JDBC driver needed).

@sfc-gh-wfateem The JDBC driver isn't removed, it is still as marked a compile dependency of the shaded ingest SDK JAR, so the client application depending on SDK will fetch JDBC as a transitive dependency. 

Anyway, I don't have anything against relocating JDBC driver, as well, if it brings more safety. I wanted to avoid situation, when customers who need both SDK and JDBC would have JDBC driver essentially twice on the classpath.

sfc-gh-wfateem:
@sfc-gh-lsembera When I checked out the main branch and built that then inspected the artifact, here's what I saw:
```
❯ jar tvf ./snowflake-ingest-sdk.jar | grep -i jdbc
  6386 Sun Jul 28 18:16:38 CDT 2019 net/snowflake/ingest/internal/apache/commons/beanutils/JDBCDynaClass.class
  5203 Thu Mar 31 14:41:24 CDT 2022 net/snowflake/ingest/internal/apache/commons/configuration2/DatabaseConfiguration$JdbcOperation.class
```

I saw the following in the pom.xml as part of #490 which to me explained why I see the final uber JAR without any of the Snowflake JDBC classes when I build the main branch, as opposed to when I build the v1.1.3 branch:
```
      <build>
        <plugins>
          <!-- Relocate all dependencies to internal to solve dependency conflict problem -->
          <plugin>
            <artifactId>maven-shade-plugin</artifactId>
            <version>3.4.1</version>
            <configuration>
              <artifactSet>
                <excludes>
                  <exclude>net.snowflake:snowflake-jdbc</exclude>
                </excludes>
              </artifactSet>
```

sfc-gh-pbennes:
The only other thing to add is that given #490 it's possible there's nothing to do here but release 1.1.4 without any shaded snowflake-jdbc driver. In the case of https://github.com/sfc-gh-pbennes/snowflake-class-conflict-example that would result in only snowflake-jdbc 3.13.18 being used and on the classpath which I would guess would work.

sfc-gh-pbennes:
> A Java application built against this library and runtime environments that only have the Ingest SDK library on the classpath will fail as soon as they upgrade (since we now removed the JDBC driver needed).

In any normal case snowflake-jdbc just moves from being a shaded dependency as part of ingest sdk to being a normal transitive dependency. Any normal build tool would download the transitive dependencies and add them to the classpath the same as any other dependency. I think the only case where this is breaking is if someone is doing something like trying to hot-upgrade or hot-reload which virtually no libraries support. If this is a constraint for our libraries it would be good to both document that and add tests for it.

sfc-gh-pbennes:
> @sfc-gh-lsembera When I checked out the main branch and built that then inspected the artifact, here's what I saw:
> 
> ```
> ❯ jar tvf ./snowflake-ingest-sdk.jar | grep -i jdbc
>   6386 Sun Jul 28 18:16:38 CDT 2019 net/snowflake/ingest/internal/apache/commons/beanutils/JDBCDynaClass.class
>   5203 Thu Mar 31 14:41:24 CDT 2022 net/snowflake/ingest/internal/apache/commons/configuration2/DatabaseConfiguration$JdbcOperation.class
> ```
> 
> I saw the following in the pom.xml as part of #490 which to me explained why I see the final uber JAR without any of the Snowflake JDBC classes when I build the main branch, as opposed to when I build the v1.1.3 branch:
> 
> ```
>       <build>
>         <plugins>
>           
>           <plugin>
>             <groupId>org.apache.maven.plugins</groupId>
>             <artifactId>maven-shade-plugin</artifactId>
>             <version>3.4.1</version>
>             <configuration>
>               <artifactSet>
>                 <excludes>
>                   <exclude>net.snowflake:snowflake-jdbc</exclude>
>                 </excludes>
>               </artifactSet>
> ```

Right, this is working as intended. In the dependency-reduced-pom.xml this produces there should still be a listed dependency for snowflake-jdbc so build tools can pull in that dependency.

sfc-gh-pbennes (approved):
Discussed with Waleed and they would like to defer the removal of shading in an abdundance of concern for our customers which seems reasonable to me: +1

sfc-gh-wfateem:
@sfc-gh-pbennes Right now, this is going to roll out in the very next release of the Ingest SDK, and we just don't have a good way of determining what the possible impact is. As I mentioned on Slack, there are scenarios where people aren't using build tools and simply rely on downloading artifacts directly from Maven. Therefore, my preference is that we still include the JDBC driver because that's how it's been working all along, and we plan to clean this up in a more controlled manner (possibly BCR).

sfc-gh-pbennes:
Agreed. BCR / 2.x release would be a good way to try to introduce no-shading/reduced-shading more gently.

sfc-gh-lsembera:
@sfc-gh-pbennes @sfc-gh-wfateem Sounds good! Tomorrow I will trigger our dew tests with this new jar and if everything is good, we can merge.

sfc-gh-pbennes:
> @sfc-gh-pbennes @sfc-gh-wfateem I executed dew and SDK integration tests (from another Maven project that depends on the shaded JAR) and they are all passing. However, this PR introduces a new potential incompatibility:
> 
> SDK users using a build system like Maven have had access to classes from the JDBC driver automatically, even without declaring the JDBC driver as an explicit dependency, both before and after #490. Before #490 the classes had been inside of the SDK jar unrelocated, after #490 the driver was pulled transitively by Maven. This PR relocates them and removes the dependency from the reduced `pom.xml`, so classes from the JDBC driver (like the entry point `net.snowflake.client.jdbc.SnowflakeDriver`) are no longer available. Once this PR is merged, users will have to pull in the JDBC driver explicitly.
> 
> @sfc-gh-tzhang We should mention it in readme and release notes.

Good point. While this is an improvement from a Java-library-best-practices PoV this is a change that might surprise people if they were relying on this behavior. +1 on a note in the release.

sfc-gh-lsembera (approved):

sfc-gh-tjones (approved):

sfc-gh-japatel:
> Once this PR is merged, users will have to pull in the JDBC driver explicitly.
This in itself is a BCR for me. We should also explicitly mention which version of the JDBC is recommended to use. Even the patch versions are sometimes incompatible because of bug fixes. 
(For instance if ingest-sdk with version 3.13.25 will have httpProxy feature but if some older version is used, it will start failing)

(Which makes me think more about @sfc-gh-pbennes comment about semver abuse, but that is a different discussion altogether)

sfc-gh-wfateem:
@sfc-gh-japatel @sfc-gh-lsembera @sfc-gh-pbennes If there's still a potential for a breaking change, then do we want to release this in 2.0 instead, or roll it back and release #490 in 2.0? It sounds like no matter how we want to go about this there will always be the possibility of breaking something one way or the other, so we might as well agree on what is the right approach to take, stick to that, and then release a BCR version.

The Snowflake Spark connector is an example of another client that depends on the JDBC driver, but we have a different approach there where we simply document the required JDBC driver version, and it's up to the user to ensure that's on the classpath.

If either way we're going to break something, then wouldn't it just make sense to choose the better/cleaner approach and highlight that there's a breaking change in 2.0?

sfc-gh-wfateem:
@sfc-gh-japatel thoughts?

sfc-gh-japatel:
Agreed to do it altogether in 2.0 release. 
(@sfc-gh-tzhang and @sfc-gh-lsembera  are leading the GA readiness for ingest sdk so I would like to get their thoughts too.)

PR #498, opened 2023-05-17
NO-SNOW: Update MAX_CHUNK_SIZE_IN_BYTES default to be 32MB and make it configurable 

sfc-gh-tzhang:
We have seen many cases where the BDEC file is too small and causing high cost, so increasing the max chunk size from 16MB to 32MB. To avoid regressions that prevent customers to upgrade, we make the value configurable.


src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
```diff
@@ -189,7 +190,7 @@ public boolean supportsBlockSize() {
 
     @Override
     public long defaultBlockSize() {
-      return (int) MAX_CHUNK_SIZE_IN_BYTES;
+      return maxChunkSizeInBytes;
```
sfc-gh-tzhang, line 193: @sfc-gh-gdoci Any particular reason this was cast to int before?

sfc-gh-lsembera (approved):
LGTM

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
sfc-gh-gdoci: hmm, no, you can remove the cast.

sfc-gh-gdoci (approved):

PR #501, opened 2023-05-18
SNOW-758492: avoid combining channels with different schema

sfc-gh-tzhang:
We could have two channels with different schema when the older channel haven't got invalidated yet due to schema changes, avoid combining these two channels into the same chunk so the newer channel's data could be committed


src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -471,6 +467,25 @@ void distributeFlushTasks() {
     this.registerService.addBlobs(blobs);
   }
 
+  /**
+   * Check whether we should stop merging more channels into the same chunk, we need to stop in a
+   * few cases
```
sfc-gh-tjones, line 472: Nit, `few cases:`

sfc-gh-tjones (approved):

PR #503, opened 2023-05-22
Revert "SNOW-811371 Shade and Relocate Snowflake JDBC Classes (#494)"

sfc-gh-lsembera:
This PR reverts relocating `snowflake-jdbc` driver in the shaded JAR. Relocation caused test failures and didn't solve the problem it was supposed to - not having any impact on existing users. Existing users who depended on the JDBC driver without declaring it in their pom.xml, would still see errors.

We should release this change as 2.0.0, so that it is clear from the version number that some dependencies have changes. 
Also, we will be releasing an unshaded variant in every release.

To sum up, the final state is: JDBC driver is not shaded, not relocated, it will be declared as a dependency in the final reduced pom.xml.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -794,18 +794,15 @@
             <artifactId>maven-shade-plugin</artifactId>
             <version>3.4.1</version>
             <configuration>
+              <artifactSet>
+                <excludes>
+                  <exclude>net.snowflake:snowflake-jdbc</exclude>
+                </excludes>
+              </artifactSet>
               <relocations>
                 <relocation>
                   <pattern>com.nimbusds</pattern>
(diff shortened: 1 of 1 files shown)


sfc-gh-japatel (approved):
lgtm! 

What about #490 ? Are we reverting that as well before next release?

sfc-gh-wfateem (approved):
LGTM

sfc-gh-pbennes (approved):

sfc-gh-tzhang:
@sfc-gh-lsembera I thought our next release by end of this month will be 2.0.0, should we add this back?

sfc-gh-lsembera:
@sfc-gh-tzhang No, this PR isn't coming back. At the end of the month, we release 2.0.0, which will contain #490 only.

PR #504, opened 2023-05-23
SNOW-768195: [Client Side] Support GCS downscoped token for Snowpipe Streaming

sfc-gh-tzhang:


sfc-gh-japatel (approved):
lgtm, Thanks! 
Please ensure test `testGCSDownscopedToken` is passing. It's failing currently.

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
```diff
@@ -186,6 +185,7 @@ private void putRemote(String fullFilePath, byte[] data, int retryCount)
               .setStreamingIngestClientKey(this.clientPrefix)
               .setStreamingIngestClientName(this.clientName)
               .setProxyProperties(this.proxyProperties)
+              .setDestFileName(fullFilePath)
```
sfc-gh-japatel, line 188: Should this only be the file name instead of full path?

sfc-gh-tzhang: Not sure why in KC it's the file name, but I'm pretty sure we need to use fullFilePath here, and Spark Connector is using full path as well

sfc-gh-tzhang:
> lgtm, Thanks! Please ensure test `testGCSDownscopedToken` is passing. It's failing currently.

Do you think we can give SNOWPIPE_SDK_GH_ACTION accountadmin role? How did you test KC?

Insufficient privileges to operate on user 'SNOWPIPE_SDK_GH_ACTION'

sfc-gh-japatel:
> > lgtm, Thanks! Please ensure test `testGCSDownscopedToken` is passing. It's failing currently.
> 
> @sfc-gh-japatel Do you think we can give SNOWPIPE_SDK_GH_ACTION accountadmin role? How did you test KC?
> 
> Insufficient privileges to operate on user 'SNOWPIPE_SDK_GH_ACTION'

I ran the test with param enabled on the user and then re-ran it with param disabled. why is accountadmin needed.

sfc-gh-tzhang:
> > > lgtm, Thanks! Please ensure test `testGCSDownscopedToken` is passing. It's failing currently.
> > 
> > 
> > @sfc-gh-japatel Do you think we can give SNOWPIPE_SDK_GH_ACTION accountadmin role? How did you test KC?
> > Insufficient privileges to operate on user 'SNOWPIPE_SDK_GH_ACTION'
> 
> I ran the test with param enabled on the user and then re-ran it with param disabled. why is accountadmin needed.?

How did you enable the parameter in test? Accountadmin is needed because I'm getting insufficient privilege

sfc-gh-japatel:
Other way to do it is to set `alter account <accountname> set CHANGE_INTERNAL_PARAMS = true` from snowflake. this was I guess you can do it from your tests here. (making internal param external) I havent tried it though.

sfc-gh-rcheng (approved):

PR #505, opened 2023-05-23
@snow SNOW-763199 Enable StreamingIngestIT.testTableColumnEvolution

sfc-gh-azagrebin:
Server side now supports column ids for Parquet.
Re-enabling column evolution test with enabled change tracking: `StreamingIngestIT.testTableColumnEvolution`.


sfc-gh-tzhang (approved):
what's the plan of switching to use field id instead?

sfc-gh-azagrebin:
> what's the plan of switching to use field id instead?

I have not fulled reviewed this concept yet.
I thought we need this for Iceberg or something else as well?
We need to put it on road map then.

PR #512, opened 2023-05-30
Rcheng flushreason

sfc-gh-rcheng:


sfc-gh-rcheng:
@sfc-gh-xhuang We do not need it. It would be nice to have, however we'd need to wait for server side changes which I haven't started on yet. We deprioritized this work due to the recent customer issues.

src/main/java/net/snowflake/ingest/streaming/internal/BlobMetadata.java
```diff
@@ -13,11 +13,29 @@
 
 /** Metadata for a blob that sends to Snowflake as part of the register blob request */
 class BlobMetadata {
+  enum BlobCreationReason {
+    NONE("NONE"),
+    BUFFER_BYTE_SIZE("max buffer threshold hit"),
+    ENCRYPTION_KEY_ID_CHANGE("encryption key id changed"),
+    SCHEMA_CHANGE("schema changed"),
+    ;
```
sfc-gh-xhuang, line 21: do we need another reason for the new max_lag?

PR #514, opened 2023-06-02
SNOW-830862 Add slf4j-api; fix linkage errors

sfc-gh-lsembera:
This PR:

* Adds `slf4j-api` back to `compile` scope as this is a required dependency
    * It is not shaded and not relocated, just like the JDBC driver. It is standard library used by everybody with no risk of conflicts. 
* Documents that in order to have console logs in examples, users should uncomment `test` scope from slf4j-simple. 
* Fixed linkage checker errors.
* Makes sure that linkage checker only runs on unshaded JARs, which we are running on Windows.


sfc-gh-tjones (approved):

pom.xml
```diff
@@ -487,6 +485,7 @@
       <version>2.0.2</version>
       <scope>test</scope>
     </dependency>
+    <!-- Change the 'test' scope to 'runtime' to enable console logging in examples -->
```
sfc-gh-tzhang, line 488: Did you run into dependency order issue when you remove the test scope?

sfc-gh-lsembera: No, I didn't, I ran the streaming ingest example from IDE without any problems. Are you running into issues? Could you try before I merge it, please?

sfc-gh-tzhang (approved):
lgtm

pom.xml
sfc-gh-tzhang: Signed off to unblock you, but I change it to runtime and run `mvn clean install -DskipTests`, it gives the error like below
```
[ERROR] The line 448 is not considered sorted, should be '    <!-- Change the 'test' scope to 'runtime' to enable console logging in examples -->'
[ERROR] The file /home/tzhang/Snowflake/snowflake-ingest-java/pom.xml is not sorted
```

sfc-gh-pbennes: `mvn sortpom:sort` will auto-fix sorting issues

sfc-gh-lsembera: I will make checking if the sorted is sorted a merge-gate only check in a separate PR.

PR #516, opened 2023-06-05
NO-SNOW: Limit the chunk size to avoid potential OOM issue in the Parquet Reader

sfc-gh-tzhang:
Given that we have some limitations in the current Parquet Reader, this PR contains the change to limit the chunk size to avoid any potential OOM risks. It's currently set to 128MB uncompressed based on the reader expert and we also add the support to override it if needed to avoid any potential regressions.


sfc-gh-rcheng (approved):
lgtm. going to merge conflict with my flush reason PR, but this is more important so I'll wait for this to go in first

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -25,7 +25,11 @@ public class ParameterProvider {
   public static final String MAX_MEMORY_LIMIT_IN_BYTES = "MAX_MEMORY_LIMIT_IN_BYTES".toLowerCase();
   public static final String ENABLE_PARQUET_INTERNAL_BUFFERING =
       "ENABLE_PARQUET_INTERNAL_BUFFERING".toLowerCase();
+  // This is actually channel size limit at this moment until we implement the size tracking logic
+  // at table/chunk level
   public static final String MAX_CHUNK_SIZE_IN_BYTES = "MAX_CHUNK_SIZE_IN_BYTES".toLowerCase();
```
sfc-gh-gdoci, line 30: does it make sense to rename this to `MAX_CHANNEL_BUFFER_SIZE_IN_BYTES`?

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -25,7 +25,11 @@ public class ParameterProvider {
...
+  public static final String MAX_CHUNK_SIZE_IN_BYTES_TO_AVOID_OOM =
+      "MAX_CHUNK_SIZE_IN_BYTES_TO_AVOID_OOM".toLowerCase();
```
sfc-gh-gdoci, line 32: If we rename the previous one, this one be `MAX_CHUNK_SIZE_IN_BYTES`

src/test/java/net/snowflake/ingest/streaming/internal/FlushServiceTest.java
```diff
@@ -505,6 +519,40 @@ public void testBlobSplitDueToDifferentSchema() throws Exception {
...
+    List<Map<String, Object>> rows = builder.build();
+
+    channel1.insertRows(rows, "offset1");
+    channel2.insertRows(rows, "offset2");
+
+    FlushService<?> flushService = testContext.flushService;
+
+    // Force = true flushes
+    flushService.flush(true).get();
+    Mockito.verify(flushService, Mockito.times(2)).buildAndUpload(Mockito.any(), Mockito.any());
```
sfc-gh-gdoci, line 553: Can we add a check for the chunk sizes to make sure they are deterministic and as expected?

sfc-gh-gdoci (commented):
Thanks Toby, I left a few comments.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -370,28 +370,36 @@ void distributeFlushTasks() {
 
         if (!channelsDataPerTable.isEmpty()) {
           int idx = 0;
+          float totalBufferSizePerTableInBytes = 0F;
           while (idx < channelsDataPerTable.size()) {
             ChannelData<T> channelData = channelsDataPerTable.get(idx);
             // Stop processing the rest of channels when needed
             if (idx > 0
                 && shouldStopProcessing(
```
sfc-gh-lsembera, line 378: I approved by accident, sorry. I think this PR still doesn't protect us from generating too large files due to asynchronous flush. There is no guarantee that between this check and actual flush, there aren't additional rows added to the buffer, right? Didn't we want to throw an exception at flush time when we see a large blob?

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -473,15 +481,22 @@ && shouldStopProcessing(
    * Check whether we should stop merging more channels into the same chunk, we need to stop in a
    * few cases:
    *
-   * <p>When the size is larger than a certain threshold
+   * <p>When the file size is larger than a certain threshold
```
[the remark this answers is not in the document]
sfc-gh-rcheng: We are worried about file size for migration costs, but that is out of scope for this PR. I think this comment update just clarifies that we stop processing both on totalBufferSizeInBytes and totalBufferSizePerTableInBytes

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-tzhang: I believe you can reset your status. We first create a snapshot so this size won't change anymore, all additional rows will be added to a new buffer

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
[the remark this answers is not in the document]
sfc-gh-tzhang: The file only affects the upload time, and we don't want to upload a huge file which affects end2end latency

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-tzhang: Makes sense, let me update

src/test/java/net/snowflake/ingest/streaming/internal/FlushServiceTest.java
sfc-gh-tzhang: I don't think we can easily get the blob/chunk information, any suggestions? Is there way to get the response from buildAndUpload using Mockito?

PR #517, opened 2023-06-06
SNOW-768197 Attribution notice for shaded dependencies

sfc-gh-lsembera:
We need to include full license of SDK dependencies in the shaded JAR. Dependencies, which do not ship with a license file, should be mentioned with their license name in a separate file.

The license processing logic is implemented as a Python script, which is easier to write and maintain than bash and it will work on Windows, too.


sfc-gh-tjones (approved):

sfc-gh-xhuang (approved):

sfc-gh-tzhang (approved):
LGTM, could we ask @sfc-gh-pbennes to take a look?

sfc-gh-pbennes (approved):
Looks good-enough to me. Ship it.

This isn't anything I've done before so I don't have much to say otherwise. I assume we looked for an existing maven plugin to do this and didn't find one. It's unfortunate there's not already a maven plugin to do this, and we might consider making one in the future as a small improvement over the python script approach, but if it works, ship it.

PR #519, opened 2023-06-08
SNOW-720420 Enforce per-row size limits

sfc-gh-lsembera:
This PR enforces the max row limit of 64 MB.

### DIFF
--- a/src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
+++ b/src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
@@ -28,6 +28,8 @@ public class ParameterProvider {
   // This should not be needed once we have the ability to track size at table/chunk level
   public static final String MAX_CHANNEL_SIZE_IN_BYTES = "MAX_CHANNEL_SIZE_IN_BYTES".toLowerCase();
   public static final String MAX_CHUNK_SIZE_IN_BYTES = "MAX_CHUNK_SIZE_IN_BYTES".toLowerCase();
+  public static final String MAX_ALLOWED_ROW_SIZE_IN_BYTES =
+      "MAX_ALLOWED_ROW_SIZE_IN_BYTES".toLowerCase();
 
   // Default values
   public static final long BUFFER_FLUSH_INTERVAL_IN_MILLIS_DEFAULT = 1000;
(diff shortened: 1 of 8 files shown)


src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -40,6 +42,7 @@ public class ParameterProvider {
   public static final int BLOB_UPLOAD_MAX_RETRY_COUNT_DEFAULT = 24;
   public static final long MAX_MEMORY_LIMIT_IN_BYTES_DEFAULT = -1L;
   public static final long MAX_CHUNK_SIZE_IN_BYTES_DEFAULT = 32000000L;
+  public static final long MAX_ALLOWED_ROW_SIZE_IN_BYTES_DEFAULT = 32 * 1024 * 1024; // 32 MB
```
sfc-gh-tzhang, line 45: I thought we want to do 128MB to match the max chunk size?

sfc-gh-azagrebin: The concern was here that concurrent flush is most dangerous in case of huge rows because if one compressed row is close to 64MB (compressed limit), adding just one more row concurrently basically results in constant oversizing.
We could try to analyse in Snowhouse what is currently the expected compressed row size (chunk size / row count).

sfc-gh-azagrebin: This would be also inconsistent to keep it more than `MAX_CHANNEL_SIZE_IN_BYTES_DEFAULT = 32MB`, right?

sfc-gh-lsembera: Good point, @sfc-gh-tzhang we are still enforcing max channel size of 32mb? Then the row size should be less than that, right?

sfc-gh-xhuang (approved):

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-tzhang: I'm not sure why max channel size matters here, customer can insert rows more than 32MB and the query works as well. I thought the concern is that we don't want to generate too big file that may cause OOM, so 128MB is the limit we use. Remember that JDBC 32MB is compressed, i feel like 32MB uncompressed is too aggressive, WDYT?

sfc-gh-lsembera: Currently, we enforce that the max channel size is `32000000` and my PR is setting max row size to `32*1024*1024`, which is more than that, so it probably does not make sense.

sfc-gh-tzhang (approved):
lgtm, please also update the description

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-lsembera: Discussed this with @sfc-gh-tzhang offline - channel size is only a hint when to flush, it is not used for any validations, so it is fine to have it less than row size.

PR #520, opened 2023-06-08
NO-SNOW Enforce allowed DATE and TIMESTAMP range

sfc-gh-lsembera:
This PR introduces enforcement of min and max allowed dates and timestamps. For dates, we allow the range [-9999..9999], for timestamps we allow [1..9999].

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
@@ -454,6 +454,11 @@ class DataValidationUtil {
     if (trimTimezone) {
       offsetDateTime = offsetDateTime.withOffsetSameLocal(ZoneOffset.UTC);
     }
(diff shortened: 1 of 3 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -454,6 +454,11 @@ static TimestampWrapper validateAndParseTimestamp(
     if (trimTimezone) {
       offsetDateTime = offsetDateTime.withOffsetSameLocal(ZoneOffset.UTC);
     }
+    if (offsetDateTime.getYear() < 1 || offsetDateTime.getYear() > 9999) {
```
sfc-gh-tzhang, line 457: What's the current behavior if we have an invalid time?

sfc-gh-tzhang (approved):
lgtm, thanks!

sfc-gh-xhuang:
Where's the requirement for this coming from?

sfc-gh-xhuang (approved):

PR #521, opened 2023-06-08
@snow SNOW-835618 Snowpipe Streaming: send uncompressed chunk length from SDK to GS

sfc-gh-azagrebin:
[SNOW-835618](https://snowflakecomputing.atlassian.net/browse/SNOW-835618)

Currently, we send only compressed chunk length in blob metadata.
We can also send uncompressed and log in ingestion events to analyse compression ratio stats.

[SNOW-835618]: https://snowflakecomputing.atlassian.net/browse/SNOW-835618?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
@@ -114,6 +114,7 @@ class BlobBuilder {
                 // The paddedChunkLength is used because it is the actual data size used for
                 // decompression and md5 calculation on server side.
                 .setChunkLength(paddedChunkLength)
+                .setUncompressedChunkLength((int) serializedChunk.chunkEstimatedUncompressedSize)
                 .setChannelList(serializedChunk.channelsMetadataList)
                 .setChunkMD5(md5)
                 .setEncryptionKeyId(firstChannelFlushContext.getEncryptionKeyId())
@@ -132,13 +133,13 @@ class BlobBuilder {
 
         logger.logInfo(
             "Finish building chunk in blob={}, table={}, rowCount={}, startOffset={},"
-                + " uncompressedSize={}, paddedChunkLength={}, encryptedCompressedSize={},"
+                + " estimatedUncompressedSize={}, paddedChunkLength={}, encryptedCompressedSize={},"
                 + " bdecVersion={}",
             filePath,
             firstChannelFlushContext.getFullyQualifiedTableName(),
             serializedChunk.rowCount,
             startOffset,
-            serializedChunk.chunkUncompressedSize,
+            serializedChunk.chunkEstimatedUncompressedSize,
             paddedChunkLength,
             encryptedCompressedChunkDataSize,
             bdecVersion);
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ChunkMetadata.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ChunkMetadata.java
@@ -15,6 +15,7 @@ class ChunkMetadata {
   private final String tableName;
   private Long chunkStartOffset;
   private final Integer chunkLength;
+  private final Integer uncompressedChunkLength;
   private final List<ChannelMetadata> channels;
   private final String chunkMD5;
   private final EpInfo epInfo;
@@ -33,6 +34,9 @@ class ChunkMetadata {
     private String tableName;
     private Long chunkStartOffset;
     private Integer chunkLength; // compressedChunkLength
+
+    private Integer uncompressedChunkLength;
+
     private List<ChannelMetadata> channels;
     private String chunkMD5;
     private EpInfo epInfo;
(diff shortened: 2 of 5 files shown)


sfc-gh-lsembera (approved):

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -114,6 +114,7 @@ static <T> Blob constructBlobAndMetadata(
                 // The paddedChunkLength is used because it is the actual data size used for
                 // decompression and md5 calculation on server side.
                 .setChunkLength(paddedChunkLength)
+                .setUncompressedChunkLength((int) serializedChunk.chunkUncompressedSize)
```
sfc-gh-tzhang, line 117: This size will be different compared with our estimated uncompressed buffer size, right? Is that something you want?

src/main/java/net/snowflake/ingest/streaming/internal/ChunkMetadata.java
```diff
@@ -152,6 +162,11 @@ Integer getChunkLength() {
     return chunkLength;
   }
 
+  @JsonProperty("chunk_length_uncompressed")
```
sfc-gh-tzhang, line 165: I believe this will require a server side change first

sfc-gh-tzhang (commented):
Could we add a simple test?

src/main/java/net/snowflake/ingest/streaming/internal/ChunkMetadata.java
sfc-gh-azagrebin: server side change is merged

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
sfc-gh-azagrebin: I think this is what we estimate. We do not have the actual uncompressed size because it is tracked only internally in Parquet lib.

src/main/java/net/snowflake/ingest/streaming/internal/ChunkMetadata.java
sfc-gh-azagrebin: but yes, we have to wait

sfc-gh-tzhang (approved):
lgtm

sfc-gh-tzhang:
> [SNOW-835618](https://snowflakecomputing.atlassian.net/browse/SNOW-835618)

Please put a meaningful description instead of just linking the JIRA, thanks!

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
sfc-gh-rcheng: Also can we adjust the naming from 'chunkUncompressedSize' to 'chunkUncompressedSizeEstimate' to reduce this confusion?

sfc-gh-rcheng (approved):
lgtm - left a small nit and a question about actual uncompressed chunk length

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
sfc-gh-azagrebin: True, I changed the variable name and in the result of flush but not in server response.
The problem is that now server side can accept only this name for the size.

The current intention was to send the estimated size to investigate the compression ratio based on the estimation
because the way we use the parquet lib now we cannot know the actual size before we start the flush
hence we have only estimation for compressed prediction before we start the flush.

This is a good point about the actual size. I looked again into the code and I think it is possible
now to report the actual uncompressed size from the parquet lib after the flush and I added now to the
serialisation result. As I understand the expected difference should be very small and close to negligible.
Nonetheless, we could send both if we change the server size to accept both but for now
as I mentioned in the beginning, I would prefer to send the estimation.

Let me know what you think.

sfc-gh-rcheng: Makes sense, thank you for looking into this. Just want to make sure we have a vague plan for reporting the actual size in the future - can we change `chunk_length_uncompressed` to represent the actual size in the future? I'm worried about future confusion if the field changes across versions compared to the current naming.

Ie should we have: 
- `chunk_length_uncompressed` and `estimated_chunk_length_uncompressed`
- `chunk_length_uncompressed` and `actual_chunk_length_uncompressed`

sfc-gh-azagrebin: > can we change chunk_length_uncompressed to represent the actual size in the future?

This is not so easy because it is a part of client/server API.
Server already accepts `chunk_length_uncompressed` not anything else.
We can just add `actual_chunk_length_uncompressed` later and have a comment for `chunk_length_uncompressed`. I agree that naming should have been better but otherwise there are 2 ways:
- deprecating the old name that will be a hassle because of old clients sill sending the old name
- or we would have to change first server side and block this PR until server side is released. It would mean that we will not able to gather any stats for more time and customers will use older post-GA SDK versions that do not send any stats.

Eventually, I think we should have only actual size stats if we decide to refactor parquet generator or we will always use estimation.  We do not really need both. The estimation reflects the user data good enough. The actual size will just include a bit more or less parquet specific overhead that is not user data anyways.

sfc-gh-azagrebin: Moreover, if you ask about customer throughput, we should actually compute it a bit differently but call it completely differently :) Parquet encoding for page 2 (we use page 1) is like compression and it can significantly reduce the actual customer data size. I assume when you say `customer throughput` you mean the actual data bytes that customers insert and not how parquet packs them, even before the compression. Hence, it would be a different stat to report, something like `chunk_user_data_size` or something.

PR #524, opened 2023-06-09
NO-SNOW Exclude *.md and *.xml from code coverage

sfc-gh-xhuang:
Exclude *.md and *.xml from code coverage

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -564,6 +564,8 @@
             <excludes>
               <exclude>**/*SnowflakeStreamingIngestExample*</exclude>
               <exclude>**/*SnowflakeIngestBasicExample*</exclude>
+              <exclude>*.md</exclude>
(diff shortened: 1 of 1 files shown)


sfc-gh-lsembera:
I don't think these are being included in code coverage, see https://app.codecov.io/gh/snowflakedb/snowflake-ingest-java

sfc-gh-lsembera:
Could you also exclude [IngestExampleHelper.java](https://app.codecov.io/gh/snowflakedb/snowflake-ingest-java/blob/master/src%2Fmain%2Fjava%2Fnet%2Fsnowflake%2Fingest%2Fexample%2FIngestExampleHelper.java)? It is only used by the examples, so there is no need to report code coverage on it.

sfc-gh-lsembera (approved):
LGTM, Thank you!

PR #527, opened 2023-06-14
NO-SNOW Specify dependencies in public_pom.xml

sfc-gh-lsembera:


sfc-gh-tzhang:
This is good for now, for longer term we should come up with a better way to automate this instead of updating it manually

sfc-gh-japatel (commented):
What happens to the location of this dependency? Do we need relocation here?

sfc-gh-xhuang (approved):

sfc-gh-pbennes (approved):
Ohh yeah I can see how we missed this.

sfc-gh-lsembera:
> What happens to the location of this dependency? Do we need relocation here?

No, these two are not being shaded, they will be declared as transitive dependencies of the SDK in the final pom.xml uploaded to Maven Central.

sfc-gh-pbennes:
> This is good for now, for longer term we should come up with a better way to automate this instead of updating it manually

The maven-shade-plugin supports this, but the other way around. Usually jars are unshaded by default and then when shading the maven-shade-plugin will produce a separate dependency-reduced-pom.xml which doesn't include shaded dependencies in the dependencies section of the published pom.xml.

If we move to that model and align with the maven-shade-plugin that will make managing both regular and shaded jars easier to maintain I think.

PR #528, opened 2023-06-14
V2.0.1 release

sfc-gh-tzhang:
This is a patch release on top of the v2.0.0 where we fix an unexpected dependency behavior for Snowflake JDBC


pom.xml
```diff
@@ -6,7 +6,7 @@
   <!-- Arifact name and version information -->
   <groupId>net.snowflake</groupId>
   <artifactId>snowflake-ingest-sdk</artifactId>
-  <version>2.0.0</version>
+  <version>2.0.1</version>
```
sfc-gh-xhuang, line 9: should we also release 2.0.1-unshaded?

sfc-gh-lsembera (approved):

pom.xml
sfc-gh-tzhang: Yes, that will require a separate PR, at this point, I suggest let's release the shaded one first to make sure there is no other issues

sfc-gh-tjones (approved):

PR #529, opened 2023-06-15
NO-SNOW Bump version to 2.0.2-SNAPSHOT

sfc-gh-lsembera:


pom.xml
```diff
@@ -6,7 +6,7 @@
   <!-- Arifact name and version information -->
   <groupId>net.snowflake</groupId>
   <artifactId>snowflake-ingest-sdk</artifactId>
-  <version>2.0.1</version>
+  <version>2.0.2-SNAPSHOT</version>
```
sfc-gh-japatel, line 9: requestbuilder.java also has version which IMO we should update. It is sent as user agent in our rest endpoints. 
It will help us distinguish on server side.

sfc-gh-tjones (approved):

pom.xml
sfc-gh-tzhang: +1, please see the previous commit as an example

sfc-gh-tzhang (approved):

sfc-gh-asen (approved):

PR #543, opened 2023-07-03
SNOW-851351 Retry javax.net.ssl.SSLException

sfc-gh-lsembera:
This PR retries SSLException, which we have been seeing in client logs. It additionally evicts idle connections after 60s of inactivity.

### DIFF
--- a/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
+++ b/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
@@ -266,8 +266,8 @@ public class HttpUtil {
   }
 
   /**
-   * Retry handler logic. Retry at most {@link HttpUtil.MAX_RETRIES} times if any of the following exceptions is thrown by
-   * request execution:
+   * Retry handler logic. Retry at most {@link HttpUtil.MAX_RETRIES} times if any of the following
+   * exceptions is thrown by request execution:
    *
    * <ul>
    *   <li>No response from Service exception (NoHttpResponseException)
@@ -276,7 +276,7 @@ public class HttpUtil {
    *
    * @return retryHandler to add to http client.
    */
-  private static HttpRequestRetryHandler getHttpRequestRetryHandler() {
+  static HttpRequestRetryHandler getHttpRequestRetryHandler() {
     return (exception, executionCount, httpContext) -> {
       final String requestURI = getRequestUriFromContext(httpContext);
       if (executionCount > MAX_RETRIES) {
--- a/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
+++ b/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
@@ -266,7 +266,7 @@ public class HttpUtil {
   }
 
   /**
-   * Retry handler logic. Retry at most 3 times if any of the following exceptions is thrown by
+   * Retry handler logic. Retry at most {@link HttpUtil.MAX_RETRIES} times if any of the following exceptions is thrown by
    * request execution:
    *
    * <ul>
--- a/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
+++ b/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
@@ -73,6 +73,12 @@ public class HttpUtil {
   private static final int DEFAULT_CONNECTION_TIMEOUT_MINUTES = 1;
   private static final int DEFAULT_HTTP_CLIENT_SOCKET_TIMEOUT_MINUTES = 5;
 
+  /**
+   * After how many seconds of inactivity should be idle connections evicted from the connection
+   * pool.
+   */
+  private static final int DEFAULT_EVICT_IDLE_AFTER_SECONDS = 60;
+
   // Default is 2, but scaling it up to 100 to match with default_max_connections
   private static final int DEFAULT_MAX_CONNECTIONS_PER_ROUTE = 100;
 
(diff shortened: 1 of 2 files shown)


sfc-gh-wfateem:
@sfc-gh-lsembera I'm just curious about this because we made a change in the JDBC driver to explicitly not retry HTTP requests that result in SSLExceptions. If an SSLException is raised (for example [SSLHandshakeException](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLHandshakeException.html)), it's unlikely retrying the request will help. 
I just want to be sure if there's a compelling reason why we should retry a request that failed with an SSLException then we'll want to be sure the JDBC driver behaves in the same manner.

sfc-gh-wfateem:
Ok, I see this is related to stale connections, so I guess in this particular scenario it would make sense.
I think though after you added the change to evict idle connections, those reset errors shouldn't be a problem. The JDBC driver had overly aggressive retry logic so whenever a customer had SSL/TLS issues (very common) then the problem would manifest itself as a hanging problem since it was consistently retrying SSLExceptions over and over again.

sfc-gh-lsembera:
@sfc-gh-wfateem The SDK does not have aggressive retry logic, it retries 3 times only with exponential backoff. Just out of curiosity, why would customers commonly have SSL/TLS issues?

sfc-gh-wfateem:
@sfc-gh-lsembera The most commons reasons:
1.) They have a proxy server or firewall that intercepts SSL/TLS connections and they don't have the vendor's root certificate in the trust store.
2.) They're using some custom trust store that doesn't include all the CA root certificates that you would normally find by default in a JDK installation (bundled in cacerts).
3.) They're running the application on the cloud and they have a security policy blocking traffic.

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -146,6 +152,7 @@ private static void initHttpClient(String accountName) {
     HttpClientBuilder clientBuilder =
         HttpClientBuilder.create()
             .setConnectionManager(connectionManager)
+            .evictIdleConnections(DEFAULT_EVICT_IDLE_AFTER_SECONDS, TimeUnit.SECONDS)
```
sfc-gh-japatel, line 155: In retry logic, we should consume the response to not let any connection block other http requests. I will open the PR. but this will also help alleviate the problem.

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -270,16 +283,17 @@ private static HttpRequestRetryHandler getHttpRequestRetryHandler() {
...
         return false;
       }
-      if (exception instanceof NoHttpResponseException) {
+      if (exception instanceof NoHttpResponseException
+          || exception instanceof javax.net.ssl.SSLException) {
         LOGGER.info(
-            "Retrying request which caused No HttpResponse Exception with "
-                + "URI:{}, retryCount:{} and maxRetryCount:{}",
+            "Retrying request which caused {} with " + "URI:{}, retryCount:{} and maxRetryCount:{}",
+            exception.getClass().getName(),
```
sfc-gh-tjones, line 290: Codecov says it is not covered by tests.

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -270,16 +283,17 @@ private static HttpRequestRetryHandler getHttpRequestRetryHandler() {
...
             requestURI,
             executionCount,
             MAX_RETRIES);
         return true;
       }
-      LOGGER.info("No retry for URI:{} with exception", requestURI, exception);
+      LOGGER.info("No retry for URI:{} with exception {}", requestURI, exception.toString());
```
sfc-gh-tjones, line 296: Same Codecov comment here.

sfc-gh-tzhang (approved):
LGTM, please consider increase the default value in case it introduces any issues.

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -73,6 +73,12 @@ public class HttpUtil {
   private static final int DEFAULT_CONNECTION_TIMEOUT_MINUTES = 1;
   private static final int DEFAULT_HTTP_CLIENT_SOCKET_TIMEOUT_MINUTES = 5;
 
+  /**
+   * After how many seconds of inactivity should be idle connections evicted from the connection
+   * pool.
+   */
+  private static final int DEFAULT_EVICT_IDLE_AFTER_SECONDS = 60;
```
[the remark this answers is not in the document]
sfc-gh-wfateem: 60 seconds is reasonable. Our load balancers will close off idle connections way sooner than 300 seconds. Here are the current idle timeouts for each cloud provider:
AWS - 150 seconds
Azure - 120 seconds
GCP - 610 seconds

The idle timeout in GCP is pretty large and there are plans to change that and bring it down in the future.

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-lsembera: Mockito tests of `HttpRequestRetryHandler` added.

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-lsembera: Mockito tests of `HttpRequestRetryHandler` added.

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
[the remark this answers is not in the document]
sfc-gh-lsembera: I also think 60 seconds is reasonable, there is no need to keep idle connections around for too long.

PR #544, opened 2023-07-03
SNOW-855191 Race condition causing NPE

sfc-gh-lsembera:
There is a race condition, which leads to a NPE when one thread invalidates the channel, while another thread is inside of insertRows(). The problem is that channel invalidation clears fieldIndex, which may still be used in threads concurrently inserting rows.

There is no need to clear the fieldIndex during invalidation, it is enough to mark the channel as invalid and it won't be flushed. When channel is reopened, setupSchema() will clear fieldIndex and reinitialize the schema.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
@@ -309,7 +309,6 @@ public class ParquetRowBuffer extends AbstractRowBuffer<ParquetChunkData> {
   /** Close the row buffer by releasing its internal resources. */
   @Override
   void closeInternal() {
-    this.fieldIndex.clear();
     if (bdecParquetWriter != null) {
       try {
         bdecParquetWriter.close();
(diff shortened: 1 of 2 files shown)


sfc-gh-gdoci (approved):

sfc-gh-rcheng:
We call `markClosed()` on the channel not the buffer, so the `RowBufferTest.testClose` is failing because the buffer can still insert, it just should never be called. Can you add an additional channel IT test to confirm that the channel will stop inserting after close? (unsure if it already exists, might have missed it when scanning `StreamingIngestIt` and `SnowlakeStreamingIngestChannelTest`)

sfc-gh-tjones (approved):

sfc-gh-tzhang (approved):
LGTM, please fix the test failure

sfc-gh-lsembera:
> We call `markClosed()` on the channel not the buffer, so the `RowBufferTest.testClose` is failing because the buffer can still insert, it just should never be called. Can you add an additional channel IT test to confirm that the channel will stop inserting after close? (unsure if it already exists, might have missed it when scanning `StreamingIngestIt` and `SnowlakeStreamingIngestChannelTest`)

Good point, I removed the buffer test because [we already have a test](https://github.com/snowflakedb/snowflake-ingest-java/blob/a06ffe57470630112f29de26f1f524d347812fa4/src/test/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelTest.java#L180), which tests that `insertRows` throws an exception if the channel has already been invalidated: `SnowflakeStreamingIngestChannelTest#testChannelValid`.

Issue #547, opened 2023-07-12
Channel Invalidation Issues


joshtree41:
Hi Team, I'm trying to make sense of some channel invalidation errors in some pipelines that use the snowflake ingest SDK.

Here's are some specs associated with my team's pipeline:

1. Using java 17, openjdk 17.0.7.
2. Using version v2.0.1 of the SDK.
3. Using apache beam v 2.48.0 as the vehicle for the pipeline - though I don't think that this is relevant.

We have some long running jobs that run smoothly for a while but eventually high channel invalidation errors.

The sequence of events typically goes like this:

1.  We typically see some warning or errors with this messaging `message=invalidate with matched sequencer` **or** `Background flush task failed, client=snowBurst:03740cf8-7673-4f07-9881-b67f4e8abd87, exception=net.snowflake.ingest.utils.SFException: Register blob request failed: Read timed out., detail=Register blob request failed: Read timed out., trace=`.
2. The code picks up an invalid channel, so we attempt to close and commit with something like `ingestChannel.close().get()`.
3. This throws a `SFException` error - because the channel is invalidated.
4. We re-open a channel and continue streaming data, but we lose some of the data that wasn't committed prior to the channel invalidation.

Before I go further - I realize that is what the offset token and `SnowflakeStreamingIngestChannel::getLatestCommittedOffsetToken` method is for, but without going into too much detail ... in our case it's a bit tricky for us to re-open our stream at the offset token prior to the failure. We can do this from outside the pipeline, but it's much harder to do from within in. As a result, I'd like to better understand how to mitigate these issues to reduce the frequency at which we have to _rewind and re-launch_.

From our perspective, it's a bit confusing - it feels as though nothing has gone wrong: no errors on our end in processing or serialization, we have plenty of resources on our workers, etc. It looks as though something in the flush service is breaking unexpectedly, but we can't really tell why.

It seems as though a _flush_ method on the channel/client would help in these cases. Which is mentioned in this issue as well: https://github.com/snowflakedb/snowflake-ingest-java/issues/542.

Happy to provide more details or logs if needed!

sfc-gh-lsembera:
Hi @joshtree41, there are multiple reasons why a channel could be invalidated, are you always seeing the same error `Register blob request failed: Read timed out., detail=Register blob request failed: Read timed out.` or are the error messages different?

[...]

PR #550, opened 2023-07-17
NO-SNOW Retry failed requests more aggressively

sfc-gh-lsembera:
Currently, we have resilient retries when the cause of the error is unavailable service (e.g. 503 or 429 returned from the server side) - we retry every 3 seconds for up to 2 minutes in total. However, when an `IOException` is thrown by the http client, we only retry specific exceptions up to 3 times. This PR makes sure we all retry previously seen `IOException` up to 10x.

This PR should reduce the number of channel invalidations due to incorrect row sequencer (`status_code=21`) as reported in #547.

### DIFF
--- a/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
+++ b/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
@@ -39,6 +39,7 @@ import net.snowflake.client.jdbc.internal.apache.http.impl.conn.PoolingHttpClien
 import net.snowflake.client.jdbc.internal.apache.http.pool.PoolStats;
 import net.snowflake.client.jdbc.internal.apache.http.protocol.HttpContext;
 import net.snowflake.client.jdbc.internal.apache.http.ssl.SSLContexts;
+import net.snowflake.ingest.streaming.internal.StreamingIngestUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -296,6 +297,7 @@ public class HttpUtil {
           requestURI,
           executionCount,
           MAX_RETRIES);
+      StreamingIngestUtils.sleepForRetry(executionCount);
       return true;
     };
   }
--- a/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
+++ b/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
@@ -20,7 +20,6 @@ import net.snowflake.client.core.SFSessionProperty;
 import net.snowflake.client.jdbc.internal.apache.http.HttpHost;
 import net.snowflake.client.jdbc.internal.apache.http.HttpRequest;
 import net.snowflake.client.jdbc.internal.apache.http.HttpResponse;
-import net.snowflake.client.jdbc.internal.apache.http.NoHttpResponseException;
 import net.snowflake.client.jdbc.internal.apache.http.auth.AuthScope;
 import net.snowflake.client.jdbc.internal.apache.http.auth.Credentials;
 import net.snowflake.client.jdbc.internal.apache.http.auth.UsernamePasswordCredentials;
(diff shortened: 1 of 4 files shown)


sfc-gh-tjones (approved):
Nice - thanks for implementing this one!

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -283,18 +290,15 @@ static HttpRequestRetryHandler getHttpRequestRetryHandler() {
...
-            exception.getClass().getName(),
-            requestURI,
-            executionCount,
-            MAX_RETRIES);
-        return true;
-      }
-      LOGGER.info("No retry for URI:{} with exception {}", requestURI, exception.toString());
-      return false;
+
+      LOGGER.info(
```
sfc-gh-tzhang, line 294: Should we include the message as well?

sfc-gh-tzhang:
> This PR makes sure we retry all `IOException` 30x (which should add up to ≈2 minutes in total if 4s max backoff is used).

I'm not sure if we want to do this, would retry fix all 30x issues? IMO we should add each specific case as we hit them instead of this "catch and retrying all" approach.

Also I'm wondering why we need to retry such a long time for these connection issues to recover, in general retrying 2 minutes is bad for the file registration request, this means all blobs are holding in-memory. Probably customer would run into memory issues after two minutes of waiting.

sfc-gh-lsembera:
@sfc-gh-tzhang We already retry for up to 2 minutes when there are service unavailable errors from the server (429, 5xx, etc). I remember we discussed it some time ago and came to the conclusion that it is a reasonable to assume that the client can hold 2 minutes worth of data in memory. In case there is not enough memory, `insertRows()` would start throttling, right?

What would you say about the following:

1. We retry 15 times with max backoff of 4s
2. We only retry specific exceptions
3. We give users the option to opt-in for longer retries by configuration

//cc @sfc-gh-azagrebin

sfc-gh-azagrebin:
> We only retry specific exceptions
> We give users the option to opt-in for longer retries by configuration
opt-in only for the predefined specific exceptions or for all?

sfc-gh-lsembera:
@sfc-gh-azagrebin Good point, the opt-in option can be for all exceptions to unblock users seeing exceptions we may have missed.

sfc-gh-tzhang:
> @sfc-gh-tzhang We already retry for up to 2 minutes when there are service unavailable errors from the server (429, 5xx, etc). I remember we discussed it some time ago and came to the conclusion that it is a reasonable to assume that the client can hold 2 minutes worth of data in memory. In case there is not enough memory, `insertRows()` would start throttling, right?
> 
> What would you say about the following:
> 
> 1. We retry 15 times with max backoff of 4s
> 2. We only retry specific exceptions
> 3. We give users the option to opt-in for longer retries by configuration
> 
> //cc @sfc-gh-azagrebin

Yes, I think my point was that I don't think it will require 2 minutes (or even one minute) to recovery a SocketException, I feel like the issue must be somewhere else or can be fixed in a different way and adding more retry time just hides the actual issue. I think the eviction logic you added should help here so maybe we don't even need this fix. I'm also fine with 1 minute retry on all the known exceptions that we have seen. 

Regarding a new knob for customer, in general I'm not in a favor of adding knobs so I would suggest we don't do it. I think it's a good practice for them to implement the logic for failure recovery instead of coming back to us saying they have no way to recovery data that was ingested 1-2 minutes ago.

sfc-gh-tzhang (approved):
Left some comments, PTAL and see if it makes sense. Signed off to unblock you, I don't think there is a correct answer here so I'm in supportive of whatever decision you make. Thanks for the fix!

sfc-gh-lsembera:
@sfc-gh-tzhang Thanks! I adapted the PR to:

* Only retry exceptions that I found in our internal logs that occasionally happen:
    * `javax.net.ssl.SSLException`
    * `java.net.SocketException`
    * `java.net.UnknownHostException`
    * `javax.net.ssl.SSLHandshakeException`
* Retry 10 times with max backoff of 4 seconds
    * Overall retry time may be longer than 40 seconds, depending on how long does each execution take
* No knobs

sfc-gh-rcheng (approved):
lgtm - small nit, can you update the description? may be confusing in the future to have 30x retry when we actually implemented 10x retry. I'm not sure many people would look down into the comments instead of the description

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-lsembera: We are logging `exception.toString()`, which includes both exception class name and message

PR #551, opened 2023-07-17
SNOW-861190 Add warning for large batches in insertRows

sfc-gh-lsembera:
The SDK is currently unable to split the batch passed to insertRows if it is too large. This PR emits a warning if the batch is larger than 16MB and completely rejects the batch if its size is more than 128 MB. Both values are configurable.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
@@ -38,6 +38,9 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
   // side scanner
   private static final int INVALID_SERVER_SIDE_DATA_TYPE_ORDINAL = -1;
 
+  // The maximum recommended batch size of data passed into a single insertRows() call
+  private static final int INSERT_ROWS_RECOMMENDED_MAX_BATCH_SIZE_IN_BYTES = 16 * 1024 * 1024;
+
   // Snowflake table column logical type
   enum ColumnLogicalType {
     ANY,
@@ -159,6 +162,8 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
   // Metric callback to report size of inserted rows
   private final Consumer<Float> rowSizeMetric;
 
+  private final long maxChunkSizeInBytes;
+
   // State of the owning channel
   final ChannelRuntimeState channelState;
 
@@ -167,28 +172,23 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
 
   final ZoneId defaultTimezone;
 
-  private final long insertRowsRecommendedMaxSizeInBytes;
-  private final long insertRowsEnforcedMaxSizeInBytes;
-
   AbstractRowBuffer(
       OpenChannelRequest.OnErrorOption onErrorOption,
       ZoneId defaultTimezone,
       String fullyQualifiedChannelName,
       Consumer<Float> rowSizeMetric,
       ChannelRuntimeState channelRuntimeState,
-      long insertRowsRecommendedMaxSizeInBytes,
-      long insertRowsEnforcedMaxSizeInBytes) {
+      long maxChunkSizeInBytes) {
     this.onErrorOption = onErrorOption;
     this.defaultTimezone = defaultTimezone;
     this.rowSizeMetric = rowSizeMetric;
     this.channelState = channelRuntimeState;
     this.channelFullyQualifiedName = fullyQualifiedChannelName;
-    this.insertRowsRecommendedMaxSizeInBytes = insertRowsRecommendedMaxSizeInBytes;
-    this.insertRowsEnforcedMaxSizeInBytes = insertRowsEnforcedMaxSizeInBytes;
     this.nonNullableFieldNames = new HashSet<>();
     this.flushLock = new ReentrantLock();
     this.bufferedRowCount = 0;
     this.bufferSize = 0F;
+    this.maxChunkSizeInBytes = maxChunkSizeInBytes;
 
     // Initialize empty stats
     this.statsMap = new HashMap<>();
@@ -541,9 +541,7 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
       ChannelRuntimeState channelRuntimeState,
       boolean enableParquetMemoryOptimization,
       long maxChunkSizeInBytes,
-      long maxRowSizeInBytes,
-      long insertRowsRecommendedMaxSizeInBytes,
-      long insertRowsEnforcedMaxSizeInBytes) {
+      long maxRowSizeInBytes) {
     switch (bdecVersion) {
       case THREE:
         //noinspection unchecked
@@ -556,9 +554,7 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
                 channelRuntimeState,
                 enableParquetMemoryOptimization,
                 maxChunkSizeInBytes,
-                maxRowSizeInBytes,
-                insertRowsRecommendedMaxSizeInBytes,
-                insertRowsEnforcedMaxSizeInBytes);
+                maxRowSizeInBytes);
       default:
         throw new SFException(
             ErrorCode.INTERNAL_ERROR, "Unsupported BDEC format version: " + bdecVersion);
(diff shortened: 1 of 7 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -38,6 +38,9 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
   // side scanner
   private static final int INVALID_SERVER_SIDE_DATA_TYPE_ORDINAL = -1;
 
+  // Maximum number of rows we recommend to pass to insertRows()
+  private static final int RECOMMENDED_MAX_BATCH_SIZE = 10_000;
```
sfc-gh-tjones, line 42: Worth making a config option?

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -549,4 +554,14 @@ static <T> AbstractRowBuffer<T> createRowBuffer(
             ErrorCode.INTERNAL_ERROR, "Unsupported BDEC format version: " + bdecVersion);
     }
   }
+
+  private static void batchSizeWarning(long rowsPassed) {
+    if (rowsPassed > RECOMMENDED_MAX_BATCH_SIZE) {
```
sfc-gh-tzhang, line 559: I'm not sure if we want to warn on row count, should we use batch data size instead? Also this is case by case, a large batch would work fine if you have enough resources so I'm not sure we should warn on anything, it just adds noise. Updating our online doc should be sufficient?

sfc-gh-rcheng: +1, and can you add more about why we chose 10k rows?

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
sfc-gh-lsembera: Created an option

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
sfc-gh-lsembera: You are right, row count does not matter, only batch size does. We discussed it with @sfc-gh-azagrebin offline and this PR now does the following If the batch size passed to insertRows() is:

1. more than 128 MB => throw an exception
2. more than 16 MB => emit warning log

Both values are configurable by the user.

src/main/resources/net/snowflake/ingest/ingest_error_messages.properties
```diff
@@ -34,3 +34,4 @@
 0029=Data type not supported: {0}
 0030=The given row cannot be converted to the internal format due to invalid value: {0}
 0031=The given row exceeds the maximum allowed row size {0}
+0035=The given batch of rows passed to insertRows() exceeded the maximum allowed batch size of {0} bytes. Consider splitting the batch into multiple smaller batches before passing them to insertRows(). We recommend batch size does not exceed {1} bytes.
```
sfc-gh-rcheng, line 37: nit: can we make the message more strict since we are throwing an error on this?
"We recommend batch size does not exceed {1} bytes." -> "The batch size cannot exceed {1} bytes."

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -543,10 +556,32 @@ static <T> AbstractRowBuffer<T> createRowBuffer(
...
+          ErrorCode.MAX_BATCH_SIZE_EXCEEDED,
+          insertRowsEnforcedMaxSizeInBytes,
+          insertRowsRecommendedMaxSizeInBytes);
+    }
+  }
+
+  private void checkBatchSizeRecommendedMaximum(float batchSizeInBytes) {
+    if (batchSizeInBytes > insertRowsRecommendedMaxSizeInBytes) {
+      logger.logWarn(
+          "Too large batch of rows passed to 'insertRows': {} bytes. The recommended max batch size"
```
sfc-gh-rcheng, line 580: nit: since this is just a recommended error warning, can we make this message less strict? 
"Too large batch of rows passed to 'insertRows': {} bytes. The recommended max batch is {} bytes" -> "The batch of rows passed to 'insertRows' is over the recommended max batch size. Given {} bytes, recommended max batch is {} bytes."

sfc-gh-rcheng (approved):
lgtm, though I think the error/warn messages should add more emphasis on the difference between 'recommended batch size' and 'maximum allowed batch size' 

Customers may be confused why we recommend 16mb, but then have 128 for the max batch size.

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
sfc-gh-tzhang: This is fine, but I don't think we should keep adding configurable constants (at least for the warning one)

sfc-gh-tzhang (approved):
LGTM, left a few minor comments

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -47,6 +53,11 @@ public class ParameterProvider {
   public static final long MAX_CHUNK_SIZE_IN_BYTES_DEFAULT = 128000000L;
   public static final long MAX_ALLOWED_ROW_SIZE_IN_BYTES_DEFAULT = 64 * 1024 * 1024; // 64 MB
 
+  public static final long INSERT_ROWS_BATCH_SIZE_RECOMMENDED_MAX_SIZE_IN_BYTES_DEFAULT =
+      16 * 1024 * 1024;
+  public static final long INSERT_ROWS_BATCH_SIZE_ENFORCED_MAX_SIZE_IN_BYTES_DEFAULT =
```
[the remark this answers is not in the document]
sfc-gh-tzhang: And I'm not sure if 128MB is too small, looking at the logs you shared with me, the size before compression is 2GB and size after is 185MB, so if we limit on 128MB uncompressed, then the max size is only 11MB which is even smaller than the 16MB file size. Do you think if adding the warning with a bigger value should be sufficient enough?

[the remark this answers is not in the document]
sfc-gh-lsembera: These aren't hard limits on chunk size, it just tells how much data can be passed into a single execution of `insertRows()`. If customers get these warnings or errors, they just need to split the batch into multiple ones. In the ideal case, customers send <16MB (=default chunk size) in one batch and let the SDK decide when to flush. 
I am going to reuse the max chunk for the enforced limit.

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
sfc-gh-lsembera: Good point, I reworded it like this, what do you think?

Warning:
> The batch of rows passed to 'insertRows' is over the recommended max batch size. Given 8.3886112E7 bytes, recommended max batch is 16777216 bytes. For optimal performance and memory utilization, we recommend splitting large batches into multiple smaller ones and call insertRows for each smaller batch separately.

Error:
> Too large batch of rows passed to insertRows, the batch size cannot exceed 128,000,000 bytes, recommended batch size for optimal performance and memory utilization is 16,777,216 bytes. We recommend splitting large batches into multiple smaller ones and call insertRows for each smaller batch separately.

src/main/resources/net/snowflake/ingest/ingest_error_messages.properties
sfc-gh-lsembera: See my comment below.

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
sfc-gh-rcheng: Would it be possible to also include the given batch size in the Error message?

Too large batch of rows passed to insertRows, **given xxx bytes.** The batch size cannot exceed 128,000,000 bytes...

Otherwise looks good!

sfc-gh-lsembera: We don't know the full batch size because we stop processing once we exceed the limit.

PR #552, opened 2023-07-17
SNOW-870386 Add default scheme parameters for test utils when profile is not specified

sfc-gh-alhuang:
OAuth Test fails when profile path is unspecified. Adding a default scheme parameter to resolve this issue.
See [JIRA ticket](https://snowflakecomputing.atlassian.net/browse/SNOW-870386) for more info.


sfc-gh-tzhang (approved):
lgtm, thanks!

PR #559, opened 2023-07-21
reduce number of rows in BigFilesIT

sfc-gh-gdoci:
This PR reduces the number of rows in `StreamingIngestBigFilesIT` from 20k to 10k per channel. This will stop server-side alerts on oversized chunks.

### DIFF
--- a/src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestBigFilesIT.java
+++ b/src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestBigFilesIT.java
@@ -69,7 +69,7 @@ public class StreamingIngestBigFilesIT {
 
     int numTables = 2;
     int numChannels = 4; // channels are assigned round-robin to tables.
-    int batchSize = 20000;
+    int batchSize = 10000;
     int numBatches = 10; // number of rows PER CHANNEL is batchSize * numBatches
     boolean isNullable = false;
 


sfc-gh-azagrebin (approved):
Thanks Gloria!

sfc-gh-tzhang:
> This PR reduces the number of rows in `StreamingIngestBigFilesIT` from 20k to 10k per channel. This will stop server-side alerts on oversized chunks.

I assume this starts to happen recently? Do we have any idea on why? If no, please create a JIRA and we need to follow-up on the root cause, thanks!

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestBigFilesIT.java
```diff
@@ -69,7 +69,7 @@ public void testManyRowsMultipleChannelsToMultipleTable()
 
     int numTables = 2;
     int numChannels = 4; // channels are assigned round-robin to tables.
-    int batchSize = 20000;
+    int batchSize = 10000;
```
sfc-gh-tzhang, line 72: Reducing the size seems to go against the original purpose of this test, should we just disable the test while we're investigating?

sfc-gh-rcheng: ^ We don't have a limit on num rows coming in per channel, so the alerts probably expose a bug on ingest sdk where we are creating too big chunks that needs to be fixed, rather than changing this test.

sfc-gh-gdoci:
> I assume this starts to happen recently? Do we have any idea on why?

@sfc-gh-tzhang Yes. The oversized chunk metric was whitelisted on Wednesday. Even though it was introduced a long time ago, it needs to be whitelisted before alerts can fire.

The test ingests random data, so the compression ratio is bad, and we generate BDECs around 70-74MB.


 > We don't have a limit on num rows coming in per channel, so the alerts probably expose a bug on ingest sdk where we are creating too big chunks that needs to be fixed, rather than changing this test.
 
@sfc-gh-rcheng yes, @sfc-gh-lsembera pointed out that there is a fix not merged yet, coming soon.

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestBigFilesIT.java
sfc-gh-lsembera: Instead of reducing the number of rows, we could make the payload values non-random, which would improve the compression ration and hence not produce this alert.

sfc-gh-gdoci: Discussed offline that is is better to have a test with random values. We want to still produce large BDEC files (as the name of the test suggests), but a bit smaller than the server-side limit on compressed size. I can decrease the batchSize from 20k to 15k (instead of 10k).

sfc-gh-lsembera (approved):

PR #564, opened 2023-07-31
Update AWS test profile

sfc-gh-alhuang:
The old AWS test profile does not include role, while the Azure and GCP does. Add a role property to make it consistent.


sfc-gh-japatel (approved):
Awesome, thanks for adding this Alec!

sfc-gh-tzhang (approved):
lgtm, thanks!

sfc-gh-rcheng (approved):

PR #566, opened 2023-08-09
Asen remove jdbc dependency

sfc-gh-asen:


sfc-gh-xhuang:
would we no longer need jdbc to start the snowpipe streaming connection?

Issue #567, opened 2023-08-10
`registerBlobs` does not limit the size of the blobs request


jdcaperon:
Unfortunately, it looks like when a blob register request fails and the size of the blob request increases dramatically due to the backoff, the client is not pruning the request to fit within whatever size limit that is being specified by the backend.

```
  Aug 10, 2023 @ 14:38:54.678	[SF_INGEST] Start registering blobs in client=snowflake_ingest, totalBlobListSize=94, currentBlobListSize=94, idx=94
	Aug 10, 2023 @ 14:38:55.048	Retrying request to {tls}->http://******************************->https://***************************************
	Aug 10, 2023 @ 14:38:55.048	I/O exception (java.net.SocketException) caught when processing request to {tls}->http://***********************->https://*****************************: Connection reset by peer
	Aug 10, 2023 @ 14:38:55.048	Retrying request which caused java.net.SocketException with URI:/v1/streaming/channels/write/blobs/, retryCount:1 and maxRetryCount:10
	Aug 10, 2023 @ 14:38:55.199	Exceptional Status Code from STREAMING_REGISTER_BLOB: 413, requestId:
	Aug 10, 2023 @ 14:38:55.200	[SF_INGEST] Background flush task failed, client=snowflake_ingest, exception=net.snowflake.ingest.utils.SFException: Register blob request failed: HTTP Status: 413 ErrorBody: <html>
<head><title>413 Request Entity Too Large</title></head>
<body>
<center><h1>413 Request Entity Too Large</h1></center>
</body>
</html>
., detail=Register blob request failed: HTTP Status: 413 ErrorBody: <html>
<head><title>413 Request Entity Too Large</title></head>
<body>
<center><h1>413 Request Entity Too Large</h1></center>
</body>
</html>
., trace=
net.snowflake.ingest.streaming.internal.SnowflakeStreamingIngestClientInternal.registerBlobs(SnowflakeStreamingIngestClientInternal.java:462)
net.snowflake.ingest.streaming.internal.SnowflakeStreamingIngestClientInternal.registerBlobs(SnowflakeStreamingIngestClientInternal.java:417)
net.snowflake.ingest.streaming.internal.RegisterService.registerBlobs(RegisterService.java:199)
net.snowflake.ingest.streaming.internal.FlushService.lambda$registerFuture$2(FlushService.java:244)
java.base/java.util.concurrent.CompletableFuture$AsyncRun.run(CompletableFuture.java:1804)
[...]

PR #569, opened 2023-08-11
Blob number limit for reg req to avoid oversized registrations

sfc-gh-azagrebin:
Currently, if there is a connection outage, client can accumulate many blobs to register.
The blobs are serialised into a json to be sent to Snowflake.
The size of json can exceed limit of HTTP request.
To avoid this, the PR suggests to split blobs into batches of a configurable size.

See #567

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/RegisterService.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/RegisterService.java
@@ -8,6 +8,7 @@ import static net.snowflake.ingest.utils.Constants.BLOB_UPLOAD_TIMEOUT_IN_SEC;
 import static net.snowflake.ingest.utils.Utils.getStackTrace;
 
 import com.codahale.metrics.Timer;
+import com.google.common.collect.Lists;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -195,8 +196,9 @@ class RegisterService<T> {
             Timer.Context registerContext =
                 Utils.createTimerContext(this.owningClient.registerLatency);
 
+            int blobsPerReq = getBlobNumToRegisterInOneReq(blobs.size());
             // Register the blobs, and invalidate any channels that return a failure status code
-            this.owningClient.registerBlobs(blobs);
+            Lists.partition(blobs, blobsPerReq).forEach(owningClient::registerBlobs);
 
             if (registerContext != null) {
               registerContext.stop();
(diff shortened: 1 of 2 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/RegisterService.java
```diff
@@ -195,8 +196,9 @@ List<FlushService.BlobData<T>> registerBlobs(Map<String, Timer.Context> latencyT
             Timer.Context registerContext =
                 Utils.createTimerContext(this.owningClient.registerLatency);
 
+            int blobsPerReq = getBlobNumToRegisterInOneReq(blobs.size());
             // Register the blobs, and invalidate any channels that return a failure status code
-            this.owningClient.registerBlobs(blobs);
+            Lists.partition(blobs, blobsPerReq).forEach(owningClient::registerBlobs);
```
jdcaperon, line 201: Since this is a IO networking request, would it make sense to parallelize?

sfc-gh-azagrebin: I would be generally in favour if it was a common case.
Here it should be more like exception indicating that system already overloaded (added a log),
Hence I tend to not complicate the system with even more concurrency for this.
Happy to discuss more

jdcaperon: Makes sense :)

sfc-gh-azagrebin: > I would be generally in favour if it was a common case.

Actually, I take this back, we cannot do this because of ordering guarantees per channel.

jdcaperon: @sfc-gh-azagrebin then you will need to use `forEachOrdered` as `forEach` has no ordering guarantees.

src/main/java/net/snowflake/ingest/streaming/internal/RegisterService.java
```diff
@@ -216,6 +218,21 @@ List<FlushService.BlobData<T>> registerBlobs(Map<String, Timer.Context> latencyT
...
   }
 
+  /**
+   * We split blobs into batches to avoid too big payload in registration requests in case of
+   * connection hiccups if too many blobs are accumulated.
+   */
+  private int getBlobNumToRegisterInOneReq(int blobNum) {
+    int blobsPerReq = owningClient.getParameterProvider().getMaxBlobsToRegisterInOneRequest();
+    if (blobNum > blobsPerReq) {
+      logger.logWarn(
```
sfc-gh-tzhang, line 228: Could we add client name in the log as well?

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -51,6 +54,8 @@ public class ParameterProvider {
   It reduces memory consumption compared to using Java Objects for buffering.*/
   public static final boolean ENABLE_PARQUET_INTERNAL_BUFFERING_DEFAULT = false;
 
+  public static final int MAX_BLOBS_TO_REGISTER_IN_ONE_REQUEST_DEFAULT = 10;
```
sfc-gh-tzhang, line 57: 10 seems to low, should we start with something higher like 30? If we have the log in Snowhouse for tacking number of blobs in one request, we could pick the P99 as well.

sfc-gh-tzhang (approved):
Left some comments, please add tests, otherwise LGTM!

sfc-gh-lsembera:
Can we close this PR? In https://github.com/snowflakedb/snowflake-ingest-java/pull/580 I implemented limiting the number of chunks in a blob and in a registration request.

sfc-gh-azagrebin:
Sure

Issue #555, opened 2023-07-19
Retrieve channel status for multiple channels


jdcaperon:
Hello,

Would it be possible to create an interface to retrieve the channel status of many channels? Currently, for each channel I have open I have to make a separate request to SF for each channel status. Ideally I would make a single request that returns the channels status for all the channels I care about to avoid the overhead of making lots of requests.

https://github.com/snowflakedb/snowflake-ingest-java/blob/6e4eb63e56874f3bcea516894de7e0fb99392a58/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java#L382-L401

Cheers

sfc-gh-xhuang:
Hi @jdcaperon we do have a SQL command `[SHOW CHANNELS](https://docs.snowflake.com/en/sql-reference/sql/show-channels)` that currently shows the channel offset. 

Would adding a status to that command be sufficient?

jdcaperon:
> Would adding a status to that command be sufficient?

Hmmmm, ideally I wouldn't have to open a JDBC client as well, but if thats the supported path I can do that however nasty it feels. Would it be possible to extend the Client interface to have support? AFAICT the machinery already exists to query the status of multiple channels from the client code since the client is querying a list of channels on shutdown.

```java
public interface SnowflakeStreamingIngestClient extends AutoCloseable {
...
  ChannelStatus getStatusForChannels(Collection<SnowflakeStreamingIngestChannel> channels);
...
}
```

sfc-gh-tzhang:
Yes, we could take this as a small feature request, it should be fairly easy given that we already support it in the REST API

To be clear, we don't even expose the channel status for one channel, we just expose the offset token, basically what you want is a function under `SnowflakeStreamingIngestClient` to return a list of offset tokens?

jdcaperon:
> To be clear, we don't even expose the channel status for one channel, we just expose the offset token, basically what you want is a function under SnowflakeStreamingIngestClient to return a list of offset tokens?

Yeah exactly. I was thinking it might be nice to have access to things like whether the channel `isValid` as well, but committed offset tokens is the hard requirement I need right now.

sfc-gh-tzhang:
This is done, closing

PR #574, opened 2023-08-16
Add method to client to retrieve all offset tokens

jdcaperon:
Implements a solution for #555 

This is the best I could come up with to retrieve channel status for all channels with the limitations of the current interfaces. There are a number of challenges related to type erasure when working in the current code as some of the functions are typed with `<T>` and others are unknown `<?>`. Not sure what the best way to deal with this was so have made the `ChannelCache` return `<?>` to improve re-use. However, I suspect that refactoring the code to remove `<?>` might be a better long term solution. Please advise :)

Once y'all are happy with the impl I will add tests.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
@@ -96,10 +96,6 @@ class ChannelCache<T> {
     }
   }
 
-  List<SnowflakeStreamingIngestChannelInternal<?>> getChannels() {
-    return cache.values().stream().flatMap(v -> v.values().stream()).collect(Collectors.toList());
-  }
-
   /** Get the number of key-value pairs in the cache */
   int getSize() {
     return cache.size();
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
@@ -5,10 +5,8 @@
 package net.snowflake.ingest.streaming.internal;
 
 import java.util.Iterator;
-import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.stream.Collectors;
 
 /**
  * In-memory cache that stores the active channels for a given Streaming Ingest client, and the
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
@@ -5,8 +5,10 @@
 package net.snowflake.ingest.streaming.internal;
 
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
 
 /**
  * In-memory cache that stores the active channels for a given Streaming Ingest client, and the
@@ -92,6 +94,10 @@ class ChannelCache<T> {
         channel.invalidate("invalidate with matched sequencer");
       }
     }
+  }
+
+  List<SnowflakeStreamingIngestChannelInternal<?>> getChannels() {
+    return cache.values().stream().flatMap(v -> v.values().stream()).collect(Collectors.toList());
   }
 
   /** Get the number of key-value pairs in the cache */
--- a/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
@@ -268,7 +268,7 @@ public class SnowflakeStreamingIngestClientInternal<T> implements SnowflakeStrea
    * @return map of channel to the latest persisted offset token
    */
   @Override
-  public Map<SnowflakeStreamingIngestChannel, String> getALlLatestCommittedOffsetTokens() {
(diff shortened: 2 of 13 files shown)


src/test/java/net/snowflake/ingest/streaming/internal/FlushServiceTest.java
```diff
@@ -928,8 +928,12 @@ public void testShutDown() throws Exception {
...
   @Test
   public void testEncryptionDecryption()
-      throws InvalidAlgorithmParameterException, NoSuchPaddingException, IllegalBlockSizeException,
-          NoSuchAlgorithmException, BadPaddingException, InvalidKeyException {
+      throws InvalidAlgorithmParameterException,
+          NoSuchPaddingException,
+          IllegalBlockSizeException,
+          NoSuchAlgorithmException,
+          BadPaddingException,
+          InvalidKeyException {
```
jdcaperon, line 936: My GJF dislikes this but can revert 👍

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
```diff
@@ -94,6 +96,10 @@ void invalidateChannelIfSequencersMatch(
     }
   }
 
+  List<SnowflakeStreamingIngestChannelInternal<?>> getChannels() {
+    return cache.values().stream().flatMap(v -> v.values().stream()).collect(Collectors.toList());
```
sfc-gh-tzhang, line 100: We might have stall channels (channels that are not active for a certain time and will be auto cleaned up server side) or invalidated channels, so I'm not sure if it's a good idea to do this for all channels in cache, I think what we can do is to pass a list of channel objects into getAllLatestCommittedOffsetTokens?

jdcaperon: > I think what we can do is to pass a list of channel objects into getAllLatestCommittedOffsetTokens?

Are you happy to unsafe cast from `SnowflakeStreamingIngestChannel` to `SnowflakeStreamingIngestChannelInternal`?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -260,6 +261,27 @@ public String getName() {
...
+   * @return map of channel to the latest persisted offset token
+   */
+  @Override
+  public Map<SnowflakeStreamingIngestChannel, String> getLatestCommittedOffsetTokens(
+      List<SnowflakeStreamingIngestChannel> channels) {
+    List<SnowflakeStreamingIngestChannelInternal<?>> internalChannels =
+        channels.stream()
+            .map(c -> (SnowflakeStreamingIngestChannelInternal<?>) c)
+            .collect(Collectors.toList());
+    List<ChannelsStatusResponse.ChannelStatusResponseDTO> channelsStatus =
```
sfc-gh-tzhang, line 276: Should we check the statusCode first?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -260,6 +261,27 @@ public String getName() {
...
+        getChannelsStatus(internalChannels).getChannels();
+    Map<SnowflakeStreamingIngestChannel, String> result = new HashMap<>();
+    for (int idx = 0; idx < channels.size(); idx++) {
+      result.put(channels.get(idx), channelsStatus.get(idx).getPersistedOffsetToken());
+    }
+    return result;
```
sfc-gh-tzhang, line 282: In your use case, a Map makes more sense then just a list of offset tokens?

sfc-gh-tzhang (commented):
Could we add tests for this change?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
jdcaperon: Status Code is checked in `getChannelStatus` https://github.com/jdcaperon/snowflake-ingest-java/blob/069269994c2a8b59aa8b94f954bd0848e3c8ba99/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java#L417-L439

We can check it here as well, what would you like to do if the status is not valid?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
jdcaperon: Mmmm, if we guarantee ordering then a list is fine, but a map is nicer because the results don't have to be Zipped back together. Ideally we could return a list of some channel identifier to the offset token, maybe channel qualified name?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
sfc-gh-tzhang: If it's checked then it's fine

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
sfc-gh-tzhang: I'm fine with either option, but using the fully qualified channel name sounds better?

sfc-gh-tzhang (approved):
LGTM, thanks!

PR #575, opened 2023-08-16
Bump GJF to 1.17.0 to run formatter on JDK 17

jdcaperon:
View changelog here: https://github.com/google/google-java-format/releases

Anyone running latest java can't run the older versions due to breaking changes in the JDK

### DIFF
--- a/format.sh
+++ b/format.sh
@@ -4,8 +4,8 @@ set -euo pipefail
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 cd "$SCRIPT_DIR"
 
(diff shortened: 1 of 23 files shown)


sfc-gh-tzhang (approved):
I'm not sure we can do this now given that our dev environment is still using 1.8

jdcaperon:
@sfc-gh-tzhang the GJF is compile for 1.8 still, the newer versions are compatible with the latest version of the SDK :) https://github.com/google/google-java-format/blob/master/pom.xml#L88

sfc-gh-tzhang:
> @sfc-gh-tzhang the GJF is compiled for 1.8 still, the newer versions are compatible with the latest version of the SDK :) https://github.com/google/google-java-format/blob/master/pom.xml#L88
> 
> It is a safe upgrade for your dev env AFAICT

It's not my own dev env, it's the company DevVm setup

PR #579, opened 2023-08-29
SNOW-672156 support specifying compression algorithm to be used for BDEC Parquet files

sfc-gh-gdoci:
This PR adds support for specifying the compression algorithm to be used for BDEC Parquet files. The allowed value is just `GZIP` for now.

### DIFF
--- a/src/main/java/net/snowflake/ingest/utils/Constants.java
+++ b/src/main/java/net/snowflake/ingest/utils/Constants.java
@@ -138,7 +138,7 @@ public class Constants {
       }
       throw new IllegalArgumentException(
           String.format(
-              "Unsupported BDEC_PARQUET_COMPRESSION_ALGORITHM = '%d', allowed values are %s",
+              "Unsupported BDEC_PARQUET_COMPRESSION_ALGORITHM = '%s', allowed values are %s",
               name, Arrays.asList(BdecParquetCompression.values())));
     }
   }
--- a/src/main/java/net/snowflake/ingest/utils/Constants.java
+++ b/src/main/java/net/snowflake/ingest/utils/Constants.java
@@ -119,15 +119,10 @@ public class Constants {
    * CompressionCodecName, but we want to control and allow only specific values of that.
    */
   public enum BdecParquetCompression {
-    UNCOMPRESSED,
-    GZIP,
-    SNAPPY,
-    ZSTD;
+    GZIP;
 
     public CompressionCodecName getCompressionCodec() {
-      return (this == UNCOMPRESSED
-          ? CompressionCodecName.fromConf(null)
-          : CompressionCodecName.fromConf(this.name()));
+      return CompressionCodecName.fromConf(this.name());
     }
 
     public static BdecParquetCompression fromName(String name) {
--- a/src/main/java/net/snowflake/ingest/utils/Constants.java
+++ b/src/main/java/net/snowflake/ingest/utils/Constants.java
@@ -5,6 +5,7 @@
 package net.snowflake.ingest.utils;
 
 import java.util.Arrays;
+import org.apache.parquet.hadoop.metadata.CompressionCodecName;
 
 /** Contains all the constants needed for Streaming Ingest */
 public class Constants {
@@ -113,6 +114,34 @@ public class Constants {
     }
   }
 
+  /**
+   * Compression algorithm supported by BDEC Parquet Writer. It is a wrapper around Parquet's lib
+   * CompressionCodecName, but we want to control and allow only specific values of that.
+   */
+  public enum BdecParquetCompression {
+    UNCOMPRESSED,
+    GZIP,
+    SNAPPY,
+    ZSTD;
+
+    public CompressionCodecName getCompressionCodec() {
+      return (this == UNCOMPRESSED
+          ? CompressionCodecName.fromConf(null)
+          : CompressionCodecName.fromConf(this.name()));
+    }
+
+    public static BdecParquetCompression fromName(String name) {
+      for (BdecParquetCompression e : BdecParquetCompression.values()) {
+        if (e.name().toLowerCase().equals(name.toLowerCase())) {
+          return e;
+        }
+      }
+      throw new IllegalArgumentException(
+          String.format(
+              "Unsupported BDEC_PARQUET_COMPRESSION_ALGORITHM = '%d', allowed values are %s",
+              name, Arrays.asList(BdecParquetCompression.values())));
+    }
+  }
   // Parameters
   public static final boolean DISABLE_BACKGROUND_FLUSH = false;
   public static final boolean COMPRESS_BLOB_TWICE = false;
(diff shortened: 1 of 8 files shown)


sfc-gh-tjones (approved):

sfc-gh-tzhang:
> This PR adds support for specifying the compression algorithm to be used for BDEC Parquet files. The allowed values are `UNCOMPRESSED, GZIP, ZSTD, SNAPPY`.

I think the intention for this parameter is to allow customer to go back to GZIP if there is any issue with ZSTD, so I suggest we only support GZIP and ZSTD, WDYT?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
```diff
@@ -130,7 +130,10 @@ class SnowflakeStreamingIngestChannelInternal<T> implements SnowflakeStreamingIn
                 : ParameterProvider.MAX_CHUNK_SIZE_IN_BYTES_DEFAULT,
             owningClient != null
                 ? owningClient.getParameterProvider().getMaxAllowedRowSizeInBytes()
-                : ParameterProvider.MAX_ALLOWED_ROW_SIZE_IN_BYTES_DEFAULT);
+                : ParameterProvider.MAX_ALLOWED_ROW_SIZE_IN_BYTES_DEFAULT,
+            owningClient != null
```
sfc-gh-tzhang, line 134: We can't keep adding this...at some point we need to improve this :)

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -51,6 +54,9 @@ public class ParameterProvider {
   It reduces memory consumption compared to using Java Objects for buffering.*/
   public static final boolean ENABLE_PARQUET_INTERNAL_BUFFERING_DEFAULT = false;
 
+  public static final Constants.BdecParquetCompression BDEC_PARQUET_COMPRESSION_ALGORITHM_DEFAULT =
```
sfc-gh-tzhang, line 57: I thought we want ZSTD to be the default?

src/test/java/net/snowflake/ingest/streaming/internal/ParameterProviderTest.java
```diff
@@ -3,6 +3,7 @@
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
+import net.snowflake.ingest.utils.Constants;
```
sfc-gh-tzhang, line 6: Should we add an small IT test for GZIP?

sfc-gh-tzhang:
Please hold this PR until we're done with the current release, thanks!

sfc-gh-xhuang:
> > This PR adds support for specifying the compression algorithm to be used for BDEC Parquet files. The allowed values are `UNCOMPRESSED, GZIP, ZSTD, SNAPPY`.
> 
> I think the intention for this parameter is to allow customer to go back to GZIP if there is any issue with ZSTD, so I suggest we only support GZIP and ZSTD, WDYT?

I agree. Do we need to support anything else as customers do not have access to the underlying blob? 
We only need to allow compression algos that we want

src/main/java/net/snowflake/ingest/utils/Constants.java
```diff
@@ -113,6 +114,34 @@ public static BdecVersion fromInt(int val) {
...
+
+    public static BdecParquetCompression fromName(String name) {
+      for (BdecParquetCompression e : BdecParquetCompression.values()) {
+        if (e.name().toLowerCase().equals(name.toLowerCase())) {
+          return e;
+        }
+      }
+      throw new IllegalArgumentException(
+          String.format(
+              "Unsupported BDEC_PARQUET_COMPRESSION_ALGORITHM = '%d', allowed values are %s",
```
sfc-gh-azagrebin, line 141:
```suggestion
              "Unsupported BDEC_PARQUET_COMPRESSION_ALGORITHM = '%s', allowed values are %s",
```

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-azagrebin: was the ZSTD already tested with all our tests?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
sfc-gh-gdoci: Right, based on this comment, I created and merged https://github.com/snowflakedb/snowflake-ingest-java/pull/581.

src/main/java/net/snowflake/ingest/utils/Constants.java
sfc-gh-gdoci: Ack.

src/test/java/net/snowflake/ingest/streaming/internal/ParameterProviderTest.java
sfc-gh-gdoci: In which class do you mean?

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-gdoci: I have tested `GZIP, ZSTD, SNAPPY` previously, but I will run the tests again. The SDK has changed a lot since then.

sfc-gh-gdoci:
I narrowed down the scope of this PR to just make compression algorithm configurable. GZIP is the only allowed value as it has been the default value so far. I will have to do more correctness testing and performance evaluations before adding other allowed compression algorithms.

 @sfc-gh-tjones @sfc-gh-tzhang @sfc-gh-azagrebin PTAL.

src/main/java/net/snowflake/ingest/utils/Constants.java
```diff
@@ -113,6 +114,29 @@ public static BdecVersion fromInt(int val) {
...
+      return CompressionCodecName.fromConf(this.name());
+    }
+
+    public static BdecParquetCompression fromName(String name) {
+      for (BdecParquetCompression e : BdecParquetCompression.values()) {
+        if (e.name().toLowerCase().equals(name.toLowerCase())) {
+          return e;
+        }
+      }
+      throw new IllegalArgumentException(
```
sfc-gh-tzhang, line 134: could we add a test case for invalid input?

src/test/java/net/snowflake/ingest/streaming/internal/ParameterProviderTest.java
sfc-gh-tzhang: StreamingIngestIT

sfc-gh-tzhang (approved):

src/test/java/net/snowflake/ingest/streaming/internal/ParameterProviderTest.java
sfc-gh-gdoci: GZIP is the default in this PR.

PR #580, opened 2023-08-29
SNOW-902709 Limit the max allowed number of chunks in blob

### LINKED ISSUE #567: `registerBlobs` does not limit the size of the blobs request

### LINKED ISSUE #570: Only flush channels for tables with a channel that needs flushing rather than flushing all channels for all tables
DISCLAIMER: This is based on my interpretation of the SDK, still getting up to speed on the internals.

Currently the SDK attempts to flush all channels on an interval, instead, would it be possible and potentially more efficient to only flush channels where the rowbuffer has hit the expected size? In high volume/high channel count use cases the SDK has to flush a large number of blobs after every time period.

For example, if we have 100 channels, and a flush interval of 1 second where each channel receives at least 1 row per second. Then the SDK will attempt to flush 100 chunks every second. Even if the flush interval is increased to 60 seconds, if a single channel in our list of 100 channels needs flushing once per second because it is higher volume than remaining channels, then the SDK will still attempt to flush 100 chunks every second.

Instead, if the `FlushService` is refactored such that it knows which tables/channels need to be flushed then the number of chunks being pushed can be reduced. Taking the example above where there exist 100 channels, with a single channel needing to be flushed every second, with a force flush interval of 60 seconds. There should ideally be 1 chunks being pushed every second (assuming the rowbuffers stack to a single blob), rather than 100 chunks.

e2: Potentially there is scope to have the interval be per table so that regularly flushing tables don't need to be force flushed as part of the flush cycle which could potentially help with increasing size of the blobs being pushed, assuming that bigger blobs are better. This would also help avoiding the glut of requests/volume when flushing on the interval.

My understanding is that this should be okay, as ordering is only preserved within channels, so processing data out of order is okay across different tables.

I understand this might be a fairly big refactor of some of the SDK, would be keen to hear your thoughts or whether I am off track with my understanding of the code. I'm hoping this helps with some of the issues I've seen with high blob counts causing networking errors.

e: I'm also happy to provide an implementation/PR for this with guidance if this isn't going to be prioritized.

e3: Blob -> Chunk

sfc-gh-lsembera:
When one client is ingesting into many tables, we are seeing occasional timeouts from blob registrations calls. This PR introduces the limit of 20 chunks in one blob and in blob registration request.

Fixes #570
Fixes #567

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
@@ -350,6 +350,16 @@ class FlushService<T> {
         if (!leftoverChannelsDataPerTable.isEmpty()) {
           channelsDataPerTable.addAll(leftoverChannelsDataPerTable);
           leftoverChannelsDataPerTable.clear();
+        } else if (blobData.size()
+            >= this.owningClient.getParameterProvider().getMaxChunksInBlob()) {
+          // Create a new blob if the current one already contains max allowed number of chunks
+          logger.logInfo(
+              "Max allowed number of chunks in the current blob reached. chunkCount={}"
+                  + " maxChunkCount={} currentBlobPath={}",
+              blobData.size(),
+              this.owningClient.getParameterProvider().getMaxChunksInBlob(),
+              blobPath);
+          break;
         } else {
           ConcurrentHashMap<String, SnowflakeStreamingIngestChannelInternal<T>> table =
               itr.next().getValue();
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
@@ -351,13 +351,15 @@ class FlushService<T> {
           channelsDataPerTable.addAll(leftoverChannelsDataPerTable);
           leftoverChannelsDataPerTable.clear();
         } else if (blobData.size()
-            >= this.owningClient.getParameterProvider().getMaxChunksInBlob()) {
+            >= this.owningClient
+                .getParameterProvider()
+                .getMaxChunksInBlobAndRegistrationRequest()) {
           // Create a new blob if the current one already contains max allowed number of chunks
           logger.logInfo(
               "Max allowed number of chunks in the current blob reached. chunkCount={}"
                   + " maxChunkCount={} currentBlobPath={}",
               blobData.size(),
-              this.owningClient.getParameterProvider().getMaxChunksInBlob(),
+              this.owningClient.getParameterProvider().getMaxChunksInBlobAndRegistrationRequest(),
               blobPath);
           break;
         } else {
--- a/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
@@ -465,10 +465,21 @@ public class SnowflakeStreamingIngestClientInternal<T> implements SnowflakeStrea
     List<List<BlobMetadata>> result = new ArrayList<>();
     List<BlobMetadata> currentBatch = new ArrayList<>();
     int chunksInCurrentBatch = 0;
+    int maxChunksInBlobAndRegistrationRequest =
+        parameterProvider.getMaxChunksInBlobAndRegistrationRequest();
 
     for (BlobMetadata blob : blobs) {
-      if (chunksInCurrentBatch + blob.getChunks().size()
-          > parameterProvider.getMaxChunksInRegistrationRequest()) {
+      if (blob.getChunks().size() > maxChunksInBlobAndRegistrationRequest) {
+        throw new SFException(
+            ErrorCode.INTERNAL_ERROR,
+            String.format(
+                "Incorrectly generated blob detected - number of chunks in the blob is larger than"
+                    + " the max allowed number of chunks. Please report this bug to Snowflake."
+                    + " bdec=%s chunkCount=%d maxAllowedChunkCount=%d",
+                blob.getPath(), blob.getChunks().size(), maxChunksInBlobAndRegistrationRequest));
+      }
+
+      if (chunksInCurrentBatch + blob.getChunks().size() > maxChunksInBlobAndRegistrationRequest) {
         // Newly added BDEC file would exceed the max number of chunks in a single registration
         // request. We put chunks collected so far into the result list and create a new batch with
         // the current blob
--- a/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
@@ -452,7 +452,42 @@ public class SnowflakeStreamingIngestClientInternal<T> implements SnowflakeStrea
    * @param blobs list of uploaded blobs
    */
   void registerBlobs(List<BlobMetadata> blobs) {
-    this.registerBlobs(blobs, 0);
+    for (List<BlobMetadata> blobBatch : partitionBlobListForRegistrationRequest(blobs)) {
+      this.registerBlobs(blobBatch, 0);
+    }
+  }
+
+  /**
+   * Partition the collection of blobs into sub-lists, so that the total number of chunks in each
+   * sublist does not exceed the max allowed number of chunks in one registration request.
+   */
+  List<List<BlobMetadata>> partitionBlobListForRegistrationRequest(List<BlobMetadata> blobs) {
+    List<List<BlobMetadata>> result = new ArrayList<>();
+    List<BlobMetadata> currentBatch = new ArrayList<>();
+    int chunksInCurrentBatch = 0;
+
+    for (BlobMetadata blob : blobs) {
+      if (chunksInCurrentBatch + blob.getChunks().size()
+          > parameterProvider.getMaxChunksInRegistrationRequest()) {
+        // Newly added BDEC file would exceed the max number of chunks in a single registration
+        // request. We put chunks collected so far into the result list and create a new batch with
+        // the current blob
+        result.add(currentBatch);
+        currentBatch = new ArrayList<>();
+        currentBatch.add(blob);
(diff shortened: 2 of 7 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -452,7 +452,42 @@ ChannelsStatusResponse getChannelsStatus(
...
+    List<BlobMetadata> currentBatch = new ArrayList<>();
+    int chunksInCurrentBatch = 0;
+
+    for (BlobMetadata blob : blobs) {
+      if (chunksInCurrentBatch + blob.getChunks().size()
+          > parameterProvider.getMaxChunksInRegistrationRequest()) {
+        // Newly added BDEC file would exceed the max number of chunks in a single registration
+        // request. We put chunks collected so far into the result list and create a new batch with
+        // the current blob
+        result.add(currentBatch);
```
sfc-gh-asen, line 475: nit: This probably can't happen as long as `maxChunksInBlob` parameter is less than `getMaxChunksInRegistrationRequest` parameter but if it is not we will add a `currentBatch` that is empty to the `result` list.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -351,6 +351,16 @@ void distributeFlushTasks() {
...
+        } else if (blobData.size()
+            >= this.owningClient.getParameterProvider().getMaxChunksInBlob()) {
+          // Create a new blob if the current one already contains max allowed number of chunks
+          logger.logInfo(
+              "Max allowed number of chunks in the current blob reached. chunkCount={}"
+                  + " maxChunkCount={} currentBlobPath={}",
+              blobData.size(),
+              this.owningClient.getParameterProvider().getMaxChunksInBlob(),
+              blobPath);
+          break;
```
sfc-gh-asen, line 363: Should this check be in the shouldStopProcessing() method? 

if I am not wrong if the leftoverChannelsDataPerTable is not empty, then we will ignore this check

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -351,6 +351,16 @@ void distributeFlushTasks() {
         if (!leftoverChannelsDataPerTable.isEmpty()) {
           channelsDataPerTable.addAll(leftoverChannelsDataPerTable);
           leftoverChannelsDataPerTable.clear();
+        } else if (blobData.size()
```
sfc-gh-tzhang, line 354: Could the logic here combined with https://github.com/snowflakedb/snowflake-ingest-java/blob/3a3cbc8798e2e4eddab1cc672b9d48d557e48d78/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java#L407? Both are checking whether a separate blob is needed

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -46,6 +49,8 @@ public class ParameterProvider {
   public static final long MAX_CHANNEL_SIZE_IN_BYTES_DEFAULT = 32000000L;
   public static final long MAX_CHUNK_SIZE_IN_BYTES_DEFAULT = 128000000L;
   public static final long MAX_ALLOWED_ROW_SIZE_IN_BYTES_DEFAULT = 64 * 1024 * 1024; // 64 MB
+  public static final int MAX_CHUNKS_IN_BLOB_DEFAULT = 20;
+  public static final int MAX_CHUNKS_IN_REGISTRATION_REQUEST_DEFAULT = 100;
```
sfc-gh-tzhang, line 53: What is the real issue here? Is it because of the total number of chunks in one blob or the total number of the chunks in one request? If it's former, then why we need to limit the total number of chunks in one request? If it's latter, then these two values should be the same?

sfc-gh-tzhang (commented):
Left some comment, PTAL! No need to block on my approval :)

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-tzhang: I think shouldStopProcessing is checking on the channel level but this is on the chunks level, there is no correctness issue if leftoverChannelsDataPerTable is not empty, right?

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-lsembera: Fundamentally, the issue issue is the number of chunks in one request, because it is where the server-side latency and potential timeouts come from. They could be the same, the reasoning behind one being smaller than the other was that it would give the SDK an oportunity to put BDECs with fewer chunks into the same registration request. For example, if the limit of both is 100 and there is another bdec with just one chunk, it would have to go into its own registration request.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-lsembera: It could, but it could be triggered in the middle of a chunk, which is makes sense for splitting based on chunk/blob size, but not for the newly added check. The only added check only cares about the number of chunks in a blob, so I put the potential break just before it starts a new chunk.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-lsembera: Additional test added to `FlushServiceTest`

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
sfc-gh-lsembera: Good point, I added validation that `maxChunksInBlob` is always less than `getMaxChunksInRegistrationRequest`

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -30,6 +30,9 @@ public class ParameterProvider {
   public static final String MAX_CHUNK_SIZE_IN_BYTES = "MAX_CHUNK_SIZE_IN_BYTES".toLowerCase();
   public static final String MAX_ALLOWED_ROW_SIZE_IN_BYTES =
       "MAX_ALLOWED_ROW_SIZE_IN_BYTES".toLowerCase();
+  public static final String MAX_CHUNKS_IN_BLOB = "MAX_CHUNKS_IN_BDEC".toLowerCase();
```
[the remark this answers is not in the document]
sfc-gh-lsembera: Unit tests added

sfc-gh-lsembera:
@sfc-gh-tzhang I addressed your PR comments, could you re-review, please?

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-tzhang: I see, given that this situation should be rare, I'm not sure if we need to be smart here. I'm more in a favor of adding less configurable parameters with random default values, WDYT?

sfc-gh-tzhang (approved):
Left a suggestion, PTAL and feel free to merge if you disagree, thanks!

PR #586, opened 2023-09-13
SNOW-914666 Adds MAX_CLIENT_LAG configuration option

sfc-gh-tjones:
We want to expose a knob that gives users the ability to control when data is ingested. This has a material difference on the size of blobs generated and can result in fewer smaller-sized blobs which in turn affects query performance. The trade-off is higher ingest latencies. We have decided to expose this in the form of an optional `MAX_CLIENT_LAG` option that accepts inputs as the following:

- `number second` (ex: `1 second`)
- `number seconds` (ex: `2 seconds`)
- `number minute` (ex: `1 minute`)
- `number minutes` (ex: `2 minutes`)

By default we use 1 second as the maximum client lag which is the current behavior of the SDK.

Note that this dictates when a flush is triggered to cloud storage. Depending on your connection to cloud storage and cloud storage tail latencies a blob persist may take longer than expected. Therefore, it is helpful to think of this parameter as a target, rather than an absolute number.

@test Adds tests to `ParameterProviderTest`

### DIFF
--- a/src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
+++ b/src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
@@ -3,6 +3,7 @@ package net.snowflake.ingest.utils;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
+import java.util.concurrent.TimeUnit;
 
 /** Utility class to provide configurable constants */
 public class ParameterProvider {
@@ -54,10 +55,9 @@ public class ParameterProvider {
   public static final String MAX_CLIENT_LAG_DEFAULT = "1 second";
   public static final boolean MAX_CLIENT_LAG_ENABLED_DEFAULT = true;
 
-  static final long MAX_CLIENT_LAG_MS_MIN = 1000;
+  static final long MAX_CLIENT_LAG_MS_MIN = TimeUnit.SECONDS.toMillis(1);
 
-  // 10 minutes
-  static final long MAX_CLIENT_LAG_MS_MAX = 600000;
+  static final long MAX_CLIENT_LAG_MS_MAX = TimeUnit.MINUTES.toMillis(10);
   public static final long MAX_ALLOWED_ROW_SIZE_IN_BYTES_DEFAULT = 64 * 1024 * 1024; // 64 MB
 
   /* Parameter that enables using internal Parquet buffers for buffering of rows before serializing.
@@ -211,15 +211,15 @@ public class ParameterProvider {
       throw new IllegalArgumentException(
           String.format("Failed to parse MAX_CLIENT_LAG = '%s'", lagParts[0]), t);
     }
-    long computedLag = BUFFER_FLUSH_INTERVAL_IN_MILLIS_DEFAULT;
+    long computedLag;
     switch (lagParts[1].toLowerCase()) {
       case "second":
       case "seconds":
-        computedLag = lag * 1000;
+        computedLag = lag * TimeUnit.SECONDS.toMillis(1);
         break;
       case "minute":
       case "minutes":
-        computedLag = lag * 60000;
+        computedLag = lag * TimeUnit.SECONDS.toMillis(60);
         break;
       default:
         throw new IllegalArgumentException(
--- a/src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
+++ b/src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
@@ -31,6 +31,10 @@ public class ParameterProvider {
   public static final String MAX_ALLOWED_ROW_SIZE_IN_BYTES =
       "MAX_ALLOWED_ROW_SIZE_IN_BYTES".toLowerCase();
 
+  public static final String MAX_CLIENT_LAG = "MAX_CLIENT_LAG".toLowerCase();
+
+  public static final String MAX_CLIENT_LAG_ENABLED = "MAX_CLIENT_LAG_ENABLED".toLowerCase();
+
   // Default values
   public static final long BUFFER_FLUSH_INTERVAL_IN_MILLIS_DEFAULT = 1000;
   public static final long BUFFER_FLUSH_CHECK_INTERVAL_IN_MILLIS_DEFAULT = 100;
@@ -45,6 +49,10 @@ public class ParameterProvider {
   public static final long MAX_MEMORY_LIMIT_IN_BYTES_DEFAULT = -1L;
   public static final long MAX_CHANNEL_SIZE_IN_BYTES_DEFAULT = 32000000L;
   public static final long MAX_CHUNK_SIZE_IN_BYTES_DEFAULT = 128000000L;
+
+  // Lag related parameters
+  public static final String MAX_CLIENT_LAG_DEFAULT = "1 second";
+  public static final boolean MAX_CLIENT_LAG_ENABLED_DEFAULT = true;
   public static final long MAX_ALLOWED_ROW_SIZE_IN_BYTES_DEFAULT = 64 * 1024 * 1024; // 64 MB
 
   /* Parameter that enables using internal Parquet buffers for buffering of rows before serializing.
@@ -53,6 +61,9 @@ public class ParameterProvider {
 
   /** Map of parameter name to parameter value. This will be set by client/configure API Call. */
   private final Map<String, Object> parameterMap = new HashMap<>();
+
+  // Cached buffer flush interval - avoid parsing each time for quick lookup
+  private Long cachedBufferFlushIntervalMs = -1L;
 
   /**
    * Constructor. Takes properties from profile file and properties from client constructor and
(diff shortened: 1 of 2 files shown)


src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -45,6 +49,10 @@ public class ParameterProvider {
   public static final long MAX_MEMORY_LIMIT_IN_BYTES_DEFAULT = -1L;
   public static final long MAX_CHANNEL_SIZE_IN_BYTES_DEFAULT = 32000000L;
   public static final long MAX_CHUNK_SIZE_IN_BYTES_DEFAULT = 128000000L;
+
+  // Lag related parameters
+  public static final String MAX_CLIENT_LAG_DEFAULT = "1 second";
+  public static final boolean MAX_CLIENT_LAG_ENABLED_DEFAULT = true;
```
sfc-gh-xhuang, line 55: why do we need this enabled parameter too?

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -149,20 +160,74 @@ private void setParameterMap(Map<String, Object> parameterOverrides, Properties
...
+      return 1000;
+    }
+    String maxLag = (String) val;
+    String[] lagParts = maxLag.split(" ");
+    if (lagParts.length != 2
+        || (lagParts[0] == null || "".equals(lagParts[0]))
+        || (lagParts[1] == null || "".equals(lagParts[1]))) {
+      return 1000;
+    }
+    long unit;
```
sfc-gh-asen, line 202: nit: imo unit is the second part, this can be named value or lag

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -149,20 +160,74 @@ private void setParameterMap(Map<String, Object> parameterOverrides, Properties
...
+      return cachedBufferFlushIntervalMs;
+    } else {
+      Object val =
+          this.parameterMap.getOrDefault(
+              BUFFER_FLUSH_INTERVAL_IN_MILLIS, BUFFER_FLUSH_INTERVAL_IN_MILLIS_DEFAULT);
+      return (val instanceof String) ? Long.parseLong(val.toString()) : (long) val;
+    }
+  }
+
+  private long getMaxClientLagMs() {
```
sfc-gh-asen, line 190: This method might cause some confusing results if the customer enters wrong inputs.

(happy case) if the input is "1 second" the output is 1000

(not string) if the input is 1 or 5 or 1000 the result is 1000
(no unit) if the input is "1" or "5" or "1000" the result is 1000
(double space) if the input is "5 seconds" or "abc abc" the result is 1000

(non parsable value) if the input is "abc abc" then we get an IllegalArgumentException

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -149,20 +160,74 @@ private void setParameterMap(Map<String, Object> parameterOverrides, Properties
...
+            String.format("Invalid time unit supplied = '%s", lagParts[1]));
     }
-    return (long) val;
   }
 
-  /** @return Time in milliseconds between checks to see if the buffer should be flushed */
+  private boolean getMaxClientLagEnabled() {
+    Object val =
+        this.parameterMap.getOrDefault(MAX_CLIENT_LAG_ENABLED, MAX_CLIENT_LAG_ENABLED_DEFAULT);
+    return (val instanceof Boolean) ? (Boolean) val : false;
```
sfc-gh-asen, line 225: `getEnableParquetInternalBuffering` and `hasEnabledSnowpipeStreamingMetrics` checks for `val instanceof String`, should we do the same here?

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -149,20 +160,74 @@ private void setParameterMap(Map<String, Object> parameterOverrides, Properties
...
+          this.parameterMap.getOrDefault(
+              BUFFER_FLUSH_INTERVAL_IN_MILLIS, BUFFER_FLUSH_INTERVAL_IN_MILLIS_DEFAULT);
+      return (val instanceof String) ? Long.parseLong(val.toString()) : (long) val;
+    }
+  }
+
+  private long getMaxClientLagMs() {
+    Object val = this.parameterMap.getOrDefault(MAX_CLIENT_LAG, MAX_CLIENT_LAG_DEFAULT);
+    if (!(val instanceof String)) {
+      return 1000;
```
sfc-gh-asen, line 193: nit: BUFFER_FLUSH_INTERVAL_IN_MILLIS_DEFAULT instead of 1000

sfc-gh-asen (commented):
Looks good, i had couple questions only, thank you for the quick change!

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-asen: Should we also throw IllegalArgumentException for the above cases instead of reverting back to default?

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-tjones: Good question: if we find that there's a bug with this new code path we can disable it and go back to the old behavior.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-tjones: Yeah let's throw an exception here.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-tjones: Ack.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-tjones: Let's use lag.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-tjones: Yeah that seems better.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-xhuang: Makes sense to me

src/test/java/net/snowflake/ingest/streaming/internal/ParameterProviderTest.java
```diff
@@ -123,4 +131,115 @@ public void withDefaultValues() {
...
+    try {
+      parameterProvider.getBufferFlushIntervalInMs();
+      Assert.fail("Should not have succeeded");
+    } catch (IllegalArgumentException e) {
+      Assert.assertTrue(e.getMessage().startsWith("Invalid time unit"));
+    }
+  }
+
+  @Test
+  public void testMaxClientLagEnabledInvalidUnit() {
```
sfc-gh-xhuang, line 232: Should we try these 2 test cases?
parameterMap.put(ParameterProvider.MAX_CLIENT_LAG, "0 seconds");
parameterMap.put(ParameterProvider.MAX_CLIENT_LAG, "");

sfc-gh-xhuang: maybe even negative time and float times too
parameterMap.put(ParameterProvider.MAX_CLIENT_LAG, "-2 seconds");
parameterMap.put(ParameterProvider.MAX_CLIENT_LAG, "1.5 seconds");

sfc-gh-xhuang:
What's the max lag that can be set ? 
Perhaps we should enforce an upper bound, 10 minutes?

src/test/java/net/snowflake/ingest/streaming/internal/ParameterProviderTest.java
sfc-gh-tjones: What should the behavior be for 0 seconds?

sfc-gh-tjones:
> What's the max lag that can be set ? Perhaps we should enforce an upper bound, 10 minutes?

We probably should have a minimum and maximum. 10 minutes seems reasonable as an upper limit, what about for a lower limit?

src/test/java/net/snowflake/ingest/streaming/internal/ParameterProviderTest.java
sfc-gh-tjones: NVM I set a min/max.

sfc-gh-tjones:
> What's the max lag that can be set ? Perhaps we should enforce an upper bound, 10 minutes?

@sfc-gh-xhuang added minimum of 1 second, max of 10 minutes. PTAL

sfc-gh-asen (approved):
Looks great! Thank you

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -149,17 +166,80 @@ private void setParameterMap(Map<String, Object> parameterOverrides, Properties
...
+      lag = Long.parseLong(lagParts[0]);
+    } catch (Throwable t) {
+      throw new IllegalArgumentException(
+          String.format("Failed to parse MAX_CLIENT_LAG = '%s'", lagParts[0]), t);
+    }
+    long computedLag = BUFFER_FLUSH_INTERVAL_IN_MILLIS_DEFAULT;
+    switch (lagParts[1].toLowerCase()) {
+      case "second":
+      case "seconds":
+        computedLag = lag * 1000;
```
sfc-gh-japatel, line 218: nit: lets use `TimeUnit.SECONDS.toMillis(1)`

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -149,17 +166,80 @@ private void setParameterMap(Map<String, Object> parameterOverrides, Properties
...
+        break;
+      case "minute":
+      case "minutes":
+        computedLag = lag * 60000;
```
sfc-gh-japatel, line 222: similar: `lag * TimeUnit.SECONDS.toMillis(60)`

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -45,6 +49,15 @@ public class ParameterProvider {
...
   public static final long MAX_CHANNEL_SIZE_IN_BYTES_DEFAULT = 32000000L;
   public static final long MAX_CHUNK_SIZE_IN_BYTES_DEFAULT = 128000000L;
+
+  // Lag related parameters
+  public static final String MAX_CLIENT_LAG_DEFAULT = "1 second";
+  public static final boolean MAX_CLIENT_LAG_ENABLED_DEFAULT = true;
+
+  static final long MAX_CLIENT_LAG_MS_MIN = 1000;
+
+  // 10 minutes
```
sfc-gh-japatel, line 59: nit: here as well!

sfc-gh-japatel (approved):
lgtm!

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-tjones: Ack

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-tjones: Ack

PR #589, opened 2023-09-20
SNOW-919423 Surfaces Table Schema for a Channel

sfc-gh-tjones:
We've had requests to surface the table schema for a channel so that it is possible to reason about what columns need to be supplied to a given call of `insertRow` or `insertRows`. This surfaces a map of Column Name to Column Properties that are normally surfaced in the output of `SHOW COLUMNS`.

@test adds test to `SnowflakeStreamingIngestChannelTest.java`

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
+++ b/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
@@ -8,6 +8,7 @@ import java.time.ZoneId;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import javax.annotation.Nullable;
+import net.snowflake.ingest.streaming.internal.ColumnProperties;
 
 /**
  * A logical partition that represents a connection to a single Snowflake table, data will be
@@ -253,4 +254,11 @@ public interface SnowflakeStreamingIngestChannel {
    */
   @Nullable
   String getLatestCommittedOffsetToken();
+
+  /**
+   * Gets the table schema associated with this channel
+   *
+   * @return map representing Column Name -> Column Properties
+   */
+  Map<String, ColumnProperties> getTableSchema();
 }
--- a/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
+++ b/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
@@ -256,7 +256,9 @@ public interface SnowflakeStreamingIngestChannel {
   String getLatestCommittedOffsetToken();
 
   /**
-   * Gets the table schema associated with this channel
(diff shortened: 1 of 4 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/ColumnProperties.java
```diff
@@ -0,0 +1,59 @@
+package net.snowflake.ingest.streaming.internal;
+
+/**
```
sfc-gh-tzhang, line 3: Could you add a comment to explain the relationship with `ColumnMetdata`?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
```diff
@@ -126,6 +130,7 @@ class SnowflakeStreamingIngestChannelInternal<T> implements SnowflakeStreamingIn
         "Channel={} created for table={}",
         this.channelFlushContext.getName(),
         this.channelFlushContext.getTableName());
+    this.tableColumns = new HashMap<>();
```
sfc-gh-tzhang, line 133: wrong indentation?

src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
```diff
@@ -253,4 +254,11 @@ InsertValidationResponse insertRows(
    */
   @Nullable
   String getLatestCommittedOffsetToken();
+
+  /**
+   * Gets the table schema associated with this channel
+   *
+   * @return map representing Column Name -> Column Properties
+   */
+  Map<String, ColumnProperties> getTableSchema();
```
sfc-gh-tzhang, line 263: Could you make it clear in the comment that this is not update to date schema?

sfc-gh-tzhang (approved):
Left some comments, otherwise LGTM!

src/main/java/net/snowflake/ingest/streaming/internal/ColumnProperties.java
sfc-gh-tjones: Absolutely

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
sfc-gh-tjones: Weird! I ran the formatter. Let me fix.

src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
sfc-gh-tjones: Yeah good call out.

PR #590, opened 2023-09-20
SNOW-918949: Add row index information to all exceptions

sfc-gh-tzhang:
Add row index information to all exceptions, also unify the format so it can be easily parsed


src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -210,8 +210,8 @@ private float addRow(
       throw new SFException(
```
sfc-gh-tjones, line 210: Alternatively, is it worth having a list of longs that we store in a `SFException` that can be accessed via something like `getRowIndexes`? This will reduce the need to parse an error message which is error prone.

sfc-gh-tzhang: We already doing that with CONTINUE, this is for ABORT where we thrown an exception directly

sfc-gh-rcheng (approved):
lgtm. small nit: i think `insertRowsCurrIndex` is more descriptive for the exception and log parsing than `Row Index`

sfc-gh-xhuang (approved):

sfc-gh-japatel:
> lgtm. small nit: i think `insertRowsCurrIndex` is more descriptive for the exception and log parsing than `Row Index`

+1

sfc-gh-tzhang:
> > lgtm. small nit: i think `insertRowsCurrIndex` is more descriptive for the exception and log parsing than `Row Index`
> 
> +1

I updated everything to rowIndex, I think it's better than insertRowsCurrIndex

PR #596, opened 2023-10-04
SNOW-926149 Fix issues while using snowflake-jdbc-fips

sfc-gh-lsembera:
This PR implements support for FIPS-compliant JDBC driver `snowflake-jdbc-fips`, which some users are using instead of the default `snowflake-jdbc`. The difference is that the `snowflake-jdbc` shades bouncy castle, but `snowflake-jdbc-fips` declares FIPS compliant bouncy castle JARs in its `pom.xml`.

The existing SDK versions are not working with `snowflake-jdbc-fips` because they assume that bouncy castle is bundled in the JDBC driver and that `BouncyCastleProvider` exists on classpath, but in case of the FIPS version, the provider class name is `BouncyCastleFipsProvider`.

Documentation has been updated with instructions how to use the SDK in a FIPS-compliant way.

This PR also introduces a new external testing suite - a dedicated maven project, which does not test the SDK from "within" the SDK maven project, like our existing unit and integrations tests do, but runs a simple integration test with the SDK declared as a standard Maven dependency, installed into the local Maven repository. This suite can detect issues with JAR builds, for example. There is a wrapper script, which builds the SDK and orchestrates the end-to-end JAR tests. The following JAR variants are tested: shaded, unshaded and FIPS-compliant setups.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ The Snowflake Ingest SDK provides shaded and unshaded versions of its jar. The s
 whereas the unshaded version declares its dependencies in `pom.xml`, which are fetched as standard transitive dependencies by the build system like Maven or Gradle.
 The shaded JAR can help avoid potential dependency conflicts, but the unshaded version provides finer graned control over transitive dependencies.
 
-## Using with snowflake-jdbc-fics
+## Using with snowflake-jdbc-fips
 
 For use cases, which need to use `snowflake-jdbc-fips` instead of the default `snowflake-jdbc`, we recommend to take the following steps:
 
--- a/README.md
+++ b/README.md
@@ -65,6 +65,25 @@ dependencies {
 }
 ```
 
+## Jar Versions
+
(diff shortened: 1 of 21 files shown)


README.md
```diff
@@ -65,6 +65,25 @@ dependencies {
 }
 ```
 
+## Jar Versions
+
+The Snowflake Ingest SDK provides shaded and unshaded versions of its jar. The shaded version bundles the dependencies into its own jar,
+whereas the unshaded version declares its dependencies in `pom.xml`, which are fetched as standard transitive dependencies by the build system like Maven or Gradle.
+The shaded JAR can help avoid potential dependency conflicts, but the unshaded version provides finer graned control over transitive dependencies.
+
+## Using with snowflake-jdbc-fics
```
sfc-gh-japatel, line 74: nit: `jdbc-fips`

sfc-gh-japatel (approved):
super cool, thanks for adding these tests!

README.md
sfc-gh-lsembera: Fixed

PR #597, opened 2023-10-04
SNOW-906224: Add new ON_ERROR=SKIP_BATCH option

sfc-gh-tzhang:
Introduce a new ON_ERROR option SKIP_BATCH
- Skips the entire batch
- Returns all the erroneous rows as part of the response, along with their indexes and error messages

Pros:
- Easy implementation

Cons:
- Worse performance: data validation will be done twice for the good rows

[One pager doc](https://docs.google.com/document/d/1wa-yYOe-clIJTAaSRKSdpeRYE6CrBtiUVrC0hl-bd_w/edit#heading=h.i3tdayjco02j) for Snowflake internal review only


src/test/java/net/snowflake/ingest/streaming/internal/RowBufferTest.java
```diff
@@ -1393,10 +1403,87 @@ public void testOnErrorAbortFailures() {
     Assert.assertEquals(0, innerBuffer.bufferedRowCount);
   }
 
+  @Test
+  public void testOnErrorAbortSkipBatch() {
```
sfc-gh-tzhang, line 1407: This test needs to be updated, and we need IT tests as well

src/main/java/net/snowflake/ingest/streaming/OpenChannelRequest.java
```diff
@@ -12,6 +12,7 @@ public class OpenChannelRequest {
   public enum OnErrorOption {
     CONTINUE, // CONTINUE loading the rows, and return all the errors in the response
     ABORT, // ABORT the entire batch, and throw an exception when we hit the first error
+    SKIP_BATCH, // Skip the batch after the first error, and return all the errors in the response
```
sfc-gh-tjones, line 15: Maybe change to `If an error in the batch is detected return a response containing all error row indexes. No data is ingested`

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -356,10 +357,46 @@ public InsertValidationResponse insertRows(
                 this.statsMap.put(
                     colName,
                     RowBufferStats.getCombinedStats(stats, this.tempStatsMap.get(colName))));
-      }
+        this.channelState.setOffsetToken(offsetToken);
+      } else if (onErrorOption == OpenChannelRequest.OnErrorOption.SKIP_BATCH) {
```
sfc-gh-tjones, line 361: For readability and to reduce the size of the function should we pull the below into its own function?

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -297,76 +297,140 @@ Set<String> verifyInputColumns(
...
-        this.statsMap.forEach(
-            (colName, stats) ->
-                this.statsMap.put(
-                    colName,
-                    RowBufferStats.getCombinedStats(stats, this.tempStatsMap.get(colName))));
+        response = insertRowsHelperForContinue(rows, offsetToken);
+      } else if (onErrorOption == OpenChannelRequest.OnErrorOption.ABORT) {
+        response = insertRowsHelperForAbort(rows, offsetToken);
+      } else if (onErrorOption == OpenChannelRequest.OnErrorOption.SKIP_BATCH) {
+        response = insertRowsHelperForSkipBatch(rows, offsetToken);
```
sfc-gh-japatel, line 312: This is a good place to implement Strategy design and factory pattern. 

Being a little pedantic but this is how it would look like. (You probably would know it but just hashing it out here so there is less confusion)


```
C'tor of AbstractRowBuffer would fetch the implementation based on factory class

public class InsertionStrategyFactory {
    public static InsertionStrategy createStrategy(OnErrorOption type) {
        switch (type) {
            case CONTINUE:
                return new ContinueInsertionStrategy();
            case ABORT:
                return new AbortInsertionStrategy();
            case SKIP_BATCH:
                return new SkipBatchInsertionStrategy();
            default:
                throw new IllegalArgumentException("Unknown on error option: " + type);
        }
    }
}


public interface InsertionStrategy {
    Response insertRows(List<Row> rows, OffsetToken offsetToken);
}
// 
public class ContinueInsertionStrategy implements InsertionStrategy {

    @Override
    public Response insertRows(List<Row> rows, OffsetToken offsetToken) {
        channelState.updateInsertStats(System.currentTimeMillis(), bufferedRowCount);
        return insertRowsHelperForContinue(rows, offsetToken);
    }
}

```


Benefits:
- we can make this code more testable. 
- Abstract out different implementations. 
- In future we can add more strategies easily. 



WDYT?

sfc-gh-japatel: Might as well use abstract class if you think there is some common implementation in all three subclasses.

sfc-gh-tjones (approved):

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -581,4 +651,18 @@ private void checkBatchSizeRecommendedMaximum(float batchSizeInBytes) {
           INSERT_ROWS_RECOMMENDED_MAX_BATCH_SIZE_IN_BYTES);
     }
   }
+
+  /** Create the ingestion strategy based on the channel on_error option */
+  IngestionStrategy<T> createIngestionStrategy(OpenChannelRequest.OnErrorOption onErrorOption) {
```
sfc-gh-japatel, line 656: nice.. looks good, thanks for refactoring!

sfc-gh-japatel (approved):
lgtm! Thank you!

PR #645, opened 2023-10-24
SNOW-949967 Add an optional offset token parameter for openChannel

sfc-gh-asen:
Adds an optional offsetToken field to OpenChannelRequest.java which will overwrite existing offset token if provided.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/OpenChannelRequest.java
+++ b/src/main/java/net/snowflake/ingest/streaming/OpenChannelRequest.java
@@ -40,6 +40,9 @@ public class OpenChannelRequest {
   // Default timezone for TIMESTAMP_LTZ and TIMESTAMP_TZ columns
   private final ZoneId defaultTimezone;
 
+  private final String offsetToken;
+  private final boolean isOffsetTokenProvided;
+
   public static OpenChannelRequestBuilder builder(String channelName) {
     return new OpenChannelRequestBuilder(channelName);
   }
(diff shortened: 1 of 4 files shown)


sfc-gh-tjones (approved):

src/main/java/net/snowflake/ingest/streaming/OpenChannelRequest.java
```diff
@@ -40,6 +40,9 @@ public enum OnErrorOption {
   // Default timezone for TIMESTAMP_LTZ and TIMESTAMP_TZ columns
   private final ZoneId defaultTimezone;
 
+  private final String offsetToken;
+  private final boolean isOffsetTokenProvided;
```
sfc-gh-tzhang, line 44: Could you explain why isOffsetTokenProvided is required?

sfc-gh-tjones: I think this is to distinguish null as an offset token vs. it not being set.

sfc-gh-asen: Exactly, i want to support setting the offsetToken as null as well.

 In order to distinguish between default null string vs user provided null, I am using this extra flag.

sfc-gh-tzhang: I see, could you add a IT test for NULL?

PR #651, opened 2023-11-14
SNOW-928922 send spansMixedTables flag in blob registration requests

sfc-gh-gdoci:
This PR add a new field `spansMixedTables` in BlobMetadata, which is send in the blob registration request.
This flag will be used server-side for optimizations and sanity checks. 
After this PR, we want to simplify the retry logic so that the new register blob request has all chunks contained in a blob and attempted previously, not just failed chunks.

Both of these will serve the same goal to have correct flag value server-side. The flag is there in case there is something wrong with retries or we have to revert back to old retry logic. 

**I'll wait for the server-side change to be deployed before merging this PR.**

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/BlobMetadata.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/BlobMetadata.java
@@ -18,11 +18,18 @@ class BlobMetadata {
   private final Constants.BdecVersion bdecVersion;
   private final List<ChunkMetadata> chunks;
   private final BlobStats blobStats;
+  private final boolean spansMixedTables;
 
   // used for testing only
   @VisibleForTesting
   BlobMetadata(String path, String md5, List<ChunkMetadata> chunks, BlobStats blobStats) {
-    this(path, md5, ParameterProvider.BLOB_FORMAT_VERSION_DEFAULT, chunks, blobStats);
+    this(
+        path,
+        md5,
+        ParameterProvider.BLOB_FORMAT_VERSION_DEFAULT,
+        chunks,
+        blobStats,
+        chunks == null ? false : chunks.size() > 1);
   }
 
   BlobMetadata(
@@ -30,12 +37,14 @@ class BlobMetadata {
       String md5,
       Constants.BdecVersion bdecVersion,
       List<ChunkMetadata> chunks,
-      BlobStats blobStats) {
+      BlobStats blobStats,
+      boolean spansMixedTables) {
     this.path = path;
     this.md5 = md5;
     this.bdecVersion = bdecVersion;
     this.chunks = chunks;
     this.blobStats = blobStats;
+    this.spansMixedTables = spansMixedTables;
   }
 
   @JsonIgnore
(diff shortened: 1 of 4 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/BlobMetadata.java
```diff
@@ -18,24 +18,33 @@ class BlobMetadata {
...
   @VisibleForTesting
   BlobMetadata(String path, String md5, List<ChunkMetadata> chunks, BlobStats blobStats) {
-    this(path, md5, ParameterProvider.BLOB_FORMAT_VERSION_DEFAULT, chunks, blobStats);
+    this(
+        path,
+        md5,
+        ParameterProvider.BLOB_FORMAT_VERSION_DEFAULT,
+        chunks,
+        blobStats,
+        chunks == null ? false : chunks.size() > 1);
```
sfc-gh-tzhang, line 32: can `chunks` be null here?

src/main/java/net/snowflake/ingest/streaming/internal/BlobMetadata.java
```diff
@@ -18,24 +18,33 @@ class BlobMetadata {
   private final Constants.BdecVersion bdecVersion;
   private final List<ChunkMetadata> chunks;
   private final BlobStats blobStats;
+  private final boolean spansMixedTables;
```
sfc-gh-tzhang, line 21: want to add a comment explain this? The name is not straight forward like other fields

sfc-gh-tzhang: Or I suggest we call it something like `hasMultipleTables`

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -568,8 +568,15 @@ BlobMetadata upload(
         blob.length,
         System.currentTimeMillis() - startTime);
 
+    // at this point we know for sure if the BDEC file has data for more than one chunk, i.e.
```
sfc-gh-tzhang, line 571: nit
```suggestion
    // At this point we know for sure if the BDEC file has data for more than one chunk, i.e.
```

sfc-gh-tzhang (approved):
Left some comments, otherwise LGTM! And I assume the server side change is now rollback safe?

src/main/java/net/snowflake/ingest/streaming/internal/BlobMetadata.java
sfc-gh-gdoci: Yes, because that constructor is used in tests (only) and `RegisterServiceTest` calls it with `chunks null`.

src/main/java/net/snowflake/ingest/streaming/internal/BlobMetadata.java
sfc-gh-gdoci: I added a comment.

sfc-gh-gdoci:
> And I assume the server side change is now rollback safe?

Yeah, it will work with old SDKs. The default value server-side if the flag is missing is true.

sfc-gh-xhuang (approved):

PR #654, opened 2023-12-05
SNOW-983635 Allow ZSTD compression algorithm

sfc-gh-lthiede:
Test setting ZSTD as default compression algorithm


sfc-gh-xhuang (approved):

sfc-gh-lthiede:
Maybe add zstd-jni as dependency in README.md

sfc-gh-lthiede:
### DEW testing

DEW tests using a local release of the Snowpipe Streaming SDK with my changes and ZSTD as the default compression algorithm passed.
To run successfully I had to add a dependency on zstd-jni to the DEW  dependencies ([PR](https://github.com/snowflakedb/snowflake/pull/147870))

sfc-gh-lthiede:
### How I ran the precommits
I ran a [precommit](http://sfc-eng-jenkins-dpe-ci.snowflakecomputing.com.s3-website-us-west-2.amazonaws.com/floeci/45ec1b4dee634f31b9fb376ba04ff389/index.html#) with my branch rebased on sdk release 2.0.5.
I ran a second [precommit](http://sfc-eng-jenkins-dpe-ci.snowflakecomputing.com.s3-website-us-west-2.amazonaws.com/floeci/481fc5db05244cb39382a3e55834a6fc/index.html#) with sdk release 2.0.5 to check which test failures are actually caused by my changes.
I ran both precommits with some changes to the BDEC file generator and its build in RegressionTests. ([PR](https://github.com/snowflakedb/snowflake/pull/149492))
The changes fixed the file generator for SDK release 2.0.5 and improved the its build.
### Results
Expected failures:
- t_replication_streaming_ingest:mixed_table_replication.py same failure on release 2.0.5
- t_replication_streaming_ingest:rg_failover_mixed_table_replication.py same failure on release 2.0.5
- t_streaming:basic_changetracking_on_mixed_table_register_after_changetracking_enabled.py on AZURE, CORE, GCS only reduced byte counts
- t_streaming_2:mixed_table.py on AZURE, CORE, GCS only reduced byte counts
- t_streaming_2:mixed_table_bdec_blacklist.py on AZURE, CORE, GCS same failure on release 2.0.5
- t_streaming_2:mixed_table_dcs_checker.py on AZURE, CORE, GCS only reduced byte count
- t_streaming_2:mixed_table_dml_patch.py on AZURE, CORE, GCS only reduced byte counts
- t_streaming_2:mixed_table_ep_in_fdb_scan.py on AZURE, CORE, GCS same failure on release 2.0.5 and reduced byte count
- t_streaming_2:mixed_table_null_full_name.py on AZURE, CORE, GCS only reduced byte counts
- t_streaming_2:mixed_table_real_zero_sign.py on AZURE, CORE, GCS only reduced byte count
- t_streaming_2:mixed_table_ep_in_fdb_bg_migration.py on CORE only reduced byte counts
- t_streaming_2:mixed_table_accounting.py on AZURE, CORE, GCS an assertion fails due to reduced byte counts

Unexpected faulures:
- t_streaming_2:mixed_table_ep_scan.py on AZURE, CORE, GCS increased byte count 58576 to 58588
- t_streaming_2:mixed_table_dml.py on AZURE unexpected failure

Is zstd sometimes leading to a higher byte count a problem?
@sfc-gh-gdoci suggested that the failure of t_streaming_2:mixed_table_dml.py only on AZURE is probably flaky. I ran [another precommit](http://sfc-eng-jenkins-dpe-ci.snowflakecomputing.com.s3-website-us-west-2.amazonaws.com/floeci/d11617fc42af448797915d28719b1cf9/index.html#) with only t_streaming_2 on AZURE and the failure didn't show up again.

README.md
```diff
@@ -17,6 +17,7 @@ The Snowflake Ingest Service SDK depends on the following libraries:
 
 * snowflake-jdbc (3.13.30 to 3.13.33) 
 * slf4j-api
+* com.github.luben:zstd-jni (1.5.0-1)
```
sfc-gh-tzhang, line 20: Why we decide to treat this as one of the special dependencies? @sfc-gh-lsembera has more context here if you need more help

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
```diff
@@ -8,6 +8,7 @@
 import java.io.IOException;
 import java.util.List;
 import java.util.Map;
+
```
sfc-gh-tzhang, line 11: is this intended?

sfc-gh-tzhang (dismissed):
Thanks for the change! I think we need more test coverage here, for example: we need coverage to make sure it works on other data types. We probably have most of them which you can just add a parameter like what you did in StreamingIngestBigFilesIT.java

sfc-gh-tzhang:
What's our plan of making ZSTD the default?

README.md
sfc-gh-lthiede: Hi, the problem is that zstd-jni can't be shaded. Because of that we can't include it in our sdk's shaded jar. It has to be provided separately when customers want to build their application with our sdk. I thought this was reason enough to list it with our dependencies.

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
sfc-gh-lthiede: not really, I'll revert it

sfc-gh-lthiede:
@sfc-gh-tzhang In the protocol of our Streaming Ingest Repeating Eng Sync on Dec 5th we wrote

> [@sfc-gh-gdoci /Lorenz] Rolling out (risky) changes in the SDK, e.g. switching from Parquet page version v1 to v2
> - A way to control client parameters from Snowflake server-side to rollback in case of issues before customers are asked to upgrade to SDK version with the fix, e.g. override parameters in clientConfigure or openChannel request
> - [Toby] Wanted to do smth similar for Kafka connector [SNOW-980722](https://snowflakecomputing.atlassian.net/browse/SNOW-980722) 
> - [Toby] IMO makes sense to do, it’s a priority question. Could be a good starter task for Alec when he returns 

I think this also applies to making ZSTD the default

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
sfc-gh-lthiede, line 1: @sfc-gh-tzhang I parameterised this file to run with GZIP and ZSTD. It tests ingesting the different data types (and many other things). Is there something else you'd like to be tested?

sfc-gh-gdoci (approved):
Thank you.

README.md
sfc-gh-tzhang: @sfc-gh-lsembera Any idea if we can do better here? Or any potential issue? I'm concerned that we keep adding to this list

public_pom.xml
```diff
@@ -48,5 +48,10 @@
             <version>1.7.36</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>com.github.luben</groupId>
+            <artifactId>zstd-jni</artifactId>
+            <version>1.5.0-1</version>
```
sfc-gh-tzhang, line 54: should we add compile here similar to the other two?

sfc-gh-tzhang (commented):
Please consider me as approved if you can check with @sfc-gh-lsembera to see if we can do better on the ZSTD dependency, thanks!

sfc-gh-tzhang:
> @sfc-gh-tzhang In the protocol of our Streaming Ingest Repeating Eng Sync on Dec 5th we wrote
> 
> > [@sfc-gh-gdoci /Lorenz] Rolling out (risky) changes in the SDK, e.g. switching from Parquet page version v1 to v2
> > 
> > * A way to control client parameters from Snowflake server-side to rollback in case of issues before customers are asked to upgrade to SDK version with the fix, e.g. override parameters in clientConfigure or openChannel request
> > * [Toby] Wanted to do smth similar for Kafka connector [SNOW-980722](https://snowflakecomputing.atlassian.net/browse/SNOW-980722)
> > * [Toby] IMO makes sense to do, it’s a priority question. Could be a good starter task for Alec when he returns
> 
> I think this also applies to making ZSTD the default

Ah, thanks for the reminder, we can do this, is this something you want to start if you have free time?

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
sfc-gh-tzhang: Thanks! I looked through the tests, looks like we have a bunch of data type related IT tests under streaming.internal.datatypes, do you think it worths to update them as well?

pom.xml
```diff
@@ -691,6 +694,9 @@
             <configuration>
               <failOnWarning>true</failOnWarning>
               <ignoreNonCompile>true</ignoreNonCompile>
+              <ignoredDependencies>
+                <ignoredDependency>com.github.luben:zstd-jni</ignoredDependency>
```
sfc-gh-lsembera, line 698: Why is this needed?

pom.xml
```diff
@@ -1092,6 +1095,12 @@
                     <exclude>org/slf4j/**</exclude>
                   </excludes>
                 </filter>
+                <filter>
```
sfc-gh-lsembera, line 1098: We are not shading this dependency, why is this filter needed?

public_pom.xml
sfc-gh-lsembera: `compile` is the default, so in order to be consistent, we should either remove it everywhere and add it everywhere.

README.md
sfc-gh-lsembera: I think it is good if shade fewer dependencies, and this change goes in that direction. Another advantage of the unshaded approach here is that if the customer does not want to use zstd and wants smaller distribution, they can exclude the zstd jar from their maven project.

pom.xml
sfc-gh-lthiede: We declare zstd-jni as a dependency even though it's not directly used in our code. If we didn't ignore zstd-jni here analyze-only would fail due to unused dependency.

sfc-gh-lsembera: In that case, we can use the `runtime` maven scope?

sfc-gh-lthiede:
> Ah, thanks for the reminder, we can do this, is this something you want to start if you have free time?
@sfc-gh-tzhang Currently I'm busy with other parts of my project

sfc-gh-lsembera (approved):

sfc-gh-asen (approved):

README.md
sfc-gh-tzhang: Looks like we stop relocating `com.github.luben`, is there any concern that this is a behavior change?

sfc-gh-tzhang (approved):
lgtm, thanks!

PR #656, opened 2023-12-09
SNOW-984877 Update to support customized url and add customer name in request header

sfc-gh-dwang:
SNOW-984877  idea
Update to support customized url and option to add account name in request http header


sfc-gh-dwang:
@sfc-gh-tzhang @sfc-gh-tjones can you take a look if this change is ok?
basically we have 2 changes needed to support localhost:
1. SnowhouseURL to support localhost as url, no account name in the url path
2. we need to add the snowflake-account header in the requestbuilder
I tested in local devvm, and it works, but likely need to test in temptest env
Can we deploy a snapshot unshaded jar?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -151,13 +148,36 @@ public class SnowflakeStreamingIngestClientInternal<T> implements SnowflakeStrea
...
-      boolean isTestMode,
-      RequestBuilder requestBuilder,
-      Map<String, Object> parameterOverrides) {
+          String name,
+          SnowflakeURL accountURL,
+          Properties prop,
+          CloseableHttpClient httpClient,
+          boolean isTestMode,
+          RequestBuilder requestBuilder,
+          Map<String, Object> parameterOverrides) {
```
sfc-gh-tjones, line 157: Formatting looks wrong

sfc-gh-tjones (approved):

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
sfc-gh-dwang: updated

sfc-gh-dwang:
@sfc-gh-tjones can you take another check?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -59,10 +59,7 @@
 import javax.management.ObjectName;
 import net.snowflake.client.core.SFSessionProperty;
 import net.snowflake.client.jdbc.internal.apache.http.impl.client.CloseableHttpClient;
-import net.snowflake.ingest.connection.IngestResponseException;
-import net.snowflake.ingest.connection.OAuthCredential;
-import net.snowflake.ingest.connection.RequestBuilder;
-import net.snowflake.ingest.connection.TelemetryService;
+import net.snowflake.ingest.connection.*;
```
sfc-gh-tjones, line 62: Nit, no wildcard imports.

sfc-gh-tjones (approved):

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
sfc-gh-dwang: did not notice this was changed to wildcard, will fix, thanks for the catching

PR #657, opened 2023-12-13
Allow clients to drop channel on close or blindly

sfc-gh-psaha:
This change exposes a new API to drop channels. It can be done either on channel close (verifies that no other client has raced to reopen the channel) or blindly with just a table name.
Note that dropping a channel blindly discards any pending data and will invalidate any already opened channel - so use with caution.


sfc-gh-psaha:
Note that tests will fail until the API is enabled on the server side.

src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClient.java
```diff
@@ -25,6 +25,20 @@ public interface SnowflakeStreamingIngestClient extends AutoCloseable {
...
+   * <p>Note that {@link DropChannelRequest.DropChannelRequestBuilder#setClientSequencer(Long)}} can
+   * be used to drop a specific version of the channel and prevent accidentally dropping a channel
+   * concurrently opened by another client. If it is not specified, this call will blindly drop the
+   * latest version of the channel and any pending data will be lost. Also see {@link
+   * OpenChannelRequest.OpenChannelRequestBuilder#setDropOnClose(boolean)} to automatically drop
+   * channels on close.
+   *
+   * @param request the drop channel request
+   */
+  void dropChannel(DropChannelRequest request);
```
sfc-gh-tjones, line 40: Is this a public docstring?

sfc-gh-xhuang:
@sfc-gh-psaha do we want to allow this only on primary or both primary and secondary? 
I think primary only for replication is standard as we disallow writes and deletes on secondary 
https://docs.snowflake.com/en/user-guide/account-replication-considerations#replication-and-snowpipe-streaming

src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClient.java
sfc-gh-psaha: Yes, it is. Any suggestions to improve it?

sfc-gh-psaha:
> @sfc-gh-psaha do we want to allow this only on primary or both primary and secondary? I think primary only for replication is standard as we disallow writes and deletes on secondary https://docs.snowflake.com/en/user-guide/account-replication-considerations#replication-and-snowpipe-streaming

Currently, we allow it only on the primary - same as open channel. This check is performed server-side.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
```diff
@@ -291,6 +297,19 @@ public CompletableFuture<Void> close() {
                         .map(SnowflakeStreamingIngestChannelInternal::getFullyQualifiedName)
                         .collect(Collectors.toList()));
               }
+              if (this.dropOnClose) {
```
sfc-gh-tzhang, line 300: should we drop the channel even if it hits the exception above?

src/test/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelTest.java
```diff
@@ -808,6 +818,46 @@ public void testClose() throws Exception {
 
     // Calling close again on closed channel shouldn't fail
     channel.close().get();
+    Mockito.verify(client, Mockito.times(0)).dropChannel(Mockito.any());
+  }
+
+  @Test
+  public void testDropOnClose() throws Exception {
```
sfc-gh-tzhang, line 825: could you add a test on drop on invalid?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
sfc-gh-tzhang: In theory, I don't see an issue of dropping a closed channel, should this be removed?

sfc-gh-tzhang: Also what about when a channel becomes invalid, currently we can't close an invalid channel, but we should allow dropping them?

sfc-gh-tzhang (commented):
Thanks for the change! My biggest discussion point is how to expose the API to the customer. To me, it seems more intuitive to support a new channel.close(boolean ifDrop) function instead of specifying it as part of open channel, WDYT? I'm not sure if we have ask this to the customers that requested this feature

src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClient.java
sfc-gh-psaha: Good point. I have removed references to the sequencer from the public api now. The internal code still supports sequencer when invoked via `close` but that's not visible externally anymore.

src/main/java/net/snowflake/ingest/streaming/OpenChannelRequest.java
```diff
@@ -43,6 +43,10 @@ public enum OnErrorOption {
   private final String offsetToken;
   private final boolean isOffsetTokenProvided;
 
+  // If true, the channel will be dropped when it is closed after any pending data is fully
+  // committed.
+  private final boolean dropOnClose;
```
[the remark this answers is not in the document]
sfc-gh-psaha: The original reason I did it this way was because I thought Channel extends `AutoCloseable` and therefore, we cannot change the signature of `close` but I was mistaken - only the client extends it. Since users of this class has to call `close` explicitly anyways, I think your suggestions is the less-intrusive approach. There is still a corner case of someone using try-with using the Client and closing channels that way implicitly without any option to drop them but I think that's likely a rare corner case. We can always go and implement my idea if that is requested.

src/main/java/net/snowflake/ingest/streaming/internal/DropChannelResponse.java
```diff
@@ -0,0 +1,68 @@
+package net.snowflake.ingest.streaming.internal;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
```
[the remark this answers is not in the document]
sfc-gh-psaha: Done.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
```diff
@@ -291,6 +297,19 @@ public CompletableFuture<Void> close() {
...
+                this.owningClient.dropChannel(
+                    DropChannelRequest.builder(this.getChannelContext().getName())
+                        .setDBName(this.getDBName())
+                        .setTableName(this.getTableName())
+                        .setSchemaName(this.getSchemaName())
+                        .setClientSequencer(this.getChannelSequencer())
+                        .build());
+                System.out.println(
```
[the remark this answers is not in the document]
sfc-gh-psaha: Good eye. Done.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
sfc-gh-psaha: I think dropping after failing to commit would be a surprising behavior. Typically, when `close` fails, the caller is supposed to handle it e.g. by retrying etc. Dropping in this specific case would lose uncommitted data.

src/test/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelTest.java
sfc-gh-psaha: Done. Note that invalidation check is done at the start of the `close` function now and therefore, we never get to flushing or dropping the channel.

sfc-gh-tzhang (approved):
LGTM, thanks!

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
sfc-gh-tzhang: I think it depends, I can imagine a use case where customer always open a new channel with an unique identifier every time the old channel gets invalidated. I guess we can come back to this if needed in the future, basically right now we won't drop any closed and invalid channels.

sfc-gh-xhuang (approved):
let me know if this is ready

sfc-gh-psaha:
@sfc-gh-tzhang Please take another look.

sfc-gh-tzhang (approved):
LGTM, thanks!

sfc-gh-tzhang:
The change itself looks good to me, it's just the impact of accidentally dropping the wrong channel worries me as the offset token info will be lost. We need to be careful about documenting this, maybe this is something customer wants to reach out to support before doing this. cc: @sfc-gh-xhuang

PR #659, opened 2024-01-04
NO-SNOW: Deprecate BUFFER_FLUSH_INTERVAL_IN_MILLIS parameter

sfc-gh-tzhang:
- After `MAX_CLIENT_LAG` was introduced, it's not backward compatible with the current logic since `MAX_CLIENT_LAG_ENABLED` is true by default and we won't even check for `BUFFER_FLUSH_INTERVAL_IN_MILLIS`. Instead of supporting backward compatible, I decide to remove `BUFFER_FLUSH_INTERVAL_IN_MILLIS` completely since that is an undocumented parameter. We only support `MAX_CLIENT_LAG` going forward and using `BUFFER_FLUSH_INTERVAL_IN_MILLIS` will result in an exception. 
- To match the behavior of `BUFFER_FLUSH_INTERVAL_IN_MILLIS` so the transition is smooth, `MAX_CLIENT_LAG` now also supports LONG input and input without a time unit (default will be milliseconds)
- Some code format changes done automatically by the Google Java Format
- Reenable some tests disabled long time ago

### DIFF
--- a/src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
+++ b/src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
@@ -295,7 +295,8 @@ public class RequestBuilder {
       SecurityManager securityManager,
       CloseableHttpClient httpClient,
       String clientName) {
-    this(accountName,
+    this(
+        accountName,
         userName,
         credential,
         schemeName,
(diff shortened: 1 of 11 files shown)


sfc-gh-tjones (approved):

sfc-gh-japatel (approved):
lgtm! Thank you

sfc-gh-xhuang (approved):

Issue #599, opened 2023-10-05
NoSuchMethodError when using snowflake-ingest-sdk 2.0.3


sfc-gh-kterada:
Sharing for awareness: https://github.com/snowflakedb/snowflake-jdbc/issues/1524

sfc-gh-kterada:
Until there's a fix available, please use a [supported](https://github.com/snowflakedb/snowflake-ingest-java#dependencies) 3.13.x version of the Snowflake JDBC driver.

sfc-gh-xhuang:
JDBC 3.14.3 was recently released and fixes this.

sfc-gh-xhuang:
@sfc-gh-tzhang @sfc-gh-psaha Can you of you upgrade the dependency to JDBC 3.14.3 for the next release?

sfc-gh-xhuang:
This has been resolved in the latest 2.1.0 sdk release as we have upgraded to JDBC 3.14.5

PR #660, opened 2024-01-05
SNOW-990469 Update snowjdbc version to 3.14.4

sfc-gh-psaha:
Updating to the latest version to address: https://github.com/snowflakedb/snowflake-ingest-java/issues/599

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -65,7 +65,7 @@
     <shadeBase>net.snowflake.ingest.internal</shadeBase>
     <slf4j.version>1.7.36</slf4j.version>
     <snappy.version>1.1.10.4</snappy.version>
-    <snowjdbc.version>3.13.30</snowjdbc.version>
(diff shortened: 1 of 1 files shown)


sfc-gh-tzhang:
@sfc-gh-psaha We would need to root cause SNOW-987122 before upgrading JDBC version, this is being identified as part of upgrading JDBC version in KC and I assume we would hit the same issue for the Ingest SDK

sfc-gh-psaha:
> @sfc-gh-psaha We would need to root cause SNOW-987122 before upgrading JDBC version, this is being identified as part of upgrading JDBC version in KC and I assume we would hit the same issue for the Ingest SDK

Ack. I'll mark this blocked on that issue then and wait for now.

sfc-gh-xhuang:
blocked by https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/819

PR #662, opened 2024-01-12
NO-SNOW: Revert one change that updates public API for internal use case

sfc-gh-tzhang:
Revert https://github.com/snowflakedb/snowflake-ingest-java/commit/2b702db9918f26eaf59a64346bab06462e10186f

This change updates the public API for an internal special use case only which is not ideal, looks like we keep updating public APIs in https://github.com/snowflakedb/snowflake-ingest-java/pull/661 as well. Given that we're still in a POC phase, I suggest we do everything in another branch instead, hence removing this from the next public release and could discuss what's the best way to support this internal use case once the POC is done.

### DIFF
--- a/src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
+++ b/src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
@@ -54,10 +54,6 @@ public class RequestBuilder {
   // whatever the actual host is
   private final String host;
 
-  private final String accountName;
-
-  private final boolean addAccountNameInRequest;
-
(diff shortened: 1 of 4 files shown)


sfc-gh-dwang (approved):
LGTM, I can create a new PR after POC if necessary

sfc-gh-xhuang (approved):
thanks, best to remove before next sdk release

sfc-gh-japatel (approved):
lgtm! 
Please have a meaning description in the title PR to not have to open commits from change logs. 
Thanks

PR #670, opened 2024-01-27
v2.0.5-unshaded release

sfc-gh-tzhang:


pom.xml
```diff
@@ -779,9 +779,9 @@
         <configuration>
           <errorRemedy>failFast</errorRemedy>
           <!--
-                               The list of allowed licenses. If you see the build failing due to "There are some forbidden licenses used, please
-                              check your dependencies", verify the conditions of the license and add the reference to it here.
-                            -->
+                                         The list of allowed licenses. If you see the build failing due to "There are some forbidden licenses used, please
```
sfc-gh-lsembera, line 782: ?

src/test/java/net/snowflake/ingest/connection/UserAgentTest.java
```diff
@@ -5,10 +5,12 @@
 import java.util.Properties;
 import net.snowflake.ingest.SimpleIngestManager;
 import org.junit.Assert;
+import org.junit.Ignore;
 import org.junit.Test;
 
 public class UserAgentTest {
   @Test
+  @Ignore
```
sfc-gh-lsembera, line 13: Why aren't we changing also `RequestBuilder.DEFAULT_VERSION`?

sfc-gh-lsembera: Ah, we want the version in user agent without `-unshaded`, right?

pom.xml
sfc-gh-tzhang: This is done by running the format.sh

src/test/java/net/snowflake/ingest/connection/UserAgentTest.java
sfc-gh-tzhang: yup, at least that's what we did for the past few versions and we don't have an issue so far

sfc-gh-xhuang:
I think we can skip 2.0.5-unshaded and release 2.0.6-unshaded

sfc-gh-lsembera (approved):

PR #673, opened 2024-02-02
SNOW-1032874: Fix an Overflow Issue for Integer-store Timestamp 

sfc-gh-tzhang:
According to https://docs.snowflake.com/en/sql-reference/functions/to_timestamp#usage-notes:

- If the integer is less than 31536000000 (the number of milliseconds in a year), then the value is treated as a number of seconds.
- If the value is greater than or equal to 31536000000 and less than 31536000000000, then the value is treated as milliseconds.
- If the value is greater than or equal to 31536000000000 and less than 31536000000000000, then the value is treated as microseconds.
- If the value is greater than or equal to 31536000000000000, then the value is treated as nanoseconds.

Due to the multiplication logic in the code, we run into an overflow issue when the input is bigger than a certain value (for example: 31535999999999) and it causes **silent corrupted values** in the table, this change updates the code to use BigInteger instead of long to resolve the overflow issue.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
@@ -746,12 +746,6 @@ class DataValidationUtil {
       epochNanos = new BigInteger(input);
     }
 
-    long a = epochNanos.divide(BigInteger.valueOf(Power10.intTable[9])).longValue();
-    long b = epochNanos.remainder(BigInteger.valueOf(Power10.intTable[9])).longValue();
-    Instant c = Instant.ofEpochSecond(a, b);
-    LocalDateTime d = LocalDateTime.ofInstant(c, ZoneOffset.UTC);
-    LocalTime e = d.toLocalTime();
-
     return Instant.ofEpochSecond(
         epochNanos.divide(BigInteger.valueOf(Power10.intTable[9])).longValue(),
         epochNanos.remainder(BigInteger.valueOf(Power10.intTable[9])).longValue());
--- a/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
@@ -733,11 +733,11 @@ class DataValidationUtil {
       long val = Long.parseLong(input);
 
       if (val > -SECONDS_LIMIT_FOR_EPOCH && val < SECONDS_LIMIT_FOR_EPOCH) {
-        epochNanos = BigInteger.valueOf(val).multiply(BigInteger.valueOf(Power10.intTable[9]));
+        epochNanos = BigInteger.valueOf(val).multiply(Power10.sb16Table[9]);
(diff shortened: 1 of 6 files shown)


src/test/java/net/snowflake/ingest/streaming/internal/DataValidationUtilTest.java
```diff
@@ -251,20 +253,47 @@ public void testValidateAndParseTime() {
...
     assertEquals(1500, wrapper.getTimeZoneIndex());
 
     wrapper = validateAndParseTimestamp("COL", "  2021-01-01T01:00:00.123 \t\n", 9, UTC, true, 0);
-    Assert.assertEquals(1609462800, wrapper.getEpoch());
+    Assert.assertEquals(1609462800, wrapper.getEpochSecond());
     Assert.assertEquals(123000000, wrapper.getFraction());
     Assert.assertEquals(new BigInteger("1609462800123000000"), wrapper.toBinary(false));
 
+    // Test integer-stored time and scale guessing
+    SimpleDateFormat df = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS");
```
sfc-gh-lsembera, line 271: I think we also need integration test coverage, which would test JDBC compatibility. We have a bunch of IT tests for scale guessing in [DateTimeIT.java](https://github.com/snowflakedb/snowflake-ingest-java/blob/master/src/test/java/net/snowflake/ingest/streaming/internal/datatypes/DateTimeIT.java), just not for these edge cases. Could you add tests along the lines of [these](https://github.com/snowflakedb/snowflake-ingest-java/blob/master/src/test/java/net/snowflake/ingest/streaming/internal/datatypes/DateTimeIT.java#L382)?

sfc-gh-tzhang: added

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -728,20 +728,27 @@ static BigInteger validateAndParseTime(
...
-    } else if (val > -MICROSECONDS_LIMIT_FOR_EPOCH && val < MICROSECONDS_LIMIT_FOR_EPOCH) {
-      epochNanos = val * Power10.intTable[3];
-    } else {
-      epochNanos = val;
+    BigInteger epochNanos;
+    try {
+      long val = Long.parseLong(input);
+
+      if (val > -SECONDS_LIMIT_FOR_EPOCH && val < SECONDS_LIMIT_FOR_EPOCH) {
+        epochNanos = BigInteger.valueOf(val).multiply(BigInteger.valueOf(Power10.intTable[9]));
```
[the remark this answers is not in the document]
sfc-gh-tzhang: updated

sfc-gh-japatel (approved):
lgtm! thank you

sfc-gh-xhuang (approved):

sfc-gh-lsembera (approved):
Thank you!

sfc-gh-tzhang:
Will merge this one first, we might need to update the logic again once we hear back from SQL folks about the behavior for negative timestamps.

PR #674, opened 2024-02-07
Snowpipe Streaming: send column ordinal back to GS with blob EPs to register

sfc-gh-azagrebin:
This PR sends column ordinals additionally to column names to server side for cross-check of columns in case of table schema changes where columns are recreated with the same name but different data type.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
@@ -8,7 +8,7 @@ import java.util.Objects;
 
 /** Audit register endpoint/FileColumnPropertyDTO property list. */
 class FileColumnProperties {
-
+  private int columnOrdinal;
   private String minStrValue;
 
   private String maxStrValue;
@@ -45,6 +45,7 @@ class FileColumnProperties {
   public static final Double DEFAULT_MIN_MAX_REAL_VAL_FOR_EP = 0d;
 
   FileColumnProperties(RowBufferStats stats) {
+    this.setColumnOrdinal(stats.getOrdinal());
     this.setCollation(stats.getCollationDefinitionString());
     this.setMaxIntValue(
         stats.getCurrentMaxIntValue() == null
@@ -81,6 +82,15 @@ class FileColumnProperties {
 
     this.setNullCount(stats.getCurrentNullCount());
     this.setDistinctValues(stats.getDistinctValues());
+  }
+
+  @JsonProperty("columnId")
+  public int getColumnOrdinal() {
+    return columnOrdinal;
+  }
+
+  public void setColumnOrdinal(int columnOrdinal) {
+    this.columnOrdinal = columnOrdinal;
   }
 
   // Annotation required in order to have package private fields serialized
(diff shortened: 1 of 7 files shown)


sfc-gh-tjones (approved):

sfc-gh-lsembera (approved):
Thanks!

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
```diff
@@ -83,6 +84,15 @@ class FileColumnProperties {
     this.setDistinctValues(stats.getDistinctValues());
   }
 
+  @JsonProperty("columnId")
```
sfc-gh-japatel, line 87: Should this be `columnOrdinal`?

sfc-gh-japatel (approved):
lgtm! Thank you

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
sfc-gh-azagrebin: I thought about this but then server side fails to deserialise with unknown field.
We can still detect if it is sent on server side by making it nullable.

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
```diff
@@ -220,7 +231,8 @@ public boolean equals(Object o) {
     if (this == o) return true;
     if (o == null || getClass() != o.getClass()) return false;
     FileColumnProperties that = (FileColumnProperties) o;
-    return distinctValues == that.distinctValues
+    return Objects.equals(columnOrdinal,that.columnOrdinal)
```
sfc-gh-tzhang, line 234: what's the reason we're using objects.equals for int?

src/main/java/net/snowflake/ingest/streaming/internal/RowBufferStats.java
```diff
@@ -29,14 +30,15 @@ class RowBufferStats {
...
+  RowBufferStats(String columnDisplayName, String collationDefinitionString, int ordinal) {
     this.columnDisplayName = columnDisplayName;
     this.collationDefinitionString = collationDefinitionString;
+    this.ordinal= ordinal;
     reset();
   }
 
   RowBufferStats(String columnDisplayName) {
-    this(columnDisplayName, null);
+    this(columnDisplayName, null, 0);
```
sfc-gh-tzhang, line 41: This is just for tests? what about -1?

sfc-gh-tzhang (approved):
LGTM, left some minor comments, ship it!

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
```diff
@@ -8,7 +8,7 @@
 
 /** Audit register endpoint/FileColumnPropertyDTO property list. */
 class FileColumnProperties {
-
+  private int columnOrdinal;
```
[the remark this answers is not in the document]
sfc-gh-azagrebin: I have not done any research on this.

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
sfc-gh-azagrebin: yes but it looks to me misleading tbh

PR #675, opened 2024-02-07
SNOW-967235 Sends start and end offsets from the Client SDK

sfc-gh-tjones:
This sends the start and end offset tokens for a channel when sending a blob to Snowflake. Right now only the end offset is sent which makes reasoning about ranges difficult.

@test verified via existing tests in `RowBufferTest` and `SnowflakeStreamingIngestChannelTest`

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
@@ -470,6 +470,7 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
       float oldBufferSize = 0F;
       long oldRowSequencer = 0;
       String oldOffsetToken = null;
+      String oldStartOffsetToken = null;
       Map<String, RowBufferStats> oldColumnEps = null;
       Pair<Long, Long> oldMinMaxInsertTimeInMs = null;
 
@@ -484,6 +485,7 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
           oldBufferSize = this.bufferSize;
           oldRowSequencer = this.channelState.incrementAndGetRowSequencer();
           oldOffsetToken = this.channelState.getEndOffsetToken();
+          oldStartOffsetToken = this.channelState.getStartOffsetToken();
           oldColumnEps = new HashMap<>(this.statsMap);
           oldMinMaxInsertTimeInMs =
               new Pair<>(
@@ -508,6 +510,7 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
         data.setBufferSize(oldBufferSize);
         data.setRowSequencer(oldRowSequencer);
         data.setOffsetToken(oldOffsetToken);
+        data.setStartOffsetToken(oldStartOffsetToken);
         data.setColumnEps(oldColumnEps);
         data.setMinMaxInsertTimeInMs(oldMinMaxInsertTimeInMs);
         data.setFlusherFactory(this::createFlusher);
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ChannelMetadata.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ChannelMetadata.java
@@ -17,6 +17,7 @@ class ChannelMetadata {
   private final Long clientSequencer;
   private final Long rowSequencer;
   @Nullable private final String offsetToken;
+  @Nullable private final String startOffsetToken;
 
   static Builder builder() {
     return new Builder();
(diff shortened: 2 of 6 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/ChannelRuntimeState.java
```diff
@@ -56,12 +61,21 @@ long getRowSequencer() {
...
+  }
+
   /**
    * Updates the channel's offset token.
    *
    * @param offsetToken new offset token
    */
   void setOffsetToken(String offsetToken) {
+    if (!startOffsetTokenSet) {
+      startOffsetToken = offsetToken;
```
sfc-gh-tzhang, line 76: Not sure why we're setting the startOffsetToken to offsetToken, offsetToken is the last offset of the first batch, so basically we're sending (last offset of first batch, last offset of the chunk) which won't be continuous from the previous chunk, also the row count won't be matched.

sfc-gh-tzhang (commented):
Thanks for taking this JIRA! I believe your logic needs to based on my [PR](https://github.com/snowflakedb/snowflake-ingest-java/pull/676) which will allow customer to input the start offset of a batch, otherwise the offset we send to server side will never be continuous. And glad we're thinking about doing the same thing :)

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -457,6 +457,7 @@ public ChannelData<T> flush(final String filePath) {
       float oldBufferSize = 0F;
       long oldRowSequencer = 0;
       String oldOffsetToken = null;
```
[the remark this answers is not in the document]
sfc-gh-tjones: I don't think we want to do that. I thought about it and in order to minimize changes server side while preserving backwards compatibility and ensuring a safe rollout we need to keep `offset_token` and treat `offset_token` as `end_offset_token`. This ensures that both old clients and new ones continue to work.

src/main/java/net/snowflake/ingest/streaming/internal/ChannelRuntimeState.java
sfc-gh-tjones: This is where we get offset token for a batch, no? We copy the data here from `ChannelRuntimeState` for a `RowBuffer` into a `ChannelDTO` or something further down the line. So with `startOffsetTokenSet` we essentially grab the FIRST offset token provided in a call to `insertRows` for a batch and store that.

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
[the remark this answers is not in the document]
sfc-gh-tzhang: Oh, I mean update offsetToken to endOffsetToken, then when passing it to server side, both offsetToken and endOffsetToken should use endOffsetToken to support backwards compatibility, happy to chat if it's still confusing

src/main/java/net/snowflake/ingest/streaming/internal/ChannelRuntimeState.java
sfc-gh-tzhang: If you see how ChannelRuntimeState is being constructed, it's only being constructed once in the SnowflakeStreamingIngestChannelInternal and never being reseted between flushes. You can run a test to verify, but I think this means the startOffsetToken will always be the same for all chunks.

sfc-gh-tzhang:
@sfc-gh-tjones https://github.com/snowflakedb/snowflake-ingest-java/pull/676 is merged and should unblock your PR

src/main/java/net/snowflake/ingest/streaming/internal/ChannelRuntimeState.java
sfc-gh-tzhang: I updated the logic in `ChannelRuntimeState` already to store the correct start and end offset, so this is not needed anymore

sfc-gh-asen (approved):

src/main/java/net/snowflake/ingest/streaming/internal/ChannelMetadata.java
```diff
@@ -17,6 +17,7 @@ class ChannelMetadata {
   private final Long clientSequencer;
   private final Long rowSequencer;
   @Nullable private final String offsetToken;
```
sfc-gh-azagrebin, line 19:
```suggestion
  @Nullable private final String endOffsetToken;
```

this is basically an internal field name, does it make sense to rename it as well for better readability?

PR #676, opened 2024-02-08
SNOW-979849: Add start offset token at batch level in the Client SDK

sfc-gh-tzhang:
Support start offset token to be part of the interface so that it can be used to log and verify offset related behavior in upcoming PRs

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
+++ b/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
@@ -260,7 +260,7 @@ public interface SnowflakeStreamingIngestChannel {
    * time of a channel open event. The schema may be changed on the Snowflake side in which case
    * this will continue to show an old schema version until the channel is re-opened.
    *
-   * @return map representing Column Name -> Column Properties
+   * @return map representing Column Name to Column Properties
    */
   Map<String, ColumnProperties> getTableSchema();
 }
--- a/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
+++ b/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
@@ -66,14 +66,10 @@ public interface SnowflakeStreamingIngestChannel {
    */
   String getFullyQualifiedTableName();
 
-  /**
-   * @return a boolean which indicates whether the channel is valid
-   */
+  /** @return a boolean which indicates whether the channel is valid */
   boolean isValid();
 
-  /**
-   * @return a boolean which indicates whether the channel is closed
-   */
+  /** @return a boolean which indicates whether the channel is closed */
   boolean isClosed();
 
   /**
--- a/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
+++ b/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
@@ -66,10 +66,14 @@ public interface SnowflakeStreamingIngestChannel {
    */
   String getFullyQualifiedTableName();
 
-  /** @return a boolean which indicates whether the channel is valid */
+  /**
+   * @return a boolean which indicates whether the channel is valid
+   */
   boolean isValid();
 
-  /** @return a boolean which indicates whether the channel is closed */
+  /**
+   * @return a boolean which indicates whether the channel is closed
+   */
   boolean isClosed();
 
   /**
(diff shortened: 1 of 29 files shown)


sfc-gh-tjones (commented):
Should this be done in the Client? I think we're better off doing this server side for KC specific pipes.

sfc-gh-tzhang:
> Should this be done in the Client? I think we're better off doing this server side for KC specific pipes.

@sfc-gh-tjones I don't see how we can verify batch level offset at server side, happy for any suggestions

sfc-gh-tjones:
> > Should this be done in the Client? I think we're better off doing this server side for KC specific pipes.
> 
> @sfc-gh-tjones I don't see how we can verify batch level offset at server side, happy for any suggestions

Assuming start and end offset are sent server side, you should be able to check that start offset = previously committed end offset unless null.

sfc-gh-tzhang:
> > > Should this be done in the Client? I think we're better off doing this server side for KC specific pipes.
> > 
> > 
> > @sfc-gh-tjones I don't see how we can verify batch level offset at server side, happy for any suggestions
> 
> Assuming start and end offset are sent server side, you should be able to check that start offset = previously committed end offset unless null.

That's won't be at batch level, right? That is chunk level

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -653,6 +677,42 @@ private void checkBatchSizeRecommendedMaximum(float batchSizeInBytes) {
...
+          }
+        }
+      } catch (NumberFormatException ignored) {
+        // Do nothing since we can't compare the offset
+      }
+
+      if (reportMismatch) {
+        this.telemetryService.reportBatchOffsetMismatch(
+            channelFullyQualifiedName, prevEndOffset, curStartOffset, curEndOffset, rowCount);
+      }
```
sfc-gh-tjones, line 712: We should put `if (!reportMismatch) return` at the top of this method because otherwise we're doing a bunch of work for nothing.

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -653,6 +677,42 @@ private void checkBatchSizeRecommendedMaximum(float batchSizeInBytes) {
...
+        // We verify that the end_offset - start_offset + 1 = row_count
+        if (curEnd - curStart + 1 != rowCount) {
+          reportMismatch = true;
+        }
+
+        // We verify that start_offset_of_current_batch = end_offset_of_previous_batch+1
+        if (prevEndOffset != null) {
+          long prevEnd = Long.parseLong(prevEndOffset);
+          if (curStart != prevEnd + 1) {
+            reportMismatch = true;
```
sfc-gh-tjones, line 702: This should really be renamed `mismatchDetected` because technically reporting a mismatch is determined by having telemetry and/or logging enabled

src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
```diff
@@ -240,10 +240,20 @@ public interface SnowflakeStreamingIngestChannel {
...
-   *     could be null if you don't plan on replaying or can't replay
+   * @param startOffsetToken start offset of the batch/row-set
+   * @param endOffsetToken end offset of the batch/row-set, used for replay in case of failures, *
+   *     It could be null if you don't plan on replaying or can't replay
    * @return insert response that possibly contains errors because of insertion failures
    */
+  InsertValidationResponse insertRows(
+      Iterable<Map<String, Object>> rows,
+      @Nullable String startOffsetToken,
+      @Nullable String endOffsetToken);
```
sfc-gh-tjones, line 251: Is this transparently invoked by the connector? There's nothing user facing here that has to change right?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
```diff
@@ -369,6 +378,16 @@ public InsertValidationResponse insertRows(
...
   }
 
+  /**
+   * Insert a batch of rows into the channel with the end offset token only, please see {@link
+   * SnowflakeStreamingIngestChannel#insertRows(Iterable, String, String)} for more information.
+   */
+  @Override
+  public InsertValidationResponse insertRows(
+      Iterable<Map<String, Object>> rows, String offsetToken) {
+    return insertRows(rows, null, offsetToken);
```
sfc-gh-tjones, line 388: When is this invoked.

src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
sfc-gh-tzhang: No, this is a new interface and the old one is still there for backward compatibility

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
sfc-gh-tzhang: I think the github diff is making things confusing, this is the old interface that customer currently used to insert rows with the rows and the end_offset_token, the new interface added is to allow them pass start_offset_token as well

sfc-gh-tjones (approved):
Toby and I chatted about this one IRL and I think its fine to go in as-is. We think that longer term the right solution is to have per-row offset tokens (`RECORD$CHANNEL_METADATA` or similar that could act as a variant column with `CHANNEL_NAME` and `OFFSET_TOKEN` fields respectively). However, since that warrants more discussion we feel like this is an acceptable intermediary solution

sfc-gh-japatel:
Thanks for this effort, 
Can you update the PR description, it is a bit confusing, it says 

> and verify offset related behavior in upcoming PRs
I dont think we want to verify any offset behavior right? start and end offset are still opaque to the sdk.

PR #677, opened 2024-02-08
SNOW-987122 Upgrade JDBC to 3.14.5 and Catch new exception type for r…

sfc-gh-japatel:
…enewing expired S3 token

### Changes

- Here is the reason why we were not able to run for more than 1 hour when updating JDBC to latest available version: https://github.com/snowflakedb/snowflake-jdbc/pull/1473/files
- @sfc-gh-lsembera did add a retry logic but that only compatible with version 3.13.30
- Few changes are because of formatter. 
- `isCredentialsExpiredException` function is removed! 
  - we will retry on all exceptions and run metadata refresh on first exception. 
- Fixed a failing test and will fix it once next version of JDBC is released: https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/819
  - The issue is only while JDBC query reads a column containing large values, not writing to table. 

- Ran longRunningTest present in E2E Jar. Look at this commit's E2E runs: (Took> 2 hours to run) 35c8468029dcdbd243bf2e73710833798ceb8f68

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ authentication. Currently, we support ingestion through the following APIs:
 
 The Snowflake Ingest Service SDK depends on the following libraries:
 
-* snowflake-jdbc (3.13.30 to 3.13.33) 
+* snowflake-jdbc (3.13.30 to 3.14.5) 
 * slf4j-api
 * com.github.luben:zstd-jni (1.5.0-1)
 
@@ -24,7 +24,7 @@ using a build system, please make sure these dependencies are on the classpath.
 
 # Prerequisites
 
-**If your project depends on the Snowflake JDBC driver, as well, please make sure the JDBC driver version is 3.13.30 to 3.13.33. JDBC driver version 3.14.0 or higher is currently not supported. **
+**If your project depends on the Snowflake JDBC driver, as well, please make sure the JDBC driver version is 3.13.30 to 3.14.5.
 
 ## Java 8+
 
--- a/e2e-jar-test/fips/src/test/java/net/snowflake/FipsIngestE2ETest.java
+++ b/e2e-jar-test/fips/src/test/java/net/snowflake/FipsIngestE2ETest.java
@@ -33,6 +33,7 @@ public class FipsIngestE2ETest {
   }
 
   @Test
+  @Ignore("Takes too long to run")
   public void longRunningTest() throws InterruptedException {
     ingestTestUtils.runLongRunningTest(Duration.of(80, ChronoUnit.MINUTES));
   }
--- a/e2e-jar-test/fips/src/test/java/net/snowflake/FipsIngestE2ETest.java
+++ b/e2e-jar-test/fips/src/test/java/net/snowflake/FipsIngestE2ETest.java
@@ -33,7 +33,6 @@ public class FipsIngestE2ETest {
   }
 
   @Test
-  @Ignore("Takes too long to run")
   public void longRunningTest() throws InterruptedException {
     ingestTestUtils.runLongRunningTest(Duration.of(80, ChronoUnit.MINUTES));
   }
--- a/e2e-jar-test/standard/src/test/java/net/snowflake/StandardIngestE2ETest.java
+++ b/e2e-jar-test/standard/src/test/java/net/snowflake/StandardIngestE2ETest.java
@@ -28,6 +28,7 @@ public class StandardIngestE2ETest {
   }
 
   @Test
+  @Ignore("Takes too long to run")
   public void longRunningTest() throws InterruptedException {
     ingestTestUtils.runLongRunningTest(Duration.of(80, ChronoUnit.MINUTES));
   }
--- a/e2e-jar-test/standard/src/test/java/net/snowflake/StandardIngestE2ETest.java
+++ b/e2e-jar-test/standard/src/test/java/net/snowflake/StandardIngestE2ETest.java
@@ -28,7 +28,6 @@ public class StandardIngestE2ETest {
   }
 
   @Test
-  @Ignore("Takes too long to run")
   public void longRunningTest() throws InterruptedException {
     ingestTestUtils.runLongRunningTest(Duration.of(80, ChronoUnit.MINUTES));
   }
--- a/pom.xml
+++ b/pom.xml
@@ -711,6 +711,9 @@
             <configuration>
               <rules>
                 <banDuplicateClasses>
+                  <ignoreClasses>
+                    <ignoreClass>META-INF/versions/*/org/bouncycastle/*</ignoreClass>
+                  </ignoreClasses>
                   <findAllDuplicates>true</findAllDuplicates>
                   <ignoreWhenIdentical>true</ignoreWhenIdentical>
                 </banDuplicateClasses>
--- a/pom.xml
+++ b/pom.xml
@@ -65,7 +65,7 @@
     <shadeBase>net.snowflake.ingest.internal</shadeBase>
     <slf4j.version>1.7.36</slf4j.version>
     <snappy.version>1.1.10.4</snappy.version>
-    <snowjdbc.version>3.13.30</snowjdbc.version>
+    <snowjdbc.version>3.14.5</snowjdbc.version>
     <yetus.version>0.13.0</yetus.version>
   </properties>
 
--- a/src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
@@ -195,6 +195,12 @@ class StreamingIngestStage {
               .setDestFileName(fullFilePath)
               .build());
     } catch (Exception e) {
+      if (retryCount == 0) {
+        // for the first exception, we always perform a metadata refresh.
+        logger.logInfo(
+                "Stage metadata need to be refreshed due to upload error: {} on first retry attempt", e.getMessage());
+        this.refreshSnowflakeMetadata();
+      }
       if (retryCount >= maxUploadRetries) {
         logger.logError(
             "Failed to upload to stage, retry attempts exhausted ({}), client={}, message={}",
@@ -202,12 +208,6 @@ class StreamingIngestStage {
             clientName,
             e.getMessage());
         throw new SFException(e, ErrorCode.IO_ERROR);
-      }
-
-      if (isCredentialsExpiredException(e)) {
-        logger.logInfo(
-            "Stage metadata need to be refreshed due to upload error: {}", e.getMessage());
-        this.refreshSnowflakeMetadata();
       }
       retryCount++;
       StreamingIngestUtils.sleepForRetry(retryCount);
--- a/src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
@@ -5,6 +5,7 @@
 package net.snowflake.ingest.streaming.internal;
 
 import static net.snowflake.client.core.Constants.CLOUD_STORAGE_CREDENTIALS_EXPIRED;
+import static net.snowflake.client.jdbc.ErrorCode.S3_OPERATION_ERROR;
 import static net.snowflake.ingest.connection.ServiceResponseHandler.ApiName.STREAMING_CLIENT_CONFIGURE;
 import static net.snowflake.ingest.streaming.internal.StreamingIngestUtils.executeWithRetries;
 import static net.snowflake.ingest.utils.Constants.CLIENT_CONFIGURE_ENDPOINT;
@@ -28,6 +29,7 @@ import net.snowflake.client.jdbc.SnowflakeFileTransferAgent;
 import net.snowflake.client.jdbc.SnowflakeFileTransferConfig;
 import net.snowflake.client.jdbc.SnowflakeFileTransferMetadataV1;
 import net.snowflake.client.jdbc.SnowflakeSQLException;
+import net.snowflake.client.jdbc.SnowflakeSQLLoggedException;
 import net.snowflake.client.jdbc.cloud.storage.StageInfo;
 import net.snowflake.client.jdbc.internal.apache.commons.io.FileUtils;
 import net.snowflake.client.jdbc.internal.apache.http.impl.client.CloseableHttpClient;
@@ -225,7 +227,10 @@ class StreamingIngestStage {
       return false;
     }
 
-    if (e instanceof SnowflakeSQLException) {
+    if (e instanceof SnowflakeSQLLoggedException) {
+      return ((SnowflakeSQLLoggedException) e).getErrorCode()
+          == S3_OPERATION_ERROR.getMessageCode();
+    } else if (e instanceof SnowflakeSQLException) {
       return ((SnowflakeSQLException) e).getErrorCode() == CLOUD_STORAGE_CREDENTIALS_EXPIRED;
     } else if (e instanceof StorageException) {
       return ((StorageException) e).getCode() == 401;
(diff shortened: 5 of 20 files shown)


sfc-gh-psaha:
I expect large values test to fail until SNOW-1003775 is fixed but you can work around by changing the tests to read back the length of values instead of the values themselves. Note that gh actions will just crash for these failures and you won't see a proper test failure in the logs. @sfc-gh-japatel

e2e-jar-test/standard/src/test/java/net/snowflake/StandardIngestE2ETest.java
```diff
@@ -28,7 +28,6 @@ public void basicTest() throws InterruptedException {
   }
 
   @Test
-  @Ignore("Takes too long to run")
```
sfc-gh-japatel, line 31: I will remove in next commit before pushing!

sfc-gh-japatel:
Snyk checks can be ignored since it is complaining pom of e2e library 
Screenshot 2024-02-12 at 4 50 07 PM

README.md
```diff
@@ -24,7 +24,7 @@ using a build system, please make sure these dependencies are on the classpath.
 
 # Prerequisites
 
-**If your project depends on the Snowflake JDBC driver, as well, please make sure the JDBC driver version is 3.13.30 to 3.13.33. JDBC driver version 3.14.0 or higher is currently not supported. **
+**If your project depends on the Snowflake JDBC driver, as well, please make sure the JDBC driver version is 3.13.30 to 3.14.5.
```
sfc-gh-tzhang, line 27:
```suggestion
**If your project depends on the Snowflake JDBC driver, as well, please make sure the JDBC driver version is 3.13.30 or above.
```

pom.xml
```diff
@@ -711,6 +711,9 @@
             <configuration>
               <rules>
                 <banDuplicateClasses>
+                  <ignoreClasses>
+                    <ignoreClass>META-INF/versions/*/org/bouncycastle/*</ignoreClass>
+                  </ignoreClasses>
```
sfc-gh-tzhang, line 716: will be any behavior change with this?

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestBigFilesIT.java
```diff
@@ -40,8 +40,7 @@ public static Object[] compressionAlgorithms() {
     return new Object[] {"GZIP", "ZSTD"};
   }
 
-  @Parameter
-  public String compressionAlgorithm;
+  @Parameter public String compressionAlgorithm;
```
sfc-gh-tzhang, line 43: could you run format.sh to see if this will be reverted back?

src/test/java/net/snowflake/ingest/streaming/internal/DataValidationUtilTest.java
```diff
@@ -269,6 +270,7 @@ public void testValidateAndParseTimestamp() throws ParseException {
 
     // Test integer-stored time and scale guessing
     SimpleDateFormat df = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS");
+    df.setTimeZone(TimeZone.getTimeZone("UTC"));
```
sfc-gh-tzhang, line 273: This is interesting, is this a new behavior change in JDBC?

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
```diff
@@ -225,7 +227,10 @@ static boolean isCredentialsExpiredException(Exception e) {
       return false;
     }
 
-    if (e instanceof SnowflakeSQLException) {
+    if (e instanceof SnowflakeSQLLoggedException) {
+      return ((SnowflakeSQLLoggedException) e).getErrorCode()
+          == S3_OPERATION_ERROR.getMessageCode();
+    } else if (e instanceof SnowflakeSQLException) {
```
sfc-gh-tzhang, line 233: This is highly error-prone, the reason we need to fix it again is because the JDBC is throwing a new exception, and we have no control on whether they will update the exception type or add a new code pass that throughs a new type of exception, I think we should just catch all exceptions and always retry a few times. + @sfc-gh-lsembera to see if he has a better idea

sfc-gh-tzhang (commented):
Left some comments, PTAL, thanks!

sfc-gh-tzhang:
> Snyk checks can be ignored since it is complaining pom of e2e library Screenshot 2024-02-12 at 4 50 07 PM

I don't think this can be ignored since we don't see this issue before, and I assume Snyk will keep failing because of this

README.md
sfc-gh-xhuang: I would be explicit about the versions we support. Sometimes a new JDBC version comes out and we have not tried it.

This happened with a previous JDBC version and I made a change here to call out the explicit versions

sfc-gh-japatel: +1

pom.xml
sfc-gh-japatel: No, this is only for maven build

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
sfc-gh-japatel: I dont think catching all exception is good idea here. 
I agree that this is error prone and not safe for future versions but ideally the JDBC interface should not change its response. (I think if it as a non backward compatible change)
If you check JDBC code (`handleS3Exception`), it catches all ClientException and ServerException. 


It's better we catch specific exception and work with JDBC team! 
Thoughts?

src/test/java/net/snowflake/ingest/streaming/internal/DataValidationUtilTest.java
sfc-gh-japatel: the tests were not passing because of various TZ, I doubt it has anything to do with JDBC. 
I ran it on mac, someone else might run on devvm?

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestBigFilesIT.java
sfc-gh-japatel: `format.sh` suggested these changes.

README.md
sfc-gh-tzhang: yeah, but this is less ideal, what's the point of NOT shading the JDBC jar if we keep running into issues with other JDBC versions? If it's shaded, then we can always use a version without any issue

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
sfc-gh-tzhang: > ideally the JDBC interface should not change its response.

Agree that catch all is not ideal, but how would you guarantee that there won't be any non-compatible change in the future? We don't have anyone who look into every JDBC PRs and make sure this won't happen. In fact, the same issues happens twice in a few months, the first fix (by Lukas) is done because they update the exception type, the second fix (this PR) is because they add a new place with a different exception

src/test/java/net/snowflake/ingest/streaming/internal/DataValidationUtilTest.java
sfc-gh-tzhang: ah, might be the case, i use devvm, merge gate all passed without this line as well

README.md
sfc-gh-japatel: but still we use some JDBC with an ingest SDK right? which we feel is safe to use?

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
sfc-gh-japatel: > but how would you guarantee that there won't be any non-compatible change in the future?
I think this is something that the library owners should worry about. We as a consumer would have to go through these set of changes everytime we upgrade because a version we want to use fixes other issues. 

having that said, I dont have a strong preference of either of them, I just feel catch all is suppressing all exceptions without knowing what the root cause is.

e2e-jar-test/standard/src/test/java/net/snowflake/StandardIngestE2ETest.java
sfc-gh-lsembera: There is an equivalent long running test for FIPS JDBC driver in [FipsIngestE2ETest.java](https://github.com/snowflakedb/snowflake-ingest-java/blob/master/e2e-jar-test/fips/src/test/java/net/snowflake/FipsIngestE2ETest.java#L36), could you run it, as well?

pom.xml
```diff
@@ -65,7 +65,7 @@
     <shadeBase>net.snowflake.ingest.internal</shadeBase>
     <slf4j.version>1.7.36</slf4j.version>
     <snappy.version>1.1.10.4</snappy.version>
-    <snowjdbc.version>3.13.30</snowjdbc.version>
+    <snowjdbc.version>3.14.5</snowjdbc.version>
```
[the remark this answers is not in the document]
sfc-gh-japatel: yeah, we can wait now that it is pushed. but we will have to wait til next release of JDBC. 
are we okay with that @sfc-gh-xhuang ?

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
sfc-gh-japatel: or more like refresh the token regardless of an exception on the final attempt?

sfc-gh-lsembera: The benefit of doing it in the final attempt would be we would issue fewer token renewals, but on the other hand it would introduce latency because the token would only be renewed on 5th retry. I don't know which one is a bigger concern.

pom.xml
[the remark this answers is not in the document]
sfc-gh-xhuang: Let me see when next release of JDBC is.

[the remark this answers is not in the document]
sfc-gh-xhuang: Even if they release next week, there's no guarantees it will be a simple upgrade either. 
The large strings issue only affects reads and this test, it doesn't affect write behavior which is the only thing we use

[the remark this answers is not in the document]
sfc-gh-psaha: +1. We are just a bystander for that bug. It is not in our code and they were able to reproduce it themselves. I support not waiting for that fix to land.

e2e-jar-test/standard/src/test/java/net/snowflake/StandardIngestE2ETest.java
sfc-gh-japatel: Will do!

pom.xml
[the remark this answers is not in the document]
sfc-gh-xhuang: JDBC plans to release next week but please consider the trade off of trying to use 3.14.6 vs today's 3.14.5

src/test/java/net/snowflake/ingest/streaming/internal/ParameterProviderTest.java
```diff
@@ -310,15 +310,15 @@ public void testValidCompressionAlgorithmsAndWithUppercaseLowerCase() {
         });
     List<String> zstdValues = Arrays.asList("ZSTD", "zstd", "Zstd", "zStd");
```
sfc-gh-xhuang, line 311: unrelated unhelpful question but I don't understand the need for this test. 
Do we not just normalize the zstdValue input? Why do we only test 4 variant cases of ZSTD? There's ZStd, zSTd, etc

sfc-gh-japatel:
Folks, updated the PR with:
1. Kept using jdbc 3.13.5, modify test
2. No special exception handling, catch all exception and retry on very first exception. 

Thanks, PTAL!

README.md
sfc-gh-tzhang: I suggest we postponed this discussion and keep this as is

e2e-jar-test/fips/src/test/java/net/snowflake/FipsIngestE2ETest.java
```diff
@@ -33,7 +33,6 @@ public void basicTest() throws InterruptedException {
   }
 
   @Test
-  @Ignore("Takes too long to run")
```
sfc-gh-tzhang, line 36: add a comment in case you forget to add it back

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
```diff
@@ -201,12 +209,6 @@ private void putRemote(String fullFilePath, byte[] data, int retryCount)
             e.getMessage());
         throw new SFException(e, ErrorCode.IO_ERROR);
       }
-
-      if (isCredentialsExpiredException(e)) {
```
sfc-gh-tzhang, line 205: I might miss something, but this function is not used anymore?

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
sfc-gh-tzhang: +1 on always refresh the token on the first exception, and then retry 5 times if the exception persisted

e2e-jar-test/fips/src/test/java/net/snowflake/FipsIngestE2ETest.java
sfc-gh-japatel: I will remove it in next commit!

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
sfc-gh-japatel: yes, as discussed in slack, we will attempt to refresh token on first exception. so this method is not needed anymore.

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
sfc-gh-japatel: Yes, its implemented, PTAL! thanks

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
```diff
@@ -193,6 +195,12 @@ private void putRemote(String fullFilePath, byte[] data, int retryCount)
               .setDestFileName(fullFilePath)
               .build());
     } catch (Exception e) {
+      if (retryCount == 0) {
```
sfc-gh-japatel, line 198: CC: @sfc-gh-tzhang

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
sfc-gh-tzhang: Then the entire `isCredentialsExpiredException` function can be removed? Basically we say we always refresh the credentials on the first retry

README.md
sfc-gh-tzhang: keep what you have in the PR :)

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
sfc-gh-japatel: yess. its removed. I might miss something but I am confused. 
`refreshSnowflakeMetadata` is called above

sfc-gh-japatel: I realised i removed it but only from my mind, will remove the function here. thanks for calling it out!

sfc-gh-tzhang (approved):
LGTM, left one minor comment, thanks!

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
```diff
@@ -193,6 +191,13 @@ private void putRemote(String fullFilePath, byte[] data, int retryCount)
...
+        // for the first exception, we always perform a metadata refresh.
+        logger.logInfo(
+            "Stage metadata need to be refreshed due to upload error: {} on first retry attempt",
+            e.getMessage());
```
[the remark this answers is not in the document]
sfc-gh-japatel: This is preferable since this is indicating it is refreshing metadata. I can add stacktrace

PR #680, opened 2024-02-21
SNOW-979849: Support user to pass a user defined offset token verification logic as part of channel creation

sfc-gh-tzhang:
With this PR, user could pass in a user defined offset token verification function that is used to verify offset behaviors, if the verification failed, we will log an error and report it to SF


src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -669,6 +694,29 @@ private void checkBatchSizeRecommendedMaximum(float batchSizeInBytes) {
...
+      logger.logError(
+          "The channel {} might have an offset token mismatch based on the provided offset token"
+              + " verification logic, preEndOffset={}, curStartOffset={}, curEndOffset={},"
+              + " rowCount={}.",
+          channelFullyQualifiedName,
+          prevEndOffset,
+          curStartOffset,
+          curEndOffset,
+          rowCount);
+      telemetryService.reportBatchOffsetMismatch(
```
sfc-gh-tzhang, line 715: Consider adding a rate limiting to avoid customer intentionally flooded our telemetry service

sfc-gh-tzhang: rate limiter added

sfc-gh-xhuang (approved):

sfc-gh-psaha (changes requested):

src/main/java/net/snowflake/ingest/streaming/OffsetTokenVerificationFunction.java
```diff
@@ -0,0 +1,60 @@
...
+ *
+ *             // We verify that start_offset_of_current_batch = end_offset_of_previous_batch+1
+ *             if (prevBatchEndOffset != null) {
+ *               long prevEnd = Long.parseLong(prevBatchEndOffset);
+ *               if (curStart != prevEnd + 1) {
+ *                 hasMismatch = true;
+ *               }
+ *             }
+ *           } catch (NumberFormatException ignored) {
+ *             // Do nothing since we can't compare the offset
```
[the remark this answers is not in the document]
sfc-gh-tzhang: We actually allow offset token to be any arbitrary String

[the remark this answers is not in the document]
sfc-gh-psaha: I get that. I was only referring to this specific example where the user knows that it should be a parseable as a number.

src/main/java/net/snowflake/ingest/streaming/OffsetTokenVerificationFunction.java
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright (c) 2024 Snowflake Computing Inc. All rights reserved.
+ */
+
+package net.snowflake.ingest.streaming;
+
+/**
```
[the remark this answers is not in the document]
sfc-gh-tzhang: Added a sentence, PTAL, I don't think we have any documentation on this.

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
```diff
@@ -78,6 +79,35 @@ public static Object[] compressionAlgorithms() {
 
   @Parameter public String compressionAlgorithm;
 
+  private static final OffsetTokenVerificationFunction offsetTokenVerificationFunction =
```
sfc-gh-gdoci, line 82: Should we verify the behaviour of the this verification function by test cases that have/don't have a mismatch?

src/main/java/net/snowflake/ingest/streaming/OffsetTokenVerificationFunction.java
```diff
@@ -5,16 +5,18 @@
 package net.snowflake.ingest.streaming;
 
 /**
- * Interface to create a user defined offset token verification function that is used to verify
- * offset behaviors, if the verification failed, we will log a warning and report it to SF
+ * Interface to provide a custom offset verification logic. If specified, verification failures will
+ * be logged as warnings and reported to Snowflake. This interface could be used when there are
+ * certain assumption about the offset token behavior and please reach out to Snowflake if you
```
sfc-gh-psaha, line 10:
```suggestion
 * certain assumption about the offset token behavior. Please reach out to Snowflake if you
```

sfc-gh-psaha (approved):

src/main/java/net/snowflake/ingest/connection/TelemetryService.java
```diff
@@ -66,6 +69,8 @@ public String toString() {
   TelemetryService(CloseableHttpClient httpClient, String clientName, String url) {
     this.clientName = clientName;
     this.telemetry = (TelemetryClient) TelemetryClient.createSessionlessTelemetry(httpClient, url);
+    // At most once every second
+    this.rateLimiter = RateLimiter.create(1.0);
```
sfc-gh-japatel, line 73: isnt this too low? essentially it is one per second across all channels IIUC?

src/main/java/net/snowflake/ingest/connection/TelemetryService.java
```diff
@@ -122,6 +127,26 @@ public void reportCpuMemoryUsage(Histogram cpuUsage) {
     }
   }
 
+  public void reportBatchOffsetMismatch(
```
sfc-gh-japatel, line 130: nit: Javadocs on public method

src/main/java/net/snowflake/ingest/streaming/OffsetTokenVerificationFunction.java
```diff
@@ -0,0 +1,62 @@
...
+   * @param rowCount number of rows in the current batch
+   * @return a boolean indicates whether the verification passed or not, if not, we will log a
+   *     warning and report it to SF
+   */
+  boolean verify(
+      String prevBatchEndOffset,
+      String curBatchStartOffset,
+      String curBatchEndOffset,
+      long rowCount);
+}
```
sfc-gh-japatel, line 62: I am thinking out loud here and I dont have good suggestion. 
But, why do we need an interface here? I cant think of any more implementations of this method. 

I was mainly thinking of using something similar to a Predicate<E> where users can provide their custom implementation as this is too tied to 4 things accepted in this method.

src/main/java/net/snowflake/ingest/connection/TelemetryService.java
sfc-gh-tzhang: Good point, what about we create one RateLimiter per channel?

sfc-gh-japatel: yeah.. one per channel is okay!

src/main/java/net/snowflake/ingest/streaming/OffsetTokenVerificationFunction.java
sfc-gh-tzhang: Took a look and I'm not sure how you would use Predicate in this case, do you have an example? These 4 things are the only 4 things we track at batch level related to offset so I'm not sure what else you can do.

sfc-gh-japatel (approved):
lgtm! 
left a suggestion on ratelimiter.

src/main/java/net/snowflake/ingest/streaming/OffsetTokenVerificationFunction.java
sfc-gh-japatel: Yeah, I think its okay for us to go with this.. It's too tightly coupled with rowBuffer to make it more generalized.

src/main/java/net/snowflake/ingest/connection/TelemetryService.java
```diff
@@ -122,6 +127,26 @@ public void reportCpuMemoryUsage(Histogram cpuUsage) {
...
+      String channelName,
+      String prevBatchEndOffset,
+      String startOffset,
+      String endOffset,
+      long rowCount) {
+    if (rateLimiter.tryAcquire()) {
```
sfc-gh-japatel, line 136: also this is a blocking call, should we do this async or may be use 
tryAcquire(long timeout,
                  unit)
with a timeout?

sfc-gh-tzhang: Not sure if I understand, but tryAcquire() means the timeout is 0 and it will immediately return if it can't get it which is better than adding a timeout?

sfc-gh-japatel: got it, yeah this is better than adding a timeout.

PR #681, opened 2024-02-22
Add more detailed javadoc for the drop related api to warn about the …

sfc-gh-psaha:
…pitfalls

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
+++ b/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
@@ -82,6 +82,12 @@ public interface SnowflakeStreamingIngestChannel {
   /**
    * Close the channel, this function will make sure all the data in this channel is committed
    *
+   * <p>Note that this call with drop=true will delete <a
+   * href=https://docs.snowflake.com/en/user-guide/data-load-snowpipe-streaming-overview#offset-tokens>Offset
+   * Token</a> information from Snowflake servers unless the channel has already been opened in
+   * another client. So only use it if you are completely done ingesting data for this channel. If
+   * you open a channel with the same name in the future, it will behave like a new channel.
+   *
    * @param drop if true, the channel will be dropped after all data is successfully committed.
    * @return a completable future which will be completed when the channel is closed
    */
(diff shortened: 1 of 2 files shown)


src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
```diff
@@ -82,6 +82,12 @@ public interface SnowflakeStreamingIngestChannel {
   /**
    * Close the channel, this function will make sure all the data in this channel is committed
    *
+   * <p>Note that this call with drop=true will delete <a
+   * href=https://docs.snowflake.com/en/user-guide/data-load-snowpipe-streaming-overview#offset-tokens>Offset
+   * Token</a> and other state from Snowflake servers unless the channel has already been opened in
```
sfc-gh-xhuang, line 87: > unless the channel has already been opened in another client.

Can you explain this? Even if the channel was open in another client, it would drop it right?
Or was there a check that doesn't allow a non-client owner to drop?

src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClient.java
```diff
@@ -28,9 +28,15 @@ public interface SnowflakeStreamingIngestClient extends AutoCloseable {
...
-   * on close. That approach will drop the local version of the channel and if the channel has been
+   * <p>Note that this call will <em>blindly</em> drop the latest version of the channel and any
+   * pending data will be lost. It will also delete <a
+   * href=https://docs.snowflake.com/en/user-guide/data-load-snowpipe-streaming-overview#offset-tokens>Offset
+   * Token</a> and other state from Snowflake servers. So only use it if you are completely done
+   * ingesting data for this channel. If you open a channel with the same name in the future, it
+   * will behave like a new channel.
+   *
+   * <p>Also see {@link SnowflakeStreamingIngestChannel#close(boolean)} to drop channels on close.
+   * That approach will drop the local version of the channel and if the channel has been
```
sfc-gh-xhuang, line 39: I thought we only have 1 version of a channel regardless of which client opens?

sfc-gh-xhuang (approved):

src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClient.java
sfc-gh-psaha: Resolving based on our offline convo.

sfc-gh-tzhang (approved):

PR #682, opened 2024-02-23
V2.1.0 Release

sfc-gh-tzhang:
Let's do a minor version upgrade since:
- We upgrade the JDBC version
- We have a [change](https://github.com/snowflakedb/snowflake-ingest-java/pull/654) which might introduce some behavior changes


pom.xml
```diff
@@ -786,9 +786,9 @@
         <configuration>
           <errorRemedy>failFast</errorRemedy>
           <!--
-                               The list of allowed licenses. If you see the build failing due to "There are some forbidden licenses used, please
-                              check your dependencies", verify the conditions of the license and add the reference to it here.
-                            -->
+                                         The list of allowed licenses. If you see the build failing due to "There are some forbidden licenses used, please
```
sfc-gh-rcheng, line 789: nit: i believe that this change has happened in the past couple of releases, can we check that the formatter isn't doing anything strange?

sfc-gh-rcheng (approved):

sfc-gh-xhuang:
How does the zstd compression introduce a BCR? 
BDEC is an internal implementation

pom.xml
sfc-gh-tzhang: good question, I believe this is done by the plugin we use, I don't think the issue is in our logic

`mvn com.github.ekryd.sortpom:sortpom-maven-plugin:sort`

sfc-gh-tzhang:
> How does the zstd compression introduce a BCR? BDEC is an internal implementation

@sfc-gh-xhuang if you look at https://github.com/snowflakedb/snowflake-ingest-java/pull/654, it actually adds a new dependency and unshaded com.github.luben

sfc-gh-xhuang (approved):

PR #683, opened 2024-02-27
NO-SNOW Robustly determine tag tpcds setup

sfc-gh-lthiede:
Instead of expecting Jenkins to correctly identify a new tag, determine the latest tag using a python script and trigger a job to ingest the tpc-ds dataset using the tagged version. If the dataset already exists, the setup job doesn't reingest the dataset.

The trigger job runs for every push to master


scripts/latest_tag/requirements.txt
```diff
@@ -0,0 +1 @@
+snowflake-connector-python
```
sfc-gh-gdoci, line 1: discussed that this will be removed.

sfc-gh-gdoci (approved):

PR #684, opened 2024-02-27
[SNOW-988870] Update commons-compress version for vulnerability

sfc-gh-rcheng:
Received report for CVE-2024-26308 vulnerability on org.apache.commons:commons-compress, so upgrading to 1.26 as suggested


pom.xml
```diff
@@ -205,6 +205,10 @@
             <groupId>org.apache.avro</groupId>
             <artifactId>avro</artifactId>
           </exclusion>
+          <exclusion>
```
sfc-gh-tzhang, line 208: What's the reason we exclude this?

sfc-gh-tzhang (approved):
LGTM, please ask @sfc-gh-lsembera to take a look, thanks!

pom.xml
sfc-gh-lsembera: Let's not exclude from hadoop, but instead let's udpate also the hadoop transitive dependencies, which would violate [requireUpperBoundDeps](https://maven.apache.org/enforcer/enforcer-rules/requireUpperBoundDeps.html) enforcer rule. I opened a PR that updates all vulnerable dependencies: https://github.com/snowflakedb/snowflake-ingest-java/pull/693

sfc-gh-lsembera (changes requested):

PR #715, opened 2024-03-06
v2.1.0-unshaded release

sfc-gh-tzhang:


sfc-gh-xhuang (commented):
LGTM, lower priority item but we should figure out how to automate this such that when we release, we can do both at the same time

sfc-gh-xhuang (approved):

sfc-gh-tzhang:
Looks like the branch is created on master, let me create a new PR with the correct base

sfc-gh-tzhang:
https://github.com/snowflakedb/snowflake-ingest-java/pull/716 created

PR #719, opened 2024-03-21
SNOW-1255447: Fix leaked connection due to error response

sfc-gh-tzhang:
We're leaking HTTP connection when we get an error response from server, and if the client keeps retrying, we will run out connections from the connection pool. This PR contains the change to consume the HTTP response even under error case which will release the connection internally.


sfc-gh-japatel (approved):
Nice catch! Lgtm!

src/main/java/net/snowflake/ingest/connection/ServiceResponseHandler.java
```diff
@@ -263,11 +266,12 @@ private static HttpResponse handleExceptionalStatus(
           requestBuilder.addToken(request);
           response = httpClient.execute(request);
           if (!isStatusOK(response.getStatusLine())) {
```
sfc-gh-gdoci, line 268: Is the if-condition needed here? It looks like we are already inside the same condition (line 247).

sfc-gh-tzhang: ah, see line 262-265, we refresh the token and retry the request again

sfc-gh-xhuang (approved):

PR #721, opened 2024-03-21
SNOW-1000828 Expose specific error messages as part of the channel invalidation exception

sfc-gh-alhuang:
When a channel is invalidated, the original SDK does not throw a visible enough error. This PR makes SDK show more specific details for user when `INVALID_CHANNEL` is thrown.

[JIRA](https://snowflakecomputing.atlassian.net/browse/SNOW-1000828)

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
@@ -37,9 +37,9 @@ class ChannelCache<T> {
         channels.put(channel.getName(), channel);
     // Invalidate old channel if it exits to block new inserts and return error to users earlier
     if (oldChannel != null) {
-      String errorMessage =
+      String invalidationCause =
           String.format("Old channel removed from cache, channelName=%s", channel.getName());
-      oldChannel.invalidate("removed from cache", errorMessage);
+      oldChannel.invalidate("removed from cache", invalidationCause);
     }
   }
 
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
@@ -37,7 +37,9 @@ class ChannelCache<T> {
         channels.put(channel.getName(), channel);
     // Invalidate old channel if it exits to block new inserts and return error to users earlier
     if (oldChannel != null) {
-      oldChannel.invalidate("removed from cache");
+      String errorMessage =
+          String.format("Old channel removed from cache, channelName=%s", channel.getName());
+      oldChannel.invalidate("removed from cache", errorMessage);
     }
   }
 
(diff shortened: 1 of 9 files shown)


src/main/resources/net/snowflake/ingest/ingest_error_messages.properties
```diff
@@ -15,7 +15,7 @@
 0010=Missing {0} in config file.
 0011=Failed to upload blob.
 0012=Failed to cleanup resources during {0}.
-0013=Channel {0} is invalid and might contain uncommitted rows, please consider reopening the channel to restart.
+0013=Channel {0} is invalid and might contain uncommitted rows, please consider reopening the channel to restart. Channel invalidated cause: "{1}".
```
sfc-gh-xhuang, line 18: Should we say `Channel invalidation cause: {1}` ?

sfc-gh-alhuang: Make sense. Fixed.

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
```diff
@@ -82,14 +84,15 @@ void invalidateChannelIfSequencersMatch(
       String schemaName,
       String tableName,
       String channelName,
-      Long channelSequencer) {
+      Long channelSequencer,
+      String invalidationCause) {
```
sfc-gh-asen, line 88: nit: in some methods it is invalidationCause and others it becomes errorMessage, would it be possible to reuse the same name so it is easier to follow?

sfc-gh-asen (commented):
LGTM, just had a small nit about naming

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
```diff
@@ -37,7 +37,9 @@ void addChannel(SnowflakeStreamingIngestChannelInternal<T> channel) {
         channels.put(channel.getName(), channel);
     // Invalidate old channel if it exits to block new inserts and return error to users earlier
     if (oldChannel != null) {
-      oldChannel.invalidate("removed from cache");
+      String errorMessage =
```
sfc-gh-tzhang, line 40: Removing from cache is not an error, maybe let's call it invalidationReason or invalidationCause?

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -657,7 +657,7 @@ String getBlobPath(Calendar calendar, String clientPrefix) {
    *
    * @param blobData list of channels that belongs to the blob
    */
-  <CD> void invalidateAllChannelsInBlob(List<List<ChannelData<CD>>> blobData) {
+  <CD> void invalidateAllChannelsInBlob(List<List<ChannelData<CD>>> blobData, String errorMessage) {
```
sfc-gh-tzhang, line 660: invalidationReason or invalidationCause instead of errorMessage?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
```diff
@@ -58,6 +58,9 @@ class SnowflakeStreamingIngestChannelInternal<T> implements SnowflakeStreamingIn
   // Internal map of column name -> column properties
   private final Map<String, ColumnProperties> tableColumns;
 
+  // The cause of channel invalidation
+  private String invalidationCause;
```
sfc-gh-tzhang, line 62: Could we make it clear in the comment that this is the latest cause that we see since a channel  could get invalidated multiple times?

sfc-gh-tzhang (approved):
Left some minor comments, otherwise LGTM, thanks!

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
sfc-gh-tzhang: Agree that we should use invalidationCause or invalidationReason in the function signature, but sometimes a channel gets invalidated is because of an error, so it's ok to call it errorMessage for such cases when passing it into the function

sfc-gh-asen (approved):
Thanks!

sfc-gh-tjones (approved):

PR #731, opened 2024-03-26
SNOW-1057812: Support Open Rowset API Channel

sfc-gh-tzhang:
This PR contains the change to allow creation of a Rowset API channel:
- No behavior change on existing CLOUD_STORAGE interface
- Add an option to existing OpenChannelRequest to allow the creation of a Rowset API channel with ChannelType=ROWSET_API
- The OpenChannelRequest will be converted internally to a new REST API added for Rowset API, and we have a new implementation of the channel class (SnowflakeStreamingIngestChannelRowset) which is much simpler than the one for CLOUD_STORAGE

Note that we're merging into a feature branch instead of master


src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
```diff
@@ -247,6 +262,8 @@ public interface SnowflakeStreamingIngestChannel {
    */
   InsertValidationResponse insertRow(Map<String, Object> row, @Nullable String offsetToken);
 
+  InsertValidationResponse insertRow(Map<String, Object> row);
```
sfc-gh-asen, line 265: in which cases will a customers use insertRow without an offsetToken?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelRowset.java
```diff
@@ -0,0 +1,280 @@
...
+            "%s.%s.%s.%s", dbName, schemaName, pipeName == null ? tableName : pipeName, name);
+    this.offsetToken = offsetToken;
+    this.continuationToken = continuationToken;
+  }
+
+  /**
+   * Get the fully qualified channel name
+   *
+   * @return fully qualified name of the channel, in the format of
+   *     dbName.schemaName.tableName.channelName
```
sfc-gh-asen, line 64: nit : or dbName.schemaName.pipeName.channelName

src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
```diff
@@ -271,6 +288,8 @@ InsertValidationResponse insertRows(
   InsertValidationResponse insertRows(
       Iterable<Map<String, Object>> rows, @Nullable String offsetToken);
```
sfc-gh-asen, line 289: we might need to add something like

 insertRows(Iterable<Map<String, Object>> rows,  List<String> offsetTokens)
 
 for partial ingestion

src/main/java/net/snowflake/ingest/utils/Utils.java
```diff
@@ -88,26 +90,25 @@ private static Provider instantiateSecurityProvider() {
...
 
-  /**
-   * Assert when the value is null
-   *
-   * @param name
-   * @param value
-   * @throws SFException
-   */
+  /** Assert when all the Strings are either null or Empty */
+  public static void assertStringsNotNullOrEmpty(String name, Collection<String> values)
```
sfc-gh-asen, line 101: nit: when I first saw this method, I assumed this was asserting that all the strings in the collection should be not null, maybe we can rename it to something like assertAtLeastOneStringNotNullOrEmpty to be extra clear?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
```diff
@@ -441,6 +472,17 @@ public Map<String, ColumnProperties> getTableSchema() {
...
+   * Get the type of channel, please referring to {@link
+   * net.snowflake.ingest.streaming.OpenChannelRequest.ChannelType} for the supported channel type
+   *
+   * @return type of the channel
+   */
+  @Override
+  public OpenChannelRequest.ChannelType getType() {
+    return OpenChannelRequest.ChannelType.CLOUD_STORAGE;
+  }
+
```
sfc-gh-asen, line 485: Should we rename this class to SnowflakeStreamingIngestChannelCloud to be parallel with the rowset?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -299,76 +305,162 @@ public SnowflakeStreamingIngestChannelInternal<?> openChannel(OpenChannelRequest
...
 
-      return channel;
-    } catch (IOException | IngestResponseException e) {
-      throw new SFException(e, ErrorCode.OPEN_CHANNEL_FAILURE, e.getMessage());
-    }
+    // Channel is now registered, add it to the in-memory channel pool
+    return SnowflakeStreamingIngestChannelFactory.<T>builder(request.getChannelName())
+        .setDBName(request.getDBName())
+        .setSchemaName(request.getSchemaName())
+        .setTableName(request.getTableName())
```
sfc-gh-asen, line 455: we need to setPipeName as well

sfc-gh-asen (commented):
Thanks for the changes

src/main/java/net/snowflake/ingest/streaming/OpenChannelRequest.java
```diff
@@ -16,6 +17,11 @@ public enum OnErrorOption {
     // indexes. No data is ingested
   }
 
+  public enum ChannelType {
+    CLOUD_STORAGE,
+    ROWSET_API,
+  }
```
sfc-gh-tjones, line 23: I think this exists already? If you look at `StreamingResource` there's a check somewhere that compares against an enum.

src/main/java/net/snowflake/ingest/streaming/OpenChannelRequest.java
```diff
@@ -114,19 +135,23 @@ private OpenChannelRequest(OpenChannelRequestBuilder builder) {
     Utils.assertStringNotNullOrEmpty("channel name", builder.channelName);
     Utils.assertStringNotNullOrEmpty("database name", builder.dbName);
     Utils.assertStringNotNullOrEmpty("schema name", builder.schemaName);
-    Utils.assertStringNotNullOrEmpty("table name", builder.tableName);
+    Utils.assertStringsNotNullOrEmpty(
```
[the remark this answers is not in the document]
sfc-gh-tjones: Yeah this is a good edge case check to add.

src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
sfc-gh-tjones: Yeah shouldn't this be ` InsertValidationResponse insertRow(Map<String, Object> row, @Nullable String offsetToken);`?

src/main/java/net/snowflake/ingest/streaming/OpenChannelRequest.java
sfc-gh-tzhang: Yes, but the one used before was hided in the SDK, now this is a public enum in the public class

src/main/java/net/snowflake/ingest/streaming/OpenChannelRequest.java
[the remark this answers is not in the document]
sfc-gh-tzhang: The behavior right now is that the table name will be ignored if there is a pipe name, I thought about assert if both are provided but I think it's too restricted, it's perfect valid if customer inputs both and pipe is created on that table

src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
sfc-gh-tzhang: remember that we say the offset_token will be per row for Rowset API?

src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
sfc-gh-tzhang: Yes, that's the one below, offset_token will be part of the row now.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
sfc-gh-tzhang: Yes, I can do that in a separate PR

Issue #671, opened 2024-01-30
Move to using snowjdbc thin jar instead of the fat jar.


sfc-gh-psaha:
See https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/820 for detailed rationale.

sfc-gh-tzhang:
We switch to use jdbc-thin jar since 4.0.0

PR #733, opened 2024-04-09
SNOW-1257851 Move to using snowjdbc thin jar instead of the fat jar

sfc-gh-alhuang:
https://github.com/snowflakedb/snowflake-ingest-java/issues/671, move from jdbc fat jar to thin jar.

JIRA: 
https://snowflakecomputing.atlassian.net/browse/SNOW-1257851
https://snowflakecomputing.atlassian.net/browse/SNOW-818172


sfc-gh-tzhang:
@sfc-gh-lsembera @sfc-gh-alhuang I want to revisit our decision of **NOT shading the JDBC dependency** after this change, we have seen too many issues that customer is using an older version of JDBC with the latest SDK and something is not working due to bugs in the older versions of JDBC. Usually these issues are hard to debug as well since it requires JDBC logs and we don't know if the issues are in SDK or JDBC. I believe one of the main reason to not shading was due to the user jar size, which will be addressed in this PR, WDYT?

pom.xml
```diff
@@ -78,11 +84,31 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+      <dependency>
```
sfc-gh-lsembera, line 87: Why do we need to pull all these and not get them transitively via jdbc-thin?

scripts/process_licenses.py
```diff
@@ -61,6 +61,64 @@
     "org.bouncycastle:bcpkix-jdk18on": BOUNCY_CASTLE_LICENSE,
     "org.bouncycastle:bcutil-jdk18on": BOUNCY_CASTLE_LICENSE,
     "org.bouncycastle:bcprov-jdk18on": BOUNCY_CASTLE_LICENSE,
+    "com.amazonaws:aws-java-sdk-core": APACHE_LICENSE,
```
sfc-gh-lsembera, line 64: Did you check all of these don't ship with a license?

sfc-gh-lsembera (approved):
This looks fantastic, thanks @sfc-gh-alhuang! I added a few minor comments.

Since this is a huge PR, and it may have an impact on customers' setup, we should release it as 2.2.0. Before releasing, I'd also suggest to first upgrade everything internally to snapshot version from nexus to be 100% sure there are no regressions.

pom.xml
sfc-gh-alhuang: For `com.amazonaws:aws-java-sdk-core`, I got `Used undeclared dependencies found` error from `maven-dependency-plugin` if it's not explicitly specified in dependencies.

sfc-gh-alhuang: For other dependencies like ` com.google.code.gson`, `Failed while enforcing releasability.` errors is raised by `maven-enforcer-plugin` if it's not specified in dependencies.

scripts/process_licenses.py
sfc-gh-alhuang: Yes, I checked the artifact directory to see if there's a license file. If not, I used the output from `license:add-third-party` directly.

sfc-gh-japatel:
May be this is already discussed, but can we confirm from jdbc team this is okay to use in production? 
https://docs.snowflake.com/en/release-notes/clients-drivers/jdbc-2024
release notes says, it is an experimental feature - not sure what it means. 

but if we have confirmed this is okay to use, please ignore my above comment.

sfc-gh-tzhang:
> May be this is already discussed, but can we confirm from jdbc team this is okay to use in production? https://docs.snowflake.com/en/release-notes/clients-drivers/jdbc-2024 release notes says, it is an experimental feature - not sure what it means -/ (PrPr?)
> 
> but if we have confirmed this is okay to use, please ignore my above comment.

I believe we did check with them, but I didn't see the note, it's always good to double check again :)

sfc-gh-tzhang:
Thanks @sfc-gh-alhuang ! No concerns from me as long as @sfc-gh-lsembera signed off, could you wait for release 2.1.1 to go out before merging this change?

sfc-gh-lsembera:
@sfc-gh-alhuang @sfc-gh-tzhang Are we going to merge this one? If so, we should do it early, so we can internally switch to the new SNAPSHOT version, so it gets thoroughly tested before the release.

sfc-gh-dhandermann:
Closing based on implementation changes merged in #947. Thanks for providing the foundation in this pull request @sfc-gh-alhuang!

PR #739, opened 2024-04-15
SNOW-1167007: Validate no null values for not null cols

sfc-gh-wtrefon:
Validate if row contains null values for non-nullable columns. If yes, add such column names to InsertValidationResponse error as _NullValueForNotNullColNames_ so it can be used for schema evolution.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
@@ -390,7 +390,7 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
    * Verify that the input row columns are all valid.
    *
    * <p>Checks that the columns, specified in the row, are present in the table and values for all
-   * non-nullable columns are specified.
+   * non-nullable columns are specified and different from null.
    *
    * @param row the input row
    * @param error the insert error that we return to the customer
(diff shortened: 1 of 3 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -442,6 +442,26 @@ Set<String> verifyInputColumns(
...
     }
 
+    // Check for null values in not-nullable columns in the row
+    List<String> nullValueNotNullCols = new ArrayList<>();
+    for (String columnName : this.nonNullableFieldNames) {
+      if (inputColNamesMap.containsKey(columnName)
+          && row.get(inputColNamesMap.get(columnName)) == null) {
+        nullValueNotNullCols.add(statsMap.get(columnName).getColumnDisplayName());
+      }
+    }
```
sfc-gh-tzhang, line 452: We already throw an exception at the end of `parseColumnValueToParquet` when we insert null into a non-nullable column, should be just catch that and add the col name to the nullValueNotNullCols list?

sfc-gh-wtrefon: I'm not sure if it is the right place. The `verifyInputColumns` is called before and sounds like the right place for such validation. `verifyInputColumns` is also in the abstract class, the `parseColumnValueToParquet` only in one implementation (Parquet).

If we catched the exception in insertRows function we would only add single column to the nullValueNotNullCols even if there is more columns with null values. We would need to pass the error object into the `addRow` function to catch all the exception that may be thrown in a loop.

Therefore i think the `verifyInputColumns` is much better place for the validation

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -442,6 +442,26 @@ Set<String> verifyInputColumns(
               "Values for all non-nullable columns must be specified, rowIndex:%d", rowIndex));
     }
 
+    // Check for null values in not-nullable columns in the row
+    List<String> nullValueNotNullCols = new ArrayList<>();
```
sfc-gh-lshcharbaty, line 446: Does it make sense to differentiate between `missingCols` and `nullValueNotNullCols`?
Is there a conceptual difference between the cases when a column is present with `null` value and when it isn't present at all.

sfc-gh-wtrefon: i was thinking about this, the only thing is the naming of missingCols. But if we do like this than we don't need to adapt anything in kafka push code. What u think @sfc-gh-tzhang ?

sfc-gh-tzhang: I'm fine with combing it and we will need to update the name to make it clear, the only concern is that we're updating a public API, but I highly doubt any customer is using this.

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
sfc-gh-tzhang: I see, makes sense to me, my concern was that in the worst case, we're iterating through all the data again to check this which is not ideal so I was hoping that we can do this as part of data validation.

sfc-gh-tzhang (approved):
LGTM, thanks!

PR #759, opened 2024-05-01
SNOW-1357377 Add request Id in all streaming ingest APIs

sfc-gh-japatel:
- Add requestId as a Query Parameter. 
- Remove requestId from payload since it is not used anywhere on server side. 

Along with this, our response should also include requestId but that can come later since they can still rely on offsetToken API. (Wont work for other APIs but there hasn't been any request)

This is for internal logging purposes.

### DIFF
--- a/src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
+++ b/src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
@@ -678,18 +678,23 @@ public class RequestBuilder {
    */
   public HttpPost generateStreamingIngestPostRequest(
       String payload, String endPoint, String message) {
-    LOGGER.debug("Generate Snowpipe streaming request: endpoint={}, payload={}", endPoint, payload);
+    final String requestId = UUID.randomUUID().toString();
+    LOGGER.debug(
+        "Generate Snowpipe streaming request: endpoint={}, payload={}, requestId={}",
+        endPoint,
+        payload,
+        requestId);
     // Make the corresponding URI
     URI uri = null;
     try {
       uri =
           new URIBuilder()
-                  .setScheme(scheme)
-                  .setHost(host)
-                  .setPort(port)
-                  .setPath(endPoint)
-                  .setParameter(REQUEST_ID, UUID.randomUUID().toString())
-                  .build();
+              .setScheme(scheme)
+              .setHost(host)
+              .setPort(port)
+              .setPath(endPoint)
+              .setParameter(REQUEST_ID, requestId)
+              .build();
     } catch (URISyntaxException e) {
       throw new SFException(e, ErrorCode.BUILD_REQUEST_FAILURE, message);
     }
--- a/src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
+++ b/src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
@@ -683,7 +683,13 @@ public class RequestBuilder {
     URI uri = null;
     try {
       uri =
-          new URIBuilder().setScheme(scheme).setHost(host).setPort(port).setPath(endPoint).build();
+          new URIBuilder()
+                  .setScheme(scheme)
+                  .setHost(host)
+                  .setPort(port)
+                  .setPath(endPoint)
+                  .setParameter(REQUEST_ID, UUID.randomUUID().toString())
+                  .build();
     } catch (URISyntaxException e) {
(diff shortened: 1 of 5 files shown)


sfc-gh-asen (approved):
Thanks!

src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
```diff
@@ -683,7 +683,13 @@ public HttpPost generateStreamingIngestPostRequest(
     URI uri = null;
     try {
       uri =
-          new URIBuilder().setScheme(scheme).setHost(host).setPort(port).setPath(endPoint).build();
+          new URIBuilder()
+                  .setScheme(scheme)
+                  .setHost(host)
+                  .setPort(port)
+                  .setPath(endPoint)
+                  .setParameter(REQUEST_ID, UUID.randomUUID().toString())
```
sfc-gh-rcheng, line 691: should we log this requestid?

sfc-gh-rcheng (approved):
lgtm! small question on if we want to log the requestid

src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
sfc-gh-japatel: Valid point. added requestId in above debug log! we dont need everywhere IMO. this requestId is mainly for internal debugging.

src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
```diff
@@ -678,12 +678,23 @@ public HttpGet generateHistoryRangeRequest(
    */
   public HttpPost generateStreamingIngestPostRequest(
       String payload, String endPoint, String message) {
-    LOGGER.debug("Generate Snowpipe streaming request: endpoint={}, payload={}", endPoint, payload);
+    final String requestId = UUID.randomUUID().toString();
```
sfc-gh-tzhang, line 681: My assumption is that the request id will remain the same with retry?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -493,7 +493,6 @@ ChannelsStatusResponse getChannelsStatus(
               .collect(Collectors.toList());
       request.setChannels(requestDTOs);
       request.setRole(this.role);
-      request.setRequestId(this.flushService.getClientPrefix() + "_" + counter.getAndIncrement());
```
sfc-gh-tzhang, line 496: We should remove request_id in the request as well

src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
sfc-gh-japatel: thats true!

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
sfc-gh-japatel: you mean in the POJO?

sfc-gh-tzhang: Yes

src/test/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelTest.java
```diff
@@ -1,12 +1,7 @@
 package net.snowflake.ingest.streaming.internal;
 
 import static java.time.ZoneOffset.UTC;
-import static net.snowflake.ingest.utils.Constants.ACCOUNT_URL;
-import static net.snowflake.ingest.utils.Constants.OPEN_CHANNEL_ENDPOINT;
-import static net.snowflake.ingest.utils.Constants.PRIVATE_KEY;
-import static net.snowflake.ingest.utils.Constants.RESPONSE_SUCCESS;
-import static net.snowflake.ingest.utils.Constants.ROLE;
-import static net.snowflake.ingest.utils.Constants.USER;
+import static net.snowflake.ingest.utils.Constants.*;
```
sfc-gh-tzhang, line 4: no star import

sfc-gh-tzhang (approved):
Let's remove the request_id in the ChannelsStatusRequest class, otherwise LGTM, thanks

PR #760, opened 2024-05-02
Add error log for flush task scheduling

sfc-gh-alhuang:
Add error log for flush task scheduling. This could happen when [getCachedMaxClientLagInMs](https://github.com/snowflakedb/snowflake-ingest-java/blob/eb87281a6a1f5ee3e2e959594bb571b40b770b90/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java#L268C65-L268C92) throw cast exception when provided parameter is `int` not `long`.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
@@ -305,7 +305,8 @@ class FlushService<T> {
           } catch (Exception e) {
             String errorMessage =
                 String.format(
-                    "Failed to schedule a flush task, client=%s, exception=%s, detail=%s, trace=%s.",
+                    "Failed to schedule a flush task, client=%s, exception=%s, detail=%s,"
+                        + " trace=%s.",
                     this.owningClient.getName(),
                     e.getClass().getName(),
                     e.getMessage(),
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
@@ -312,6 +312,11 @@ class FlushService<T> {
                     e.getMessage(),
                     getStackTrace(e));
             logger.logError(errorMessage);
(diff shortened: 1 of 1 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -282,25 +282,37 @@ private void createWorkers() {
...
+          } catch (Exception e) {
+            String errorMessage =
+                String.format(
+                    "Failed to schedule a flush task, client=%s, exception=%s, detail=%s,"
+                        + " trace=%s.",
+                    this.owningClient.getName(),
+                    e.getClass().getName(),
+                    e.getMessage(),
+                    getStackTrace(e));
+            logger.logError(errorMessage);
```
sfc-gh-tzhang, line 314: Have you verify that we're able to see the error in the log?

sfc-gh-tzhang: Also could we re-throw it in the end?

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -282,25 +282,37 @@ private void createWorkers() {
...
+          }
```
sfc-gh-tzhang, line 315: Consider also send it to SF similar to the logic above
```
                      if (this.owningClient.getTelemetryService() != null) {
                        this.owningClient
                            .getTelemetryService()
                            .reportClientFailure(this.getClass().getSimpleName(), errorMessage);
                      }
```

sfc-gh-tzhang (approved):
Left some comments, otherwise lgtm, thanks!

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-alhuang: Verified and added re-throw logic.

PR #763, opened 2024-05-23
SNOW-1437885  Disable blob interleaving when in Iceberg mode

sfc-gh-alhuang:
* Added a method to `SnowflakeStreamingIngestClientInternal` to activate/deactivate streaming to Iceberg tables.
* A blob should only contain one chunk (parquet) under Iceberg mode. Disable blob interleaving under Iceberg mode.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClientFactory.java
+++ b/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClientFactory.java
@@ -49,7 +49,7 @@ public class SnowflakeStreamingIngestClientFactory {
       return this;
     }
 
-    public Builder setIcebergMode(boolean isIcebergMode) {
+    public Builder setIsIceberg(boolean isIcebergMode) {
       this.isIcebergMode = isIcebergMode;
       return this;
     }
--- a/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClientFactory.java
+++ b/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClientFactory.java
@@ -28,6 +28,9 @@ public class SnowflakeStreamingIngestClientFactory {
     // Allows client to override some default parameter values
     private Map<String, Object> parameterOverrides;
 
+    // Indicates whether it's streaming to Iceberg tables
+    private boolean isIcebergMode;
+
     // Indicates whether it's under test mode
     private boolean isTestMode;
 
@@ -45,6 +48,11 @@ public class SnowflakeStreamingIngestClientFactory {
       return this;
     }
 
+    public Builder setIcebergMode(boolean isIcebergMode) {
+      this.isIcebergMode = isIcebergMode;
+      return this;
+    }
+
     public Builder setIsTestMode(boolean isTestMode) {
       this.isTestMode = isTestMode;
       return this;
@@ -58,7 +66,12 @@ public class SnowflakeStreamingIngestClientFactory {
       SnowflakeURL accountURL = new SnowflakeURL(prop.getProperty(Constants.ACCOUNT_URL));
 
       return new SnowflakeStreamingIngestClientInternal<>(
-          this.name, accountURL, prop, this.parameterOverrides, this.isTestMode);
+          this.name,
+          accountURL,
+          prop,
+          this.parameterOverrides,
+          this.isIcebergMode,
+          this.isTestMode);
     }
   }
 }
--- a/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClientFactory.java
+++ b/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClientFactory.java
@@ -28,7 +28,8 @@ public class SnowflakeStreamingIngestClientFactory {
     // Allows client to override some default parameter values
     private Map<String, Object> parameterOverrides;
 
-    // Indicates whether it's streaming to Iceberg tables
+    // Indicates whether it's streaming to Iceberg tables. Open channels on regular tables should
+    // fail in this mode.
     private boolean isIcebergMode;
 
     // Indicates whether it's under test mode
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
@@ -123,10 +123,6 @@ class FlushService<T> {
   // blob encoding version
   private final Constants.BdecVersion bdecVersion;
 
-  // Indicates if it's flushing to Iceberg tables, a blob could only contain one chunk under Iceberg
-  // mode
-  private final boolean isIcebergMode;
-
   /**
    * Constructor for TESTING that takes (usually mocked) StreamingIngestStage
    *
@@ -138,7 +134,6 @@ class FlushService<T> {
       SnowflakeStreamingIngestClientInternal<T> client,
       ChannelCache<T> cache,
       StreamingIngestStage targetStage, // For testing
-      boolean isIcebergMode,
       boolean isTestMode) {
     this.owningClient = client;
     this.channelCache = cache;
@@ -147,7 +142,6 @@ class FlushService<T> {
     this.registerService = new RegisterService<>(client, isTestMode);
     this.isNeedFlush = false;
     this.lastFlushTime = System.currentTimeMillis();
-    this.isIcebergMode = isIcebergMode;
     this.isTestMode = isTestMode;
     this.latencyTimerContextMap = new ConcurrentHashMap<>();
     this.bdecVersion = this.owningClient.getParameterProvider().getBlobFormatVersion();
@@ -162,10 +156,7 @@ class FlushService<T> {
    * @param isTestMode
    */
   FlushService(
-      SnowflakeStreamingIngestClientInternal<T> client,
-      ChannelCache<T> cache,
-      boolean isIcebergMode,
-      boolean isTestMode) {
+      SnowflakeStreamingIngestClientInternal<T> client, ChannelCache<T> cache, boolean isTestMode) {
     this.owningClient = client;
     this.channelCache = cache;
     try {
@@ -185,7 +176,6 @@ class FlushService<T> {
     this.counter = new AtomicLong(0);
     this.isNeedFlush = false;
     this.lastFlushTime = System.currentTimeMillis();
-    this.isIcebergMode = isIcebergMode;
     this.isTestMode = isTestMode;
     this.latencyTimerContextMap = new ConcurrentHashMap<>();
     this.bdecVersion = this.owningClient.getParameterProvider().getBlobFormatVersion();
@@ -447,7 +437,7 @@ class FlushService<T> {
           }
           // Add processed channels to the current blob, stop if we need to create a new blob
           blobData.add(channelsDataPerTable.subList(0, idx));
-          if (idx != channelsDataPerTable.size() || isIcebergMode) {
+          if (idx != channelsDataPerTable.size()) {
             break;
           }
         }
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
@@ -123,6 +123,10 @@ class FlushService<T> {
   // blob encoding version
   private final Constants.BdecVersion bdecVersion;
 
+  // Indicates if it's flushing to Iceberg tables, a blob could only contain one chunk under Iceberg
+  // mode
+  private final boolean isIcebergMode;
+
   /**
    * Constructor for TESTING that takes (usually mocked) StreamingIngestStage
    *
@@ -134,6 +138,7 @@ class FlushService<T> {
       SnowflakeStreamingIngestClientInternal<T> client,
       ChannelCache<T> cache,
       StreamingIngestStage targetStage, // For testing
+      boolean isIcebergMode,
       boolean isTestMode) {
     this.owningClient = client;
     this.channelCache = cache;
@@ -142,6 +147,7 @@ class FlushService<T> {
     this.registerService = new RegisterService<>(client, isTestMode);
     this.isNeedFlush = false;
     this.lastFlushTime = System.currentTimeMillis();
+    this.isIcebergMode = isIcebergMode;
     this.isTestMode = isTestMode;
     this.latencyTimerContextMap = new ConcurrentHashMap<>();
     this.bdecVersion = this.owningClient.getParameterProvider().getBlobFormatVersion();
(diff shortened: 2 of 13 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -123,6 +123,10 @@ List<List<ChannelData<T>>> getData() {
   // blob encoding version
   private final Constants.BdecVersion bdecVersion;
 
+  // Indicates if it's flushing to Iceberg tables, a blob could only contain one chunk under Iceberg
+  // mode
+  private final boolean isIcebergMode;
```
sfc-gh-gdoci, line 128: Can this have a more general name, e.g. `isNonInterleavedMode`?

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -437,7 +447,7 @@ && shouldStopProcessing(
           }
           // Add processed channels to the current blob, stop if we need to create a new blob
           blobData.add(channelsDataPerTable.subList(0, idx));
-          if (idx != channelsDataPerTable.size()) {
+          if (idx != channelsDataPerTable.size() || isIcebergMode) {
```
sfc-gh-hmadan, line 450: I want to guard against the possibility of having isIcebergMode peppered across 10 different places in code, making it that much harder to maintain the client.

Ideally we should have one place in the whole client that does `if (isIcebergMode) { /* initialize booleans / ints / settings one way */ } else { /* initialize another way */ }` - it seems that the ParameterProvider is the best place to do this, or we introduce a separate ClientSettings class that holds all these settings that control different behaviors in different places in the client.

I also see there are three parameters on the parameter provider that need different values for iceberg - MAX_CHUNK_SIZE_IN_BYTES (set to 512 MB rn) and MAX_CHUNKS_IN_BLOB_AND_REGISTRATION_REQUEST_DEFAULT (set to 100 rn) and MAX_BLOB_SIZE_IN_BYTES (set to 1 GB rn). This is in addition to MAX_CLIENT_LAG needing an iceberg-specific value.

It looks like if parameterProvider.getMaxChunksInBlobAndRegistrationRequest() returns 1 then you'll get this same desired behavior ?

Note - in the parameterProvider you'll need to validate that customers can't override MAX_CHUNKS_IN_BLOB_AND_REGISTRATION_REQUEST and it can only ever be 1.

src/test/java/net/snowflake/ingest/streaming/internal/FlushServiceTest.java
```diff
@@ -422,7 +427,7 @@ public void testGetFilePath() {
 
   @Test
   public void testFlush() throws Exception {
-    TestContext<?> testContext = testContextFactory.create();
+    TestContext<?> testContext = testContextFactory.create(false);
```
sfc-gh-hmadan, line 430: as suggested in another thread please using JUnit parameters / somehow parameterize this isIcebergMode so that all tests run with isIcebergMode=on/off, instead of only testing the non-iceberg usecases.

feel free to close this thread out so we just have one thread on this!

sfc-gh-hmadan (changes requested):
Left a couple comments, PTAL.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-tzhang: Yeah, we can make this more general for the internal names since we disable interleaved mode due to other reasons in the future, not only Iceberg

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -258,16 +265,17 @@ public SnowflakeStreamingIngestClientInternal(
...
+    this(name, accountURL, prop, null, isIcebergMode, isTestMode, null, parameterOverrides);
   }
 
   /*** Constructor for TEST ONLY
    *
    * @param name the name of the client
    */
   SnowflakeStreamingIngestClientInternal(String name) {
-    this(name, null, null, null, true, null, new HashMap<>());
+    this(name, null, null, null, false, true, null, new HashMap<>());
```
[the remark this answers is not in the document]
sfc-gh-tzhang: agree, you can search for `Parameterized` and see how to use that

sfc-gh-tzhang (commented):
I think there're a few things missing in this PR:
- Block creation of regular channels on Iceberg client and creation of iceberg channels on regular client
- No need to call the client/configure endpoint for Iceberg client

Do you plan to do that in a separate PR?

src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClientFactory.java
```diff
@@ -28,6 +28,9 @@ public static class Builder {
     // Allows client to override some default parameter values
     private Map<String, Object> parameterOverrides;
 
+    // Indicates whether it's streaming to Iceberg tables
+    private boolean isIcebergMode;
```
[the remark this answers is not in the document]
sfc-gh-hmadan: I am adding this to the server side validation in another PR already; do we typically do open-channel validations on both sides?

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-alhuang: Removed the non-interleave parameter for now as we could control this via `MAX_CHUNKS_IN_BLOB_AND_REGISTRATION_REQUEST`

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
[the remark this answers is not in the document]
sfc-gh-alhuang: Added parameters for unit tests.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-alhuang: Deleted the `isIcebergMode` param and use `MAX_CHUNKS_IN_BLOB_AND_REGISTRATION_REQUEST` for now.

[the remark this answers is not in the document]
sfc-gh-alhuang: Removed the non-interleave parameter for now as we could control this via MAX_CHUNKS_IN_BLOB_AND_REGISTRATION_REQUEST

sfc-gh-alhuang:
Thanks for the comments and tips! Added `icebergMode` on/off params for some unit tests. Some tests included blob registration / authentication only test with non regular mode for now. @sfc-gh-tzhang I plan to delete/modify the client configure call in later PRs as it's only called by `StreamingIngestStage` and we might not using this class in the iceberg mode later. The Iceberg client/server check is done in the server side ([ref](https://github.com/snowflakedb/snowflake-ingest-java/pull/763#discussion_r1617601137)), should we also added check in the client?

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -188,6 +200,13 @@ private void setParameterMap(Map<String, Object> parameterOverrides, Properties
         BDEC_PARQUET_COMPRESSION_ALGORITHM_DEFAULT,
         parameterOverrides,
         props);
+
+    // Required parameter override for Iceberg mode
+    if (this.isIcebergMode) {
+      this.parameterMap.put(
+          MAX_CHUNKS_IN_BLOB_AND_REGISTRATION_REQUEST,
```
sfc-gh-hmadan, line 207: why not call this.updateValue here ? That's the only method that's updating parameterMap before this block was added, lets keep it that way..?

sfc-gh-hmadan (approved):
LGTM, left one small comment that you can take in the next PR too as you have other changes waiting for this to go in.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-alhuang: The `updateValue` considers props and paramOverrides before default value. While under iceberg mode we should not allow user to set the chunk number > 1, that's why I use the `put` method directly. I think we can use `updateValue(key, value, null, null)` to achieve the same result, wdyt?

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -64,6 +64,12 @@ public class ParameterProvider {
   public static final Constants.BdecParquetCompression BDEC_PARQUET_COMPRESSION_ALGORITHM_DEFAULT =
       Constants.BdecParquetCompression.GZIP;
 
+  // Iceberg mode default parameters
+  public static final int MAX_CHUNKS_IN_BLOB_AND_REGISTRATION_REQUEST_ICEBERG_MODE_DEFAULT = 1;
+
+  // If the provided parameters need to be verified and modified to meet Iceberg mode
+  private final boolean isIcebergMode;
```
sfc-gh-tzhang, line 71: what happens if customer create a Iceberg client but override this value to false?

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -64,6 +64,12 @@ public class ParameterProvider {
   public static final Constants.BdecParquetCompression BDEC_PARQUET_COMPRESSION_ALGORITHM_DEFAULT =
       Constants.BdecParquetCompression.GZIP;
 
+  // Iceberg mode default parameters
+  public static final int MAX_CHUNKS_IN_BLOB_AND_REGISTRATION_REQUEST_ICEBERG_MODE_DEFAULT = 1;
```
sfc-gh-tzhang, line 68: This sounds like we're allowing to change it to something other than 1, but it has to be one for Iceberg, right? We should put this into Constants.java, everything in this file means that we're allow to override it.

src/test/java/net/snowflake/ingest/streaming/internal/FlushServiceTest.java
```diff
@@ -48,11 +48,22 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
 import org.mockito.ArgumentCaptor;
 import org.mockito.ArgumentMatchers;
 import org.mockito.Mockito;
 
+@RunWith(Parameterized.class)
 public class FlushServiceTest {
+
+  @Parameterized.Parameters(name = "isIcebergMode: {0}")
```
sfc-gh-tzhang, line 60: I thought what we should be doing is to test the isIcebergMode with {true, false}

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-alhuang: I don't think client can inject `ParameterProvider` directly, instantiate is called by `ClientInternal` at [here](https://github.com/snowflakedb/snowflake-ingest-java/blob/6c9bc9b3e3b9bd687898aeb59cb0d30940b4c8c9/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java#L172).

src/test/java/net/snowflake/ingest/streaming/internal/FlushServiceTest.java
sfc-gh-alhuang: This line is the format of the test name, where the first parameter is either `false` or `true` from [here](https://github.com/snowflakedb/snowflake-ingest-java/blob/6c9bc9b3e3b9bd687898aeb59cb0d30940b4c8c9/src/test/java/net/snowflake/ingest/streaming/internal/FlushServiceTest.java#L62).

src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClientFactory.java
```diff
@@ -45,6 +49,11 @@ public Builder setParameterOverrides(Map<String, Object> parameterOverrides) {
       return this;
     }
 
+    public Builder setIcebergMode(boolean isIcebergMode) {
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Renamed.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-alhuang: Yes. Moved to `Constants.java`.

sfc-gh-alhuang:
Per offline discussion with @sfc-gh-tzhang , moved `MAX_CHUNKS_IN_BLOB_AND_REGISTRATION_REQUEST_ICEBERG_MODE_DEFAULT` back to ParameterProvider.java.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -258,16 +264,17 @@ public SnowflakeStreamingIngestClientInternal(
       SnowflakeURL accountURL,
       Properties prop,
       Map<String, Object> parameterOverrides,
+      boolean isIcebergMode,
```
sfc-gh-tzhang, line 267: Let's make sure this is being added to the comment as well

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -64,6 +66,9 @@ public class ParameterProvider {
   public static final Constants.BdecParquetCompression BDEC_PARQUET_COMPRESSION_ALGORITHM_DEFAULT =
       Constants.BdecParquetCompression.GZIP;
 
+  // Iceberg mode parameters
```
sfc-gh-tzhang, line 69: Let's extend this comment explain why we need a different default for iceberg

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -411,4 +433,15 @@ public Constants.BdecParquetCompression getBdecParquetCompressionAlgorithm() {
   public String toString() {
     return "ParameterProvider{" + "parameterMap=" + parameterMap + '}';
   }
+
+  private void icebergModeValidation(String key, Object expected) {
+    Object val = this.parameterMap.get(key);
+    if (!val.equals(expected)) {
```
sfc-gh-tzhang, line 439: We should throw an exception instead of just a warning

sfc-gh-tzhang (approved):
Left some comments, PTAL, otherwise LGTM! One request I have is could we merge all the Iceberg related changes to a feature branch instead of `main`? We can move everything to `main` once we ready, otherwise the new release will contain all the WIP changes and customer might do something with it.

PR #768, opened 2024-06-08
SNOW-1437884 Flush service supports uploading file to Iceberg tables

sfc-gh-alhuang:
Modify flush service to support uploading file to external volume. This PR includes following change:

- Add default parameter `MAX_CLIENT_LAG` for Iceberg mode
- Introduce new parameter `DISABLE_CHUNK_ENCRYPTION` to disable chunk encryption in Iceberg mode
- Move stage from client level to table level when using Iceberg mode, see: `StreamingIngestExternalVolume.java`
- Move stage/volume configure logic to channel level when using Iceberg
- Fix `FlushServiceTest` and `StreamingIngestStage`, more test cases is expected

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/AbstractCloudStorage.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/AbstractCloudStorage.java
@@ -253,8 +253,7 @@ abstract class AbstractCloudStorage {
     payload.put("file_name", fileName);
     Map<String, Object> response = this.makeConfigureCall(payload);
 
-    JsonNode responseNode =
-        this.parseStageLocation(mapper.valueToTree(response).get("stage_location"));
+    JsonNode responseNode = this.parseConfigureResponse(response);
 
     SnowflakeFileTransferMetadataV1 metadata =
         (SnowflakeFileTransferMetadataV1)
@@ -264,14 +263,15 @@ abstract class AbstractCloudStorage {
     return metadata;
   }
 
-  protected JsonNode parseStageLocation(JsonNode stageLocation) {
-    ObjectNode responseNode = mapper.createObjectNode();
+  protected JsonNode parseConfigureResponse(Map<String, Object> response) {
+    JsonNode responseNode = mapper.valueToTree(response);
 
     // Currently there are a few mismatches between the client/configure response and what
     // SnowflakeFileTransferAgent expects
-    responseNode.putObject("data");
-    ObjectNode dataNode = (ObjectNode) responseNode.get("data");
-    dataNode.set("stageInfo", stageLocation);
+    ObjectNode mutable = (ObjectNode) responseNode;
+    mutable.putObject("data");
+    ObjectNode dataNode = (ObjectNode) mutable.get("data");
+    dataNode.set("stageInfo", responseNode.get("stage_location"));
 
     // JDBC expects this field which maps to presignedFileUrlName.  We will set this later
     dataNode.putArray("src_locations").add("placeholder");
@@ -300,8 +300,7 @@ abstract class AbstractCloudStorage {
     Map<Object, Object> payload = getConfigurePayload();
     Map<String, Object> response = this.makeConfigureCall(payload);
 
-    JsonNode responseNode =
-        this.parseStageLocation(mapper.valueToTree(response).get("stage_location"));
+    JsonNode responseNode = this.parseConfigureResponse(response);
     // Do not change the prefix everytime we have to refresh credentials
     if (Utils.isNullOrEmpty(this.clientPrefix)) {
       this.clientPrefix = createClientPrefix(responseNode);
--- a/src/main/java/net/snowflake/ingest/streaming/internal/AbstractCloudStorage.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/AbstractCloudStorage.java
@@ -1,0 +1,373 @@
+/*
+ * Copyright (c) 2021 Snowflake Computing Inc. All rights reserved.
+ */
+
+package net.snowflake.ingest.streaming.internal;
+
+import static net.snowflake.ingest.connection.ServiceResponseHandler.ApiName.STREAMING_CLIENT_CONFIGURE;
+import static net.snowflake.ingest.streaming.internal.StreamingIngestUtils.executeWithRetries;
+import static net.snowflake.ingest.utils.Constants.RESPONSE_SUCCESS;
+import static net.snowflake.ingest.utils.HttpUtil.generateProxyPropertiesForJDBC;
+import static net.snowflake.ingest.utils.Utils.getStackTrace;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Properties;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
+import net.snowflake.client.core.OCSPMode;
+import net.snowflake.client.jdbc.SnowflakeFileTransferAgent;
+import net.snowflake.client.jdbc.SnowflakeFileTransferConfig;
+import net.snowflake.client.jdbc.SnowflakeFileTransferMetadataV1;
+import net.snowflake.client.jdbc.SnowflakeSQLException;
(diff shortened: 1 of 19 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -83,25 +87,29 @@ static <T> Blob constructBlobAndMetadata(
           flusher.serialize(channelsDataPerTable, filePath);
 
       if (!serializedChunk.channelsMetadataList.isEmpty()) {
-        ByteArrayOutputStream chunkData = serializedChunk.chunkData;
-        Pair<byte[], Integer> paddedChunk =
-            padChunk(chunkData, Constants.ENCRYPTION_ALGORITHM_BLOCK_SIZE_BYTES);
-        byte[] paddedChunkData = paddedChunk.getFirst();
-        int paddedChunkLength = paddedChunk.getSecond();
+        byte[] chunkData = serializedChunk.chunkData.toByteArray();
```
sfc-gh-hmadan, line 90: this allocation introduces unnecessary GC pressure  when disableChunkEncryption is false. please do this in an else block only when encryption is disabled.

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -61,10 +61,14 @@ class BlobBuilder {
...
    * @param bdecVersion version of blob
+   * @param disableChunkEncryption If the encryption is disabled
    * @return {@link Blob} data
    */
   static <T> Blob constructBlobAndMetadata(
-      String filePath, List<List<ChannelData<T>>> blobData, Constants.BdecVersion bdecVersion)
+      String filePath,
+      List<List<ChannelData<T>>> blobData,
+      Constants.BdecVersion bdecVersion,
+      boolean disableChunkEncryption)
```
sfc-gh-hmadan, line 71: nit: suggest calling it `boolean encrypt` and pass in true for non-iceberg / false for iceberg.

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -127,9 +135,9 @@ static <T> Blob constructBlobAndMetadata(
...
         chunksMetadataList.add(chunkMetadata);
-        chunksDataList.add(encryptedCompressedChunkData);
-        curDataSize += encryptedCompressedChunkDataSize;
-        crc.update(encryptedCompressedChunkData, 0, encryptedCompressedChunkDataSize);
+        chunksDataList.add(chunkData);
+        curDataSize += compressedChunkDataSize;
+        crc.update(chunkData, 0, compressedChunkDataSize);
 
         logger.logInfo(
             "Finish building chunk in blob={}, table={}, rowCount={}, startOffset={},"
```
sfc-gh-hmadan, line 143: add whether encryptChunk was passed in as true or false, to this log stmt.

src/main/java/net/snowflake/ingest/streaming/internal/AbstractCloudStorage.java
```diff
@@ -0,0 +1,372 @@
...
+import net.snowflake.client.jdbc.internal.fasterxml.jackson.databind.node.ObjectNode;
+import net.snowflake.ingest.connection.IngestResponseException;
+import net.snowflake.ingest.connection.RequestBuilder;
+import net.snowflake.ingest.utils.ErrorCode;
+import net.snowflake.ingest.utils.Logging;
+import net.snowflake.ingest.utils.SFException;
+import net.snowflake.ingest.utils.Utils;
+
+/** Handles uploading files to the Snowflake Streaming Ingest Cloud Storage */
+abstract class AbstractCloudStorage {
```
sfc-gh-hmadan, line 39: i'm not fully bought into this class split as it looks to be based on code reuse and not a clear responsibility split between classes. Will read old/new code to make a suggestion (or get behind your current setup!). JFYI.

In general I'd prefer not having an inheritance hierarchy, feel free to read up on inheritance vs. composition. Its much more maintainable to have a class (C1) implement some common behavior, and have two other classes C2 and C3 that need to use that behavior instantiate C1 inside themselves (that is C2 and C3 compose over C1). Versus have a base class C1 and make C2 and C3 inherit from C1, as you have it right now.

src/main/java/net/snowflake/ingest/streaming/internal/OpenChannelResponse.java
```diff
@@ -130,4 +132,13 @@ void setEncryptionKeyId(Long encryptionKeyId) {
   Long getEncryptionKeyId() {
     return this.encryptionKeyId;
   }
+
+  @JsonProperty("stage_location")
+  void setStageLocation(JsonNode stageLocation) {
```
sfc-gh-hmadan, line 137: Can we use a proper java contract here instead of JsonNode? JsonNode gives us no compile time guarantees on what object was passed in, makes code much harder to read, and makes it easy to introduce corner case bugs where under some specific conditions the json node is constructed differently.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -337,6 +338,7 @@ public SnowflakeStreamingIngestChannelInternal<?> openChannel(OpenChannelRequest
       payload.put("schema", request.getSchemaName());
       payload.put("write_mode", Constants.WriteMode.CLOUD_STORAGE.name());
       payload.put("role", this.role);
+      payload.put("is_iceberg_client", isIcebergMode);
```
sfc-gh-hmadan, line 341: I believe Toby asked for a different name on the contract, can you pl check my pr? Can find it via go/prby/hmadan in chrome.
Also note that drop channel needs this property for iceberg clients.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -394,6 +396,10 @@ public SnowflakeStreamingIngestChannelInternal<?> openChannel(OpenChannelRequest
       // Add channel to the channel cache
       this.channelCache.addChannel(channel);
 
+      if (isIcebergMode) {
+        this.flushService.addExternalVolume(request.getFullyQualifiedTableName(), response);
```
sfc-gh-hmadan, line 400: Have only skimmed FlushService but the changes there look like we need something more intentional:
1. Stage management in a separate class, instead of in FlushService
2. Pass that StageManager into flushService, like the channelCache is passed in
3. IngestClientInternal (this place) directly adds the stage info to the stagemanager, instead of to flushService.

Think of it as - flush service is responsible for flushing to a stage, but not really managing addition / removal of stages, or of renewing the tokens for a stage.

As an aside, this is a good set of concepts esp the S and the O - https://en.wikipedia.org/wiki/SOLID

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -205,11 +214,21 @@ private void setParameterMap(
         parameterOverrides,
         props);
 
+    this.updateValue(
+        DISABLE_CHUNK_ENCRYPTION,
+        isIcebergMode
+            ? DISABLE_CHUNK_ENCRYPTION_ICEBERG_MODE_DEFAULT
+            : DISABLE_CHUNK_ENCRYPTION_DEFAULT,
+        parameterOverrides,
```
sfc-gh-hmadan, line 222: this allows SDK customers to specify whether chunk encryption is enabled or disabled, why do we want to do that?

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -430,6 +449,13 @@ public Constants.BdecParquetCompression getBdecParquetCompressionAlgorithm() {
     return Constants.BdecParquetCompression.fromName((String) val);
   }
 
+  /** @return If the chunk encryption is disabled */
+  public boolean getDisableChunkEncryption() {
```
sfc-gh-hmadan, line 453: public == customers can call this? Why?

PR #770, opened 2024-06-11
DO NOT MERGE: Make binary string encoding configurable, add support for base64

sfc-gh-kgaputis:
Addresses issue where streaming client receives base64-encoded strings for binary.  

Adds new parameter `BINARY_STRING_ENCODING` which supports values `hex` and `base64`.   The default value is `hex`.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
@@ -24,17 +24,14 @@ import java.time.ZoneId;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeParseException;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
 import java.util.function.Supplier;
 import net.snowflake.client.jdbc.internal.google.common.collect.Sets;
 import net.snowflake.client.jdbc.internal.snowflake.common.core.SnowflakeDateTimeFormat;
 import net.snowflake.client.jdbc.internal.snowflake.common.util.Power10;
 import net.snowflake.ingest.streaming.internal.serialization.ByteArraySerializer;
 import net.snowflake.ingest.streaming.internal.serialization.ZonedDateTimeSerializer;
+import net.snowflake.ingest.utils.Constants;
 import net.snowflake.ingest.utils.ErrorCode;
 import net.snowflake.ingest.utils.SFException;
 import org.apache.commons.codec.DecoderException;
@@ -615,7 +612,7 @@ class DataValidationUtil {
    * @return Validated array
    */
   static byte[] validateAndParseBinary(
-      String columnName, Object input, Optional<Integer> maxLengthOptional, long insertRowIndex) {
+      String columnName, Object input, Optional<Integer> maxLengthOptional, long insertRowIndex, Constants.BinaryStringEncoding binaryStringEncoding) {
     byte[] output;
     if (input instanceof byte[]) {
       // byte[] is a mutable object, we need to create a defensive copy to protect against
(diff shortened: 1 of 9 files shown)


sfc-gh-tzhang:
@sfc-gh-lsembera what do you think about supporting base64 as well? 

Also, could we do it automatically instead of adding a parameter?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -625,12 +622,30 @@ static byte[] validateAndParseBinary(
...
       System.arraycopy(originalInputArray, 0, output, 0, originalInputArray.length);
     } else if (input instanceof String) {
-      try {
-        String stringInput = ((String) input).trim();
-        output = Hex.decodeHex(stringInput);
-      } catch (DecoderException e) {
+      if(binaryStringEncoding == Constants.BinaryStringEncoding.BASE64) {
+        try {
+          String stringInput = ((String) input).trim();
+          // Remove double quotes if present
```
sfc-gh-lsembera, line 628: We should not strip surrounding quotes. We don't do it for hex and neither does it work in Snowflake Worksheets.

src/test/java/net/snowflake/ingest/streaming/internal/DataValidationUtilTest.java
```diff
@@ -888,76 +890,76 @@ public void testValidateAndParseBinary() throws DecoderException {
     assertArrayEquals(
         "honk".getBytes(StandardCharsets.UTF_8),
         validateAndParseBinary(
-            "COL", "honk".getBytes(StandardCharsets.UTF_8), Optional.empty(), 0));
+            "COL", "honk".getBytes(StandardCharsets.UTF_8), Optional.empty(), 0, Constants.BinaryStringEncoding.HEX));
```
sfc-gh-lsembera, line 893: We need the same test coverage for base64, as we have for hex.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -24,17 +24,14 @@
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeParseException;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
```
sfc-gh-lsembera, line 27: no wildcard imports.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -64,6 +67,9 @@ public class ParameterProvider {
   public static final Constants.BdecParquetCompression BDEC_PARQUET_COMPRESSION_ALGORITHM_DEFAULT =
       Constants.BdecParquetCompression.GZIP;
 
+  public static final Constants.BinaryStringEncoding BINARY_STRING_ENCODING_DEFAULT =
```
sfc-gh-lsembera, line 70: Please add tests to `ParameterProviderTest`.

sfc-gh-lsembera (commented):
> @sfc-gh-lsembera what do you think about supporting base64 as well?

I am fine with supporting base64, as well. I just don't like the removal of surrounding quotes - we don't do it for any other data type, and I think we shouldn't do it for base64 either. 

> Also, could we do it automatically instead of adding a parameter?

We cannot do it automatically. Without additional information, we cannot tell if a string is base64- or hex-encoded.

sfc-gh-tzhang:
> > Also, could we do it automatically instead of adding a parameter?
> 
> We cannot do it automatically. Without additional information, we cannot tell if a string is base64- or hex-encoded.

Thanks for confirming, then I think this should be a channel level configuration instead of the whole client, take a look at what we did for `setDefaultTimezone`

sfc-gh-lsembera:
> Thanks for confirming, then I think this should be a channel level configuration instead of the whole client, take a look at what we did for `setDefaultTimezone`

The problem with having the configuration per-channel is that it is not as easily configurable from KC.

sfc-gh-kgaputis:
@sfc-gh-tzhang @sfc-gh-lsembera,    based on the requirement to work with kafka connector,  I think this config should apply  to the whole client.     

Also,  I suspect that most streaming use cases with embedded binary in Avro messages will be either 100% hex encoded or 100% base64 encoded,  based on the particular tooling used in the customer's environment, so the need to support mixed encodings would be less likely.

sfc-gh-tzhang:
Closing old PRs, feel free to reopen if needed

PR #773, opened 2024-06-14
SNOW-1483230 Disable blob encryption / Add class for storage metadata & configure response

sfc-gh-alhuang:
Aim to support uploading file to external volume in Iceberg mode. This PR includes following change:

- Add default parameter MAX_CLIENT_LAG for Iceberg mode
- Introduce new parameter DISABLE_CHUNK_ENCRYPTION to disable chunk encryption in Iceberg mode
- Create `FileLocationInfo` and `ConfigureResponse` class to hold the json response from `OpenChannel` in Iceberg mode


src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -430,6 +448,13 @@ public Constants.BdecParquetCompression getBdecParquetCompressionAlgorithm() {
     return Constants.BdecParquetCompression.fromName((String) val);
   }
 
+  /** @return If the chunk encryption is enabled */
+  public boolean getEnableChunkEncryption() {
```
sfc-gh-alhuang, line 452: Used [here](https://github.com/snowflakedb/snowflake-ingest-java/blob/9b69386ab4ff855b0bed52fc516aba75b9d3d8c6/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java#L559)

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -83,25 +87,30 @@ static <T> Blob constructBlobAndMetadata(
           flusher.serialize(channelsDataPerTable, filePath);
 
       if (!serializedChunk.channelsMetadataList.isEmpty()) {
-        ByteArrayOutputStream chunkData = serializedChunk.chunkData;
-        Pair<byte[], Integer> paddedChunk =
-            padChunk(chunkData, Constants.ENCRYPTION_ALGORITHM_BLOCK_SIZE_BYTES);
-        byte[] paddedChunkData = paddedChunk.getFirst();
-        int paddedChunkLength = paddedChunk.getSecond();
+        byte[] chunkData;
+        int actualChunkLength = serializedChunk.chunkData.size();
```
sfc-gh-hmadan, line 91: chunkLength should be set to paddedChunk.getSecond() when encrypt == true. Lets not try to optimize for one line of code sharing between these two code paths just because they happen to be identical.

Suggest:
```
byte[] chunkData;
int chunkLength;
if (encrypt) {
    chunkData = ...;
    chunkLength = ...;
} else {
    chunkData = serializedChunk.chunkData.toByteArray();
    chunkLength = chunkData.size();
}
```

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -200,28 +210,24 @@ static byte[] buildBlob(
                 + BLOB_CHECKSUM_SIZE_IN_BYTES
```
sfc-gh-hmadan, line 210: all the changes in this method (buildBlob) look unnecessary, is something wrong with the old logic?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -337,6 +337,7 @@ public SnowflakeStreamingIngestChannelInternal<?> openChannel(OpenChannelRequest
       payload.put("schema", request.getSchemaName());
       payload.put("write_mode", Constants.WriteMode.CLOUD_STORAGE.name());
       payload.put("role", this.role);
+      payload.put("is_iceberg", isIcebergMode);
```
sfc-gh-hmadan, line 340: no request payload test coverage that needed to be updated?

src/main/java/net/snowflake/ingest/streaming/internal/StageMetadata.java
```diff
@@ -0,0 +1,117 @@
+package net.snowflake.ingest.streaming.internal;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Map;
+
+public class StageMetadata {
```
sfc-gh-hmadan, line 6: mind sharing where this contract is coming from / how did you decide to come up with it ?

src/main/java/net/snowflake/ingest/streaming/internal/StageMetadata.java
```diff
@@ -0,0 +1,117 @@
...
+  @JsonProperty(value = "locationType", required = true)
+  private String stageType; // The stage type
+
+  @JsonProperty(value = "location", required = true)
+  private String location; // The container or bucket
+
+  @JsonProperty(value = "creds", required = true)
```
sfc-gh-hmadan, line 13: nit: I'd much rather do required field validation manually after deserialization instead of relying on the deserializer's validation via this annotation. With manual validation, the error message can be made more descriptive and have more context (like which channel was this request sent for? what was the request id?) than anything jackson can generate.

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
sfc-gh-alhuang: Sure, resolved.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
sfc-gh-alhuang: I don't think we have unit test for this. Tried delete some entries and still pass unit tests. As for IT, I think we should wait for the change on sfctest0.

src/main/java/net/snowflake/ingest/streaming/internal/StageMetadata.java
sfc-gh-alhuang: It's from client configure ([ref](https://github.com/snowflakedb/snowflake/blob/46f48668a72e8f063bce1ac152616e16fd70d876/GlobalServices/src/main/java/com/snowflake/resources/streaming/StreamingResource.java#L1528))

e2e-jar-test/core/src/main/java/net/snowflake/IngestTestUtils.java
```diff
@@ -24,6 +22,8 @@
 import java.util.Map;
 import java.util.Properties;
 import java.util.UUID;
+import net.snowflake.client.jdbc.internal.fasterxml.jackson.databind.JsonNode;
```
sfc-gh-alhuang, line 25: Since we are using JDBC's method [here](https://github.com/snowflakedb/snowflake-ingest-java/blob/000a13912c1827c34365cbc8148fef744a943fee/src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java#L276) after adding `ConfigResponse`. Aligning Jackson version is needed.

pom.xml
```diff
@@ -45,7 +45,6 @@
     <commonslang3.version>3.14.0</commonslang3.version>
     <commonslogging.version>1.3.1</commonslogging.version>
     <commonstext.version>1.11.0</commonstext.version>
-    <fasterxml.version>2.16.1</fasterxml.version>
```
sfc-gh-alhuang, line 48: Deleted as we are using shaded Jackson from JDBC.

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
sfc-gh-alhuang: Merged by mistake, reverted.

src/main/java/net/snowflake/ingest/streaming/internal/StageMetadata.java
sfc-gh-alhuang: Removed the check logic here.

src/main/java/net/snowflake/ingest/streaming/internal/StageMetadata.java
```diff
@@ -0,0 +1,127 @@
+package net.snowflake.ingest.streaming.internal;
```
sfc-gh-tzhang, line 1: copyright

src/main/java/net/snowflake/ingest/streaming/internal/StageMetadata.java
```diff
@@ -0,0 +1,127 @@
...
+
+import java.util.Map;
+import net.snowflake.client.jdbc.internal.fasterxml.jackson.annotation.JsonProperty;
+
+class StageMetadata {
```
sfc-gh-tzhang, line 6: comment please

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
```diff
@@ -353,25 +348,23 @@ private JsonNode parseClientConfigureResponse(Map<String, Object> response) {
     return responseNode;
   }
 
-  private Map<String, Object> makeClientConfigureCall(Map<Object, Object> payload)
+  private ConfigureResponse makeClientConfigureCall(Map<Object, Object> payload)
```
sfc-gh-tzhang, line 351: Thanks, I like this change, we know we didn't do the right thing back then

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
```diff
@@ -255,21 +254,17 @@ synchronized SnowflakeFileTransferMetadataWithAge refreshSnowflakeMetadata(boole
     }
     Utils.assertStringNotNullOrEmpty("client prefix", this.clientPrefix);
 
-    if (responseNode
-        .get("data")
-        .get("stageInfo")
-        .get("locationType")
-        .toString()
+    if (response
```
sfc-gh-tzhang, line 257: Please make sure to test this on all three cloud providers

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
```diff
@@ -353,25 +348,23 @@ private JsonNode parseClientConfigureResponse(Map<String, Object> response) {
...
           executeWithRetries(
-              Map.class,
+              ConfigureResponse.class,
               CLIENT_CONFIGURE_ENDPOINT,
               mapper.writeValueAsString(payload),
               "client configure",
               STREAMING_CLIENT_CONFIGURE,
               httpClient,
-              requestBuilder,
-              statusGetter);
```
sfc-gh-tzhang, line 369: can we remove the statusGetter now or it's still used somewhere else?

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -41,6 +41,8 @@ public class ParameterProvider {
   public static final String BDEC_PARQUET_COMPRESSION_ALGORITHM =
       "BDEC_PARQUET_COMPRESSION_ALGORITHM".toLowerCase();
 
+  public static final String ENABLE_CHUNK_ENCRYPTION = "ENABLE_CHUNK_ENCRYPTION".toLowerCase();
```
sfc-gh-tzhang, line 44: I don't think we want this to be a configurable parameter

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -551,7 +551,12 @@ BlobMetadata buildAndUpload(String blobPath, List<List<ChannelData<T>>> blobData
     Timer.Context buildContext = Utils.createTimerContext(this.owningClient.buildLatency);
 
     // Construct the blob along with the metadata of the blob
-    BlobBuilder.Blob blob = BlobBuilder.constructBlobAndMetadata(blobPath, blobData, bdecVersion);
+    BlobBuilder.Blob blob =
+        BlobBuilder.constructBlobAndMetadata(
+            blobPath,
+            blobData,
+            bdecVersion,
+            this.owningClient.getParameterProvider().getEnableChunkEncryption());
```
sfc-gh-tzhang, line 559: we should just check if the client is for iceberg or not?

e2e-jar-test/core/src/main/java/net/snowflake/IngestTestUtils.java
sfc-gh-hmadan: 1. can we remove com.fasterxml as a direct dependency of our SDK's pom.xml? I don't want future PR authors to accidentally pull in com.fasterxml directly.
2. A big chunk of changes in the PR is now just import changes :/ Suggestion for future - you could try to split your commit history so that there's one commit with just such a blind change, and another commit with the actual logical changes being reviewed. ok how it is for now.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-hmadan: I asked alec to not pass around isIceberg everywhere - we want one place (in the parameter provider) that looks at isIceberg and sets up all the configuration / feature flags in that one place.
This is because over time we'll have a proliferation of `if (isIceberg)` in ten different places, making it incredibly hard to figure out what all is different for iceberg vs. fdn code paths.

At the same time I don't mean to say all these features need to be controllable by customers, I remember leavinga comment in the last round that enableChunkEncryption should not be public to customers + should not be publically settable/overrideable.

src/main/java/net/snowflake/ingest/streaming/internal/OpenChannelResponse.java
```diff
@@ -21,6 +21,7 @@ class OpenChannelResponse extends StreamingIngestResponse {
   private List<ColumnMetadata> tableColumns;
   private String encryptionKey;
   private Long encryptionKeyId;
+  private StageMetadata stageLocation;
```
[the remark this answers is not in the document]
sfc-gh-hmadan: the location metadata for internal stage vs. external volume should be the same, no? Maybe we should call it FileLocationInfo instead of StageMetadata so there's no stage vs. volume confusion in client sdk?

e2e-jar-test/core/src/main/java/net/snowflake/IngestTestUtils.java
sfc-gh-alhuang: Yes, sorry for the typo.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
sfc-gh-hmadan: Please add backlog JIRAs that we need test coverage on whichever front you've made changes but didn't have to also update any tests to make them pass.

src/main/java/net/snowflake/ingest/streaming/internal/StageMetadata.java
```diff
@@ -0,0 +1,127 @@
...
+  private String stageType; // The stage type
+
+  @JsonProperty("location")
+  private String location; // The container or bucket
+
+  @JsonProperty("path")
+  private String path; // path of the target file
+
+  @JsonProperty("creds")
+  private Map<?, ?> credentials; // the credentials required for the  stage
```
sfc-gh-hmadan, line 17: curious, why is this `Map<?,?>` ? Do we not have an explicit type for the credentials response?

src/main/java/net/snowflake/ingest/streaming/internal/StageMetadata.java
```diff
@@ -0,0 +1,127 @@
...
+  @JsonProperty("volumeHash")
+  private String volumeHash; // a unique id for volume assigned by server
+
+  String getStageType() {
+    return stageType;
+  }
+
+  void setLocationType(String stageType) {
+    this.stageType = stageType;
+  }
```
sfc-gh-hmadan, line 46: getStageType but setLocationType?
A test would've caught this, which means we don't have test coverage for this part too :)

pom.xml
sfc-gh-alhuang: The fasterxml.Jackson is only used internally in SDK for serialize & deserialize json. As long as the serializer version is the same as the deserializer, it should be fine.

src/main/java/net/snowflake/ingest/streaming/internal/ConfigureResponse.java
```diff
@@ -0,0 +1,61 @@
...
+  private String prefix;
+
+  @JsonProperty("status_code")
+  private Long statusCode;
+
+  @JsonProperty("message")
+  private String message;
+
+  @JsonProperty("stage_location")
+  private StageMetadata stageMetadata;
```
sfc-gh-hmadan, line 16: why not call the variable stageLocation?

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
```diff
@@ -246,7 +245,7 @@ synchronized SnowflakeFileTransferMetadataWithAge refreshSnowflakeMetadata(boole
 
     Map<Object, Object> payload = new HashMap<>();
     payload.put("role", this.role);
-    Map<String, Object> response = this.makeClientConfigureCall(payload);
+    ConfigureResponse response = this.makeClientConfigureCall(payload);
```
sfc-gh-hmadan, line 248: I see we still have the responseNode object around, so now there's two different clientConfigure response objects floating around.
If we can't completely remove responseNode I'm not sure what this change buys us ?

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-hmadan: +1 on not allowing customers to configure this.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-alhuang: Made non-settable to customers. For `getEnableChunkEncrption()` I think it should still be public as the `ParameterProvider` is located in different package with `FlushService`

sfc-gh-tzhang: i see, but `ParameterProvider` is supposed to contain only configurable parameters by customers, do you want to create a separate class for internal control? (and this class can be in the same package as FlushService)

src/main/java/net/snowflake/ingest/streaming/internal/OpenChannelResponse.java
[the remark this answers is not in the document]
sfc-gh-tzhang: it's good as long as we remove stage in the name

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-hmadan: also pointed this out in the previous PR: https://github.com/snowflakedb/snowflake-ingest-java/pull/768#discussion_r1637199029 !

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -83,25 +87,32 @@ static <T> Blob constructBlobAndMetadata(
...
+          // to align with decryption on the Snowflake query path.
+          // TODO: address alignment for the header SNOW-557866
+          long iv = curDataSize / Constants.ENCRYPTION_ALGORITHM_BLOCK_SIZE_BYTES;
+          compressedChunkData =
+              Cryptor.encrypt(
+                  paddedChunkData, firstChannelFlushContext.getEncryptionKey(), filePath, iv);
+        } else {
+          compressedChunkData = serializedChunk.chunkData.toByteArray();
+          chunkLength = compressedChunkData.length;
+        }
```
sfc-gh-hmadan, line 112: @sfc-gh-tzhang do you know why in the old code, computeMD5() was called with a chunk length that wasn't == encryptedCompressedChunkData.length ?

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-hmadan: Interesting, maybe we should chat f2f on this, but i don't see where ParameterProvider object is returned out to customers?
I see IngestClientInternal has a non-public getParameterProvider method, and the SnowflakeStreamingIngestClient interface also does not have any methods to expose the parameter provider.

If/when we do want to expose this object out in future, we can easily introduce an interface IParameterProvider with only the methods that we want to expose out, and add a getter on client that returns objects of this interface?

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -127,21 +137,22 @@ static <T> Blob constructBlobAndMetadata(
 
         // Add chunk metadata and data to the list
         chunksMetadataList.add(chunkMetadata);
-        chunksDataList.add(encryptedCompressedChunkData);
-        curDataSize += encryptedCompressedChunkDataSize;
-        crc.update(encryptedCompressedChunkData, 0, encryptedCompressedChunkDataSize);
+        chunksDataList.add(chunkData);
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Renamed

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
sfc-gh-alhuang: Sure, removed.

src/main/java/net/snowflake/ingest/streaming/internal/StageMetadata.java
```diff
@@ -0,0 +1,127 @@
+package net.snowflake.ingest.streaming.internal;
+
+import java.util.Map;
+import net.snowflake.client.jdbc.internal.fasterxml.jackson.annotation.JsonProperty;
+
+class StageMetadata {
+  @JsonProperty("locationType")
+  private String stageType; // The stage type
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Renamed.

src/main/java/net/snowflake/ingest/streaming/internal/StageMetadata.java
sfc-gh-alhuang: Renamed. The serialization / deserialization was tested in `StreamingIngestStageTest.testRefreshSnowflakeMetadataRemote`. It was passed because the `@JsonProperty` override the key set by setter/getter method name, which makes the method name doesn't matter.

e2e-jar-test/core/src/main/java/net/snowflake/IngestTestUtils.java
sfc-gh-alhuang: Removed in pom.xml.

src/main/java/net/snowflake/ingest/streaming/internal/StageMetadata.java
sfc-gh-alhuang: Referenced from [here](https://github.com/snowflakedb/snowflake-jdbc/blob/6a95dccda713969cc8681f2c3bfa5791b4d8b97e/src/main/java/net/snowflake/client/jdbc/SnowflakeFileTransferAgent.java#L1322). Just saw they are using String -> String mapper, changed to Map<String, String>.

src/main/java/net/snowflake/ingest/streaming/internal/ConfigureResponse.java
sfc-gh-alhuang: Renamed.

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
sfc-gh-alhuang: Removed `responseNode` for `createClientPrefix`, but we still need to convert to JsonNode [here](https://github.com/snowflakedb/snowflake-ingest-java/blob/45e705f2b56f6d3f0fb0adb8a71e84f2d8bac09f/src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java#L275).

src/main/java/net/snowflake/ingest/streaming/internal/OpenChannelResponse.java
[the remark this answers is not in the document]
sfc-gh-alhuang: Renamed.

e2e-jar-test/core/src/main/java/net/snowflake/IngestTestUtils.java
sfc-gh-hmadan: still not convinced about this change - two different classes with the name JsonNode in completely different packages / namespaces don't really need to be the same unless we have some pieces of code using package1.JsonNode interacting with other pieces of code that expect package2.JsonNode.

I'm not sure where we have this situation, the only place where jdbc...JsonNode is used is `StreamingIngestStage.java` - there is no outgoing or incoming call that needs every other usage of JsonNode across the project to be using jdbc...JsonNode.

If this is a pure cleanup that you want to do that's not really needed for what you were trying to do in this PR, it should not be mixed in with feature changes IMHO. LMK if I misunderstood your changes in some way?

pom.xml
sfc-gh-hmadan: Left a comment on another thread for basically the same thing:
 https://github.com/snowflakedb/snowflake-ingest-java/pull/773#discussion_r1645244609

e2e-jar-test/core/src/main/java/net/snowflake/IngestTestUtils.java
sfc-gh-alhuang: The problem I faced was that in [makeClientConfigureCall->executeWithRetries](https://github.com/snowflakedb/snowflake-ingest-java/blob/abe040568fdb7202952859b2e02e80e5b7d9dfa7/src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java#L342), the [executeWithRetires](https://github.com/snowflakedb/snowflake-ingest-java/blob/188632ad70e1bc15de4ef076961d7069b57d8af6/src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestUtils.java#L7) call [unmarshallStreamingIngestResponse](https://github.com/snowflakedb/snowflake-ingest-java/blob/188632ad70e1bc15de4ef076961d7069b57d8af6/src/main/java/net/snowflake/ingest/connection/ServiceResponseHandler.java#L200) which use Jackson object mappers. Though there's no clear limitation that the Jackson annotation and mapper should be using the same version, I found that if we tried to use different jackson (com.fasterxml) for object mapper and (jdbc.fasterxml) for annotation in `ConfigureResponse` it'll fail. This implies that all classes using `executeWithRetires` should be using the same Jackson package, as there're all using the same mapper in `unmarshallStreamingIngestResponse`.

Since `parseClientConfigureResponse` should be returning jdbc..JsonNode, and it also requires mapper to unmarshall the response, I thought that aligning the Jackson package will make life easier. Instead facing this issue in future as there's no warning for annotation & mapper version mismatch but just returning an empty node which I spent plenty of time to figure out this issue.


We can still use String as intermediate object to convert com....JsonNode to com...JsonNode without aligning Jackson version:
```
com....Object mapper = new com....ObjectMapper();
String jsonString = mapper.writeValueAsString(configureResponse);
net.snowflake.client.jdbc....Object jdbcMapper = new net.snowflake.client.jdbc....ObjectMapper();
net.snowflake.client.jdbc....JsonNode jsonNode = jdbcMapper.readTree(jsonString);
```

What do you think?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
sfc-gh-alhuang: Added [here](https://snowflakecomputing.atlassian.net/browse/SNOW-1490151)

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-tzhang: We're not exposing ParameterProvider directly, but if you see how the ParameterProvider is created, it will take an user input property/parameter overrides as part of the client creation.

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
sfc-gh-tzhang: @sfc-gh-azagrebin do you remember this?

sfc-gh-tzhang: Do we even use MD5 at server side today?

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -67,8 +67,8 @@ public class ParameterProvider {
       Constants.BdecParquetCompression.GZIP;
 
   /* Iceberg mode parameters: When streaming to Iceberg mode, different default parameters are required because it generates Parquet files instead of BDEC files. */
-  public static final int MAX_CHUNKS_IN_BLOB_AND_REGISTRATION_REQUEST_ICEBERG_MODE_DEFAULT =
-      1; // 1 parquet file per blob
+  public static final int MAX_CHUNKS_IN_BLOB_AND_REGISTRATION_REQUEST_ICEBERG_MODE_DEFAULT = 1;
+  public static final long MAX_CLIENT_LAG_ICEBERG_MODE_DEFAULT = 30000;
```
sfc-gh-tzhang, line 71: Could you move this to the "lag related parameters" section?

src/main/java/net/snowflake/ingest/streaming/internal/ConstantParameterProvider.java
```diff
@@ -0,0 +1,30 @@
...
+
+  ConstantParameterProvider(boolean isIcebergMode) {
+    this.enableChunkEncryption =
+        isIcebergMode
+            ? IcebergModeConstants.ENABLE_CHUNK_ENCRYPTION
+            : NonIcebergModeConstants.ENABLE_CHUNK_ENCRYPTION;
+  }
+
+  boolean getEnableChunkEncryption() {
+    return this.enableChunkEncryption;
```
sfc-gh-tzhang, line 28: could we just check isIcebergMode here?

src/main/java/net/snowflake/ingest/streaming/internal/ConstantParameterProvider.java
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright (c) 2024 Snowflake Computing Inc. All rights reserved.
+ */
+
+package net.snowflake.ingest.streaming.internal;
+
+/** A class to provide non-configurable constants depends on Iceberg or non-Iceberg mode */
+class ConstantParameterProvider {
```
sfc-gh-tzhang, line 8: What about InternalParameterProvider?

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -111,6 +115,17 @@ private void updateValue(
     } else {
       this.parameterMap.put(key, defaultValue);
     }
+
+    if (enforceDefault) {
```
sfc-gh-tzhang, line 119: Should we put this at the beginning of the function?

sfc-gh-tzhang (approved):
Left some comments, PTAL, otherwise LGTM, thanks!

src/main/java/net/snowflake/ingest/streaming/internal/ConstantParameterProvider.java
sfc-gh-alhuang: Yes, modified.

src/main/java/net/snowflake/ingest/streaming/internal/ConstantParameterProvider.java
sfc-gh-alhuang: Renamed.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-alhuang: Make sense, resolved.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-alhuang: Moved.

src/main/java/net/snowflake/ingest/streaming/internal/ConstantParameterProvider.java
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  static class NonIcebergModeConstants {
+    static final boolean ENABLE_CHUNK_ENCRYPTION = true;
+  }
+
+  private final boolean enableChunkEncryption;
+
+  ConstantParameterProvider(boolean isIcebergMode) {
+    this.enableChunkEncryption =
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Sure, added.

src/main/java/net/snowflake/ingest/streaming/internal/ConfigureResponse.java
```diff
@@ -0,0 +1,66 @@
...
+  @JsonProperty("prefix")
+  private String prefix;
+
+  @JsonProperty("status_code")
+  private Long statusCode;
+
+  @JsonProperty("message")
+  private String message;
+
+  @JsonProperty("stage_location")
```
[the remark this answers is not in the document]
sfc-gh-alhuang: @sfc-gh-hmadan This also required change on server side. Added to backlog [jira](https://snowflakecomputing.atlassian.net/browse/SNOW-1490151).

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -964,6 +970,15 @@ ParameterProvider getParameterProvider() {
     return parameterProvider;
   }
 
+  /**
+   * Get ConstantParameterProvider with constants parameters
+   *
+   * @return {@link InternalParameterProvider} used by the client
+   */
+  InternalParameterProvider getConstantParameterProvider() {
```
sfc-gh-hmadan, line 978: getInternalParameterProvider

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-hmadan: unclear why this should be at the beginning of the function? You are missing out on validation of any override passed in via parameterOverrides or via props.

sfc-gh-hmadan (approved):
Left two comments - one rename and one about validation at the start or the end in ParameterProvider, otherwise LGTM!

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-alhuang: Ah yes, reverted to original order.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
sfc-gh-alhuang: Renamed.

Issue #753, opened 2024-04-22
Refresh stage metadata automatically in the background to avoid exceptions encountered during file upload in uploadWithoutConnection


ademfreshbooks:
Snowflake Ingest SDK: 2.1.0

For long running clients in streaming mode, It seems ~ every hour, session expires for StreamingIngestStage (used by FlushService), that raises:

```
Exception encountered during file upload in uploadWithoutConnection
```

Then StreamingIngestStage retries request again, by default 5 times. 

Usually this does the job, however for app observability sake, this is not perfect as this creates false positives.

Also, there does not appear to be any configuration options available to avoid this behaviour, or options to override some of the things.

**Can you please try to add auto refresh for stage metadata to avoid this situation?**

sfc-gh-lsembera:
Thanks for the suggestion, @ademfreshbooks. We do attempt to auto-refresh stage metadata when uploads start failing, but we can improve the situation by not logging the initial exception when metadata was immediately successfully refreshed.

sfc-gh-lsembera:
Yeah, the SDK doesn't log the first upload error, but the library we are using to upload data does. Thanks for reporting the issue, we will post updates here once the issue gets prioritized.

sfc-gh-tzhang:
This should be fixed in 2.1.2 release, please give it a try and report back if needed, thanks!

PR #780, opened 2024-06-21
SNOW-1373151: Proactively refresh token to avoid token expiration exception from JDBC

sfc-gh-tzhang:
We have seen many complains from customer about having false positive ERROR log from JDBC, since we will refresh the token only if it expires. This change switches to actively refresh the token before it expires to avoid the ERROR log.
- https://github.com/snowflakedb/snowflake-ingest-java/issues/753

Testing is covered by existing long running end2end test

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
@@ -17,6 +17,8 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.Paths;
+import java.time.Duration;
+import java.time.Instant;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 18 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
```diff
@@ -47,6 +49,11 @@ class StreamingIngestStage {
   private static final long REFRESH_THRESHOLD_IN_MS =
       TimeUnit.MILLISECONDS.convert(1, TimeUnit.MINUTES);
 
+  // Stage credential refresh interval, currently the token will expire in 1hr for GCS and 2hr for
```
sfc-gh-psaha, line 52: Should we pass back the credential duration in the client configure response instead of hardcoding it here? That would be more future-proof.

sfc-gh-tzhang: This seems to be an overkill for me, because
- This is just to fix a complain from customer that they don't like to see the ERROR log in their log file. Even if the frequency gets smaller in the future, everything will continue to work, except customers will see the log again.
- The parameter not user configurable, and the last time it gets updated was back in 2021
- Even if we support this, they will get the new parameter values only if they restart the client

In general, I agree it's better to do what you suggested. We have discussed in the past to support something like this for all other parameters so that we can control other behaviors of the SDK without a client side upgrade/downgrade, but we decided to NOT do it and focus on ROWSET API. WDYT?

sfc-gh-psaha (approved):

sfc-gh-tjones (approved):

PR #782, opened 2024-06-24
Various performance improvements in the `insertRows` path

sfc-gh-psaha:
Here are the changes in order of high impact to low:
- Use pre-computed powers of 10 instead of computing it fresh for every insertRows call. 
- Periodically compute free memory in the system instead of on-demand.
- Use a for-loop instead of stream and collect in `verifyInputColumns`
- Cache numAvailableProcessors in FlushService.
- Cache parameters used in the hot path in `insertRows`.

I wrote a simple JMH benchmark to measure these changes - see `InsertRowsBenchmarkTest`. This benchmark inserts 1M rows in a tight loop and each row has a single sb16 column. Overall, 35-40% improvement in this benchmark observed. The benchmark results should carry over to real world workloads that have sb16 columns and the improvement in the memory tracking will transfer over to all workloads. 

Before:
```
# Warmup Iteration   1: 1822464.171 us/op
# Warmup Iteration   2: 1835477.606 us/op
Iteration   1: 1592362.119 us/op
Iteration   2: 1675821.530 us/op
Iteration   3: 1698952.582 us/op
Iteration   4: 1728100.495 us/op
Iteration   5: 1527890.615 us/op
Iteration   6: 1530390.389 us/op
Iteration   7: 2039346.448 us/op
Iteration   8: 1846764.755 us/op
Iteration   9: 1514480.481 us/op
Iteration  10: 1510466.306 us/op


Result "net.snowflake.ingest.streaming.internal.InsertRowsBenchmarkTest.testInsertRow":
  1666457.572 ±(99.9%) 260468.722 us/op [Average]
  (min, avg, max) = (1510466.306, 1666457.572, 2039346.448), stdev = 172283.933
  CI (99.9%): [1405988.850, 1926926.294] (assumes normal distribution)

# Run complete. Total time: 00:01:07

Benchmark                              (numRows)  Mode  Cnt        Score        Error  Units
InsertRowsBenchmarkTest.testInsertRow    1000000  avgt   10  1666457.572 ± 260468.722  us/op
```

After:

```
# Warmup Iteration   1: 1139006.469 us/op
# Warmup Iteration   2: 1132933.357 us/op
Iteration   1: 987257.653 us/op
Iteration   2: 1070616.846 us/op
Iteration   3: 982547.952 us/op
Iteration   4: 1225079.558 us/op
Iteration   5: 1198002.424 us/op
Iteration   6: 976386.714 us/op
Iteration   7: 964823.936 us/op
Iteration   8: 1141437.334 us/op
Iteration   9: 1292684.431 us/op
Iteration  10: 940295.781 us/op


Result "net.snowflake.ingest.streaming.internal.InsertRowsBenchmarkTest.testInsertRow":
  1077913.263 ±(99.9%) 192324.330 us/op [Average]
  (min, avg, max) = (940295.781, 1077913.263, 1292684.431), stdev = 127210.636
  CI (99.9%): [885588.933, 1270237.593] (assumes normal distribution)


# Run complete. Total time: 00:00:51

Benchmark                              (numRows)  Mode  Cnt        Score        Error  Units
InsertRowsBenchmarkTest.testInsertRow    1000000  avgt   10  1077913.263 ± 192324.330  us/op
```

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
@@ -16,7 +16,6 @@ import java.util.Set;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Consumer;
-import java.util.stream.Collectors;
 import net.snowflake.ingest.connection.TelemetryService;
 import net.snowflake.ingest.streaming.InsertValidationResponse;
 import net.snowflake.ingest.streaming.OffsetTokenVerificationFunction;
@@ -403,7 +402,7 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
     Set<String> originalKeys = row.keySet();
     Map<String, String> inputColNamesMap = new HashMap<>();
     for (String key : originalKeys) {
-        inputColNamesMap.put(LiteralQuoteUtils.unquoteColumnName(key), key);
+      inputColNamesMap.put(LiteralQuoteUtils.unquoteColumnName(key), key);
     }
     // Check for extra columns in the row
     List<String> extraCols = new ArrayList<>();
--- a/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
@@ -401,9 +401,8 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
     // Map of unquoted column name -> original column name
     Set<String> originalKeys = row.keySet();
     Map<String, String> inputColNamesMap = new HashMap<>();
-    for (String key : originalKeys) {
-      inputColNamesMap.put(LiteralQuoteUtils.unquoteColumnName(key), key);
-    }
+    originalKeys.forEach(
+        key -> inputColNamesMap.put(LiteralQuoteUtils.unquoteColumnName(key), key));
     // Check for extra columns in the row
     List<String> extraCols = new ArrayList<>();
     for (String columnName : inputColNamesMap.keySet()) {
--- a/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
@@ -400,10 +400,11 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
   Set<String> verifyInputColumns(
       Map<String, Object> row, InsertValidationResponse.InsertError error, int rowIndex) {
     // Map of unquoted column name -> original column name
-    Map<String, String> inputColNamesMap =
-        row.keySet().stream()
-            .collect(Collectors.toMap(LiteralQuoteUtils::unquoteColumnName, value -> value));
-
+    Set<String> originalKeys = row.keySet();
+    Map<String, String> inputColNamesMap = new HashMap<>();
+    for (String key : originalKeys) {
+        inputColNamesMap.put(LiteralQuoteUtils.unquoteColumnName(key), key);
+    }
     // Check for extra columns in the row
     List<String> extraCols = new ArrayList<>();
     for (String columnName : inputColNamesMap.keySet()) {
--- a/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
@@ -86,6 +86,8 @@ class DataValidationUtil {
     objectMapper.registerModule(module);
   }
 
+  // Caching the powers of 10 that are used for checking the range of numbers because computing them
+  // on-demand is expensive.
   private static final BigDecimal[] POWER_10 = makePower10Table();
 
   private static BigDecimal[] makePower10Table() {
--- a/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
@@ -833,9 +833,9 @@ class DataValidationUtil {
 
   static void checkValueInRange(
       BigDecimal bigDecimalValue, int scale, int precision, final long insertRowIndex) {
-//    if (bigDecimalValue.abs().compareTo(BigDecimal.TEN.pow(precision - scale)) >= 0) {
-      if (bigDecimalValue.abs().compareTo(POWER_10[precision - scale]) >= 0) {
-        throw new SFException(
+    //    if (bigDecimalValue.abs().compareTo(BigDecimal.TEN.pow(precision - scale)) >= 0) {
+    if (bigDecimalValue.abs().compareTo(POWER_10[precision - scale]) >= 0) {
+      throw new SFException(
           ErrorCode.INVALID_FORMAT_ROW,
           String.format(
               "Number out of representable exclusive range of (-1e%s..1e%s), rowIndex:%d",
--- a/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
@@ -833,7 +833,11 @@ class DataValidationUtil {
 
   static void checkValueInRange(
       BigDecimal bigDecimalValue, int scale, int precision, final long insertRowIndex) {
-    if (bigDecimalValue.abs().compareTo(POWER_10[precision - scale]) >= 0) {
+    BigDecimal comparand =
+        (precision >= scale) && (precision - scale) < POWER_10.length
+            ? POWER_10[precision - scale]
+            : BigDecimal.TEN.pow(precision - scale);
+    if (bigDecimalValue.abs().compareTo(comparand) >= 0) {
       throw new SFException(
           ErrorCode.INVALID_FORMAT_ROW,
           String.format(
--- a/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
@@ -833,7 +833,6 @@ class DataValidationUtil {
 
   static void checkValueInRange(
       BigDecimal bigDecimalValue, int scale, int precision, final long insertRowIndex) {
-    //    if (bigDecimalValue.abs().compareTo(BigDecimal.TEN.pow(precision - scale)) >= 0) {
     if (bigDecimalValue.abs().compareTo(POWER_10[precision - scale]) >= 0) {
       throw new SFException(
           ErrorCode.INVALID_FORMAT_ROW,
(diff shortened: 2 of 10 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -86,6 +86,16 @@ class DataValidationUtil {
     objectMapper.registerModule(module);
   }
 
+  private static final BigDecimal[] POWER_10 = makePower10Table();
```
sfc-gh-tjones, line 89: Nice!

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -823,7 +833,7 @@ static int validateAndParseBoolean(String columnName, Object input, long insertR
 
   static void checkValueInRange(
```
sfc-gh-tjones, line 834: Although unlikely might want to add a `@throws IndexOutOfBoundsException` if `precision-scale` exceeds `Power10.sb16Size`

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -360,6 +361,9 @@ void distributeFlushTasks() {
     List<Pair<BlobData<T>, CompletableFuture<BlobMetadata>>> blobs = new ArrayList<>();
     List<ChannelData<T>> leftoverChannelsDataPerTable = new ArrayList<>();
 
+    // The API states that the number of available processors reported can change and therefore, we
+    // should poll it occasionally.
```
sfc-gh-tjones, line 365: Out of curiosity under what circumstances can this change?

src/main/java/net/snowflake/ingest/streaming/internal/MemoryInfoProviderFromRuntime.java
```diff
@@ -4,20 +4,44 @@
...
 public class MemoryInfoProviderFromRuntime implements MemoryInfoProvider {
-  @Override
-  public long getMaxMemory() {
-    return Runtime.getRuntime().maxMemory();
+  private final long maxMemory;
+  private volatile long totalFreeMemory;
+  private final ScheduledExecutorService executorService;
+
+  public MemoryInfoProviderFromRuntime(long freeMemoryUpdateIntervalMs) {
+    maxMemory = Runtime.getRuntime().maxMemory();
```
sfc-gh-tjones, line 18: How is `maxMemory` calculated? Can this change over time or is this fixed at JVM start?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
```diff
@@ -61,6 +65,9 @@ class SnowflakeStreamingIngestChannelInternal<T> implements SnowflakeStreamingIn
   // The latest cause of channel invalidation
   private String invalidationCause;
 
+  private final MemoryInfoProvider memoryInfoProvider;
+  private volatile long freeMemory = 0;
```
sfc-gh-tjones, line 69: Nit, let's add a unit here, `freeMemoryInBytes`

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-psaha: Good call. It made me realize that I was changing the behavior a bit. I have now changed it to fallback to old behavior if we get unexpected input.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-psaha: I had the same question and looked around a bit but could not find any concrete answers. So this is mostly defensive in case in the future there is a use-case for this.

src/main/java/net/snowflake/ingest/streaming/internal/MemoryInfoProviderFromRuntime.java
sfc-gh-psaha: IIUC, this comes from the `-Xmx` parameter of the JVM and therefore, cannot change.

src/main/java/net/snowflake/ingest/streaming/internal/MemoryInfoProviderFromRuntime.java
```diff
@@ -4,20 +4,51 @@
...
 /** Reads memory information from JVM runtime */
 public class MemoryInfoProviderFromRuntime implements MemoryInfoProvider {
-  @Override
-  public long getMaxMemory() {
-    return Runtime.getRuntime().maxMemory();
+  private final long maxMemory;
+  private volatile long totalFreeMemory;
+  private final ScheduledExecutorService executorService;
+  private static final long FREE_MEMORY_UPDATE_INTERVAL_MS = 1000;
+  private static final MemoryInfoProviderFromRuntime INSTANCE =
```
sfc-gh-psaha, line 17: @sfc-gh-tjones note that I have now made it a singleton since we don't want multiple instances of this in the same JVM all polling for free memory. One downside is that we cannot make the polling interval configurable easily anymore but that does not seem too bad to do it once a second.

sfc-gh-tzhang:
> Overall, 35-40% improvement in the benchmark observed.

Wow, this is huge! Could you add in the describe what types of benchmark did you do? Is it targeted benchmarks for these cases or we're seeing this across all use cases?

sfc-gh-psaha:
> > Overall, 35-40% improvement in the benchmark observed.
> 
> Wow, this is huge! Could you add in the describe what types of benchmark did you do? Is it targeted benchmarks for these cases or we're seeing this across all use cases?

I have the benchmark in this PR and now updated the PR description to describe the benchmark and where the impact will be seen mostly. I think there are probably other areas to improve in the data validation and also in the ParquetRowBuffer code but that's another fight for another day :)

src/main/java/net/snowflake/ingest/streaming/internal/MemoryInfoProviderFromRuntime.java
```diff
@@ -4,20 +4,51 @@
...
+        Runtime.getRuntime().freeMemory() + (maxMemory - Runtime.getRuntime().totalMemory());
+    executorService =
+        new ScheduledThreadPoolExecutor(
+            1,
+            r -> {
+              Thread th = new Thread(r, "MemoryInfoProviderFromRuntime");
+              th.setDaemon(true);
+              return th;
+            });
+    executorService.scheduleAtFixedRate(
```
sfc-gh-tzhang, line 32: I'm not sure updating the memory usage every second is enough, the SDK is memory intensive and we have run into many OOM issues before adding this check. We basically want to stop accepting new rows completely ASAP. I feel like we can ask DeltaStream to call insertRows instead of insertRow to fix this issue, WDYT?

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -400,10 +399,10 @@ public float getSize() {
   Set<String> verifyInputColumns(
       Map<String, Object> row, InsertValidationResponse.InsertError error, int rowIndex) {
     // Map of unquoted column name -> original column name
-    Map<String, String> inputColNamesMap =
-        row.keySet().stream()
-            .collect(Collectors.toMap(LiteralQuoteUtils::unquoteColumnName, value -> value));
-
+    Set<String> originalKeys = row.keySet();
+    Map<String, String> inputColNamesMap = new HashMap<>();
+    originalKeys.forEach(
```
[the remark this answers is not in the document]
sfc-gh-psaha: Yes, esp. for small number of elements [1]. JMH actually showed this clearly. If you know of other places where we use stream operations in a tight loop like we do for `insertRows` then those would be good candidates for optimization as well.

[1] https://www.diva-portal.org/smash/get/diva2:1783234/FULLTEXT01.pdf

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -278,7 +279,7 @@ CompletableFuture<Void> flush(boolean isForce) {
   private void createWorkers() {
     // Create thread for checking and scheduling flush job
     ThreadFactory flushThreadFactory =
-        new ThreadFactoryBuilder().setNameFormat("ingest-flush-thread").build();
+        new ThreadFactoryBuilder().setNameFormat("ingest-flush-thread").setDaemon(true).build();
```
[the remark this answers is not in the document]
sfc-gh-psaha: I have reverted this change. This would have been fine but unneeded since we already shutdown the threadpool when the client is closed. I was having some exceptions in the new benchmark that prevented the shutdown call and I thought we never shutdown the thread pool.

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -400,10 +399,11 @@ public float getSize() {
   Set<String> verifyInputColumns(
       Map<String, Object> row, InsertValidationResponse.InsertError error, int rowIndex) {
     // Map of unquoted column name -> original column name
-    Map<String, String> inputColNamesMap =
-        row.keySet().stream()
-            .collect(Collectors.toMap(LiteralQuoteUtils::unquoteColumnName, value -> value));
-
+    Set<String> originalKeys = row.keySet();
+    Map<String, String> inputColNamesMap = new HashMap<>();
+    for (String key : originalKeys) {
```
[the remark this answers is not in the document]
sfc-gh-psaha: Done!

src/main/java/net/snowflake/ingest/streaming/internal/MemoryInfoProviderFromRuntime.java
sfc-gh-tzhang: My gut feeling tells me that this will introduce some regression, and then we will need to tweak the parameter default values, but let's see :)

sfc-gh-tzhang: I'm wondering what's the common practice for this? sync vs async? We would need to do similar things in Buffering Tier as well, cc: @sfc-gh-japatel

sfc-gh-tzhang (approved):
LGTM, thanks! Have some concerns for the low memory check being async now, but I don't have a better idea.

src/main/java/net/snowflake/ingest/streaming/internal/MemoryInfoProviderFromRuntime.java
sfc-gh-psaha: It is much easier in C++. There is no garbage collection and you can track exactly how much memory is being used by keeping rowset buffers in memory without having to ask the runtime.

sfc-gh-psaha: I don't see any problem with the thread continuing to run - it is fairly cheap at this frequency. Even with per-client singleton, if they don't close the client this and other threads would keep running anyways. I'll merge this but if you feel strongly about this, let's chat and I can do a follow-up.

jdcaperon:
@sfc-gh-psaha if you would like real world samples as well for performance I might be able to provide some data from our production environment.

PR #783, opened 2024-06-24
SNOW-1497358 Support multiple storage for Iceberg mode

sfc-gh-alhuang:
This PR includes several refactoring for `FlushService` and `StreamingIngestInternalStage` to support Iceberg/Non-Iceberg mode file upload

- Create `ConfigureCallHandler` for different configure call includes per client (Non-Iceberg) and per channel (Iceberg mode).
- Create `StorageManager` to manage storage and blob path generation.

Tests for Iceberg mode will be added once the refactor review is done.


src/main/java/net/snowflake/ingest/streaming/internal/StorageManager.java
```diff
@@ -0,0 +1,54 @@
...
+
+  /**
+   * Get the blob path given time and client prefix, used for testing only
+   *
+   * @param calendar the time
+   * @param clientPrefix the client prefix
+   * @return the blob path
+   */
+  @VisibleForTesting
+  String getBlobPath(Calendar calendar, String clientPrefix);
```
sfc-gh-hmadan, line 46: Calendar looks like an odd API to latch onto, will reply to this thread once ive read through the rest of the files in case I have a better suggestion!

src/main/java/net/snowflake/ingest/streaming/internal/ConfigureCallHandler.java
```diff
@@ -0,0 +1,188 @@
...
+  static ConfigureCallHandlerBuilder builder(
+      CloseableHttpClient httpClient,
+      RequestBuilder requestBuilder,
+      ServiceResponseHandler.ApiName apiName,
+      String configureEndpoint) {
+    return new ConfigureCallHandlerBuilder(httpClient, requestBuilder, apiName, configureEndpoint);
+  }
+
+  /** Builder class to build a {@link ConfigureCallHandler} */
+  static class ConfigureCallHandlerBuilder {
```
sfc-gh-hmadan, line 57: A builder pattern is typically used when the object you're trying to build has many many optional arguments at construction time, which results in a lot of constructor signature combinations.
In this case all the arguments are mandatory, so you can just have one ConfigureCallHandler (public) constructor and take all arguments over there, instead of using a builder.

src/main/java/net/snowflake/ingest/streaming/internal/ConfigureCallHandler.java
```diff
@@ -0,0 +1,188 @@
...
+import net.snowflake.ingest.utils.ErrorCode;
+import net.snowflake.ingest.utils.SFException;
+import net.snowflake.ingest.utils.Utils;
+
+/** A class that is used to configure a Snowflake Ingest Storage */
+class ConfigureCallHandler {
+
+  // Object mapper for creating payload, ignore null fields
+  private static final ObjectMapper mapper =
+      new ObjectMapper().setSerializationInclusion(JsonInclude.Include.NON_NULL);
```
sfc-gh-hmadan, line 28: nit: can we reuse object mappers across the process, via some JsonUtils class or something?

src/main/java/net/snowflake/ingest/streaming/internal/ConfigureCallHandler.java
```diff
@@ -0,0 +1,188 @@
...
+   * @return the configure response
+   * @throws IOException
+   */
+  ConfigureResponse makeConfigureCall(String fileName) throws IOException {
+    Map<String, String> payload = makeConfigurePayload();
+    payload.put("file_name", fileName);
+    return makeConfigureCall(payload);
+  }
+
+  private ConfigureResponse makeConfigureCall(Map<String, String> payload) throws IOException {
```
sfc-gh-hmadan, line 158: its really this one API call thats being wrapped by this call. Suggest the following:
- Introduce a SnowflakeServiceClient.java class that takes a CloseableHttpClient as a ctor argument
- Have one method per server API in this class, with the method's signature taking in a strongly typed object (ConfigureClientRequest in this case) as the method argument
- Respond with strongly-typed response objects (ConfigureClientResponse)
- hide all error handling and request serialization and response serialization inside this one class

Since each API call is pretty simple we don't need a separate class/handler object for each API call.

Eventually we should only be passing around this SnowflakeServiceClient.java object around instead of the current practice of passing around CloseableHttpClient.

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java
```diff
@@ -0,0 +1,126 @@
...
+  /**
+   * Given a blob, return the target storage by looking up the table name from the channel context
+   *
+   * @param blobData the blob to upload
+   * @return target storage
+   */
+  @Override
+  public StreamingIngestStorage getStorage(List<List<ChannelData<T>>> blobData) {
+    // Only one chunk per blob in Iceberg mode.
+    ChannelFlushContext channelContext = blobData.get(0).get(0).getChannelContext();
```
sfc-gh-hmadan, line 49: lets pass in the channelContext on the method signature instead of assuming blobData's structure is constructed a certain way by the caller.

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java
```diff
@@ -0,0 +1,126 @@
...
+  /**
+   * Add a storage to the manager by looking up the table name from the open channel response
+   *
+   * @param openChannelResponse response from open channel
+   */
+  @Override
+  public void addStorage(OpenChannelResponse openChannelResponse) {
+    String fullyQualifiedTableName =
+        String.format(
+            "%s.%s.%s",
```
sfc-gh-hmadan, line 72: 1. do we have other places that do this concatenation? lets unify all these fully-qualified-name-construction pieces of code into one util method.
2. How is each component's encoding taken care of? There's some single quotes wrapping/unwrapping business that probably needs to be accounted for.
3. String.format behaves differently for different "cultures" (string.format for english typically behaves differently than string.format for arabic), please investigate how String.format infers the culture - is it some context on the thread? on the process? You might have to explicitly pass in "invariant" culture if such a thing exists.

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java
```diff
@@ -0,0 +1,126 @@
...
+                .build();
+        this.externalVolumeMap.put(
+            fullyQualifiedTableName,
+            new StreamingIngestStorage(
+                isTestMode,
+                configureCallHandler,
+                this.owningClient.getName(),
+                DEFAULT_MAX_UPLOAD_RETRIES));
+      } catch (SnowflakeSQLException | IOException err) {
+        throw new SFException(err, ErrorCode.UNABLE_TO_CONNECT_TO_STAGE);
```
sfc-gh-hmadan, line 97: 1. what happens when this exception is thrown? Does the customer's code get an opportunity to handle this exception, or does it go unhandled and cause a process crash?
2. Lets add retry handling to all API calls made via the new SnowflakeServiceClient I proposed in another thread. Need retry with backoff and jitter.
3. Can we log to whatever log provider is used by the sdk before throwing?

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java
```diff
@@ -0,0 +1,126 @@
...
+  @Override
+  public StreamingIngestStorage getStorage(List<List<ChannelData<T>>> blobData) {
+    // Only one chunk per blob in Iceberg mode.
+    ChannelFlushContext channelContext = blobData.get(0).get(0).getChannelContext();
+    StreamingIngestStorage stage =
+        this.externalVolumeMap.get(channelContext.getFullyQualifiedTableName());
+
+    if (stage == null) {
+      throw new SFException(
+          ErrorCode.INTERNAL_ERROR,
```
sfc-gh-hmadan, line 55: same question as another thread: does customer code get a chance to intercept this exception? Does it cause a process crash (or does it silently get ignored) ?

src/main/java/net/snowflake/ingest/streaming/internal/ConfigureCallHandler.java
sfc-gh-alhuang: The `role`, `table`, `schema`, and `database` are optional ([ref1](https://github.com/snowflakedb/snowflake-ingest-java/blob/44cc28e77b49c60c5d772a25fe00271bb675670e/src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java#L78C8-L88C26), [ref2](https://github.com/snowflakedb/snowflake-ingest-java/blob/44cc28e77b49c60c5d772a25fe00271bb675670e/src/main/java/net/snowflake/ingest/streaming/internal/InternalStageManager.java#L42C9-L49C22)), should we keep this?

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java
```diff
@@ -0,0 +1,126 @@
...
+                .setRole(this.owningClient.getRole())
+                .setDatabase(openChannelResponse.getDBName())
+                .setSchema(openChannelResponse.getSchemaName())
+                .setTable(openChannelResponse.getTableName())
+                .build();
+        this.externalVolumeMap.put(
+            fullyQualifiedTableName,
+            new StreamingIngestStorage(
+                isTestMode,
+                configureCallHandler,
```
sfc-gh-hmadan, line 93: I like the thought process here but here's the problem - you're forcing CLIENT_CONFIGURE and CHANNEL_CONFIGURE APIs to behave almost exactly the same.
- The goal is for StreamingIngestStorage to have a clean way to call the configure API without knowing whether we're in iceberg mode or not, and without knowing which API was called to get that info.
- The StorageManager abstraction is serving pretty much this purpose already
- I suggest passing in the StorageManager interface object to StreamingIngestStorage, and adding a configure() method on that interface that StreamingIngestStorage can call. Let this method return a strongly typed object, and each concrete impl of that interface will call whatever it wants on the service side to get latest tokens / figs ids / etc.

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java
```diff
@@ -0,0 +1,126 @@
...
+  private final boolean isTestMode;
+
+  /**
+   * Constructor for ExternalVolumeManager
+   *
+   * @param isTestMode whether the manager in test mode
+   * @param client the owning client
+   */
+  ExternalVolumeManager(boolean isTestMode, SnowflakeStreamingIngestClientInternal<T> client) {
+    this.owningClient = client;
```
sfc-gh-hmadan, line 35: nit: can replace this with the SnowflakeServiceClient object suggested in another thread.

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java
```diff
@@ -0,0 +1,126 @@
...
+                    this.owningClient.getHttpClient(),
+                    this.owningClient.getRequestBuilder(),
+                    STREAMING_CHANNEL_CONFIGURE,
+                    CHANNEL_CONFIGURE_ENDPOINT)
+                .setRole(this.owningClient.getRole())
+                .setDatabase(openChannelResponse.getDBName())
+                .setSchema(openChannelResponse.getSchemaName())
+                .setTable(openChannelResponse.getTableName())
+                .build();
+        this.externalVolumeMap.put(
```
sfc-gh-hmadan, line 89: what if there were two concurrent addStorage calls for the same table? Suggest taking a lock after if (!containsKey) in line 76 so that we only ever construct one StreamingIngestStorage object.

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java
```diff
@@ -0,0 +1,126 @@
...
+   * Get the client prefix from first external volume in the map
+   *
+   * @return the client prefix
+   */
+  @Override
+  public String getClientPrefix() {
+    if (this.externalVolumeMap.isEmpty()) {
+      return null;
+    }
+    return this.externalVolumeMap.values().iterator().next().getClientPrefix();
```
sfc-gh-hmadan, line 124: lets chat about this, am not clear on what's being done here. Couple issues I see here:
1. return type should be Optional if there's a chance we'll return null
2. why is the first external volume "special"?
3. map.values doesn't typically guarantee what order it iterates in (for some map implementations if not all), so there's no guarantee that every call to map.values.iterator.next will give you the same object.
4. It appears that clientPrefix is a static value that isn't meant to change, so we should pass it in via the ctor if possible?

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -359,11 +317,17 @@ void distributeFlushTasks() {
...
     List<Pair<BlobData<T>, CompletableFuture<BlobMetadata>>> blobs = new ArrayList<>();
     List<ChannelData<T>> leftoverChannelsDataPerTable = new ArrayList<>();
+    boolean isBlobCreated = true;
+    String currentBlobPath = null;
 
     while (itr.hasNext() || !leftoverChannelsDataPerTable.isEmpty()) {
       List<List<ChannelData<T>>> blobData = new ArrayList<>();
       float totalBufferSizeInBytes = 0F;
-      final String blobPath = getBlobPath(this.targetStage.getClientPrefix());
+      if (isBlobCreated) {
```
sfc-gh-hmadan, line 326: you could just call generateBlobPath from outside the while loop in line 321 directly.

can add `if (this.channelCache.size() == 0) { return; }` block at the start of this method to make sure generate is only called when necessary.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -444,11 +408,8 @@ && shouldStopProcessing(
       }
 
       // Kick off a build job
```
sfc-gh-hmadan, line 410: The counter.decrement is gone, which manes blob names can have gaps now?

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -700,7 +632,7 @@ <CD> void invalidateAllChannelsInBlob(
 
   /** Get the server generated unique prefix for this client */
   String getClientPrefix() {
```
sfc-gh-hmadan, line 634: does flushService still need this method which is just passing the call onto storageManager? Or can callers directly call storageManager.getClientPrefix instead?

src/main/java/net/snowflake/ingest/streaming/internal/StorageManager.java
```diff
@@ -0,0 +1,54 @@
...
+  String generateBlobPath();
+
+  /**
+   * Get the blob path given time and client prefix, used for testing only
+   *
+   * @param calendar the time
+   * @param clientPrefix the client prefix
+   * @return the blob path
+   */
+  @VisibleForTesting
```
sfc-gh-hmadan, line 45: Can we remove this method from the interface and have testcases directly call InternalStageManager's getBlobPath method? The testcase can safely do a hard cast to that type?

src/main/java/net/snowflake/ingest/streaming/internal/StorageManager.java
sfc-gh-hmadan: I see this is not a decision you're making and is how things already behave (+we're not changing anything for non-iceberg functionality). Resolving.

src/main/java/net/snowflake/ingest/streaming/internal/InternalStageManager.java
```diff
@@ -0,0 +1,133 @@
...
+   */
+  @Override
+  @SuppressWarnings("unused")
+  public StreamingIngestStorage getStorage(List<List<ChannelData<T>>> blobData) {
+    // There's always only one stage for the client in non-iceberg mode
+    return targetStage;
+  }
+
+  /**
+   * Add storage to the manager. Should not be called in non-iceberg mode.
```
sfc-gh-hmadan, line 76: comment appears untrue; it will still be called with the open channel API's response handling in IngestClientInternal.openChannel?

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStorage.java
```diff
@@ -116,26 +111,43 @@ state to record unknown age.
     }
   }
 
+  /**
+   * Default constructor for external volume
```
sfc-gh-hmadan, line 115: Hmm this looks odd, having different constructors for different flows when we already are trying to abstract away the differences with the StorageManager interface.
Am guessing what's happening here is that in iceberg mode, we first get the fileLocation from open_channel but subsequently call channel_configure for token renewal. Whereas for iceberg mode we get file location and renewal tokens from the same CLIENT_CONFIGURE api call, thus you're doing it this way? If so I have a suggestion here:
1. As suggested in another thread - ExternalVolumemanager's [line 93](https://github.com/snowflakedb/snowflake-ingest-java/pull/783#discussion_r1653348767) - the StreamingIngestStorage class should just take in an object that knows how to renew tokens and doesn't impose client_configure / channel_configure should have the same call handler.
2. Always take in a FileLocationInfo object for both iceberg and non-iceberg modes, so that the "boostrap" codepath is also identical

this means for non-iceberg, you'll have to make an explicit call to client_configure at the callsite of StreamingIngestStorage.ctor to get the FileLocationInfo for the internal stage, which IMO is okay.

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStorage.java
```diff
@@ -250,26 +262,29 @@ synchronized SnowflakeFileTransferMetadataWithAge refreshSnowflakeMetadata(boole
...
     }
 
-    Map<Object, Object> payload = new HashMap<>();
-    payload.put("role", this.role);
-    ConfigureResponse response = this.makeClientConfigureCall(payload);
-
+    ConfigureResponse response = this.configureCallHandler.makeConfigureCall();
     // Do not change the prefix everytime we have to refresh credentials
     if (Utils.isNullOrEmpty(this.clientPrefix)) {
       this.clientPrefix = createClientPrefix(response);
```
sfc-gh-hmadan, line 268: lets get the clientPrefix from StorageManager interface?

sfc-gh-hmadan: and not "create" it here?

src/main/java/net/snowflake/ingest/streaming/internal/StorageManager.java
```diff
@@ -0,0 +1,54 @@
...
+   * @return target stage
+   */
+  StreamingIngestStorage getStorage(List<List<ChannelData<T>>> blobData);
+
+  /**
+   * Add a storage to the manager
+   *
+   * @param openChannelResponse response from open channel
+   */
+  void addStorage(OpenChannelResponse openChannelResponse);
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Changed the signature to `addStorage(dbName, schemaName, tableName, channelContext)`

src/main/java/net/snowflake/ingest/streaming/internal/ConfigureCallHandler.java
sfc-gh-alhuang: It's possible to create JsonUtils with a pool of object mappers to balance between concurrent mapping and memory. Could we add it to backlog and address this later?

src/main/java/net/snowflake/ingest/streaming/internal/ConfigureCallHandler.java
sfc-gh-alhuang: Created `SnowflakeServiceClent`

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java
sfc-gh-alhuang: Created unify method for fully-qualified-name in `StreamingIngestUtils`. I don't think we deal with the cultures issue explicitly now. I think if the naming pattern follows the snowflake rule we should be fine.

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java
sfc-gh-alhuang: Added log for this and retry logic is in `executeWithRetries`. If the error is thrown, the process stop immediately.

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java
sfc-gh-alhuang: This cause a process crash. I don't think this is suppose to happens given that client seems impossible to ingest to a channel without open it via `StreamingIngestClient`

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java
sfc-gh-alhuang: Added `owningManager` for this issue in `StreamingIngestStorage`.

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java
sfc-gh-alhuang: Changed this to `putIfAbsent` with `ConcurrentHashMap` to address concurrent acces.

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java
sfc-gh-alhuang: Moved the `clientPrefix` logic to first configure call.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-alhuang: Removed and call directly from a `StorageManager`

src/main/java/net/snowflake/ingest/streaming/internal/InternalStageManager.java
sfc-gh-alhuang: Fixed.

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStorage.java
sfc-gh-alhuang: Refactored to the same ctor.

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStorage.java
sfc-gh-alhuang: Moved to `StorageManager` level.

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestResponse.java
sfc-gh-alhuang, line 1: Should we moved these API request and response class to a separate package/directory?

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestRequest.java
```diff
@@ -0,0 +1,11 @@
...
+ * Copyright (c) 2024 Snowflake Computing Inc. All rights reserved.
+ */
+
+package net.snowflake.ingest.streaming.internal;
+
+/**
+ * The StreamingIngestRequest interface is a marker interface used for type safety in the {@link
+ * SnowflakeServiceClient} for streaming ingest API request.
+ */
+public interface StreamingIngestRequest {}
```
sfc-gh-hmadan, line 11: Do all inheriting class have a request id field that can be moved into this interface?

sfc-gh-alhuang: No, the configure request doesn't have request id.

src/main/java/net/snowflake/ingest/streaming/internal/ConfigureRequest.java
```diff
@@ -0,0 +1,84 @@
...
+  @JsonProperty("database")
+  private String database;
+
+  @JsonInclude(JsonInclude.Include.NON_NULL)
+  @JsonProperty("schema")
+  private String schema;
+
+  @JsonInclude(JsonInclude.Include.NON_NULL)
+  @JsonProperty("table")
+  private String table;
```
sfc-gh-hmadan, line 26: ClientConfigure does not use these fields, not sure why they're here? If you're thinking we should use one configure requests for both client configure and channel configure, then IMO these should be completely separate contracts. there is no need to have a common contract for these two APIs.

I believe i left a similar comment in the last round but might be mistaken, will review old comments once im done iwth a pass here.

src/main/java/net/snowflake/ingest/streaming/internal/DropChannelRequestInternal.java
```diff
@@ -0,0 +1,88 @@
...
+
+package net.snowflake.ingest.streaming.internal;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import net.snowflake.ingest.streaming.DropChannelRequest;
+import net.snowflake.ingest.utils.Utils;
+
+/** Class used to serialize the {@link DropChannelRequest} */
+public class DropChannelRequestInternal implements StreamingIngestRequest {
```
sfc-gh-hmadan, line 13: nit: just DropChannelRequest is enough? why the internal suffix?

sfc-gh-hmadan: I see there is already a DropChannelRequest class, curious why we need two classes for the same thing?

src/main/java/net/snowflake/ingest/streaming/internal/OpenChannelRequestInternal.java
```diff
@@ -0,0 +1,101 @@
...
+package net.snowflake.ingest.streaming.internal;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import net.snowflake.ingest.streaming.OpenChannelRequest;
+import net.snowflake.ingest.utils.Constants;
+import net.snowflake.ingest.utils.Utils;
+
+/** Class used to serialize the {@link OpenChannelRequest} */
+public class OpenChannelRequestInternal implements StreamingIngestRequest {
```
sfc-gh-hmadan, line 14: same comment as DropChannelRequestInternal - why do we need an OpenChannelRequest and OpenChannelRequestInternal as two separate classes?

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestUtils.java
```diff
@@ -77,7 +80,7 @@ public static void sleepForRetry(int executionCount) {
   static <T extends StreamingIngestResponse> T executeWithRetries(
       Class<T> targetClass,
       String endpoint,
-      Map<Object, Object> payload,
+      StreamingIngestRequest payload,
```
sfc-gh-hmadan, line 83: since its an empty interface, unless you plan to add something to that common interface you can also do the following:

```
 static <TRequest, TResponse extends StreamingIngestResponse> TResponse executeWithRetries(
      Class<TResponse> targetClass,
      String endpoint,
      TRequest payload,
      String message,
      ServiceResponseHandler.ApiName apiName,
      CloseableHttpClient httpClient,
      RequestBuilder requestBuilder)
      throws IOException, IngestResponseException {
```

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeServiceClient.java
```diff
@@ -0,0 +1,226 @@
...
+    return response;
+  }
+
+  private <T extends StreamingIngestResponse> T executeApiRequestWithRetries(
+      StreamingIngestRequest request,
+      Class<T> responseClass,
+      String endpoint,
+      String operation,
+      ServiceResponseHandler.ApiName apiName,
+      ErrorCode errorCode) {
```
sfc-gh-hmadan, line 212: unclear what this errorcode is doing in the request signature. mind adding a javadoc to the method ?

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestResponse.java
```diff
@@ -1,9 +1,16 @@
...
 
+/**
+ * The StreamingIngestResponse class is an abstract class that represents a response from the
+ * Snowflake streaming ingest API. This class provides a common structure for all types of responses
+ * that can be received from the {@link SnowflakeServiceClient}.
+ */
 abstract class StreamingIngestResponse {
   abstract Long getStatusCode();
+
+  abstract String getMessage();
```
sfc-gh-hmadan, line 15: lets not add this to the interface. I just opened a JIRA to stop this intermediate-string business. We should directly serialize from objects to output json, and deserialize from response bytes to json.

src/main/java/net/snowflake/ingest/utils/ErrorCode.java
```diff
@@ -14,7 +14,7 @@ public enum ErrorCode {
   REGISTER_BLOB_FAILURE("0006"),
   OPEN_CHANNEL_FAILURE("0007"),
   BUILD_REQUEST_FAILURE("0008"),
-  CLIENT_CONFIGURE_FAILURE("0009"),
+  CONFIGURE_FAILURE("0009"),
```
sfc-gh-hmadan, line 17: lets not reuse error codes between client configure and channel configure, these are two independent APIs.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeServiceClient.java
sfc-gh-hmadan: ok i see why. Please remove the catch block from this method and let all callers do their own catch + rethrow with whatever errorcode they need.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeServiceClient.java
```diff
@@ -0,0 +1,226 @@
...
+            ConfigureResponse.class,
+            CHANNEL_CONFIGURE_ENDPOINT,
+            "channel configure",
+            STREAMING_CHANNEL_CONFIGURE,
+            ErrorCode.CONFIGURE_FAILURE);
+
+    if (response.getStatusCode() != RESPONSE_SUCCESS) {
+      logger.logDebug(
+          "Channel configure request failed, table={}, message={}",
+          request.getFullyQualifiedTableName(),
```
sfc-gh-hmadan, line 96: why not log the request id too in all of the APIs' log calls?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeServiceClient.java
```diff
@@ -0,0 +1,226 @@
...
+            OPEN_CHANNEL_ENDPOINT,
+            "open channel",
+            STREAMING_OPEN_CHANNEL,
+            ErrorCode.OPEN_CHANNEL_FAILURE);
+
+    if (response.getStatusCode() != RESPONSE_SUCCESS) {
+      logger.logDebug(
+          "Open channel request failed, table={}, message={}",
+          request.getFullyQualifiedTableName(),
+          response.getMessage());
```
sfc-gh-hmadan, line 123: old code logs more info:
```
        logger.logDebug(
            "Open channel request failed, channel={}, table={}, client={}, message={}",
            request.getChannelName(),
            request.getFullyQualifiedTableName(),
            getName(),
            response.getMessage());
```

suggest adding a method to the StreamingIngestRequest interface, called "getStringForLogging(), forcing every request to decide which of its' properties it wants to write out to logs.

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestUtils.java
sfc-gh-hmadan: closing this thread out, just added a comment where i said we should add a common toStringForLogging() method to all request classes (and add this to the interface too)

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeServiceClient.java
```diff
@@ -0,0 +1,226 @@
...
+            DROP_CHANNEL_ENDPOINT,
+            "drop channel",
+            STREAMING_DROP_CHANNEL,
+            ErrorCode.DROP_CHANNEL_FAILURE);
+
+    if (response.getStatusCode() != RESPONSE_SUCCESS) {
+      logger.logDebug(
+          "Drop channel request failed, table={}, message={}",
+          request.getFullyQualifiedTableName(),
+          response.getMessage());
```
sfc-gh-hmadan, line 149: old code is logging more info:

```
        logger.logDebug(
            "Drop channel request failed, channel={}, table={}, client={}, message={}",
            request.getChannelName(),
            request.getFullyQualifiedTableName(),
            getName(),
            response.getMessage());
```

src/main/java/net/snowflake/ingest/streaming/internal/OpenChannelRequestInternal.java
sfc-gh-hmadan: nvm, i see how this is not meant to be a customer-exposed class thus we need this internal class.

src/main/java/net/snowflake/ingest/streaming/internal/DropChannelRequestInternal.java
sfc-gh-hmadan: nvm, i see DropChannelRequest is a publicly exposed class thus we need this internal class.

src/main/java/net/snowflake/ingest/streaming/internal/DropChannelRequestInternal.java
```diff
@@ -0,0 +1,88 @@
...
+ */
+
+package net.snowflake.ingest.streaming.internal;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import net.snowflake.ingest.streaming.DropChannelRequest;
+import net.snowflake.ingest.utils.Utils;
+
+/** Class used to serialize the {@link DropChannelRequest} */
```
sfc-gh-hmadan, line 12: should this (and OpenChannelRequestInternal) really be public classes?

src/main/java/net/snowflake/ingest/streaming/internal/RegisterBlobRequest.java
```diff
@@ -0,0 +1,38 @@
...
+ * Copyright (c) 2024 Snowflake Computing Inc. All rights reserved.
+ */
+
+package net.snowflake.ingest.streaming.internal;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.List;
+
+/** Class used to serialize the blob register request */
+public class RegisterBlobRequest implements StreamingIngestRequest {
```
sfc-gh-hmadan, line 11: public?

src/main/java/net/snowflake/ingest/streaming/internal/OpenChannelRequestInternal.java
```diff
@@ -0,0 +1,101 @@
...
+  private boolean isIceberg;
+
+  @JsonInclude(JsonInclude.Include.NON_NULL)
+  @JsonProperty("offset_token")
+  String offsetToken;
+
+  OpenChannelRequestInternal(
+      String requestId,
+      String role,
+      OpenChannelRequest request,
```
sfc-gh-hmadan, line 46: nit: lets just take all the channel/table/db/schema/offset properties as ctor arguments instead of passing in the OpenChannelRequest object. Preferable to keep public contract and internal contract as insulated from each other as possible.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -331,78 +335,55 @@ public SnowflakeStreamingIngestChannelInternal<?> openChannel(OpenChannelRequest
...
+
+    // Add storage to the storage manager, only for external volume
+    this.storageManager.addStorage(
+        response.getDBName(),
+        response.getSchemaName(),
+        response.getTableName(),
+        response.getStageLocation());
+
+    return channel;
   }
```
sfc-gh-hmadan, line 387: i'd bring back this catch block as it was covering more logic (open_channel API call + local bookkeeping) than what you have now (just the API call).

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -417,55 +398,22 @@ public void dropChannel(DropChannelRequest request) {
...
-            request.getFullyQualifiedTableName(),
-            getName(),
-            response.getMessage());
-        throw new SFException(ErrorCode.DROP_CHANNEL_FAILURE, response.getMessage());
-      }
+    DropChannelRequestInternal dropChannelRequest =
+        new DropChannelRequestInternal(
+            this.storageManager.getClientPrefix() + "_" + counter.getAndIncrement(),
+            this.role,
+            request,
```
sfc-gh-hmadan, line 405: same here, lets pass all the request fields directly instead of passing the request object.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -499,56 +447,36 @@ public Map<String, String> getLatestCommittedOffsetTokens(
...
         }
       }
-
-      return response;
-    } catch (IOException | IngestResponseException e) {
-      throw new SFException(e, ErrorCode.CHANNEL_STATUS_FAILURE, e.getMessage());
     }
+
+    return response;
   }
```
sfc-gh-hmadan, line 480: lets bring this catch block back, as it was covering the reportClientFailure() call too.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -417,55 +398,22 @@ public void dropChannel(DropChannelRequest request) {
...
-    }
+    logger.logInfo(
+        "Drop channel request succeeded, channel={}, table={}, clientSequencer={} client={}",
+        request.getChannelName(),
+        request.getFullyQualifiedTableName(),
+        request instanceof DropChannelVersionRequest
+            ? ((DropChannelVersionRequest) request).getClientSequencer()
+            : null,
+        getName());
   }
```
sfc-gh-hmadan, line 417: lets bring back this catch block?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -619,37 +547,13 @@ void registerBlobs(List<BlobMetadata> blobs, final int executionCount) {
...
-            response.getMessage(),
-            executionCount);
-        throw new SFException(ErrorCode.REGISTER_BLOB_FAILURE, response.getMessage());
-      }
-    } catch (IOException | IngestResponseException e) {
-      throw new SFException(e, ErrorCode.REGISTER_BLOB_FAILURE, e.getMessage());
-    }
+    RegisterBlobRequest request =
+        new RegisterBlobRequest(
+            this.storageManager.getClientPrefix() + "_" + counter.getAndIncrement(),
```
sfc-gh-hmadan, line 552: nit: can we move counter inside serviceClient, instead of making all the API callsites do the requestId computation and getAndIncrement ? ok to do in a future PR / as a P2.

src/main/java/net/snowflake/ingest/streaming/internal/StorageManager.java
```diff
@@ -0,0 +1,55 @@
...
+  ConfigureResponse configure(ConfigureRequest request);
+
+  /**
+   * Generate a unique blob path and increment the blob sequencer
+   *
+   * @return the blob path
+   */
+  String generateBlobPath();
+
+  /** Decrement the blob sequencer */
```
sfc-gh-hmadan, line 46: nit: add in comments why/when this is used. it is an odd method on an interface thus deserves an explanation!

src/main/java/net/snowflake/ingest/streaming/internal/InternalStageManager.java
```diff
@@ -0,0 +1,171 @@
...
+  /**
+   * Constructor for InternalStageManager
+   *
+   * @param isTestMode whether the manager in test mode
+   * @param client the owning client
+   * @param snowflakeServiceClient the Snowflake service client to use for configure calls
+   */
+  InternalStageManager(
+      boolean isTestMode,
+      SnowflakeStreamingIngestClientInternal<T> client,
```
sfc-gh-hmadan, line 46: This client is being passed into stageManager purely for name and role. What if we passed those as explicit fields on this constructor? instead of coupling together these two classes and introducing a cyclic dependency between the classes?
InternalStageManager is a "lower layer" class that ideally should not depend on a "higher layer" class like the StreamingIngestClient.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -560,26 +518,36 @@ BlobMetadata buildAndUpload(String blobPath, List<List<ChannelData<T>>> blobData
 
     blob.blobStats.setBuildDurationMs(buildContext);
 
-    return upload(blobPath, blob.blobBytes, blob.chunksMetadataList, blob.blobStats);
+    return upload(
+        this.storageManager.getStorage(blobData.get(0).get(0).getChannelContext()),
```
sfc-gh-hmadan, line 522: lets chat about this f2f. Looks ripe for a nullref or a logical bug where the wrong channel context gets passed in.
Ideally you want to pass the channelContext into this method as an argument and not code in this get(0).get(0) assumption.

src/main/java/net/snowflake/ingest/streaming/internal/ConfigureRequest.java
sfc-gh-alhuang: Created `ClientConfigureRequest` and `ChannelConfigureRequest`

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeServiceClient.java
sfc-gh-alhuang: Moved the try catch clause to call sites.

src/main/java/net/snowflake/ingest/utils/ErrorCode.java
sfc-gh-alhuang: Created separate error codes.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeServiceClient.java
sfc-gh-alhuang: Added `getStringForLogging()` for every request.

src/main/java/net/snowflake/ingest/streaming/internal/DropChannelRequestInternal.java
sfc-gh-alhuang: Change to package private.

src/main/java/net/snowflake/ingest/streaming/internal/RegisterBlobRequest.java
sfc-gh-alhuang: Changed to package private.

src/main/java/net/snowflake/ingest/streaming/internal/OpenChannelRequestInternal.java
sfc-gh-alhuang: Exposed `database`, `schema`, `table`, `channel` as function signature.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
sfc-gh-alhuang: Reverted.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
sfc-gh-alhuang: Reverted.

src/main/java/net/snowflake/ingest/streaming/internal/StorageManager.java
sfc-gh-alhuang: Added comment.

src/main/java/net/snowflake/ingest/streaming/internal/InternalStageManager.java
sfc-gh-alhuang: Make sense, exposed `role` and `clientName` in new function parameters.

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestResponse.java
sfc-gh-alhuang: The streaming ingest response looks like `{"status_code": 0, "message": "some message"}`. This method is used to log error message from server side.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-alhuang: Moved the getting channel context logic to call site.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
sfc-gh-alhuang: Added as backlog in iceberg file path generation.

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestResponse.java
sfc-gh-hmadan: ah ok, i thought this is the whole response as a string

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java
```diff
@@ -0,0 +1,172 @@
...
+  private final String clientName;
+
+  // role of the owning client
+  private final String role;
+
+  // Reference to the Snowflake service client used for configure calls
+  private final SnowflakeServiceClient snowflakeServiceClient;
+
+  // Client prefix generated by the Snowflake server
+  private final String clientPrefix;
```
sfc-gh-tzhang, line 44: Why do we still need the client prefix for Iceberg case?

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java
```diff
@@ -0,0 +1,172 @@
...
+   *
+   * @param channelFlushContext the channel flush context containing the table name
+   * @return target storage
+   */
+  @Override
+  public StreamingIngestStorage<T, ExternalVolumeLocation> getStorage(
+      ChannelFlushContext channelFlushContext) {
+    // Only one chunk per blob in Iceberg mode.
+    StreamingIngestStorage<T, ExternalVolumeLocation> stage =
+        this.externalVolumeMap.get(channelFlushContext.getFullyQualifiedTableName());
```
sfc-gh-tzhang, line 86: Does `channelFlushContext.getFullyQualifiedTableName` take care of the name resolution? For example, A.B.C and a.b.c are the same table

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java
```diff
@@ -0,0 +1,172 @@
...
+
+    if (stage == null) {
+      throw new SFException(
+          ErrorCode.INTERNAL_ERROR,
+          String.format(
+              "No storage found for table %s", channelFlushContext.getFullyQualifiedTableName()));
```
sfc-gh-tzhang, line 92: nit
```suggestion
              "No external volume found for table %s", channelFlushContext.getFullyQualifiedTableName()));
```

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java
```diff
@@ -0,0 +1,172 @@
...
+
+  /**
+   * Given a channel context, return the target storage by looking up the table name
+   *
+   * @param channelFlushContext the channel flush context containing the table name
+   * @return target storage
+   */
+  @Override
+  public StreamingIngestStorage<T, ExternalVolumeLocation> getStorage(
+      ChannelFlushContext channelFlushContext) {
```
sfc-gh-tzhang, line 83: could we only pass in the table name instead? Looks like that only what we needed

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java
```diff
@@ -0,0 +1,172 @@
...
+
+  public ExternalVolumeLocation(String dbName, String schemaName, String tableName) {
+    this.dbName = dbName;
+    this.schemaName = schemaName;
+    this.tableName = tableName;
+  }
+}
+
+/** Class to manage multiple external volumes */
+class ExternalVolumeManager<T> implements StorageManager<T, ExternalVolumeLocation> {
```
sfc-gh-tzhang, line 30: Where do we remove the entires from the cache?

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java
```diff
@@ -0,0 +1,172 @@
...
+   *
+   * @param dbName the database name
+   * @param schemaName the schema name
+   * @param tableName the table name
+   * @param fileLocationInfo response from open channel
+   */
+  @Override
+  public void addStorage(
+      String dbName, String schemaName, String tableName, FileLocationInfo fileLocationInfo) {
+    String fullyQualifiedTableName = String.format("%s.%s.%s", dbName, schemaName, tableName);
```
sfc-gh-tzhang, line 109: I believe you added an util function for this?

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java
```diff
@@ -0,0 +1,172 @@
...
+
+    try {
+      this.externalVolumeMap.putIfAbsent(
```
sfc-gh-tzhang, line 112: Should we use computeIfAbsent? Also this means that we're that going to update the external volume if there is an entry? Is that correct? Think about a case where customer drops and recreates the same table name with a different external volume, then this would continue upload to the old volume?

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java
```diff
@@ -0,0 +1,172 @@
...
+  /**
+   * Gets the latest file location info (with a renewed short-lived access token) for the specified
+   * location
+   *
+   * @param location A reference to the target location
+   * @param fileName optional filename for single-file signed URL fetch from server
+   * @return the new location information
+   */
+  @Override
+  public FileLocationInfo refreshLocation(
```
sfc-gh-tzhang, line 134: nit
```suggestion
  public FileLocationInfo getRefreshedLocation(
```

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -459,7 +417,13 @@ && shouldStopProcessing(
...
                     () -> {
                       try {
-                        BlobMetadata blobMetadata = buildAndUpload(blobPath, blobData);
+                        // Get the channel flush context from the first channel in the blob. This
+                        // only matters when the client is in Iceberg mode. In Iceberg mode, all
+                        // channels in the blob belong to the same table.
+                        ChannelFlushContext channelFlushContext =
+                            blobData.get(0).get(0).getChannelContext();
+                        BlobMetadata blobMetadata =
+                            buildAndUpload(blobPath, blobData, channelFlushContext);
```
sfc-gh-tzhang, line 426: again, looks like all we need is the table name?

src/main/java/net/snowflake/ingest/streaming/internal/OpenChannelRequestInternal.java
```diff
@@ -0,0 +1,111 @@
...
+
+  @JsonProperty("write_mode")
+  private String writeMode;
+
+  @JsonProperty("is_iceberg")
+  private boolean isIceberg;
+
+  @JsonInclude(JsonInclude.Include.NON_NULL)
+  @JsonProperty("offset_token")
+  String offsetToken;
```
sfc-gh-tzhang, line 41: private?

src/main/java/net/snowflake/ingest/streaming/internal/OpenChannelRequestInternal.java
```diff
@@ -0,0 +1,111 @@
...
+
+  String getFullyQualifiedTableName() {
+    return Utils.getFullyQualifiedTableName(database, schema, table);
+  }
+
+  @Override
+  public String getStringForLogging() {
+    return String.format(
+        "OpenChannelRequestInternal(requestId=%s, role=%s, db=%s, schema=%s, table=%s, channel=%s,"
+            + " writeMode=%s, isIceberg=%s, offsetToken=%s)",
```
sfc-gh-tzhang, line 108: not sure we log offset token in other places. if not, suggest to remove it as it can contain sensitive data

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeServiceClient.java
```diff
@@ -0,0 +1,222 @@
...
+            request,
+            CLIENT_CONFIGURE_ENDPOINT,
+            "client configure",
+            STREAMING_CLIENT_CONFIGURE);
+    if (response.getStatusCode() != RESPONSE_SUCCESS) {
+      logger.logDebug(
+          "Client configure request failed, request={}, message={}",
+          request.getStringForLogging(),
+          response.getMessage());
+      throw new SFException(ErrorCode.CLIENT_CONFIGURE_FAILURE, response.getMessage());
```
sfc-gh-tzhang, line 75: why do we need both the log and the exception?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeServiceClient.java
```diff
@@ -0,0 +1,222 @@
...
+import net.snowflake.ingest.connection.RequestBuilder;
+import net.snowflake.ingest.connection.ServiceResponseHandler;
+import net.snowflake.ingest.utils.ErrorCode;
+import net.snowflake.ingest.utils.Logging;
+import net.snowflake.ingest.utils.SFException;
+
+/**
+ * The SnowflakeServiceClient class is responsible for making API requests to the Snowflake service.
+ */
+class SnowflakeServiceClient {
```
sfc-gh-tzhang, line 35: It looks to me that there're many duplicate logic for each function, is there any room for consolidation?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -332,40 +340,18 @@ public SnowflakeStreamingIngestChannelInternal<?> openChannel(OpenChannelRequest
...
-        logger.logDebug(
-            "Open channel request failed, channel={}, table={}, client={}, message={}",
-            request.getChannelName(),
-            request.getFullyQualifiedTableName(),
-            getName(),
-            response.getMessage());
-        throw new SFException(ErrorCode.OPEN_CHANNEL_FAILURE, response.getMessage());
-      }
+      OpenChannelRequestInternal openChannelRequest =
+          new OpenChannelRequestInternal(
```
sfc-gh-tzhang, line 344: looks like you miss isOffsetTokenProvided?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeServiceClient.java
sfc-gh-tzhang: ok, looks like we also did both before, but the log has more info, like table/channel name, could we make sure to keep the original logic? My assumption is that this is a pure refactoring so let's not change any logic, thanks!

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
sfc-gh-alhuang: The `request.getOffsetToken()` is okay to be null. As `@JsonInclude(JsonInclude.Include.NON_NULL)` in `OpenChannelRequest` will ignore it automatically.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeServiceClient.java
sfc-gh-alhuang: All information including table/channel name are capsulated in `getStringForLogging()`.

src/main/java/net/snowflake/ingest/streaming/internal/OpenChannelRequestInternal.java
sfc-gh-alhuang: Existed [here](https://github.com/snowflakedb/snowflake-ingest-java/blob/f5669d162603f621990538dd82eba5714a1b8faa/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java#L357) in the old logic. Should we remove it?

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java
sfc-gh-alhuang: `getFullyQualifiedTableName` uses the information directly from server response. As long as the table name from server is consist, it should be fine.

src/main/java/net/snowflake/ingest/streaming/internal/OpenChannelRequestInternal.java
sfc-gh-alhuang: Resolved.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-alhuang: Changed signature.

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java
sfc-gh-alhuang: Yes, resolved.

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java
sfc-gh-alhuang: Added `removeStorage` method.

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java
sfc-gh-alhuang: We use it for request id tracking.

src/main/java/net/snowflake/ingest/streaming/internal/ChannelConfigureRequest.java
```diff
@@ -0,0 +1,76 @@
...
+
+package net.snowflake.ingest.streaming.internal;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/** Class used to serialize the client / channel configure request. */
+class ChannelConfigureRequest implements ConfigureRequest {
+  @JsonProperty("role")
+  private String role;
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Moved to base class.

src/main/java/net/snowflake/ingest/streaming/internal/ConfigureRequest.java
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright (c) 2024 Snowflake Computing Inc. All rights reserved.
+ */
+
+package net.snowflake.ingest.streaming.internal;
+
+/** Interface for {@link ChannelConfigureRequest} and {@link ClientConfigureRequest} */
+interface ConfigureRequest extends StreamingIngestRequest {
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Added `role` and `fileName`. Changed to class.

src/main/java/net/snowflake/ingest/streaming/internal/ChannelConfigureRequest.java
```diff
@@ -0,0 +1,76 @@
...
+
+  @JsonProperty("schema")
+  private String schema;
+
+  @JsonProperty("table")
+  private String table;
+
+  @JsonInclude(JsonInclude.Include.NON_NULL)
+  @JsonProperty("file_name")
+  private String fileName;
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Added.

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java
sfc-gh-alhuang: Ah yes this is a problem. If a customer drops and recreates a table with the same name but a different external volume, should all channel flushing to that table be invalidated or continue to flush to the new storage? If we choose to flush to the new storage, should we clear all the data that hasn't been flushed yet?

cc: @sfc-gh-hmadan

sfc-gh-hmadan: good catch! lets track this with a JIRA? I'm leaning towards invalidating old data as a customer doing a drop and recreate is doing a destructive operation in the first place.

sfc-gh-alhuang: [Jira](https://snowflakecomputing.atlassian.net/browse/SNOW-1527938) created. Changed to `put` now as the SDK currently continue streaming to the new table without invalidate old channels or clear channels buffer.

sfc-gh-hmadan (approved):
lgtm!

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolumeManager.java
sfc-gh-tzhang: > As long as the table name from server is consist, it should be fine.

Have you verify this? Basically do an experiment with A.B.C and a.b.c, the fully qualified name should be A.B.C for both case

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
```diff
@@ -101,4 +101,11 @@ void invalidateChannelIfSequencersMatch(
   int getSize() {
     return cache.size();
   }
+
+  /** Get the number of channels for a given table */
+  int getSizePerTable(String fullyQualifiedTableName) {
```
sfc-gh-tzhang, line 106:
```suggestion
  int getChannelCountForTable(String fullyQualifiedTableName) {
```

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
```diff
@@ -101,4 +101,11 @@ void invalidateChannelIfSequencersMatch(
...
+    ConcurrentHashMap<String, SnowflakeStreamingIngestChannelInternal<T>> channelsMapPerTable =
+        cache.get(fullyQualifiedTableName);
```
sfc-gh-tzhang, line 108: Is this thread safe? It's possible that the count changes between these two calls, right?

src/test/java/net/snowflake/ingest/streaming/internal/FlushServiceTest.java
```diff
@@ -93,25 +93,28 @@ private abstract static class TestContext<T> implements AutoCloseable {
     ChannelCache<T> channelCache;
     final Map<String, SnowflakeStreamingIngestChannelInternal<T>> channels = new HashMap<>();
     FlushService<T> flushService;
-    StreamingIngestStage stage;
+    StorageManager<T, InternalStageLocation> storageManager;
```
sfc-gh-tzhang, line 96: Do we have tests for both the storage types?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
sfc-gh-alhuang: It was used to avoid putting `null` value in payload string. As we are changing to json serialization now, using `@JsonInclude(JsonInclude.Include.NON_NULL)` will ignore the `null` fields. Removed `isOffsetTokenProvided` in `OpenChannelRequest`.

sfc-gh-tzhang (approved):
Discussed offline with @sfc-gh-alhuang, thanks!

PR #784, opened 2024-06-25
SNOW-1465503 Check row count in Parquet footer before committing

sfc-gh-lsembera:
This PR implements an additional safety check if the number of rows in the Parquet footer matches the number of metadata rows we've collected. If not, an internal exception is thrown.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
@@ -9,6 +9,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 import net.snowflake.ingest.utils.Constants;
 import net.snowflake.ingest.utils.ErrorCode;
 import net.snowflake.ingest.utils.Logging;
@@ -124,6 +125,8 @@ public class ParquetFlusher implements Flusher<ParquetChunkData> {
 
     if (mergedChannelWriter != null) {
       mergedChannelWriter.close();
+      this.verifyRowCounts(
+          "serializeFromParquetWriteBuffers", mergedChannelWriter, channelsDataPerTable, -1);
     }
     return new SerializationResult(
         channelsMetadataList,
@@ -216,6 +219,9 @@ public class ParquetFlusher implements Flusher<ParquetChunkData> {
     rows.forEach(parquetWriter::writeRow);
     parquetWriter.close();
 
+    this.verifyRowCounts(
+        "serializeFromJavaObjects", parquetWriter, channelsDataPerTable, rows.size());
+
     return new SerializationResult(
         channelsMetadataList,
         columnEpStatsMapCombined,
@@ -223,5 +229,65 @@ public class ParquetFlusher implements Flusher<ParquetChunkData> {
         chunkEstimatedUncompressedSize,
         mergedData,
         chunkMinMaxInsertTimeInMs);
+  }
+
+  /**
+   * Validates that rows count in metadata matches the row count in Parquet footer and the row count
+   * written by the parquet writer
+   *
+   * @param serializationType Serialization type, used for logging purposes only
+   * @param writer Parquet writer writing the data
+   * @param channelsDataPerTable Channel data
+   * @param javaSerializationTotalRowCount Total row count when java object serialization is used.
+   *     Used only for logging purposes if there is a mismatch.
+   */
+  private void verifyRowCounts(
+      String serializationType,
+      BdecParquetWriter writer,
+      List<ChannelData<ParquetChunkData>> channelsDataPerTable,
+      long javaSerializationTotalRowCount) {
+    long parquetTotalRowsWritten = writer.getRowsWritten();
+
+    List<Long> parquetFooterRowsPerBlock = writer.getRowCountFromFooter();
+    long parquetTotalRowsInFooter = 0;
+    for (long perBlockCount : parquetFooterRowsPerBlock) parquetTotalRowsInFooter += perBlockCount;
+
+    long totalRowsInMetadata = 0;
+    for (ChannelData<ParquetChunkData> channelData : channelsDataPerTable)
+      totalRowsInMetadata += channelData.getRowCount();
+
+    if (parquetTotalRowsInFooter != totalRowsInMetadata
+        || parquetTotalRowsWritten != totalRowsInMetadata) {
+
+      final String perChannelRowCountsInMetadata =
+          channelsDataPerTable.stream()
+              .map(x -> String.valueOf(x.getRowCount()))
+              .collect(Collectors.joining(","));
+
+      final String perBlockRowCountsInFooter =
+          parquetFooterRowsPerBlock.stream().map(String::valueOf).collect(Collectors.joining(","));
+
+      final long channelsCountInMetadata = channelsDataPerTable.size();
+
+      throw new SFException(
+          ErrorCode.INTERNAL_ERROR,
+          String.format(
+              "[%s]The number of rows in Parquet does not match the number of rows in metadata. "
+                  + "parquetTotalRowsInFooter=%d "
+                  + "totalRowsInMetadata=%d "
+                  + "parquetTotalRowsWritten=%d "
+                  + "perChannelRowCountsInMetadata=%s "
+                  + "perBlockRowCountsInFooter=%s "
+                  + "channelsCountInMetadata=%d "
+                  + "countOfSerializedJavaObjects=%d",
+              serializationType,
+              parquetTotalRowsInFooter,
+              totalRowsInMetadata,
+              parquetTotalRowsWritten,
+              perChannelRowCountsInMetadata,
+              perBlockRowCountsInFooter,
+              channelsCountInMetadata,
+              javaSerializationTotalRowCount));
+    }
   }
 }
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
@@ -248,7 +248,7 @@ public class ParquetFlusher implements Flusher<ParquetChunkData> {
       long javaSerializationTotalRowCount) {
     long parquetTotalRowsWritten = writer.getRowsWritten();
 
-    List<Long> parquetFooterRowsPerBlock = writer.getRowCountFromFooter();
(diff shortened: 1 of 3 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -112,6 +115,23 @@ private SerializationResult serializeFromParquetWriteBuffers(
       }
 
       rowCount += data.getRowCount();
+
+      // We check if the number of rows collectively written for channels encountered so far matches
+      // the number of rows in metadata
+      if (mergedChannelWriter.getRowsWritten() != rowCount) {
```
sfc-gh-psaha, line 121: Can this be potentially racy i.e. other threads writing data via the writer concurrently? Or is the `mergedChannelWriter` only used from this current thread?

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
```diff
@@ -35,6 +36,7 @@
 public class BdecParquetWriter implements AutoCloseable {
   private final InternalParquetRecordWriter<List<Object>> writer;
   private final CodecFactory codecFactory;
+  private final AtomicLong rowsWritten = new AtomicLong(0);
```
sfc-gh-psaha, line 39: Can it be just a `volatile`? This goes back to my previous question about races. If there is no concurrent access (but non-overlapping access from different threads), then `volatile` would be sufficient.

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -152,6 +160,50 @@ static <T> Blob constructBlobAndMetadata(
...
 
+  /**
+   * Safety check to verify whether the number of rows in the parquet footer matches the number of
+   * rows in metadata
+   */
+  static <T> void verifyRowCounts(
+      Flusher.SerializationResult serializationResult,
+      List<ChannelData<T>> channelsDataPerTable,
+      byte[] paddedChunkData,
+      int chunkLength) {
```
sfc-gh-gdoci, line 171: Another idea would be to add the check in the Parquet flusher inside `serializeFromX` methods right before we return the serialization result. We can get the per block row counts from the `parquetWriter` (through a new method) directly since the writer has access to the footer. So we wouldn't need to read and create a new `BdecInputFile`.

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
sfc-gh-lsembera: In fact, in this case, we don't even need `volatile`. The field is accessed either within a lock, or in a BG thread with happens-before relationship with the spawning thread.

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
sfc-gh-lsembera: Thanks, this simplifies the code pretty significantly. I updated the PR.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -224,4 +230,64 @@ private SerializationResult serializeFromJavaObjects(
...
   }
+
+  /**
+   * Validates that rows count in metadata matches the row count in Parquet footer and the row count
+   * written by the parquet writer
+   *
+   * @param serializationType Serialization type, used for logging purposes only
+   * @param writer Parquet writer writing the data
+   * @param channelsDataPerTable Channel data
+   * @param javaSerializationTotalRowCount Total row count when java object serialization is used.
```
sfc-gh-tjones, line 241: What do you mean by `java object serialization is used` - looks like above this is just the total row count?

sfc-gh-lsembera: The SDK implements two types of row buffering. The default one, which buffers rows as lists of java objects in memory and during flush it iterates over all of them and writes them to Parquet. There is also an alternative implementation called internal buffering implemented by @sfc-gh-azagrebin, which doesn't collect any Java values, but which instead directly serializes rows to Parquet. So, for the default buffering method (java objects), we have 4 sources of row counts that all have to match:

1. size of the buffered `List<List<Object>>`
2. number of rows written by the Parquet writer
3. number of rows in Parquet footer
4. number of rows collected for metadata

For the internal buffering optimization, we only have 2nd, 3rd and 4th.

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
```diff
@@ -100,14 +103,28 @@ public BdecParquetWriter(
             encodingProps);
   }
 
+  /** @return List of row counts per block stored in the parquet footer */
+  public List<Long> getRowCountFromFooter() {
```
sfc-gh-gdoci, line 107: nit: `rowCount(s)`.

sfc-gh-gdoci (approved):
Thank you, lgtm.

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
sfc-gh-lsembera: Fixed

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -224,4 +230,64 @@ private SerializationResult serializeFromJavaObjects(
...
+    for (long perBlockCount : parquetFooterRowsPerBlock) parquetTotalRowsInFooter += perBlockCount;
+
+    long totalRowsInMetadata = 0;
+    for (ChannelData<ParquetChunkData> channelData : channelsDataPerTable)
+      totalRowsInMetadata += channelData.getRowCount();
+
+    if (parquetTotalRowsInFooter != totalRowsInMetadata
+        || parquetTotalRowsWritten != totalRowsInMetadata) {
+
+      final String perChannelRowCountsInMetadata =
```
sfc-gh-azagrebin, line 262: would it make sense to dump the whole `List<ChannelData<ParquetChunkData>>`, if ChannelData/ParquetChunkData get the proper toString methods? because it will be not clear what channel misbehaved.

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
```diff
@@ -35,6 +37,7 @@
 public class BdecParquetWriter implements AutoCloseable {
   private final InternalParquetRecordWriter<List<Object>> writer;
   private final CodecFactory codecFactory;
+  private long rowsWritten = 0;
```
sfc-gh-azagrebin, line 40:
```
rowsWritten = 0;
```
also in close?

sfc-gh-azagrebin (approved):
Thanks Lukas!

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -216,6 +219,9 @@ private SerializationResult serializeFromJavaObjects(
     rows.forEach(parquetWriter::writeRow);
     parquetWriter.close();
 
+    this.verifyRowCounts(
+        "serializeFromJavaObjects", parquetWriter, channelsDataPerTable, rows.size());
```
sfc-gh-azagrebin, line 223: it would be good to use `rowCount` calculated here outside of `verifyRowCounts` for the check.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -224,4 +230,64 @@ private SerializationResult serializeFromJavaObjects(
...
+          channelsDataPerTable.stream()
+              .map(x -> String.valueOf(x.getRowCount()))
+              .collect(Collectors.joining(","));
+
+      final String perBlockRowCountsInFooter =
+          parquetFooterRowsPerBlock.stream().map(String::valueOf).collect(Collectors.joining(","));
+
+      final long channelsCountInMetadata = channelsDataPerTable.size();
+
+      throw new SFException(
```
sfc-gh-tzhang, line 272: Could you make sure this is being sent to Snowflake?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -224,4 +230,64 @@ private SerializationResult serializeFromJavaObjects(
...
+      long javaSerializationTotalRowCount) {
+    long parquetTotalRowsWritten = writer.getRowsWritten();
+
+    List<Long> parquetFooterRowsPerBlock = writer.getRowCountsFromFooter();
+    long parquetTotalRowsInFooter = 0;
+    for (long perBlockCount : parquetFooterRowsPerBlock) parquetTotalRowsInFooter += perBlockCount;
+
+    long totalRowsInMetadata = 0;
+    for (ChannelData<ParquetChunkData> channelData : channelsDataPerTable)
+      totalRowsInMetadata += channelData.getRowCount();
```
sfc-gh-tzhang, line 257: IIRC the best practice recommends to always add curly brackets
```suggestion
    for (ChannelData<ParquetChunkData> channelData : channelsDataPerTable)
    {
      totalRowsInMetadata += channelData.getRowCount();
      }
```

sfc-gh-tzhang (approved):
LGTM, please make sure the error is being sent to SF to avoid the case where customer doesn't have the log, thanks!

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
sfc-gh-lsembera: What do you mean by sent to Snowflake? Is it possible via telemetry?

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
sfc-gh-lsembera: Closing the writer doesn't clean up the written rows, so se should reset the counter. For example, the Parquet footer is only accessible after closing the writer.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
sfc-gh-lsembera: I added a log line with channel names. We should not log full channel data because it contains user data.

PR #785, opened 2024-06-25
[DO NOT REVIEW] Add Microbenchmark for Insert rows

sfc-gh-psaha:


sfc-gh-japatel (commented):
not sure how quick these tests are, but this tests can be added in end to end jar package which doesnt impact existing jar. also keeps the current jar size unaltered after adding dependency.

PR #788, opened 2024-07-03
SNOW-1512047 Introduce independent per-table flushes when interleaving is disabled

sfc-gh-alhuang:
The current SDK behavior flushes all channels simultaneously when any buffer reaches its limit, potentially causing unnecessary small file flushes if interleaving is disabled (`MAX_CHUNKS_IN_BLOB_AND_REGISTRATION_REQUEST = 1`) and the ingestion throughput between tables is uneven.

As the WIP streaming to Iceberg table feature set `MAX_CHUNKS_IN_BLOB_AND_REGISTRATION_REQUEST = 1`. This PR introduces per-table flushing to avoid the above issue.

[JIRA](https://snowflakecomputing.atlassian.net/browse/SNOW-1512047)


src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
```diff
@@ -23,6 +24,12 @@ class ChannelCache<T> {
           String, ConcurrentHashMap<String, SnowflakeStreamingIngestChannelInternal<T>>>
       cache = new ConcurrentHashMap<>();
 
+  // Last flush time for each table, the key is FullyQualifiedTableName.
```
sfc-gh-hmadan, line 27: Lets introduce a TableCache to track per-table information. Most of the changes in this class will end up moving over.

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
```diff
@@ -43,6 +56,46 @@ void addChannel(SnowflakeStreamingIngestChannelInternal<T> channel) {
     }
   }
 
+  /**
+   * Get the last flush time for a table
+   *
+   * @param fullyQualifiedTableName fully qualified table name
+   * @return last flush time in milliseconds
+   */
+  Long getLastFlushTime(String fullyQualifiedTableName) {
```
sfc-gh-hmadan, line 65: Optional<Long> to force the caller to make sure that what's returned is not null?

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
```diff
@@ -53,6 +106,20 @@ void addChannel(SnowflakeStreamingIngestChannelInternal<T> channel) {
...
 
+  /**
+   * Returns an iterator over the (table, channels) in this map, filtered by the given table name
+   * set
+   *
+   * @param tableNames the set of table names to filter
+   * @return
+   */
+  Iterator<Map.Entry<String, ConcurrentHashMap<String, SnowflakeStreamingIngestChannelInternal<T>>>>
+      iterator(Set<String> tableNames) {
```
sfc-gh-hmadan, line 117: Do you expect a lot of places to be doing this filtered iteration? If its just one callsite you could just do the if check at the callsite? Esp since the iterator is already exposed in the previous method, we don't need to add a method here for every kind of usage of the iterator.

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
sfc-gh-hmadan: nvm, read through the whole thing and what you have looks okay.

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
```diff
@@ -101,4 +168,10 @@ void invalidateChannelIfSequencersMatch(
   int getSize() {
     return cache.size();
   }
+
```
sfc-gh-hmadan, line 171: 1. javadoc missing
2. can you place this method next to the iterator() method?
3. Once you expose the entrySet(), the caller can just call .iterator() on top of the returned entrySet. Thus can probably remove the iterator() method from this class now.
4. The exposed Set<> is mutable. We should not allow callsites to mutate the internal state of this class; Can you wrap the returned object in a `Collections.unmodifiableSet(cache.entrySet())` ?

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
```diff
@@ -43,6 +56,46 @@ void addChannel(SnowflakeStreamingIngestChannelInternal<T> channel) {
...
+  }
+
+  /**
+   * Set the last flush time for a table as the current time
+   *
+   * @param fullyQualifiedTableName fully qualified table name
+   * @param lastFlushTime last flush time in milliseconds
+   */
+  void setLastFlushTime(String fullyQualifiedTableName, Long lastFlushTime) {
+    this.lastFlushTime.put(fullyQualifiedTableName, lastFlushTime);
```
sfc-gh-hmadan, line 76: I see jitter when adding to the map in addChannel, but nothing here. 
1. Is there any other place where jitter was added?
2. Can you call setLastFlushTime() from addChannel() so there's only one place where lastFlushTIme.put happens?

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -204,36 +201,43 @@ private CompletableFuture<Void> statsFuture() {
...
+  private CompletableFuture<Void> distributeFlush(
+      boolean isForce, Set<String> tablesToFlush, Long timeDiffMillis) {
     return CompletableFuture.runAsync(
         () -> {
-          logFlushTask(isForce, timeDiffMillis);
-          distributeFlushTasks();
-          this.isNeedFlush = false;
+          logFlushTask(isForce, tablesToFlush, timeDiffMillis);
+          distributeFlushTasks(tablesToFlush);
           this.lastFlushTime = System.currentTimeMillis();
```
sfc-gh-hmadan, line 214: curious, why are we still maintaining this.lastFlushTime in this class? Can you update the comment where the member is declared and clarify how there's a per-table lastFlushTime in ChannelCache + how this is different from that?

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -249,27 +253,57 @@ private CompletableFuture<Void> registerFuture() {
...
-    long timeDiffMillis = System.currentTimeMillis() - this.lastFlushTime;
+    long currentTime = System.currentTimeMillis();
+    long timeDiffMillis = currentTime - this.lastFlushTime;
+
+    Set<String> tablesToFlush =
+        this.channelCache.entrySet().stream()
+            .filter(
+                entry ->
+                    isForce
+                        || currentTime - this.channelCache.getLastFlushTime(entry.getKey())
```
sfc-gh-hmadan, line 290: handle null / Optional.empty

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -249,27 +253,57 @@ private CompletableFuture<Void> registerFuture() {
...
+                        || this.channelCache.getNeedFlush(entry.getKey()))
+            .map(Map.Entry::getKey)
+            .collect(Collectors.toSet());
+
+    // Flush every table together when it's interleaving chunk is allowed
+    if (this.owningClient.getParameterProvider().getMaxChunksInBlobAndRegistrationRequest() != 1
+        && !tablesToFlush.isEmpty()) {
+      tablesToFlush.addAll(
+          this.channelCache.entrySet().stream().map(Map.Entry::getKey).collect(Collectors.toSet()));
+    }
```
sfc-gh-hmadan, line 301: Why do we need to do this, If the previous code block already picked up the minimal set of tables needing flush?

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
```diff
@@ -33,6 +40,12 @@ void addChannel(SnowflakeStreamingIngestChannelInternal<T> channel) {
         this.cache.computeIfAbsent(
             channel.getFullyQualifiedTableName(), v -> new ConcurrentHashMap<>());
 
+    // Update the last flush time for the table, add jitter to avoid all channels flush at the same
+    // time when the blobs are not interleaved
+    this.lastFlushTime.putIfAbsent(
```
sfc-gh-hmadan, line 45: not sure i understand what this helps with. If someone does addChannel, and doesn't add any data for a minute, the first row that they add will trigger a flush since we'll mistakenly think its been a long time since the last flush.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-asen: +1 This looks like it is only used for logging the time since the last flush but now the flush delta is per table

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -249,27 +253,57 @@ private CompletableFuture<Void> registerFuture() {
...
    * @return Completable future that will return when the blobs are registered successfully, or null
    *     if none of the conditions is met above
    */
   CompletableFuture<Void> flush(boolean isForce) {
-    long timeDiffMillis = System.currentTimeMillis() - this.lastFlushTime;
+    long currentTime = System.currentTimeMillis();
+    long timeDiffMillis = currentTime - this.lastFlushTime;
+
+    Set<String> tablesToFlush =
+        this.channelCache.entrySet().stream()
```
sfc-gh-asen, line 286: nit: this.channelCache.keySet() would simplify the logic since you don't use the values

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-asen: +1, even if interleaving is enabled, I'd prefer to keep the above logic for flushing and wait until the MaxClientLag for each channel

sfc-gh-asen (commented):
LGTM, had small comments on top of Hitesh's comments

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
sfc-gh-alhuang: Changed to `Optional<Long>` and verified at caller site.

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
sfc-gh-alhuang: Use `entrySet()` and `keySet()`.

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
sfc-gh-alhuang: Added.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-alhuang: Removed and logged `flushTimeDiffMill` per table

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-alhuang: Resolved.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-alhuang: Modified, thanks for the tip!

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-alhuang: I aimed to maintain the original interleaving behavior, where all channels are flushed if any channel needs it. With independent flushing intervals, we might miss the chance to combine multiple chunks into the same BDEC. A potential workaround is to discretize timestamps and reduce jitter on `lastFlushTime` in interleaving mode. This can increase the chances of combining multiple chunks into the same blob. What do you think?

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
sfc-gh-alhuang: Yes. Should we change the logic to following?
1. Set `lastFlushTime` to null when create a channel.
2. When calling `putRow` or `putRows`, if `lastFlushTime` is null, set to current time.
3. Whenever a table is flushed, set `lastFlushTime` to null, go to step 2.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -204,36 +201,43 @@ private CompletableFuture<Void> statsFuture() {
...
+    boolean isNeedFlush = tablesToFlush.stream().anyMatch(channelCache::getNeedFlush);
+
     final String flushTaskLogFormat =
         String.format(
             "Submit forced or ad-hoc flush task on client=%s, isForce=%s,"
                 + " isNeedFlush=%s, timeDiffMillis=%s, currentDiffMillis=%s",
             this.owningClient.getName(),
             isForce,
-            this.isNeedFlush,
+            isNeedFlush,
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Added

sfc-gh-hmadan (approved):
LGTM, please get Alkin's or Toby's signoff before merging!

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
sfc-gh-hmadan: as discussed pl track this with a JIRA so we don't forget about it.

sfc-gh-alhuang: [Jira](https://snowflakecomputing.atlassian.net/browse/SNOW-1531922) created.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -249,27 +260,56 @@ private CompletableFuture<Void> registerFuture() {
...
+                            && flushStartTime - this.channelCache.getLastFlushTime(key).get()
+                                >= this.owningClient
+                                    .getParameterProvider()
+                                    .getCachedMaxClientLagInMs())
+                        || this.channelCache.getNeedFlush(key))
+            .collect(Collectors.toSet());
+
+    // Flush every table together when it's interleaving chunk is allowed
+    if (this.owningClient.getParameterProvider().getMaxChunksInBlobAndRegistrationRequest() != 1
+        && !tablesToFlush.isEmpty()) {
```
sfc-gh-tzhang, line 306: Can we do the check before populating `tablesToFlush`?

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -204,36 +196,55 @@ private CompletableFuture<Void> statsFuture() {
...
     return CompletableFuture.runAsync(
         () -> {
-          logFlushTask(isForce, timeDiffMillis);
-          distributeFlushTasks();
-          this.isNeedFlush = false;
-          this.lastFlushTime = System.currentTimeMillis();
-          return;
+          logFlushTask(isForce, tablesToFlush, flushStartTime);
+          distributeFlushTasks(tablesToFlush);
+          long flushEndTime = System.currentTimeMillis();
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Renamed to `prevFlushEndTime`.

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
```diff
@@ -23,6 +25,12 @@ class ChannelCache<T> {
           String, ConcurrentHashMap<String, SnowflakeStreamingIngestChannelInternal<T>>>
       cache = new ConcurrentHashMap<>();
 
+  // Last flush time for each table, the key is FullyQualifiedTableName.
+  private final ConcurrentHashMap<String, Long> lastFlushTime = new ConcurrentHashMap<>();
+
+  // Need flush flag for each table, the key is FullyQualifiedTableName.
+  private final ConcurrentHashMap<String, Boolean> needFlush = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Discussed offline, created a class for flush information.

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
```diff
@@ -101,4 +155,8 @@ void invalidateChannelIfSequencersMatch(
   int getSize() {
     return cache.size();
   }
+
+  private long getRandomFlushJitter() {
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Comment added.

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
```diff
@@ -44,13 +57,54 @@ void addChannel(SnowflakeStreamingIngestChannelInternal<T> channel) {
...
 
   /**
-   * Returns an iterator over the (table, channels) in this map.
+   * Get the last flush time for a table
+   *
+   * @param fullyQualifiedTableName fully qualified table name
+   * @return last flush time in milliseconds
+   */
+  Optional<Long> getLastFlushTime(String fullyQualifiedTableName) {
+    return Optional.ofNullable(this.lastFlushTime.get(fullyQualifiedTableName));
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Added assertion in the method.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-alhuang: Discussed offline. Preserve the client level `lastFlushTime` and `isNeedFlush` to avoid checking table level flush info when interleaving is enabled which might cause performance change. Preserve old logging format when interleaving is enabled to avoid logging too much information.

cc: @sfc-gh-hmadan

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
```diff
@@ -44,13 +67,73 @@ void addChannel(SnowflakeStreamingIngestChannelInternal<T> channel) {
...
+  /**
+   * Set the last flush time for a table as the current time
+   *
+   * @param fullyQualifiedTableName fully qualified table name
+   * @param lastFlushTime last flush time in milliseconds
+   */
+  void setLastFlushTime(String fullyQualifiedTableName, Long lastFlushTime) {
+    this.tableFlushInfo.compute(
+        fullyQualifiedTableName,
+        (k, v) -> new FlushInfo(lastFlushTime + getRandomFlushJitter(), v != null && v.needFlush));
```
sfc-gh-tzhang, line 94: v should never be null, right?

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
```diff
@@ -44,13 +67,73 @@ void addChannel(SnowflakeStreamingIngestChannelInternal<T> channel) {
...
+   * @return last flush time in milliseconds
+   */
+  Long getLastFlushTime(String fullyQualifiedTableName) {
+    FlushInfo tableFlushInfo = this.tableFlushInfo.get(fullyQualifiedTableName);
+    if (tableFlushInfo == null) {
+      throw new SFException(
+          ErrorCode.INTERNAL_ERROR,
+          String.format("Last flush time for table %s not found", fullyQualifiedTableName));
+    }
+    return tableFlushInfo.lastFlushTime;
```
sfc-gh-tzhang, line 82: Do we need to guarantee synchronization between get and set flush time?

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
```diff
@@ -44,13 +67,73 @@ void addChannel(SnowflakeStreamingIngestChannelInternal<T> channel) {
...
    */
-  Iterator<Map.Entry<String, ConcurrentHashMap<String, SnowflakeStreamingIngestChannelInternal<T>>>>
-      iterator() {
-    return this.cache.entrySet().iterator();
+  void setNeedFlush(String fullyQualifiedTableName, Boolean needFlush) {
+    this.tableFlushInfo.compute(
+        fullyQualifiedTableName,
+        (k, v) ->
+            new FlushInfo(
+                v != null ? v.lastFlushTime : System.currentTimeMillis() + getRandomFlushJitter(),
```
sfc-gh-tzhang, line 124: same question here, can the value be null?

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
```diff
@@ -44,13 +67,73 @@ void addChannel(SnowflakeStreamingIngestChannelInternal<T> channel) {
...
+   * Set need flush flag for a table
    *
-   * @return
+   * @param fullyQualifiedTableName fully qualified table name
+   * @param needFlush need flush flag
    */
-  Iterator<Map.Entry<String, ConcurrentHashMap<String, SnowflakeStreamingIngestChannelInternal<T>>>>
-      iterator() {
-    return this.cache.entrySet().iterator();
+  void setNeedFlush(String fullyQualifiedTableName, Boolean needFlush) {
```
sfc-gh-tzhang, line 119: why needFlush is Boolean instead of boolean?

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
```diff
@@ -101,4 +184,9 @@ void invalidateChannelIfSequencersMatch(
   int getSize() {
     return cache.size();
   }
+
+  /** Generate a random flush jitter to avoid all channels flush at the same time */
+  private long getRandomFlushJitter() {
```
sfc-gh-tzhang, line 189: Discussed offline, let's remove this

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -162,36 +167,65 @@ private CompletableFuture<Void> statsFuture() {
...
-          this.lastFlushTime = System.currentTimeMillis();
-          return;
+          logFlushTask(isForce, tablesToFlush, flushStartTime);
+          distributeFlushTasks(tablesToFlush);
+          long prevFlushEndTime = System.currentTimeMillis();
+          tablesToFlush.forEach(
+              table -> {
+                this.lastFlushTime = prevFlushEndTime;
+                this.channelCache.setLastFlushTime(table, prevFlushEndTime);
+                this.isNeedFlush = false;
```
sfc-gh-tzhang, line 185: Is this correct, isNeedFlush is not at table level, right?

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
```diff
@@ -276,7 +276,7 @@ public void testDropChannel() throws Exception {
   @Test
   public void testParameterOverrides() throws Exception {
     Map<String, Object> parameterMap = new HashMap<>();
-    parameterMap.put(ParameterProvider.MAX_CLIENT_LAG, "3 sec");
```
sfc-gh-tzhang, line 279: why this is changed?

sfc-gh-tzhang (approved):
Left some comments, otherwise LGTM

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
sfc-gh-alhuang: Modified to null assertion.

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
sfc-gh-alhuang: Yes.The results of `get` will not be modified from other threads as we are creating new `FlushInfo` every time when calling `set`.

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
sfc-gh-alhuang: Modified to null assertion.

src/main/java/net/snowflake/ingest/streaming/internal/ChannelCache.java
sfc-gh-alhuang: Both works, change to boolean.

src/test/java/net/snowflake/ingest/streaming/internal/StreamingIngestIT.java
sfc-gh-alhuang: The ParameterProvider does not support "sec" ([ref](https://github.com/snowflakedb/snowflake-ingest-java/blob/f69259a81d849948dabf72af88d48f7e1d94e7f6/src/main/java/net/snowflake/ingest/utils/ParameterProvider.java#L225C10-L228C19)). The old code somehow ignore the exception thrown in thread.

jdcaperon:
Thanks for solving this issue :D https://github.com/snowflakedb/snowflake-ingest-java/issues/570

PR #789, opened 2024-07-05
SNOW-1512935 Retry SocketTimeoutException

sfc-gh-lsembera:
This PR makes the SDK retry requests that fail with `SocketTimeoutException`.

### DIFF
--- a/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
+++ b/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
@@ -294,7 +294,8 @@ public class HttpUtil {
       if (exception instanceof NoHttpResponseException
           || exception instanceof javax.net.ssl.SSLException
           || exception instanceof java.net.SocketException
-          || exception instanceof java.net.UnknownHostException) {
+          || exception instanceof java.net.UnknownHostException
+          || exception instanceof java.net.SocketTimeoutException) {
         LOGGER.info(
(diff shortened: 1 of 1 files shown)


src/main/java/net/snowflake/ingest/utils/HttpUtil.java
```diff
@@ -294,7 +294,8 @@ static HttpRequestRetryHandler getHttpRequestRetryHandler() {
       if (exception instanceof NoHttpResponseException
           || exception instanceof javax.net.ssl.SSLException
           || exception instanceof java.net.SocketException
-          || exception instanceof java.net.UnknownHostException) {
+          || exception instanceof java.net.UnknownHostException
+          || exception instanceof java.net.SocketTimeoutException) {
```
sfc-gh-tzhang, line 298: Do we have a good reason to add this? Just want to make sure we are not blindly adding this that masks any bugs in the code.

sfc-gh-tzhang (approved):
LGTM, just want to make sure we're not masking any potential bugs in the code by adding retries, thanks!

src/main/java/net/snowflake/ingest/utils/HttpUtil.java
sfc-gh-lsembera: Yeah, this is a network connectivity issue, which we are not retrying, which leads to channel invalidation.

PR #794, opened 2024-07-15
Reject new stage metadata if the deployment id does not match what the client was created with

sfc-gh-psaha:
This prevents issues when there is transparent failover using Client Redirect or another similar mechanism. Along the way, I found some missing/misconfigured error message resource and fixed them.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
@@ -25,6 +25,7 @@ import java.util.Optional;
 import java.util.Properties;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
+import javax.annotation.Nullable;
 import net.snowflake.client.core.OCSPMode;
 import net.snowflake.client.jdbc.SnowflakeFileTransferAgent;
 import net.snowflake.client.jdbc.SnowflakeFileTransferConfig;
@@ -91,6 +92,7 @@ class StreamingIngestStage {
   private final String role;
   private final String clientName;
   private String clientPrefix;
+  private Long deploymentId;
 
   private final int maxUploadRetries;
 
@@ -258,9 +260,11 @@ class StreamingIngestStage {
     payload.put("role", this.role);
     Map<String, Object> response = this.makeClientConfigureCall(payload);
 
-    JsonNode responseNode = this.parseClientConfigureResponse(response);
+    JsonNode responseNode = this.parseClientConfigureResponse(response, this.deploymentId);
     // Do not change the prefix everytime we have to refresh credentials
     if (Utils.isNullOrEmpty(this.clientPrefix)) {
+      this.deploymentId =
+          responseNode.has("deployment_id") ? responseNode.get("deployment_id").longValue() : null;
       this.clientPrefix = createClientPrefix(responseNode);
     }
     Utils.assertStringNotNullOrEmpty("client prefix", this.clientPrefix);
@@ -326,7 +330,7 @@ class StreamingIngestStage {
     payload.put("file_name", fileName);
     Map<String, Object> response = this.makeClientConfigureCall(payload);
 
-    JsonNode responseNode = this.parseClientConfigureResponse(response);
+    JsonNode responseNode = this.parseClientConfigureResponse(response, this.deploymentId);
 
     SnowflakeFileTransferMetadataV1 metadata =
         (SnowflakeFileTransferMetadataV1)
@@ -350,7 +354,8 @@ class StreamingIngestStage {
 
   private static final MapStatusGetter statusGetter = new MapStatusGetter();
 
-  private JsonNode parseClientConfigureResponse(Map<String, Object> response) {
+  private JsonNode parseClientConfigureResponse(
+      Map<String, Object> response, @Nullable Long expectedDeploymentId) {
     JsonNode responseNode = mapper.valueToTree(response);
 
     // Currently there are a few mismatches between the client/configure response and what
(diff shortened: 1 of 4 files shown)


sfc-gh-rcheng (approved):
lgtm thanks!

sfc-gh-asen (approved):
Thanks!

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
```diff
@@ -350,6 +355,14 @@ private JsonNode parseClientConfigureResponse(Map<String, Object> response) {
 
     // JDBC expects this field which maps to presignedFileUrlName.  We will set this later
     dataNode.putArray("src_locations").add("placeholder");
+    if (expectedDeploymentId != null) {
+      Long actualDeploymentId =
+          responseNode.has("deployment_id") ? responseNode.get("deployment_id").longValue() : null;
+      if (actualDeploymentId != null && !actualDeploymentId.equals(expectedDeploymentId)) {
+        throw new SFException(
+            ErrorCode.CLIENT_DEPLOYMENT_ID_MISMATCH, expectedDeploymentId, actualDeploymentId);
```
sfc-gh-tzhang, line 363: Could we add client name?

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
```diff
@@ -350,6 +355,14 @@ private JsonNode parseClientConfigureResponse(Map<String, Object> response) {
 
     // JDBC expects this field which maps to presignedFileUrlName.  We will set this later
     dataNode.putArray("src_locations").add("placeholder");
+    if (expectedDeploymentId != null) {
+      Long actualDeploymentId =
+          responseNode.has("deployment_id") ? responseNode.get("deployment_id").longValue() : null;
```
sfc-gh-tzhang, line 360: Now I'm thinking that we have the assumption that the deployment id for a deployment will never change, is that a correct assumption?

src/main/java/net/snowflake/ingest/utils/ErrorCode.java
```diff
@@ -41,7 +41,9 @@ public enum ErrorCode {
   OAUTH_REFRESH_TOKEN_ERROR("0033"),
   INVALID_CONFIG_PARAMETER("0034"),
   CRYPTO_PROVIDER_ERROR("0035"),
-  DROP_CHANNEL_FAILURE("0036");
+  DROP_CHANNEL_FAILURE("0036"),
+
```
sfc-gh-tzhang, line 45: nit: extra space?

src/main/resources/net/snowflake/ingest/ingest_error_messages.properties
```diff
@@ -37,5 +37,6 @@
 0032=URI builder fail to build url: {0}
 0033=OAuth token refresh failure: {0}
 0034=Invalid config parameter: {0}
-0035=Too large batch of rows passed to insertRows, the batch size cannot exceed {0} bytes, recommended batch size for optimal performance and memory utilization is {1} bytes. We recommend splitting large batches into multiple smaller ones and call insertRows for each smaller batch separately.
-0036=Failed to load {0}. If you use FIPS, import BouncyCastleFipsProvider in the application: {1}
\ No newline at end of file
+0035=Failed to load {0}. If you use FIPS, import BouncyCastleFipsProvider in the application: {1}
```
sfc-gh-tzhang, line 40: @sfc-gh-alhuang FYI, let's make sure to update the error message as well when removing an error, or it might be a good idea to never reuse error codes as customer might have custom dependency on them :)

sfc-gh-tzhang (approved):

src/main/resources/net/snowflake/ingest/ingest_error_messages.properties
sfc-gh-psaha: +1 never remove previously used error codes - maybe just rename them if needed to indicate that they are not used anymore.

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
sfc-gh-psaha: Done.

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
sfc-gh-psaha: Asked on #metadata-issues-discuss to confirm.

src/main/java/net/snowflake/ingest/utils/ErrorCode.java
sfc-gh-psaha: Done.

src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestStage.java
sfc-gh-psaha: Confirmed.

src/main/resources/net/snowflake/ingest/ingest_error_messages.properties
```diff
@@ -37,5 +37,6 @@
...
+0036=Failed to drop channel: {0}
+0037=Deployment ID mismatch, Client was created on: {0}, Got upload location for: {1}. Please restart client: {2}.
```
sfc-gh-japatel, line 42: should this be `please recreate client?`

sfc-gh-japatel (approved):
lgtm!

src/main/resources/net/snowflake/ingest/ingest_error_messages.properties
sfc-gh-psaha: I meant it as "shut down your application and restart it".

PR #730, opened 2024-03-25
SNOW-1258064 Remove/Relax the limitations in order to generate bigger files

sfc-gh-alhuang:
Relax file size limitation to deal with longer client flush lag which might produce larger file.

[JIRA](snowflakecomputing.atlassian.net/browse/SNOW-1258064)

PR #796, opened 2024-07-19
SNOW-1545879 Reduce the max channel/chunk sizes

sfc-gh-lsembera:
We [previously](https://github.com/snowflakedb/snowflake-ingest-java/pull/730) increased the limits 4 times, which turned out to cause OOMs in some cases where memory is limited, so this PR reduces the limits by 1/2, which is still double of the original values. 

This PR also fixes a flaky test.


sfc-gh-tzhang (approved):
Thanks!

PR #798, opened 2024-07-25
Structured data type support

sfc-gh-alhuang:
Support structured data type for streaming to iceberg. Refer [here](https://docs.google.com/document/d/1MdGP6LoUhjXb5jZfpCvk5cEitEn34PGD6IYEiU6WtH4/edit?usp=sharing) for all supported data type. Test is expected to add.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
@@ -803,8 +803,7 @@ class DataValidationUtil {
   }
 
   /**
-   * Validates and parses input to a specific Iceberg column number type including INT, LONG, FLOAT,
-   * and DOUBLE. Allowed Java types:
+   * Validates and parses input Iceberg INT column. Allowed Java types:
    *
    * <ul>
    *   <li>Number
@@ -814,69 +813,49 @@ class DataValidationUtil {
    * @param columnName Column name, used in validation error messages
    * @param input Object to validate and parse
    * @param insertRowIndex Row index for error reporting
-   * @param targetType The type to which the input should be parsed ("INT", "LONG", "FLOAT",
-   *     "DOUBLE")
-   * @return Parsed number as Number to be casted by the caller
+   * @return Parsed integer
    */
-  static <T extends Number> T validateAndParseIcebergNumber(
-      String columnName, Object input, long insertRowIndex, Class<T> targetType) {
+  static int validateAndParseIcebergInt(String columnName, Object input, long insertRowIndex) {
     if (input instanceof Number) {
-      if (targetType.equals(Integer.class)) {
-        return targetType.cast(((Number) input).intValue());
-      } else if (targetType.equals(Long.class)) {
-        return targetType.cast(((Number) input).longValue());
-      } else if (targetType.equals(Float.class)) {
-        return targetType.cast(((Number) input).floatValue());
-      } else if (targetType.equals(Double.class)) {
-        return targetType.cast(((Number) input).doubleValue());
-      }
+      return ((Number) input).intValue();
     } else if (input instanceof String) {
-      String stringInput = ((String) input).trim().toLowerCase();
       try {
-        if (targetType.equals(int.class)) {
-          return targetType.cast(Integer.parseInt(stringInput));
-        } else if (targetType.equals(long.class)) {
-          return targetType.cast(Long.parseLong(stringInput));
-        } else if (targetType.equals(float.class)) {
-          return targetType.cast(Float.parseFloat(stringInput));
-        } else if (targetType.equals(double.class)) {
-          return targetType.cast(Double.parseDouble(stringInput));
-        }
+        return Integer.parseInt(((String) input).trim());
       } catch (NumberFormatException e) {
-        if (targetType.equals(float.class)) {
-          switch (stringInput) {
-            case "nan":
-              return targetType.cast(Float.NaN);
-            case "inf":
-              return targetType.cast(Float.POSITIVE_INFINITY);
-            case "-inf":
-              return targetType.cast(Float.NEGATIVE_INFINITY);
-          }
-        }
-        if (targetType.equals(double.class)) {
-          switch (stringInput) {
-            case "nan":
-              return targetType.cast(Double.NaN);
-            case "inf":
-              return targetType.cast(Double.POSITIVE_INFINITY);
-            case "-inf":
-              return targetType.cast(Double.NEGATIVE_INFINITY);
-          }
-        }
         throw valueFormatNotAllowedException(
-            columnName,
-            targetType.getSimpleName().toUpperCase(),
-            "Not a valid number",
-            insertRowIndex);
+            columnName, "INT", "Not a valid integer", insertRowIndex);
       }
     }
-
     throw typeNotAllowedException(
-        columnName,
-        input.getClass(),
-        targetType.getSimpleName().toUpperCase(),
-        new String[] {"Number", "String"},
-        insertRowIndex);
+        columnName, input.getClass(), "INT", new String[] {"Number", "String"}, insertRowIndex);
+  }
+
+  /**
+   * Validates and parses input Iceberg LONG column. Allowed Java types:
+   *
+   * <ul>
+   *   <li>Number
+   *   <li>String
+   * </ul>
+   *
+   * @param columnName Column name, used in validation error messages
+   * @param input Object to validate and parse
+   * @param insertRowIndex Row index for error reporting
+   * @return Parsed long
+   */
+  static long validateAndParseIcebergLong(String columnName, Object input, long insertRowIndex) {
+    if (input instanceof Number) {
+      return ((Number) input).longValue();
+    } else if (input instanceof String) {
+      try {
+        return Long.parseLong(((String) input).trim());
+      } catch (NumberFormatException e) {
+        throw valueFormatNotAllowedException(
+            columnName, "LONG", "Not a valid long", insertRowIndex);
+      }
+    }
+    throw typeNotAllowedException(
+        columnName, input.getClass(), "LONG", new String[] {"Number", "String"}, insertRowIndex);
   }
 
   /**
--- a/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
@@ -803,6 +803,83 @@ class DataValidationUtil {
   }
 
   /**
+   * Validates and parses input to a specific Iceberg column number type including INT, LONG, FLOAT,
+   * and DOUBLE. Allowed Java types:
+   *
+   * <ul>
+   *   <li>Number
+   *   <li>String
+   * </ul>
+   *
+   * @param columnName Column name, used in validation error messages
+   * @param input Object to validate and parse
+   * @param insertRowIndex Row index for error reporting
+   * @param targetType The type to which the input should be parsed ("INT", "LONG", "FLOAT",
+   *     "DOUBLE")
+   * @return Parsed number as Number to be casted by the caller
+   */
+  static <T extends Number> T validateAndParseIcebergNumber(
+      String columnName, Object input, long insertRowIndex, Class<T> targetType) {
+    if (input instanceof Number) {
+      if (targetType.equals(Integer.class)) {
+        return targetType.cast(((Number) input).intValue());
+      } else if (targetType.equals(Long.class)) {
+        return targetType.cast(((Number) input).longValue());
+      } else if (targetType.equals(Float.class)) {
+        return targetType.cast(((Number) input).floatValue());
+      } else if (targetType.equals(Double.class)) {
+        return targetType.cast(((Number) input).doubleValue());
+      }
+    } else if (input instanceof String) {
+      String stringInput = ((String) input).trim().toLowerCase();
+      try {
+        if (targetType.equals(int.class)) {
+          return targetType.cast(Integer.parseInt(stringInput));
+        } else if (targetType.equals(long.class)) {
+          return targetType.cast(Long.parseLong(stringInput));
+        } else if (targetType.equals(float.class)) {
+          return targetType.cast(Float.parseFloat(stringInput));
+        } else if (targetType.equals(double.class)) {
+          return targetType.cast(Double.parseDouble(stringInput));
+        }
+      } catch (NumberFormatException e) {
+        if (targetType.equals(float.class)) {
+          switch (stringInput) {
+            case "nan":
+              return targetType.cast(Float.NaN);
+            case "inf":
+              return targetType.cast(Float.POSITIVE_INFINITY);
+            case "-inf":
+              return targetType.cast(Float.NEGATIVE_INFINITY);
+          }
+        }
+        if (targetType.equals(double.class)) {
+          switch (stringInput) {
+            case "nan":
+              return targetType.cast(Double.NaN);
+            case "inf":
+              return targetType.cast(Double.POSITIVE_INFINITY);
+            case "-inf":
+              return targetType.cast(Double.NEGATIVE_INFINITY);
+          }
+        }
+        throw valueFormatNotAllowedException(
+            columnName,
+            targetType.getSimpleName().toUpperCase(),
+            "Not a valid number",
+            insertRowIndex);
+      }
+    }
+
+    throw typeNotAllowedException(
+        columnName,
+        input.getClass(),
+        targetType.getSimpleName().toUpperCase(),
+        new String[] {"Number", "String"},
+        insertRowIndex);
+  }
+
+  /**
    * Validate and parse input to integer output, 1=true, 0=false. String values converted to boolean
    * according to https://docs.snowflake.com/en/sql-reference/functions/to_boolean.html#usage-notes
    * Allowed Java types:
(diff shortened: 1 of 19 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/ParquetTypeGenerator.java
```diff
@@ -4,13 +4,16 @@
...
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import net.snowflake.ingest.utils.ErrorCode;
 import net.snowflake.ingest.utils.SFException;
+import org.apache.iceberg.parquet.TypeToMessageType;
 import org.apache.parquet.schema.LogicalTypeAnnotation;
 import org.apache.parquet.schema.PrimitiveType;
 import org.apache.parquet.schema.Type;
```
sfc-gh-hmadan, line 19: nit: please move ParquetTypeInfo to a separate file.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -205,9 +204,17 @@ private float addRow(
...
+          (column.getSourceIcebergDataType() == null
+              ? SnowflakeParquetValueParser.parseColumnValueToParquet(
+                  value,
+                  column,
+                  parquetColumn.type.asPrimitiveType().getPrimitiveTypeName(),
+                  forkedStats,
+                  defaultTimezone,
+                  insertRowsCurrIndex)
+              : IcebergParquetValueParser.parseColumnValueToParquet(
+                  value, parquetColumn.type, forkedStats, defaultTimezone, insertRowsCurrIndex));
```
sfc-gh-hmadan, line 217: no need to pass in columnMetadata / columnMetadata.sourceIcebergDataType?

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
```diff
@@ -278,51 +279,70 @@ public void write(List<Object> values) {
...
+    }
+
+    private void write(List<Object> values, GroupType type) {
+      List<Type> cols = type.getFields();
       for (int i = 0; i < cols.size(); ++i) {
         Object val = values.get(i);
-        // val.length() == 0 indicates a NULL value.
         if (val != null) {
-          String fieldName = cols.get(i).getPath()[0];
+          String fieldName = cols.get(i).getName();
```
sfc-gh-hmadan, line 292: while this may be functionally equivalent, I'd not mix this into your current PR. Feel free to send a separate PR to replace getPath[0] with getName.

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
```diff
@@ -278,51 +279,70 @@ public void write(List<Object> values) {
...
+              default:
+                throw new ParquetEncodingException(
+                    "Unsupported column type: " + cols.get(i).asPrimitiveType());
+            }
+          } else {
+            if (cols.get(i).isRepetition(Type.Repetition.REPEATED)) {
+              for (Object o : values) {
+                recordConsumer.startGroup();
+                if (o != null) {
+                  write((List<Object>) o, cols.get(i).asGroupType());
```
sfc-gh-hmadan, line 333: should validate recursion level and fail beyond a threshold?

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -0,0 +1,440 @@
...
+   */
+  static ParquetBufferValue parseColumnValueToParquet(
+      Object value,
+      Type type,
+      RowBufferStats stats,
+      ZoneId defaultTimezone,
+      long insertRowsCurrIndex) {
+    Utils.assertNotNull("Parquet column stats", stats);
+    float estimatedParquetSize = 0F;
+    estimatedParquetSize += ParquetBufferValue.DEFINITION_LEVEL_ENCODING_BYTE_LEN;
```
sfc-gh-hmadan, line 57: 1. checkout the comment on this constant; we are now going to have repetition level so does that change the size estimation logic here?
2. Note that FDN tables are ALSO adding support for structured data types; arguably we'll have to add support in this SDK for them too. Our changes need to treat "iceberg data types" and "structured data type support" as separate but related concepts and not tie them up too tightly. That is, keep the door open for FDN data types + structured data type nesting.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -0,0 +1,440 @@
...
+    if (value != null) {
+      if (type.isPrimitive()) {
+        PrimitiveType primitiveType = type.asPrimitiveType();
+        switch (type.asPrimitiveType().getPrimitiveTypeName()) {
```
sfc-gh-hmadan, line 61: can replace `type.asPrimitiveType()` with just `primitiveType` ?

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -0,0 +1,440 @@
...
+                primitiveType.getPrimitiveTypeName());
+        }
+      } else {
+        return getGroupValue(
+            value, type.asGroupType(), stats, defaultTimezone, insertRowsCurrIndex);
+      }
+    }
+
+    if (value == null) {
+      if (!type.isRepetition(Repetition.REQUIRED)) {
```
sfc-gh-hmadan, line 126: what about primitive types, that are non-nullable, and value is still null?

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -0,0 +1,440 @@
...
+   */
+  private static int getInt32Value(
+      Object value, PrimitiveType type, final long insertRowsCurrIndex) {
+    LogicalTypeAnnotation logicalTypeAnnotation = type.getLogicalTypeAnnotation();
+    if (logicalTypeAnnotation == null) {
+      return DataValidationUtil.validateAndParseIcebergNumber(
+          type.getName(), value, insertRowsCurrIndex, Integer.class);
+    }
+    if (logicalTypeAnnotation instanceof DecimalLogicalTypeAnnotation) {
+      return getDecimalValue(value, type, insertRowsCurrIndex).unscaledValue().intValue();
```
sfc-gh-hmadan, line 152: why not use columnMetadata.scale ?

sfc-gh-hmadan: also, columnMetadata.precision isn't needed here?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -802,6 +802,83 @@ static double validateAndParseReal(String columnName, Object input, long insertR
...
+        return targetType.cast(((Number) input).intValue());
+      } else if (targetType.equals(Long.class)) {
+        return targetType.cast(((Number) input).longValue());
+      } else if (targetType.equals(Float.class)) {
+        return targetType.cast(((Number) input).floatValue());
+      } else if (targetType.equals(Double.class)) {
+        return targetType.cast(((Number) input).doubleValue());
+      }
+    } else if (input instanceof String) {
+      String stringInput = ((String) input).trim().toLowerCase();
```
sfc-gh-hmadan, line 834: why toLowerCase?

sfc-gh-hmadan: only do this when truly necessary (in the catch block I believe)

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -802,6 +802,83 @@ static double validateAndParseReal(String columnName, Object input, long insertR
...
+   * @param insertRowIndex Row index for error reporting
+   * @param targetType The type to which the input should be parsed ("INT", "LONG", "FLOAT",
+   *     "DOUBLE")
+   * @return Parsed number as Number to be casted by the caller
+   */
+  static <T extends Number> T validateAndParseIcebergNumber(
+      String columnName, Object input, long insertRowIndex, Class<T> targetType) {
+    if (input instanceof Number) {
+      if (targetType.equals(Integer.class)) {
+        return targetType.cast(((Number) input).intValue());
```
sfc-gh-hmadan, line 825: Can you check what's the difference between type.class.cast() and Number.intValue / Number.longValue / Number.floatValue / Number.doubleValue ?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -802,6 +802,83 @@ static double validateAndParseReal(String columnName, Object input, long insertR
...
+   *
+   * @param columnName Column name, used in validation error messages
+   * @param input Object to validate and parse
+   * @param insertRowIndex Row index for error reporting
+   * @param targetType The type to which the input should be parsed ("INT", "LONG", "FLOAT",
+   *     "DOUBLE")
+   * @return Parsed number as Number to be casted by the caller
+   */
+  static <T extends Number> T validateAndParseIcebergNumber(
+      String columnName, Object input, long insertRowIndex, Class<T> targetType) {
```
sfc-gh-hmadan, line 822: its easier to understand if you add different methods for the four types being passed in here. There is no commonality between the four classes being passed in here for them to share the same method other than the if (Number) else if (String) else throw structure to the method.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-hmadan: yep, checked documentation we do support custom precision and scale to be specified for managed iceberg tables' decimal fields. See here: https://docs.snowflake.com/en/user-guide/tables-iceberg-data-types

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -0,0 +1,440 @@
...
+   * @return parsed int64 value
+   */
+  private static long getInt64Value(
+      Object value, PrimitiveType type, ZoneId defaultTimezone, final long insertRowsCurrIndex) {
+    LogicalTypeAnnotation logicalTypeAnnotation = type.getLogicalTypeAnnotation();
+    if (logicalTypeAnnotation == null) {
+      return DataValidationUtil.validateAndParseIcebergNumber(
+          type.getName(), value, insertRowsCurrIndex, Long.class);
+    }
+    if (logicalTypeAnnotation instanceof DecimalLogicalTypeAnnotation) {
```
sfc-gh-hmadan, line 176: nit: line 154 uses a different style for a similar kind of check: `logicalTypeAnnotation == LogicalTypeAnnotation.dateType`
Lets do one style in all places (instanceof lgtm) ?

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -0,0 +1,440 @@
...
+      return getDecimalValue(value, type, insertRowsCurrIndex).unscaledValue().longValue();
```
sfc-gh-hmadan, line 177: .longValue() here and .intValue() in the previous method are guaranteed to not throw a casting exception?

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-hmadan: also applies to long decimal..

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -0,0 +1,440 @@
...
+      boolean includeTimeZone =
+          ((TimestampLogicalTypeAnnotation) logicalTypeAnnotation).isAdjustedToUTC();
+      return DataValidationUtil.validateAndParseTimestamp(
+              type.getName(),
+              value,
+              timeUnitToScale(((TimestampLogicalTypeAnnotation) logicalTypeAnnotation).getUnit()),
+              defaultTimezone,
+              !includeTimeZone,
+              insertRowsCurrIndex)
+          .toBinary(false)
```
sfc-gh-hmadan, line 197: we are removing timezone in all cases, is that correct behavior?

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -0,0 +1,440 @@
...
+   * @param insertRowsCurrIndex Used for logging the row of index given in insertRows API
+   * @return string representation
+   */
+  private static byte[] getBinaryValue(
+      Object value, PrimitiveType type, RowBufferStats stats, final long insertRowsCurrIndex) {
+    LogicalTypeAnnotation logicalTypeAnnotation = type.getLogicalTypeAnnotation();
+    if (logicalTypeAnnotation == null) {
+      byte[] bytes =
+          DataValidationUtil.validateAndParseBinary(
+              type.getName(), value, Optional.empty(), insertRowsCurrIndex);
```
sfc-gh-hmadan, line 221: why not validate against maxLength here?

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -0,0 +1,440 @@
...
+      final long insertRowsCurrIndex) {
+    Map<String, Object> structVal =
+        DataValidationUtil.validateAndParseIcebergStruct(
+            type.getName(), value, insertRowsCurrIndex);
+    List<Object> listVal = new ArrayList<>(type.getFieldCount());
+    float estimatedParquetSize = ParquetBufferValue.DEFINITION_LEVEL_ENCODING_BYTE_LEN;
+    for (int i = 0; i < type.getFieldCount(); i++) {
+      ParquetBufferValue parsedValue =
+          parseColumnValueToParquet(
+              structVal.getOrDefault(type.getFieldName(i), null),
```
sfc-gh-hmadan, line 324: Is null a properly handled value for all types of fields..?

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -0,0 +1,440 @@
...
+      return DataValidationUtil.validateAndParseTime(
+              type.getName(),
+              value,
+              timeUnitToScale(((TimeLogicalTypeAnnotation) logicalTypeAnnotation).getUnit()),
+              insertRowsCurrIndex)
+          .longValue();
+    }
+    if (logicalTypeAnnotation instanceof TimestampLogicalTypeAnnotation) {
+      boolean includeTimeZone =
+          ((TimestampLogicalTypeAnnotation) logicalTypeAnnotation).isAdjustedToUTC();
```
sfc-gh-hmadan, line 189: I see timestamp and timestamptz as two separate types in this document - are we handling both?
https://docs.snowflake.com/en/user-guide/tables-iceberg-data-types

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -0,0 +1,440 @@
...
+        DataValidationUtil.validateAndParseIcebergStruct(
+            type.getName(), value, insertRowsCurrIndex);
+    List<Object> listVal = new ArrayList<>(type.getFieldCount());
+    float estimatedParquetSize = ParquetBufferValue.DEFINITION_LEVEL_ENCODING_BYTE_LEN;
+    for (int i = 0; i < type.getFieldCount(); i++) {
+      ParquetBufferValue parsedValue =
+          parseColumnValueToParquet(
+              structVal.getOrDefault(type.getFieldName(i), null),
+              type.getType(i),
+              stats,
```
sfc-gh-hmadan, line 326: we need the nested object's "metadata" to also be passed in for stuff like precision/scale/byte[] maxlength/etc.
However we won't have a ColumnMetadata object in that situation, which means we now need an abstraction over ColumnMetadata for use in the value parsers (I'm leaving another comment soon that we should look to have one ParquetValueParser class as there's too many similarities for them to be completely independent classes. Especially since we need to support structured data types for FDN types too some day)

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: I didn't use the column metadata as `parquetColumn.type` already include all information we need including precision, scale, length, etc.

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
sfc-gh-alhuang: The `.getPath()[0]` was used against `ColumnDescriptor`. The `.getName()` is used on `Parquet.Type` as we need the information in `Parquet.Type`.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: Added repetition level encoding size estimation.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: Replaced.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetTypeGenerator.java
sfc-gh-alhuang: Moved.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: The logic was incorrect, fixed.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: The purpose of this function is to parse java type to parquet bytes. The actual bytes of a parquet decimal column is an unscaled integer. The scanner later use the precision and scale in schema to infer the correct information.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-alhuang: Removed this function and use `validateAndParseReal` for float and double parsing.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-alhuang: Tye `type.class.cast` is used for the return type check and it's no longer needed after we split it into 4 different methods. The `Number.xValue` do narrow casting if the source type cannot fit into target type.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-alhuang: Split into different methods.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: Changed to instanceof.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: According [here](https://github.com/snowflakedb/snowflake-ingest-java/blob/e808bbd6a215daaae9e5a49ec4006978afbeca8d/src/main/java/net/snowflake/ingest/streaming/internal/ParquetValueParser.java#L319C7-L324C30) and `Iceberg.timetamptz` is using `Snowflake.timestampltz` ([ref](https://github.com/snowflakedb/snowflake-ingest-java/blob/e808bbd6a215daaae9e5a49ec4006978afbeca8d/src/main/java/net/snowflake/ingest/streaming/internal/ParquetValueParser.java#L319C7-L324C30)).

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: According [here](https://github.com/openjdk-mirror/jdk7u-jdk/blob/f4d80957e89a19a29bb9f9807d2a28351ed7f7df/src/share/classes/java/math/BigInteger.java#L2720C4-L2738C6), `longValue` and `intValue` returns last 4/8 bytes. I don't think exception will be thrown here.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: Added max length validation, we don't need the max length in column meta data as the length of binary and string are constants.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: Afaik, the leaf primitive type will always check if it's nullable. Will add test to verify it.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: Yes, `Iceberg.timestamp` and `Iceberg.timestamptz` use the same `LogicalTypeAnnotation` with different `adjustToUtc`. [Ref](https://iceberg.apache.org/spec/#parquet)

PR #800, opened 2024-07-26
NO_SNOW Upgrade dependency versions

sfc-gh-xhuang:
Change release version back to snapshot after release
upgrade JDBC and other dependencies

### DIFF
--- a/e2e-jar-test/pom.xml
+++ b/e2e-jar-test/pom.xml
@@ -23,19 +23,19 @@
       <dependency>
         <groupId>com.fasterxml.jackson.core</groupId>
         <artifactId>jackson-databind</artifactId>
-        <version>2.14.0</version>
+        <version>2.17.2</version>
       </dependency>
 
       <dependency>
         <groupId>net.snowflake</groupId>
         <artifactId>snowflake-ingest-sdk</artifactId>
-        <version>2.1.2</version>
+        <version>2.2.0-SNAPSHOT</version>
       </dependency>
 
       <dependency>
         <groupId>net.snowflake</groupId>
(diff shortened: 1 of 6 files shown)


e2e-jar-test/pom.xml
```diff
@@ -23,19 +23,19 @@
...
         <artifactId>jackson-databind</artifactId>
-        <version>2.14.0</version>
+        <version>2.17.2</version>
       </dependency>
 
       <dependency>
         <groupId>net.snowflake</groupId>
         <artifactId>snowflake-ingest-sdk</artifactId>
-        <version>2.1.2</version>
+        <version>2.2.0-SNAPSHOT</version>
```
sfc-gh-tzhang, line 32: why do we need a minor version upgrade?

sfc-gh-tzhang (commented):
Please enable the long running test to make sure the JDBC version upgrade works

e2e-jar-test/pom.xml
sfc-gh-tzhang: there is no behavior change for these PRs merging into main, I asked them to have a feature branch which will contain all the user interface changes and behavior changes

sfc-gh-hmadan: Correct, 2.1.3 should work.
Do we have tooling that diffs the public API surface of our binary between releases? If not automated, we manually compare the public API surface between what's already released and what's being released now?

sfc-gh-xhuang:
Long running tests were successful

sfc-gh-xhuang:
also updated readme
https://github.com/snowflakedb/snowflake-ingest-java/issues/804

PR #801, opened 2024-07-26
SNOW-1483230 Parameter support & disable blob encryption for new table format

sfc-gh-alhuang:
Split parameter `maxChunksInBlobAndRegistrationRequest` into `maxChunksInBlobs` and `maxChunksInRegistrationRequest`
Merging change of #763 and #773 into main branch.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClientFactory.java
+++ b/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClientFactory.java
@@ -31,10 +31,6 @@ public class SnowflakeStreamingIngestClientFactory {
     // Indicates whether it's under test mode
     private boolean isTestMode;
 
-    // Indicates whether it's streaming to Iceberg tables. Open channels on regular tables should
-    // fail in this mode.
-    private boolean isIcebergMode;
-
     private Builder(String name) {
       this.name = name;
     }
@@ -62,12 +58,7 @@ public class SnowflakeStreamingIngestClientFactory {
       SnowflakeURL accountURL = new SnowflakeURL(prop.getProperty(Constants.ACCOUNT_URL));
 
       return new SnowflakeStreamingIngestClientInternal<>(
-          this.name,
-          accountURL,
-          prop,
-          this.parameterOverrides,
-          this.isIcebergMode,
-          this.isTestMode);
+          this.name, accountURL, prop, this.parameterOverrides, false, this.isTestMode);
     }
   }
 }
--- a/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClientFactory.java
+++ b/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClientFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Snowflake Computing Inc. All rights reserved.
+ * Copyright (c) 2021-2024 Snowflake Computing Inc. All rights reserved.
  */
 
 package net.snowflake.ingest.streaming;
@@ -31,6 +31,10 @@ public class SnowflakeStreamingIngestClientFactory {
     // Indicates whether it's under test mode
     private boolean isTestMode;
 
+    // Indicates whether it's streaming to Iceberg tables. Open channels on regular tables should
+    // fail in this mode.
+    private boolean isIcebergMode;
+
     private Builder(String name) {
       this.name = name;
     }
@@ -50,6 +54,11 @@ public class SnowflakeStreamingIngestClientFactory {
       return this;
     }
 
+    Builder setIsIcebergMode(boolean isIcebergMode) {
+      this.isIcebergMode = isIcebergMode;
+      return this;
+    }
+
     public SnowflakeStreamingIngestClient build() {
       Utils.assertStringNotNullOrEmpty("client name", this.name);
       Utils.assertNotNull("connection properties", this.prop);
@@ -58,7 +67,12 @@ public class SnowflakeStreamingIngestClientFactory {
       SnowflakeURL accountURL = new SnowflakeURL(prop.getProperty(Constants.ACCOUNT_URL));
 
       return new SnowflakeStreamingIngestClientInternal<>(
-          this.name, accountURL, prop, this.parameterOverrides, this.isTestMode);
+          this.name,
+          accountURL,
+          prop,
+          this.parameterOverrides,
+          this.isIcebergMode,
+          this.isTestMode);
     }
   }
 }
--- a/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClientFactory.java
+++ b/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClientFactory.java
@@ -54,11 +54,6 @@ public class SnowflakeStreamingIngestClientFactory {
       return this;
     }
 
-    Builder setIsIcebergMode(boolean isIcebergMode) {
-      this.isIcebergMode = isIcebergMode;
-      return this;
-    }
-
     public SnowflakeStreamingIngestClient build() {
       Utils.assertStringNotNullOrEmpty("client name", this.name);
       Utils.assertNotNull("connection properties", this.prop);
--- a/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Snowflake Computing Inc. All rights reserved.
+ * Copyright (c) 2021-2024 Snowflake Computing Inc. All rights reserved.
  */
 
 package net.snowflake.ingest.streaming.internal;
@@ -61,10 +61,14 @@ class BlobBuilder {
    * @param blobData All the data for one blob. Assumes that all ChannelData in the inner List
    *     belongs to the same table. Will error if this is not the case
    * @param bdecVersion version of blob
+   * @param encrypt If the output chunk is encrypted or not
    * @return {@link Blob} data
    */
   static <T> Blob constructBlobAndMetadata(
-      String filePath, List<List<ChannelData<T>>> blobData, Constants.BdecVersion bdecVersion)
+      String filePath,
+      List<List<ChannelData<T>>> blobData,
+      Constants.BdecVersion bdecVersion,
+      boolean encrypt)
       throws IOException, NoSuchPaddingException, NoSuchAlgorithmException,
           InvalidAlgorithmParameterException, InvalidKeyException, IllegalBlockSizeException,
           BadPaddingException {
@@ -83,25 +87,32 @@ class BlobBuilder {
           flusher.serialize(channelsDataPerTable, filePath);
 
       if (!serializedChunk.channelsMetadataList.isEmpty()) {
-        ByteArrayOutputStream chunkData = serializedChunk.chunkData;
-        Pair<byte[], Integer> paddedChunk =
-            padChunk(chunkData, Constants.ENCRYPTION_ALGORITHM_BLOCK_SIZE_BYTES);
-        byte[] paddedChunkData = paddedChunk.getFirst();
-        int paddedChunkLength = paddedChunk.getSecond();
+        byte[] compressedChunkData;
+        int chunkLength;
 
-        // Encrypt the compressed chunk data, the encryption key is derived using the key from
-        // server with the full blob path.
-        // We need to maintain IV as a block counter for the whole file, even interleaved,
-        // to align with decryption on the Snowflake query path.
-        // TODO: address alignment for the header SNOW-557866
-        long iv = curDataSize / Constants.ENCRYPTION_ALGORITHM_BLOCK_SIZE_BYTES;
-        byte[] encryptedCompressedChunkData =
-            Cryptor.encrypt(
-                paddedChunkData, firstChannelFlushContext.getEncryptionKey(), filePath, iv);
+        if (encrypt) {
+          Pair<byte[], Integer> paddedChunk =
+              padChunk(serializedChunk.chunkData, Constants.ENCRYPTION_ALGORITHM_BLOCK_SIZE_BYTES);
+          byte[] paddedChunkData = paddedChunk.getFirst();
+          chunkLength = paddedChunk.getSecond();
+
+          // Encrypt the compressed chunk data, the encryption key is derived using the key from
+          // server with the full blob path.
+          // We need to maintain IV as a block counter for the whole file, even interleaved,
+          // to align with decryption on the Snowflake query path.
+          // TODO: address alignment for the header SNOW-557866
+          long iv = curDataSize / Constants.ENCRYPTION_ALGORITHM_BLOCK_SIZE_BYTES;
+          compressedChunkData =
+              Cryptor.encrypt(
+                  paddedChunkData, firstChannelFlushContext.getEncryptionKey(), filePath, iv);
+        } else {
+          compressedChunkData = serializedChunk.chunkData.toByteArray();
+          chunkLength = compressedChunkData.length;
+        }
 
         // Compute the md5 of the chunk data
-        String md5 = computeMD5(encryptedCompressedChunkData, paddedChunkLength);
-        int encryptedCompressedChunkDataSize = encryptedCompressedChunkData.length;
+        String md5 = computeMD5(compressedChunkData, chunkLength);
+        int compressedChunkDataSize = compressedChunkData.length;
 
         // Create chunk metadata
         long startOffset = curDataSize;
@@ -111,9 +122,9 @@ class BlobBuilder {
                 // The start offset will be updated later in BlobBuilder#build to include the blob
                 // header
                 .setChunkStartOffset(startOffset)
-                // The paddedChunkLength is used because it is the actual data size used for
+                // The chunkLength is used because it is the actual data size used for
                 // decompression and md5 calculation on server side.
-                .setChunkLength(paddedChunkLength)
+                .setChunkLength(chunkLength)
                 .setUncompressedChunkLength((int) serializedChunk.chunkEstimatedUncompressedSize)
                 .setChannelList(serializedChunk.channelsMetadataList)
                 .setChunkMD5(md5)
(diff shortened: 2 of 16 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -83,25 +87,32 @@ static <T> Blob constructBlobAndMetadata(
...
+        } else {
+          compressedChunkData = serializedChunk.chunkData.toByteArray();
+          chunkLength = compressedChunkData.length;
+        }
 
         // Compute the md5 of the chunk data
-        String md5 = computeMD5(encryptedCompressedChunkData, paddedChunkLength);
-        int encryptedCompressedChunkDataSize = encryptedCompressedChunkData.length;
+        String md5 = computeMD5(compressedChunkData, chunkLength);
+        int compressedChunkDataSize = compressedChunkData.length;
```
sfc-gh-hmadan, line 115: moving this inside the if-else block seems a little bit better?
`int compressedChunkDataSize = compressedChunkData.length;`

it'll then look like:
```
final int compressedChunkSize;
if () {
    ....
    compressedChunkSize = compressedChunkData.length; // line 108
} else {
    ....
    compressedChunkSize = chunkLength;
}
```

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -83,25 +87,32 @@ static <T> Blob constructBlobAndMetadata(
           flusher.serialize(channelsDataPerTable, filePath);
 
       if (!serializedChunk.channelsMetadataList.isEmpty()) {
-        ByteArrayOutputStream chunkData = serializedChunk.chunkData;
-        Pair<byte[], Integer> paddedChunk =
-            padChunk(chunkData, Constants.ENCRYPTION_ALGORITHM_BLOCK_SIZE_BYTES);
-        byte[] paddedChunkData = paddedChunk.getFirst();
-        int paddedChunkLength = paddedChunk.getSecond();
+        byte[] compressedChunkData;
+        int chunkLength;
```
sfc-gh-hmadan, line 91: mark both of these as `final` so its clear they're only set-once?

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -50,11 +56,12 @@ public class ParameterProvider {
   public static final int IO_TIME_CPU_RATIO_DEFAULT = 2;
   public static final int BLOB_UPLOAD_MAX_RETRY_COUNT_DEFAULT = 24;
   public static final long MAX_MEMORY_LIMIT_IN_BYTES_DEFAULT = -1L;
-  public static final long MAX_CHANNEL_SIZE_IN_BYTES_DEFAULT = 64 * 1024 * 1024;
-  public static final long MAX_CHUNK_SIZE_IN_BYTES_DEFAULT = 256 * 1024 * 1024;
+  public static final long MAX_CHANNEL_SIZE_IN_BYTES_DEFAULT = 128 * 1024 * 1024;
+  public static final long MAX_CHUNK_SIZE_IN_BYTES_DEFAULT = 512 * 1024 * 1024;
```
sfc-gh-hmadan, line 60: This is a behavior impacting change, please don't mix this in.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -80,25 +90,45 @@ public class ParameterProvider {
...
+      boolean enforceDefault) {
     if (parameterOverrides != null && props != null) {
       this.parameterMap.put(
           key, parameterOverrides.getOrDefault(key, props.getOrDefault(key, defaultValue)));
     } else if (parameterOverrides != null) {
       this.parameterMap.put(key, parameterOverrides.getOrDefault(key, defaultValue));
     } else if (props != null) {
       this.parameterMap.put(key, props.getOrDefault(key, defaultValue));
+    } else {
+      this.parameterMap.put(key, defaultValue);
```
sfc-gh-hmadan, line 120: looks like we were never doing this before your changes. How was defaulting working then in that world?

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -80,25 +90,45 @@ public class ParameterProvider {
...
+    }
+
+    if (enforceDefault) {
```
sfc-gh-hmadan, line 123: please make sure there is a testcase to validate this codepath works as expected and you get an exception of the expected type.

src/test/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientTest.java
```diff
@@ -79,6 +86,14 @@ public class SnowflakeStreamingIngestClientTest {
   SnowflakeStreamingIngestChannelInternal<StubChunkData> channel3;
   SnowflakeStreamingIngestChannelInternal<StubChunkData> channel4;
 
+  // TODO: Add IcebergMode = True after Streaming to Iceberg is supported.
```
sfc-gh-hmadan, line 89: lets make sure we return true her so that we're testing all throughout? The feature doesn't have to be released to customers / made public for testing to start?

src/test/java/net/snowflake/ingest/streaming/internal/ParameterProviderTest.java
```diff
@@ -291,8 +337,20 @@ public void testMaxChunksInBlobAndRegistrationRequest() {
...
+
+    parameterProvider = new ParameterProvider(parameterMap, prop, true);
     Assert.assertEquals(1, parameterProvider.getMaxChunksInBlobAndRegistrationRequest());
+
+    SFException e =
+        Assert.assertThrows(
+            SFException.class,
+            () -> {
+              parameterMap.put("max_chunks_in_blob_and_registration_request", 100);
+              new ParameterProvider(parameterMap, prop, true);
```
sfc-gh-hmadan, line 351: curious, why shouldn't this test class also be made into a parameterized file so that all testcases test for the isIceberg=true mode of ParameterProvider too?

src/test/java/net/snowflake/ingest/streaming/internal/InsertRowsBenchmarkTest.java
```diff
@@ -36,7 +40,7 @@ public class InsertRowsBenchmarkTest {
 
   @Setup(Level.Trial)
   public void setUpBeforeAll() {
-    client = new SnowflakeStreamingIngestClientInternal<ParquetChunkData>("client_PARQUET");
+    client = new SnowflakeStreamingIngestClientInternal<ParquetChunkData>("client_PARQUET", false);
```
sfc-gh-hmadan, line 43: please add a code comment here with `SNOW-1490151` in the comment so that whenever that JIRA is fixed we have an easy way of finding all the places that need fixes.

src/test/java/net/snowflake/ingest/streaming/internal/OAuthBasicTest.java
```diff
@@ -114,7 +118,7 @@ public void testCreateOAuthClient() throws Exception {
   @Test
   public void testSetRefreshToken() throws Exception {
     SnowflakeStreamingIngestClientInternal<StubChunkData> client =
-        new SnowflakeStreamingIngestClientInternal<>("TEST_CLIENT");
+        new SnowflakeStreamingIngestClientInternal<>("TEST_CLIENT", false);
```
sfc-gh-hmadan, line 121: please add a code comment here with `SNOW-1490151` in the comment so that whenever that JIRA is fixed we have an easy way of finding all the places that need fixes.

src/test/java/net/snowflake/ingest/streaming/internal/BlobBuilderTest.java
```diff
@@ -23,18 +27,21 @@ public void testSerializationErrors() throws Exception {
     BlobBuilder.constructBlobAndMetadata(
         "a.bdec",
         Collections.singletonList(createChannelDataPerTable(1, false)),
-        Constants.BdecVersion.THREE);
+        Constants.BdecVersion.THREE,
+        true);
```
sfc-gh-hmadan, line 31: also test with false, make the testclass into a parameterized class?

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
sfc-gh-alhuang: Refactored.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-alhuang: Reverted

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-alhuang: This line is used to set parameter with different default value for iceberg / noniceberg. ([ref](https://github.com/snowflakedb/snowflake-ingest-java/blob/a37433d3c53bc442d28003b275ca50c755ca9ea2/src/main/java/net/snowflake/ingest/utils/ParameterProvider.java#L238))

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-alhuang: Added [here](https://github.com/snowflakedb/snowflake-ingest-java/blob/a37433d3c53bc442d28003b275ca50c755ca9ea2/src/test/java/net/snowflake/ingest/streaming/internal/ParameterProviderTest.java#L346C5-L352C16)

src/test/java/net/snowflake/ingest/streaming/internal/InsertRowsBenchmarkTest.java
sfc-gh-alhuang: Added.

src/test/java/net/snowflake/ingest/streaming/internal/OAuthBasicTest.java
sfc-gh-alhuang: Added

src/test/java/net/snowflake/ingest/streaming/internal/BlobBuilderTest.java
sfc-gh-alhuang: Make to parameterized class.

src/test/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientTest.java
sfc-gh-alhuang: Added test parameter.

src/test/java/net/snowflake/ingest/streaming/internal/ParameterProviderTest.java
sfc-gh-alhuang: Parameterized test.

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
sfc-gh-alhuang: Marked final.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -31,8 +37,9 @@ public class ParameterProvider {
   public static final String MAX_CHUNK_SIZE_IN_BYTES = "MAX_CHUNK_SIZE_IN_BYTES".toLowerCase();
   public static final String MAX_ALLOWED_ROW_SIZE_IN_BYTES =
       "MAX_ALLOWED_ROW_SIZE_IN_BYTES".toLowerCase();
-  public static final String MAX_CHUNKS_IN_BLOB_AND_REGISTRATION_REQUEST =
-      "MAX_CHUNKS_IN_BLOB_AND_REGISTRATION_REQUEST".toLowerCase();
+  public static final String MAX_CHUNKS_IN_BLOB = "MAX_CHUNKS_IN_BLOB".toLowerCase();
```
sfc-gh-hmadan, line 40: How do we know no customer was overriding the old setting ? Any changes to these settings is by default a back compat breaking change that needs to be handled with extra care.
Note that these settings are also public, see here: https://javadoc.io/doc/net.snowflake/snowflake-ingest-sdk/latest/net/snowflake/ingest/utils/ParameterProvider.html

cc @sfc-gh-tzhang

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -55,15 +62,20 @@ public class ParameterProvider {
...
   // Lag related parameters
   public static final long MAX_CLIENT_LAG_DEFAULT = 1000; // 1 second
+  public static final long MAX_CLIENT_LAG_ICEBERG_MODE_DEFAULT = 30000; // 30 seconds
   static final long MAX_CLIENT_LAG_MS_MIN = TimeUnit.SECONDS.toMillis(1);
   static final long MAX_CLIENT_LAG_MS_MAX = TimeUnit.MINUTES.toMillis(10);
 
   public static final long MAX_ALLOWED_ROW_SIZE_IN_BYTES_DEFAULT = 64 * 1024 * 1024; // 64 MB
-  public static final int MAX_CHUNKS_IN_BLOB_AND_REGISTRATION_REQUEST_DEFAULT = 100;
+  public static final int MAX_CHUNKS_IN_BLOB_DEFAULT = 100;
+  public static final int MAX_CHUNKS_IN_REGISTRATION_REQUEST_DEFAULT = 100;
```
sfc-gh-hmadan, line 71: is there any validation needed to throw when max_chunks_in_blob > max_chunks_in_registration_request ?

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -384,15 +450,17 @@ public long getMaxAllowedRowSizeInBytes() {
...
   }
 
-  /**
-   * @return The max number of chunks that can be put into a single BDEC or blob registration
-   *     request.
-   */
-  public int getMaxChunksInBlobAndRegistrationRequest() {
+  /** @return The max number of chunks that can be put into a single BDEC. */
+  public int getMaxChunksInBlob() {
+    Object val = this.parameterMap.getOrDefault(MAX_CHUNKS_IN_BLOB, MAX_CHUNKS_IN_BLOB_DEFAULT);
```
sfc-gh-hmadan, line 455: we should use the iceberg_default here to make this accurate?

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-alhuang: The exception happens [here](https://github.com/snowflakedb/snowflake-ingest-java/blob/f383146763e4c7a08853b6784ba123d8f25370fb/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java#L513) if the chunks in a blob is more than `MAX_CHUNKS_IN_REGISTRATION_REQUEST`. Didn't add a check in parameter provider since the actual number of chunks in a blob might be much smaller than the`MAX_CHUNKS_IN_BLOB`.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-alhuang: The [setParameterMap()](https://github.com/snowflakedb/snowflake-ingest-java/blob/3d95d22ea67c87a5d758d5aacd1f497a025302a7/src/main/java/net/snowflake/ingest/utils/ParameterProvider.java#L239) and [checkAndUpdate()](https://github.com/snowflakedb/snowflake-ingest-java/blob/3d95d22ea67c87a5d758d5aacd1f497a025302a7/src/main/java/net/snowflake/ingest/utils/ParameterProvider.java#L122) will guarantee that the Iceberg default value is always in the map.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-tzhang: We don't know, I don't think they need to be back compat (at least for this one) since we're not documenting them, and I'm not aware of anyone who is overriding this.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-hmadan: In that case lets remove the getOrDefault and throw if get returns null?

sfc-gh-hmadan: @sfc-gh-tzhang any preferences on this?

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-hmadan: Ok. Lets remove the public access modifier from settings that we do not want customers to be exposed to?
Alec could you please run this by Xin before committing it in?
Do we have telemetry to figure out if any client in the past 6 months has had non-default values set for these parameters?

sfc-gh-tzhang: > Lets remove the public access modifier from settings that we do not want customers to be exposed to?

I think we still want to keep it as publicly accessible, I believe it was added for some incident mitigation, but these overrides are never meant to be used long term. + @sfc-gh-lsembera to confirm

> Do we have telemetry to figure out if any client in the past 6 months has had non-default values set for these parameters?

I don't think we collect this at server side unfortunately.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-tzhang: Discussed offline with Alec, I think we should keep the old behavior and not adding everything in the map, otherwise we don't know if it's actually an override or not as all these values will be in the map.

sfc-gh-hmadan: sounds good! I mainly want to ensure there's precisely one way to do defaulting, either by adding everything to the map or at read time in this get method. 

Also the new enforceDefault block won't have to change even if we don't add the default value to the map.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-alhuang: Added validation during `ParameterProvdier` initialization.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-alhuang: Changed the behavior to the following

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-lsembera: This parameter was introduced to fix timeouts when customers ingest into thousands of tables concurrently. 

We cannot know if anybody changed the parameter locally, but I am not aware of recommending to change this parameter to any customer.

sfc-gh-tzhang: ok, then I think it's fine to not support backward compatible as these are not publicly documented parameters, customer should not use parameters without our recommendation.

src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClientFactory.java
```diff
@@ -50,6 +54,11 @@ public Builder setIsTestMode(boolean isTestMode) {
       return this;
     }
 
+    Builder setIsIcebergMode(boolean isIcebergMode) {
+      this.isIcebergMode = isIcebergMode;
+      return this;
+    }
+
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Removed.

sfc-gh-tzhang (approved):
LGTM, thanks!

PR #807, opened 2024-08-08
SNOW-1618257 Fix PRIMARY_FILE_ID_KEY

sfc-gh-kkloudas:
As part of the investigation for https://snowflakecomputing.atlassian.net/browse/SNOW-1506753 we discovered that the SDK has a concurrency issue when populating the `PRIMARY_FILE_ID_KEY` and there are concurrent uploads due to high load. 

In a nutshell, the problem is that the `ParquetRowBuffer#metadata` which is a map that contains some metadata that will be written as parquet metadata in the BDEC file, is shared across all uploading threads. So if one of those threads changes it, the others will see the changes. One of those metadata is also the  `PRIMARY_FILE_ID_KEY`.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ParquetChunkData.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ParquetChunkData.java
@@ -5,6 +5,7 @@
 package net.snowflake.ingest.streaming.internal;
 
 import java.io.ByteArrayOutputStream;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import org.apache.parquet.hadoop.BdecParquetWriter;
@@ -34,6 +35,16 @@ public class ParquetChunkData {
     this.rows = rows;
     this.parquetWriter = parquetWriter;
     this.output = output;
-    this.metadata = metadata;
+    // create a defensive copy of the parameter map because the argument map passed here
+    // may currently be shared across multiple threads.
+    this.metadata = createDefensiveCopy(metadata);
+  }
+
+  private Map<String, String> createDefensiveCopy(final Map<String, String> metadata) {
+    final Map<String, String> copy = new HashMap<>(metadata);
+    for (String k : metadata.keySet()) {
+      copy.put(k, metadata.get(k));
+    }
+    return copy;
   }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/ParquetChunkData.java
```diff
@@ -34,6 +35,15 @@ public ParquetChunkData(
     this.rows = rows;
     this.parquetWriter = parquetWriter;
     this.output = output;
-    this.metadata = metadata;
+    // create a defensive copy of the parameter map.
+    this.metadata = createDefensiveCopy(metadata);
```
sfc-gh-lpapke, line 39: Should we use `Map.copyOf(metadata)` to create an immutable map to ensure no more modifcation per design?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetChunkData.java
```diff
@@ -34,6 +35,15 @@ public ParquetChunkData(
...
+  }
+
+  private Map<String, String> createDefensiveCopy(final Map<String, String> metadata) {
```
sfc-gh-tzhang, line 42: add a comment explain we need this?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetChunkData.java
sfc-gh-kkloudas: There is a problem because of the Java level that the SDK is using. Map.copyOf() we need level 10.

src/test/java/net/snowflake/ingest/streaming/internal/RowBufferTest.java
```diff
@@ -3,17 +3,14 @@
...
 import java.math.BigInteger;
 import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
sfc-gh-tzhang, line 11: please avoid using star import

sfc-gh-kkloudas: Damn, it was by accident. I did not see this.

PR #814, opened 2024-08-13
SNOW-1507007 Support schema for new table format

sfc-gh-alhuang:
We use FDN-specific logical and physical data types only today. In this PR we change to use iceberg’s data types so there is no loss of signal between the table schema on the server versus what data type conversions are done in the client.


scripts/process_licenses.py
```diff
@@ -61,6 +61,8 @@
     "org.bouncycastle:bcpkix-jdk18on": BOUNCY_CASTLE_LICENSE,
     "org.bouncycastle:bcutil-jdk18on": BOUNCY_CASTLE_LICENSE,
     "org.bouncycastle:bcprov-jdk18on": BOUNCY_CASTLE_LICENSE,
+    "org.roaringbitmap:RoaringBitmap": APACHE_LICENSE,
```
sfc-gh-hmadan, line 64: can't find this library being used anywhere, why is this change needed?

src/main/java/net/snowflake/ingest/streaming/internal/ColumnMetadata.java
```diff
@@ -128,6 +129,15 @@ public Integer getOrdinal() {
     return ordinal;
   }
 
+  @JsonProperty("source_iceberg_data_type")
```
sfc-gh-hmadan, line 132: mind adding a comment here with a sample content of what goes into this field?

src/main/java/net/snowflake/ingest/streaming/internal/OpenChannelRequestInternal.java
```diff
@@ -83,6 +89,10 @@ String getWriteMode() {
     return writeMode;
   }
 
+  boolean getIsIceberg() {
+    return isIceberg;
+  }
```
sfc-gh-hmadan, line 94: pl add isIceberg to the getStringForLogging method too.

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
```diff
@@ -264,7 +265,7 @@ public void prepareForWrite(RecordConsumer recordConsumer) {
 
     @Override
     public void write(List<Object> values) {
-      List<ColumnDescriptor> cols = schema.getColumns();
+      List<Type> cols = schema.getFields();
```
sfc-gh-hmadan, line 268: what's the difference between getColumns and getFields?

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
```diff
@@ -278,51 +279,56 @@ public void write(List<Object> values) {
...
+                recordConsumer.addBinary(binVal);
+                break;
+              case FIXED_LEN_BYTE_ARRAY:
+                Binary binary = Binary.fromConstantByteArray((byte[]) val);
+                recordConsumer.addBinary(binary);
+                break;
+              default:
+                throw new ParquetEncodingException(
+                    "Unsupported column type: " + cols.get(i).asPrimitiveType());
+            }
```
sfc-gh-hmadan, line 327: else throw new ParquetEncodingException("Unsupported column type: " ...)

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -802,6 +802,62 @@ static double validateAndParseReal(String columnName, Object input, long insertR
...
+   * </ul>
+   *
+   * @param columnName Column name, used in validation error messages
+   * @param input Object to validate and parse
+   * @param insertRowIndex Row index for error reporting
+   * @return Parsed integer
+   */
+  static int validateAndParseIcebergInt(String columnName, Object input, long insertRowIndex) {
+    if (input instanceof Number) {
+      return ((Number) input).intValue();
```
sfc-gh-hmadan, line 820: can't this cause an exception if the input were a long value / had a value > Int.maxvalue?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -802,6 +802,62 @@ static double validateAndParseReal(String columnName, Object input, long insertR
...
+   * </ul>
+   *
+   * @param columnName Column name, used in validation error messages
+   * @param input Object to validate and parse
+   * @param insertRowIndex Row index for error reporting
+   * @return Parsed long
+   */
+  static long validateAndParseIcebergLong(String columnName, Object input, long insertRowIndex) {
+    if (input instanceof Number) {
+      return ((Number) input).longValue();
```
sfc-gh-hmadan, line 848: when all can .longValue throw exceptions?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -87,14 +86,13 @@ public void setupSchema(List<ColumnMetadata> columns) {
...
-      ParquetTypeGenerator.ParquetTypeInfo typeInfo =
-          ParquetTypeGenerator.generateColumnParquetTypeInfo(column, id);
+      ParquetTypeInfo typeInfo = ParquetTypeGenerator.generateColumnParquetTypeInfo(column, id);
       parquetTypes.add(typeInfo.getParquetType());
       this.metadata.putAll(typeInfo.getMetadata());
       int columnIndex = parquetTypes.size() - 1;
       fieldIndex.put(
           column.getInternalName(),
-          new ParquetColumn(column, columnIndex, typeInfo.getPrimitiveTypeName()));
+          new ParquetColumn(column, columnIndex, typeInfo.getParquetType()));
```
sfc-gh-hmadan, line 95: where do we validate that only primitive typed typeInfos are allowed for FDN tables?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -205,9 +203,17 @@ private float addRow(
       RowBufferStats forkedStats = statsMap.get(columnName).forkEmpty();
       forkedStatsMap.put(columnName, forkedStats);
       ColumnMetadata column = parquetColumn.columnMetadata;
-      ParquetValueParser.ParquetBufferValue valueWithSize =
-          ParquetValueParser.parseColumnValueToParquet(
-              value, column, parquetColumn.type, forkedStats, defaultTimezone, insertRowsCurrIndex);
+      ParquetBufferValue valueWithSize =
+          (column.getSourceIcebergDataType() == null
```
sfc-gh-hmadan, line 207: In iceberg mode we expect every column to have sourceIcebergDataType populated.
Lets:
1. Add a property to ClientBufferParameters called isIcebergMode, and populate it via the clientInternal.getInternalParameterProvider().getIsIcebergMode().
2. use that boolean to decide which ValueParser to use here instead of column.getSourceIcebergDataType()

3. Wherever appropriate - assert that column.getSourceIcebergDataType is ALWAYS not-null when we're in iceberg mode. This will likely be in channel_open response handling

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -337,10 +343,9 @@ public Flusher<ParquetChunkData> createFlusher() {
   private static class ParquetColumn {
```
sfc-gh-hmadan, line 343: nit: move this class out to its' own file?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetTypeInfo.java
```diff
@@ -0,0 +1,38 @@
...
+  private Map<String, String> metadata;
+
+  public Type getParquetType() {
+    return this.parquetType;
+  }
+
+  public Map<String, String> getMetadata() {
+    return this.metadata;
+  }
+
```
sfc-gh-hmadan, line 26: lets introduce a constructor to this class that forces the callsite to provide both parquetType and metadata at construction time. I checked the one callsite that constructs PTI and it is always calling both setters (at the end of ParquetTypeGenerator.generateColumnParquetTypeInfo), so might as well enforce at construction time than have this randomly mutable state on this object.

Also mark the two private fields in this class as final once you do this.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetTypeGenerator.java
```diff
@@ -94,69 +59,93 @@ static ParquetTypeInfo generateColumnParquetTypeInfo(ColumnMetadata column, int
...
-      case REAL:
-        parquetType =
-            Types.primitive(PrimitiveType.PrimitiveTypeName.DOUBLE, repetition).id(id).named(name);
-        break;
-      default:
+    if (column.getSourceIcebergDataType() != null) {
+      org.apache.iceberg.types.Type icebergDataType =
+          deserializeIcebergType(column.getSourceIcebergDataType());
+      parquetType =
+          typeToMessageType.primitive(icebergDataType.asPrimitiveType(), repetition, id, name);
```
sfc-gh-hmadan, line 66: need error handling here for when a new iceberg data type is returned by service that the client doesn't know how to handle

src/main/java/net/snowflake/ingest/streaming/internal/ParquetTypeGenerator.java
```diff
@@ -1,16 +1,19 @@
 /*
- * Copyright (c) 2022 Snowflake Computing Inc. All rights reserved.
+ * Copyright (c) 2022-2024 Snowflake Computing Inc. All rights reserved.
  */
 
 package net.snowflake.ingest.streaming.internal;
 
+import static net.snowflake.ingest.utils.IcebergDataTypeParser.deserializeIcebergType;
```
sfc-gh-hmadan, line 7: nit: don't think this import static adds much value, it'd be preferable to read `IcebergDataTypeParser.deserializeIcebergType` at the callsite instead of just `deserializeIcebergType`.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetTypeGenerator.java
```diff
@@ -94,69 +59,93 @@ static ParquetTypeInfo generateColumnParquetTypeInfo(ColumnMetadata column, int
...
-            Types.primitive(PrimitiveType.PrimitiveTypeName.BOOLEAN, repetition).id(id).named(name);
-        break;
-      case REAL:
-        parquetType =
-            Types.primitive(PrimitiveType.PrimitiveTypeName.DOUBLE, repetition).id(id).named(name);
-        break;
-      default:
+    if (column.getSourceIcebergDataType() != null) {
+      org.apache.iceberg.types.Type icebergDataType =
+          deserializeIcebergType(column.getSourceIcebergDataType());
```
sfc-gh-hmadan, line 64: curious: why not make deserializeIcebergType also do the typeToMessageType lookup and return a parquetType directly? That way we don't need to have any org.apache.iceberg imports/types in this file at all, everything stays contained in IcebergDataTypeParser?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetTypeGenerator.java
```diff
@@ -94,69 +59,93 @@ static ParquetTypeInfo generateColumnParquetTypeInfo(ColumnMetadata column, int
...
+      parquetType =
```
sfc-gh-hmadan, line 65: since you're only handling primitive types here for now, add an if check that throws when a non-primitive type is encountered.

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
```diff
@@ -0,0 +1,160 @@
...
+
+/**
+ * This class is used to Iceberg data type (include primitive types and nested types) serialization
+ * and deserialization.
+ *
+ * <p>This code is modified from
+ * GlobalServices/modules/data-lake/datalake-api/src/main/java/com/snowflake/metadata/iceberg
+ * /IcebergDataTypeParser.java
+ */
+public class IcebergDataTypeParser {
```
sfc-gh-hmadan, line 27: public?

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
```diff
@@ -0,0 +1,160 @@
...
+import org.apache.iceberg.types.Types;
+import org.apache.iceberg.util.JsonUtil;
+
+/**
+ * This class is used to Iceberg data type (include primitive types and nested types) serialization
+ * and deserialization.
+ *
+ * <p>This code is modified from
+ * GlobalServices/modules/data-lake/datalake-api/src/main/java/com/snowflake/metadata/iceberg
+ * /IcebergDataTypeParser.java
```
sfc-gh-hmadan, line 25: please bring over all the tests from the snowflake repo too!

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
```diff
@@ -0,0 +1,160 @@
...
+   *
+   * @param icebergDataType string representation of Iceberg data type
+   * @return Iceberg data type
+   */
+  public static Type deserializeIcebergType(String icebergDataType) {
+    try {
+      JsonNode json = MAPPER.readTree(icebergDataType);
+      return getTypeFromJson(json);
+    } catch (IOException e) {
+      throw new SFException(ErrorCode.INTERNAL_ERROR, "Failed to deserialize Iceberg data type", e);
```
sfc-gh-hmadan, line 60: I see in FDN case we throw UNKNOWN_DATA_TYPE ?

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
```diff
@@ -0,0 +1,160 @@
...
+   * Get Iceberg data type information by deserialization.
+   *
+   * @param icebergDataType string representation of Iceberg data type
+   * @return Iceberg data type
+   */
+  public static Type deserializeIcebergType(String icebergDataType) {
+    try {
+      JsonNode json = MAPPER.readTree(icebergDataType);
+      return getTypeFromJson(json);
+    } catch (IOException e) {
```
sfc-gh-hmadan, line 59: curious, why only IOException? readTree throws JsonProcessingException, JsonMappingException.

Lets add testcases for this situation where either readTree fails on bad JSON, or getTypeFromJson encounters an unknown data type.

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
```diff
@@ -0,0 +1,160 @@
...
+      if (STRUCT.equals(type)) {
+        return structFromJson(jsonNode);
+      } else if (LIST.equals(type)) {
+        return listFromJson(jsonNode);
+      } else if (MAP.equals(type)) {
+        return mapFromJson(jsonNode);
+      }
+    }
+
+    throw new SFException(ErrorCode.INTERNAL_ERROR, "Cannot parse Iceberg type from: " + jsonNode);
```
sfc-gh-hmadan, line 84: 1. instead of reserializing jsonNode I'd pass in the original string and write that out to logs / to the error message.

2. (P2) It might be a good idea to only throw SFExceptions from "higher" layers and not from utility parsing classes like this one, IllegalArgumentException probably is good enough here; lets chat with Toby on whether there's already a defined scheme to which all levels can/should throw SFException.

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
```diff
@@ -0,0 +1,160 @@
...
+   * Get corresponding Iceberg data type from JsonNode.
+   *
+   * @param jsonNode JsonNode parsed from Iceberg type string.
+   * @return Iceberg data type
+   */
+  public static Type getTypeFromJson(@Nonnull JsonNode jsonNode) {
+    if (jsonNode.isTextual()) {
+      return Types.fromPrimitiveString(jsonNode.asText());
+    } else if (jsonNode.isObject()) {
+      String type = jsonNode.get(TYPE).asText();
```
sfc-gh-hmadan, line 74: what if there is no TYPE field on jsonNode? asText will throw a null ref exception?

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
```diff
@@ -0,0 +1,160 @@
...
+      if (STRUCT.equals(type)) {
+        return structFromJson(jsonNode);
+      } else if (LIST.equals(type)) {
+        return listFromJson(jsonNode);
+      } else if (MAP.equals(type)) {
+        return mapFromJson(jsonNode);
+      }
```
sfc-gh-hmadan, line 81: else throw a specific exception with what type was found on the node. Highly specific error messages are worth their weight in gold!

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
```diff
@@ -0,0 +1,160 @@
...
+
+  /**
+   * Get Iceberg struct type information from JsonNode.
+   *
+   * @param json JsonNode parsed from Iceberg type string.
+   * @return struct type
+   */
+  public static @Nonnull Types.StructType structFromJson(@Nonnull JsonNode json) {
+    JsonNode fieldArray = json.get(FIELDS);
+    Preconditions.checkArgument(
```
sfc-gh-hmadan, line 95: also check if fieldArray is null? Or does json.get already do that?

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
```diff
@@ -0,0 +1,160 @@
...
+        fieldArray.isArray(), "Cannot parse struct fields from non-array: %s", fieldArray);
+
+    List<Types.NestedField> fields = Lists.newArrayListWithExpectedSize(fieldArray.size());
+    Iterator<JsonNode> iterator = fieldArray.elements();
```
sfc-gh-hmadan, line 99: can elements be of size 0? Or must it be > 0?

scripts/process_licenses.py
sfc-gh-alhuang: It's a transitive dependency from `org.apache.iceberg:iceberg-core`.

```
[INFO] +- org.apache.iceberg:iceberg-core:jar:1.3.1:compile
[INFO] |  +- org.apache.iceberg:iceberg-common:jar:1.3.1:runtime
[INFO] |  +- org.apache.avro:avro:jar:1.11.1:runtime
[INFO] |  +- org.apache.httpcomponents.client5:httpclient5:jar:5.2.1:runtime
[INFO] |  |  +- org.apache.httpcomponents.core5:httpcore5:jar:5.2:runtime
[INFO] |  |  \- org.apache.httpcomponents.core5:httpcore5-h2:jar:5.2:runtime
[INFO] |  \- org.roaringbitmap:RoaringBitmap:jar:0.9.44:runtime
[INFO] |     \- org.roaringbitmap:shims:jar:0.9.44:runtime
```

src/main/java/net/snowflake/ingest/streaming/internal/ColumnMetadata.java
sfc-gh-alhuang: Added at variable declaration.

src/main/java/net/snowflake/ingest/streaming/internal/OpenChannelRequestInternal.java
sfc-gh-alhuang: Added

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
sfc-gh-alhuang: `getFields ` does not flatten the subcolumns. This change is going to be used in structured data, reverted here and will be in next PR.

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
sfc-gh-alhuang: Added

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-alhuang: Discussed via slack. Added range check and rounding.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-alhuang: Discussed via slack. Added range check and rounding.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: The FDN type -> parquetType function `generateColumnParquetTypeInfo` only returns primitive type. Should we add assertion here?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: Added `isIcebergMode` in ClientBufferParameters and added check in SnowflakeStreamingIngestClientInternal.openChannel.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: Moved.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetTypeInfo.java
sfc-gh-alhuang: Created ctor and removed setters.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetTypeGenerator.java
sfc-gh-alhuang: Added check in IcebergDataTypeParser. If a new iceberg data type is introduced in server, `Types.fromPrimitive` will throw `IllegalArgumentException`, do you think this is sufficient?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetTypeGenerator.java
sfc-gh-alhuang: Removed and call with class name as prefix.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetTypeGenerator.java
sfc-gh-alhuang: Make sense, moved all logic into `deserializeIcebergType`

src/main/java/net/snowflake/ingest/streaming/internal/ParquetTypeGenerator.java
sfc-gh-alhuang: Added in `deserializeIcebergType`

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
sfc-gh-alhuang: This class is in utils package. Remain public since we need to use it from internal package.

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
sfc-gh-alhuang: Added tests.

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
sfc-gh-alhuang: This exception is thrown cause json parse error, imo this is an internal error as this should not happen if the server side is passing valid json string.

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
sfc-gh-alhuang: Updated to `JsonProcessingException` to cover it's derived class `JsonMappingException`. Added test for invalid json format and unknown datatype.

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
sfc-gh-alhuang: Added check for `type` field.

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
sfc-gh-alhuang: Added `IllegalArgumentException` for known types.

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
sfc-gh-alhuang: Added check.

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
sfc-gh-alhuang: The elements size of a group field can be zero.

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
sfc-gh-alhuang: Changed all exception to `IllegalArgurmentException` in this class.

sfc-gh-tzhang:
@sfc-gh-azagrebin @sfc-gh-lsembera could you help with reviewing Parquet writer and data type related changes in this PR? Thanks!

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -957,6 +957,82 @@ static double validateAndParseReal(String columnName, Object input, long insertR
...
+   * </ul>
+   *
+   * @param columnName Column name, used in validation error messages
+   * @param input Object to validate and parse
+   * @param insertRowIndex Row index for error reporting
+   * @return Parsed long
+   */
+  static long validateAndParseIcebergLong(String columnName, Object input, long insertRowIndex) {
+    if (input instanceof Number) {
+      double value = ((Number) input).doubleValue();
```
sfc-gh-lsembera, line 1013: This is a narrowing conversion, which loses precision, see:

```java
    long l1 = 499999999000000001L;
    double d = l1;
    long l2 = (long) d;
    System.out.println(l1);
    System.out.println(d);
    System.out.println(l2);
```
yields:
```plain
499999999000000001
4.99999999E17
499999999000000000
```

It is better to use `BigDecimal`.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -957,6 +957,82 @@ static double validateAndParseReal(String columnName, Object input, long insertR
...
+      double roundedDouble = (value > 0) ? Math.floor(value + 0.5) : Math.ceil(value - 0.5);
```
sfc-gh-lsembera, line 1014: Why do we have different rounding logic for ints and for longs?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -957,6 +957,82 @@ static double validateAndParseReal(String columnName, Object input, long insertR
...
+      if (roundedDouble > Long.MAX_VALUE || roundedDouble < Long.MIN_VALUE || Double.isNaN(value)) {
```
sfc-gh-lsembera, line 1015: What about `Double.NEGATIVE_INFINITY` and `Double.POSITIVE_INFINITY`, are they allowed?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -957,6 +957,82 @@ static double validateAndParseReal(String columnName, Object input, long insertR
...
+   *   <li>String
+   * </ul>
+   *
+   * @param columnName Column name, used in validation error messages
+   * @param input Object to validate and parse
+   * @param insertRowIndex Row index for error reporting
+   * @return Parsed long
+   */
+  static long validateAndParseIcebergLong(String columnName, Object input, long insertRowIndex) {
+    if (input instanceof Number) {
```
sfc-gh-lsembera, line 1012: Should we return right away for integer types without going via double?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -957,6 +957,82 @@ static double validateAndParseReal(String columnName, Object input, long insertR
...
+   *   <li>Number
+   *   <li>String
+   * </ul>
+   *
+   * @param columnName Column name, used in validation error messages
+   * @param input Object to validate and parse
+   * @param insertRowIndex Row index for error reporting
+   * @return Parsed integer
+   */
+  static int validateAndParseIcebergInt(String columnName, Object input, long insertRowIndex) {
```
sfc-gh-lsembera, line 973: Should we add unit tests for these functions, which focus on corner cases? Min and max values, Double/Float NaN+positive/negative infinity, big integers and big decimals outside of the allowed range, etc.?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-alhuang: Switched to `BigDecimal`.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-alhuang: It was used to deal with the `double.round` only lies within `Long` range. Switched to `BigDecimal` and not needed anymore.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-alhuang: Fixed. Used the logic from `validateAndParseBigDecimal`.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-alhuang: Added tests [here](https://github.com/snowflakedb/snowflake-ingest-java/blob/21afa2b86f4f4523f905fab3ac5ad622c4d62cd3/src/test/java/net/snowflake/ingest/streaming/internal/DataValidationUtilTest.java#L1200).

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-alhuang: Should not be allowed. Added logic in `validateAndParseBigDecimal` to avoid this.

sfc-gh-lsembera (commented):
I reviewed the int/long validation in `DataValidationUtil` and it LGTM, left two small comments for your consideration. 

For BDEC ingestion, we have a suite of [integration tests](https://github.com/snowflakedb/snowflake-ingest-java/tree/master/src/test/java/net/snowflake/ingest/streaming/internal/datatypes) testing data types end to end, i.e. including server-side scanning. I will leave it up to you if you think you need it for Parquet ingestion, as well.

src/main/java/net/snowflake/ingest/streaming/internal/OpenChannelRequestInternal.java
```diff
@@ -32,6 +32,10 @@ class OpenChannelRequestInternal implements IStreamingIngestRequest {
   @JsonProperty("write_mode")
   private String writeMode;
 
+  @JsonInclude(JsonInclude.Include.NON_NULL)
+  @JsonProperty("is_iceberg")
+  private boolean isIceberg;
```
sfc-gh-hmadan, line 37: looks like you need to merge with my merge from monday?

sfc-gh-hmadan: ah np, ignore this was mixing up with some other change.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -345,48 +346,56 @@ public SnowflakeStreamingIngestChannelInternal<?> openChannel(OpenChannelRequest
...
-      return channel;
+      response = snowflakeServiceClient.openChannel(openChannelRequest);
     } catch (IOException | IngestResponseException e) {
       throw new SFException(e, ErrorCode.OPEN_CHANNEL_FAILURE, e.getMessage());
     }
+
+    if (isIcebergMode
+        && response.getTableColumns().stream()
+            .anyMatch(c -> c.getSourceIcebergDataType() == null)) {
+      throw new SFException(
```
sfc-gh-hmadan, line 359: for next PR: also log this out before throwing, and log the request id too.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetTypeGenerator.java
```diff
@@ -94,73 +53,94 @@ static ParquetTypeInfo generateColumnParquetTypeInfo(ColumnMetadata column, int
...
-            Types.primitive(PrimitiveType.PrimitiveTypeName.BOOLEAN, repetition).id(id).named(name);
-        break;
-      case REAL:
-        parquetType =
-            Types.primitive(PrimitiveType.PrimitiveTypeName.DOUBLE, repetition).id(id).named(name);
-        break;
-      default:
+    if (column.getSourceIcebergDataType() != null) {
+      parquetType =
+          IcebergDataTypeParser.parseIcebergDataTypeStringToParquetType(
```
sfc-gh-hmadan, line 58: for next PR: add testcase and throw proper error if an unknown data type is encountered

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -169,15 +168,18 @@ private float addRow(
...
-              value,
-              column,
-              parquetColumn.type,
-              forkedStats,
-              defaultTimezone,
-              insertRowsCurrIndex,
-              clientBufferParameters.isEnableNewJsonParsingLogic());
+      ParquetBufferValue valueWithSize =
+          (clientBufferParameters.getIsIcebergMode()
+              ? IcebergParquetValueParser.parseColumnValueToParquet(
```
sfc-gh-hmadan, line 173: any impact of the whole isEnableNewJsonParsingLogic business for iceberg parser?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -261,6 +263,10 @@ Object getVectorValueAt(String column, int index) {
     if (logicalType == ColumnLogicalType.BINARY && value != null) {
       value = value instanceof String ? ((String) value).getBytes(StandardCharsets.UTF_8) : value;
     }
+    /* Mismatch between Iceberg string & FDN String */
+    if (Objects.equals(columnMetadata.getSourceIcebergDataType(), "\"string\"")) {
+      value = value instanceof byte[] ? new String((byte[]) value, StandardCharsets.UTF_8) : value;
```
sfc-gh-hmadan, line 268: this is test-only code so np, but otherwise assuming the byte[] is UTF8 is a bit risky and shouldn't be done.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetTypeGenerator.java
```diff
@@ -94,73 +53,94 @@ static ParquetTypeInfo generateColumnParquetTypeInfo(ColumnMetadata column, int
...
-      case BOOLEAN:
-        parquetType =
-            Types.primitive(PrimitiveType.PrimitiveTypeName.BOOLEAN, repetition).id(id).named(name);
-        break;
-      case REAL:
-        parquetType =
-            Types.primitive(PrimitiveType.PrimitiveTypeName.DOUBLE, repetition).id(id).named(name);
-        break;
-      default:
+    if (column.getSourceIcebergDataType() != null) {
```
sfc-gh-hmadan, line 56: for next PR: add validation that we only ever see column.sourceIcebergDataType as non-null when isIcebergMode is true. I saw you're already checking if this field is null on any column when we're in iceberg mode, need similar check for non-iceberg-mode.

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
sfc-gh-hmadan: for next PR: This will allow customers to directly instantiate this class because its public, lets discuss.

sfc-gh-hmadan (approved):
Reviewed all files except the following (verified they're properly protected behind isIceberg flag, will get to them again later this week), signing off to unblock checkin.

IcebergParquetValueParser.java
IcebergDataTypeParser.java
DataValidationUtil.java
testcases

src/test/java/net/snowflake/ingest/streaming/internal/DataValidationUtilTest.java
```diff
@@ -1190,6 +1196,89 @@ public void testValidateAndParseBoolean() {
...
+    expectError(
+        ErrorCode.INVALID_VALUE_ROW,
+        () -> validateAndParseIcebergInt("COL", Float.NEGATIVE_INFINITY, 0));
+  }
+
+  @Test
+  public void testValidateAndParseIcebergLong() {
+    assertEquals(1L, validateAndParseIcebergLong("COL", 1, 0));
+    assertEquals(1L, validateAndParseIcebergLong("COL", 1L, 0));
+    assertEquals(1L, validateAndParseIcebergLong("COL", 1.499f, 0));
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Added test for `-.0f`. IT for all Iceberg types are expect to be added when server side Iceberg file registration is ready, backlogged to [jira](https://snowflakecomputing.atlassian.net/browse/SNOW-1490151).

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -1026,6 +1103,16 @@ private static boolean convertStringToBoolean(
         || "on".equals(normalizedInput);
   }
 
+  private static BigDecimal convertNumberToBigDecimal(Number value, final long insertRowIndex) {
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Redundant code. Removed.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: The `isEnableNewJsonParsingLogic` is used for semi structured data type, which will not impact Iceberg parsing.

src/test/java/net/snowflake/ingest/streaming/internal/IcebergDataTypeParserTest.java
```diff
@@ -0,0 +1,107 @@
...
+    dataTypesToTest.add(new DataTypeInfo("\"boolean\"", Types.BooleanType.get()));
+    dataTypesToTest.add(new DataTypeInfo("\"int\"", Types.IntegerType.get()));
+    dataTypesToTest.add(new DataTypeInfo("\"long\"", Types.LongType.get()));
+    dataTypesToTest.add(new DataTypeInfo("\"float\"", Types.FloatType.get()));
+    dataTypesToTest.add(new DataTypeInfo("\"double\"", Types.DoubleType.get()));
+    dataTypesToTest.add(new DataTypeInfo("\"string\"", Types.StringType.get()));
+    dataTypesToTest.add(new DataTypeInfo("\"date\"", Types.DateType.get()));
+    dataTypesToTest.add(new DataTypeInfo("\"time\"", Types.TimeType.get()));
+    dataTypesToTest.add(new DataTypeInfo("\"timestamptz\"", Types.TimestampType.withZone()));
+    dataTypesToTest.add(
```
sfc-gh-hmadan, line 49: the structured data type testing needs to cover more cases; and also account for nested schema parsing.

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
```diff
@@ -278,51 +280,58 @@ public void write(List<Object> values) {
...
+    }
+
+    private void writeValues(List<Object> values, GroupType type) {
+      List<Type> cols = type.getFields();
       for (int i = 0; i < cols.size(); ++i) {
         Object val = values.get(i);
-        // val.length() == 0 indicates a NULL value.
         if (val != null) {
-          String fieldName = cols.get(i).getPath()[0];
+          String fieldName = cols.get(i).getName();
```
sfc-gh-hmadan, line 293: whats the difference between getPath()[0] and getName() ?

sfc-gh-alhuang: The `getPath()[0]` always returns the root column name. While `getName()` returns the current column/subcolumn name. This does matters for primitive data type column. Is needed for structured data type.

src/test/java/net/snowflake/ingest/streaming/internal/IcebergDataTypeParserTest.java
sfc-gh-alhuang: Added to [jira](https://snowflakecomputing.atlassian.net/browse/SNOW-1490151).

PR #817, opened 2024-08-26
SNOW-1566045 Preserve numeric format for variants

sfc-gh-lsembera:
Currently, ingesting string JSON values into semi-structured types (i.e. `variant`, `object` and `array`), the SDK may modify the scientific notation of the input, i.e. convert values like `12.34` to `1.234e1` and vice versa. This PR makes sure the input format is preserved.

---

I micro-benchmarked parsing large json documents locally with JMH and the new approach also gives us ~50% speedup, it is likely due to the fact that we don't deserialize into the intermediate Jackson AST anymore:

```plain
NEW WAY
=======
Benchmark                              Mode  Cnt       Score       Error  Units
InsertRowsBenchmarkTest.testInsertRow  avgt   10  406466.888 ± 45677.292  us/op

OLD WAY
=======
Benchmark                              Mode  Cnt       Score       Error  Units
InsertRowsBenchmarkTest.testInsertRow  avgt   10  620401.749 ± 14929.263  us/op
```

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
@@ -6,11 +6,17 @@ package net.snowflake.ingest.streaming.internal;
 
 import static net.snowflake.ingest.streaming.internal.BinaryStringUtils.unicodeCharactersCount;
 
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonParseException;
+import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.JsonToken;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.module.SimpleModule;
 import com.fasterxml.jackson.databind.ser.std.ToStringSerializer;
+import java.io.IOException;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.nio.charset.StandardCharsets;
@@ -39,6 +45,7 @@ import net.snowflake.ingest.utils.ErrorCode;
 import net.snowflake.ingest.utils.SFException;
 import org.apache.commons.codec.DecoderException;
 import org.apache.commons.codec.binary.Hex;
+import org.apache.commons.io.output.StringBuilderWriter;
 
 /** Utility class for parsing and validating inputs based on Snowflake types */
 class DataValidationUtil {
@@ -69,6 +76,8 @@ class DataValidationUtil {
   static final int MAX_SEMI_STRUCTURED_LENGTH = BYTES_16_MB - 64;
 
   private static final ObjectMapper objectMapper = new ObjectMapper();
+
+  private static final JsonFactory factory = new JsonFactory();
 
   // The version of Jackson we are using does not support serialization of date objects from the
   // java.time package. Here we define a module with custom java.time serializers. Additionally, we
@@ -140,6 +149,61 @@ class DataValidationUtil {
    * see {@link DataValidationUtil#isAllowedSemiStructuredType}.
    *
    * @param input Object to validate
+   * @return Minified JSON string
+   */
+  private static String validateAndParseSemiStructured(
+      String columnName, Object input, String snowflakeType, final long insertRowIndex) {
+    if (input instanceof String) {
+      final String stringInput = (String) input;
+      verifyValidUtf8(stringInput, columnName, snowflakeType, insertRowIndex);
+      final StringBuilderWriter resultWriter = new StringBuilderWriter(stringInput.length());
+      try (final JsonParser parser = factory.createParser(stringInput);
+          final JsonGenerator generator = factory.createGenerator(resultWriter)) {
+        while (parser.nextToken() != null) {
+          final JsonToken token = parser.currentToken();
+          if (token.isNumeric()) {
+            // If the current token is a number, we cannot just copy the current event because it
+            // would write token the token from double (or big decimal), whose scientific notation
+            // may have been altered during deserialization. We want to preserve the scientific
+            // notation from the user input, so we write the current numer as text.
+            generator.writeNumber(parser.getText());
+          } else {
+            generator.copyCurrentEvent(parser);
+          }
+        }
+      } catch (JsonParseException e) {
+        throw valueFormatNotAllowedException(
+            columnName, snowflakeType, "Not a valid JSON", insertRowIndex);
+      } catch (IOException e) {
+        throw new SFException(e, ErrorCode.IO_ERROR, "Cannot create JSON Parser or JSON generator");
+      }
+      // We return the minified string from the result writer
+      return resultWriter.toString();
+    } else if (isAllowedSemiStructuredType(input)) {
+      JsonNode node = objectMapper.valueToTree(input);
+      return node.toString();
+    }
+
+    throw typeNotAllowedException(
+        columnName,
+        input.getClass(),
+        snowflakeType,
+        new String[] {
+          "String",
+          "Primitive data types and their arrays",
+          "java.time.*",
+          "List<T>",
+          "Map<String, T>",
+          "T[]"
+        },
+        insertRowIndex);
+  }
+
+  /**
+   * Validates and parses input as JSON. All types in the object tree must be valid variant types,
+   * see {@link DataValidationUtil#isAllowedSemiStructuredType}.
+   *
+   * @param input Object to validate
    * @param insertRowIndex
    * @return JSON string representing the input
    */
@@ -163,6 +227,34 @@ class DataValidationUtil {
           insertRowIndex);
     }
     return output;
+  }
+
+  /**
+   * Validates and parses input as JSON. All types in the object tree must be valid variant types,
+   * see {@link DataValidationUtil#isAllowedSemiStructuredType}.
+   *
+   * @param input Object to validate
+   * @param insertRowIndex
+   * @return JSON string representing the input
+   */
+  static String validateAndParseVariantNew(String columnName, Object input, long insertRowIndex) {
+    final String result =
+        validateAndParseSemiStructured(columnName, input, "VARIANT", insertRowIndex);
+
+    // Empty json strings are ingested as nulls
+    if (result.isEmpty()) {
+      return null;
+    }
+    int stringLength = result.getBytes(StandardCharsets.UTF_8).length;
+    if (stringLength > MAX_SEMI_STRUCTURED_LENGTH) {
+      throw valueFormatNotAllowedException(
+          columnName,
+          "VARIANT",
+          String.format(
+              "Variant too long: length=%d maxLength=%d", stringLength, MAX_SEMI_STRUCTURED_LENGTH),
+          insertRowIndex);
+    }
+    return result;
   }
 
   /**
@@ -300,6 +392,41 @@ class DataValidationUtil {
   }
 
   /**
+   * Validates and parses JSON array. Non-array types are converted into single-element arrays. All
+   * types in the array tree must be valid variant types, see {@link
+   * DataValidationUtil#isAllowedSemiStructuredType}.
+   *
+   * @param input Object to validate
+   * @param insertRowIndex
+   * @return JSON array representing the input
+   */
+  static String validateAndParseArrayNew(String columnName, Object input, long insertRowIndex) {
+    String result = validateAndParseSemiStructured(columnName, input, "ARRAY", insertRowIndex);
+    if (result.isEmpty()) {
+      // Empty input is ingested as an array of null
+      result =
+          JsonToken.START_ARRAY.asString()
+              + JsonToken.VALUE_NULL.asString()
+              + JsonToken.END_ARRAY.asString();
+    } else if (!result.startsWith(JsonToken.START_ARRAY.asString())) {
+      // Non-array values are ingested as single-element arrays, mimicking the Worksheets behavior
+      result = JsonToken.START_ARRAY.asString() + result + JsonToken.END_ARRAY.asString();
+    }
+
+    // Throw an exception if the size is too large
+    int stringLength = result.getBytes(StandardCharsets.UTF_8).length;
+    if (stringLength > MAX_SEMI_STRUCTURED_LENGTH) {
+      throw valueFormatNotAllowedException(
+          columnName,
+          "ARRAY",
+          String.format(
+              "Array too large. length=%d maxLength=%d", stringLength, MAX_SEMI_STRUCTURED_LENGTH),
+          insertRowIndex);
+    }
+    return result;
+  }
+
+  /**
    * Validates and parses JSON object. Input is rejected if the value does not represent JSON object
    * (e.g. String '{}' or Map<String, T>). All types in the object tree must be valid variant types,
    * see {@link DataValidationUtil#isAllowedSemiStructuredType}.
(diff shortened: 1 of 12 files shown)


sfc-gh-xhuang:
enable long running tests too

sfc-gh-lsembera:
Converted back to draft because I need to add parameter protection.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -103,21 +111,38 @@ private static BigDecimal[] makePower10Table() {
...
+            generator.copyCurrentEvent(parser);
+          }
+        }
+      } catch (JsonParseException e) {
         throw valueFormatNotAllowedException(
             columnName, snowflakeType, "Not a valid JSON", insertRowIndex);
+      } catch (IOException e) {
+        throw new SFException(e, ErrorCode.IO_ERROR, "Cannot create JSON Parser or JSON generator");
       }
+      return resultWriter.toString();
```
sfc-gh-hmadan, line 142: please add a comment as to why we shouldn't return the original string back?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -70,6 +76,8 @@ class DataValidationUtil {
 
   private static final ObjectMapper objectMapper = new ObjectMapper();
 
+  private static final JsonFactory factory = new JsonFactory();
```
sfc-gh-hmadan, line 79: the objectmapper is configured in a static ctor for a couple data types, does this factory object not need a similar configuration?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -103,21 +111,38 @@ private static BigDecimal[] makePower10Table() {
...
+            // would write token the token from double (or big decimal), whose scientific notation
+            // may have been altered during deserialization. We want to preserve the scientific
+            // notation from the user input, so we write the current numer as text.
+            generator.writeNumber(parser.getText());
+          } else {
+            generator.copyCurrentEvent(parser);
+          }
+        }
+      } catch (JsonParseException e) {
         throw valueFormatNotAllowedException(
```
sfc-gh-hmadan, line 137: we're losing all information about the original exception (e) in this throw. can `e` be added as an inner exception ? Or log it out before throwing?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -103,21 +111,38 @@ private static BigDecimal[] makePower10Table() {
...
+      try (final JsonParser parser = factory.createParser(stringInput);
+          final JsonGenerator generator = factory.createGenerator(resultWriter)) {
+        while (parser.nextToken() != null) {
+          final JsonToken token = parser.currentToken();
+          if (token.isNumeric()) {
+            // If the current token is a number, we cannot just copy the current event because it
+            // would write token the token from double (or big decimal), whose scientific notation
+            // may have been altered during deserialization. We want to preserve the scientific
+            // notation from the user input, so we write the current numer as text.
+            generator.writeNumber(parser.getText());
```
sfc-gh-hmadan, line 131: does datetime parsing also introduce such mutations that need to be avoided? please add tests to verify.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -103,21 +111,38 @@ private static BigDecimal[] makePower10Table() {
...
+      } catch (JsonParseException e) {
         throw valueFormatNotAllowedException(
             columnName, snowflakeType, "Not a valid JSON", insertRowIndex);
+      } catch (IOException e) {
+        throw new SFException(e, ErrorCode.IO_ERROR, "Cannot create JSON Parser or JSON generator");
       }
+      return resultWriter.toString();
     } else if (isAllowedSemiStructuredType(input)) {
-      return objectMapper.valueToTree(input);
+      JsonNode node = objectMapper.valueToTree(input);
```
sfc-gh-hmadan, line 144: curious, why don't we need a similar approach in non-string input fields? please add a comment too.

The streaming reader/writer being used just for String is apt for all kinds of data, so ideally we shouldn't need this DOM-parse of the json anywhere now?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -144,16 +169,14 @@ private static JsonNode validateAndParseSemiStructuredAsJsonTree(
...
    */
   static String validateAndParseVariant(String columnName, Object input, long insertRowIndex) {
-    JsonNode node =
-        validateAndParseSemiStructuredAsJsonTree(columnName, input, "VARIANT", insertRowIndex);
+    final String result =
+        validateAndParseSemiStructured(columnName, input, "VARIANT", insertRowIndex);
 
-    // Missing nodes are not valid json, ingest them as NULL instead
-    if (node.isMissingNode()) {
+    // Empty json strings are ingested as nulls
```
sfc-gh-hmadan, line 175: Does a "" get replaced with NULL? If so that's us corrupting data, they're not really equivalent values.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -277,17 +300,20 @@ static boolean isAllowedSemiStructuredType(Object o) {
...
-    // Non-array values are ingested as single-element arrays, mimicking the Worksheets behavior
-    if (!jsonNode.isArray()) {
-      jsonNode = objectMapper.createArrayNode().add(jsonNode);
+    String result = validateAndParseSemiStructured(columnName, input, "ARRAY", insertRowIndex);
+    if (result.isEmpty()) {
+      // Empty input is ingested as an array of null
+      result =
+          JsonToken.START_ARRAY.asString()
+              + JsonToken.VALUE_NULL.asString()
+              + JsonToken.END_ARRAY.asString();
```
sfc-gh-hmadan, line 309: Does this mean an  `"arrayKey": NULL` and/or `"arrayKey": ""` will get converted into `"arrayKey": []` ?

The latter needs to throw, and the former should stay as null (since NULL != empty array)

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -309,15 +335,13 @@ static String validateAndParseArray(String columnName, Object input, long insert
...
    */
   static String validateAndParseObject(String columnName, Object input, long insertRowIndex) {
-    JsonNode jsonNode =
-        validateAndParseSemiStructuredAsJsonTree(columnName, input, "OBJECT", insertRowIndex);
-    if (!jsonNode.isObject()) {
+    final String result =
+        validateAndParseSemiStructured(columnName, input, "OBJECT", insertRowIndex);
+    if (!result.startsWith(JsonToken.START_OBJECT.asString())) {
       throw valueFormatNotAllowedException(columnName, "OBJECT", "Not an object", insertRowIndex);
     }
```
sfc-gh-hmadan, line 342: nit: why shouldn't validateAndParse do this validation instead of expecting the callsite to do it?

e2e-jar-test/fips/src/test/java/net/snowflake/FipsIngestE2ETest.java
```diff
@@ -32,7 +31,6 @@ public void basicTest() throws InterruptedException {
   }
 
   @Test
-  @Ignore("Takes too long to run")
   public void longRunningTest() throws InterruptedException {
```
sfc-gh-hmadan, line 34: :+1:

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -103,21 +111,38 @@ private static BigDecimal[] makePower10Table() {
...
+  private static String validateAndParseSemiStructured(
       String columnName, Object input, String snowflakeType, final long insertRowIndex) {
     if (input instanceof String) {
-      String stringInput = (String) input;
+      final String stringInput = (String) input;
       verifyValidUtf8(stringInput, columnName, snowflakeType, insertRowIndex);
-      try {
-        return objectMapper.readTree(stringInput);
-      } catch (JsonProcessingException e) {
+      final StringBuilderWriter resultWriter = new StringBuilderWriter(stringInput.length());
```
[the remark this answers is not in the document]
sfc-gh-lsembera: I think it this case, it is not a good idea to have server-side controlled parameter. This is a behavior change, which customers can opt into by upgrading the SDK.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-lsembera: This code is triggered when customers pass a string into a variant column, so we never pass dates here as the input is just a JSON.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-lsembera: They are equivalent. `select parse_json('');` yields `null`. `''` and `'""'` are different variants.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-lsembera: We do the validation here to reject incorrect user input we have to protect against, otherwise we would let the customer ingest incorrect data.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-lsembera: When the users ingest a `java.lang.String` and the target column is `variant`, `array` or `object`, we have to validate it, minify it, and store it in the data file, which is what the previous block with `JsonParser` and `JsonGenerator` does. 

On the other hand, If the user passes another supported Java type, e.g. `Map<String, Object>`, `List<Object>`, `LocalDateTime` or many other types - see the list [here](https://docs.snowflake.com/user-guide/data-load-snowpipe-streaming-overview#supported-java-data-types)), we need to serialize it into a JSON string, which we then store in the data file. In this case there is nothing to parse, we have to serialize the Java object input into String, which is why we use the intermediate AST.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-lsembera: It does not, the configured serializers only serialize custom Java data types, which we don't have when we work with JSON provided as string.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-lsembera: To ingest an empty string, customers need to ingest Java string `"\"\""`.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-lsembera: This is for the `ARRAY` data type, and it means that `""` gets converted into `[null]`. It is an inconsistency with other semi-structured data types, and we should address - I created SNOW-1641130.

e2e-jar-test/fips/src/test/java/net/snowflake/FipsIngestE2ETest.java
sfc-gh-tzhang: Long running test should not be part of the merge gate, let's think about something else if we want the coverage for every PR

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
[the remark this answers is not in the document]
sfc-gh-tzhang: > do we need a server-controlled param protection setup, to selectively enable/disable such features for customers and also manage BCRs?

This topic of supporting server side parameter control has been discussed for a few times, in general I think it's a good feature to have, but we haven't find resources to do it.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-tzhang: > This is on purpose. This exception is thrown when customers pass a value, which cannot be ingested into the target column type, e.g. because the value does not match the column type, or the value is too large for the column.

Wait, why this is on purpose? It would be good to throw the original exception, then customer can look at it in the error response and fix the input?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-tzhang: Is this a new behavior or not?

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -46,6 +46,9 @@ public class ParameterProvider {
   public static final String BDEC_PARQUET_COMPRESSION_ALGORITHM =
       "BDEC_PARQUET_COMPRESSION_ALGORITHM".toLowerCase();
 
+  public static final String ENABLE_NEW_JSON_PARSING_LOGIC =
```
sfc-gh-tzhang, line 49: I don't think we need this parameter control, if this is to prevent regression, customer can go back to the previous version. 

The goal of this class is not for enable/disable parameter control, 99% of the parameters in the class are added because not single value would work for all cases.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-lsembera: It is not a new behavior, see [here](https://github.com/snowflakedb/snowflake-ingest-java/blob/master/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java#L284) for equivalent behavior in existing code. It should be addressed as a separate issue.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-lsembera: Discuss with @sfc-gh-tzhang offline, for this PR we will keep the parameter there, I created SNOW-1646229 for a cleanup.

e2e-jar-test/fips/src/test/java/net/snowflake/FipsIngestE2ETest.java
sfc-gh-tzhang: Sounds good, I'm fine as long as it won't run hours as part of merge gate and block every PR

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-tzhang: > and we must not put customer data into logs

Agree we should not put it in the log, what I thought is to include the original exception as part of the `InsertValidationResponse` so customer can audit the error and fix it if needed.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-tzhang: sg if it's not a new behavior, thanks

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-tzhang: Could you please add the TODO with JIRA number in the comment?

sfc-gh-tzhang (approved):
LGTM, thanks for fixing this!

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-lsembera: We can do that, but as a separate PR. My PR is not introducing any changes in how our validation exceptions look like.

PR #821, opened 2024-09-04
Revert "SNOW-1566045 Preserve numeric format for variants (#817)"

sfc-gh-lsembera:
This reverts commit 97bd52a1b7aaa3579447d99fb0f26197907a1fb1.

We need to temporarily revert the BCR commit, so we can do non-BCR release first.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -426,10 +426,6 @@
       <groupId>commons-codec</groupId>
       <artifactId>commons-codec</artifactId>
     </dependency>
-    <dependency>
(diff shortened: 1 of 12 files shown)


sfc-gh-xhuang (approved):

sfc-gh-tzhang:
Why we need to revert this change, we can just release on a separate branch, right?

sfc-gh-lsembera:
> Why we need to revert this change, we can just release on a separate branch, right?

Ok, I didn't know that.

PR #824, opened 2024-09-07
SNOW-1654124: Write file name to metadata at the place when we create the file

sfc-gh-tzhang:
- Write file name to metadata at the place when we create the file (only works for `serializeFromJavaObjects`)
- Looks like `serializeFromParquetWriteBuffers` is not maintained to work with `primaryFileId` as the `ParquetWriter` is created during `setupSchema`, we would better remove the code as a whole if we decide to not use it, or a bigger, separated change is required.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -357,6 +357,24 @@
         <artifactId>hadoop-mapreduce-client-core</artifactId>
         <version>${hadoop.version}</version>
         <scope>test</scope>
+        <exclusions>
+          <exclusion>
+            <groupId>jakarta.xml.bind</groupId>
+            <artifactId>jakarta.xml.bind-api</artifactId>
+          </exclusion>
+          <exclusion>
+            <groupId>javax.xml.bind</groupId>
+            <artifactId>jaxb-api</artifactId>
+          </exclusion>
+          <exclusion>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-reload4j</artifactId>
+          </exclusion>
+          <exclusion>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-simple</artifactId>
+          </exclusion>
+        </exclusions>
       </dependency>
       <dependency>
         <groupId>org.mockito</groupId>
@@ -766,8 +784,8 @@
               <ignoreNonCompile>true</ignoreNonCompile>
               <ignoredDependencies>
                 <!-- We defined these as direct dependencies (as opposed to just declaring it in dependencyManagement)
-                                                                                                                                                                                                                to workaround https://issues.apache.org/jira/browse/MNG-7982. Now the dependency analyzer complains that
-                                                                                                                                                                                                                the dependency is unused, so we ignore it here-->
+                                                                                                                                                                                                                                to workaround https://issues.apache.org/jira/browse/MNG-7982. Now the dependency analyzer complains that
+                                                                                                                                                                                                                                the dependency is unused, so we ignore it here-->
                 <ignoredDependency>org.apache.commons:commons-compress</ignoredDependency>
                 <ignoredDependency>org.apache.commons:commons-configuration2</ignoredDependency>
               </ignoredDependencies>
@@ -862,9 +880,9 @@
         <configuration>
           <errorRemedy>failFast</errorRemedy>
           <!--
-                                                                                                                                                                           The list of allowed licenses. If you see the build failing due to "There are some forbidden licenses used, please
-                                                                                                                                                                          check your dependencies", verify the conditions of the license and add the reference to it here.
-                                                                                                                                                                        -->
+                                                                                                                                                                                     The list of allowed licenses. If you see the build failing due to "There are some forbidden licenses used, please
+                                                                                                                                                                                    check your dependencies", verify the conditions of the license and add the reference to it here.
+                                                                                                                                                                                  -->
           <includedLicenses>
             <includedLicense>Apache License 2.0</includedLicense>
             <includedLicense>BSD 2-Clause License</includedLicense>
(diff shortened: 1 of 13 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -212,6 +212,10 @@ private SerializationResult serializeFromJavaObjects(
     }
 
     Map<String, String> metadata = channelsDataPerTable.get(0).getVectors().metadata;
+    // We insert the filename in the file itself as metadata so that streams can work on replicated
+    // mixed tables. For a more detailed discussion on the topic see SNOW-561447 and
+    // http://go/streams-on-replicated-mixed-tables
+    metadata.put(Constants.PRIMARY_FILE_ID_KEY, StreamingIngestUtils.getShortname(filePath));
```
sfc-gh-tzhang, line 218: This is the major change, i move the place where we put the file name

src/main/java/net/snowflake/ingest/streaming/internal/ParquetChunkData.java
```diff
@@ -35,16 +34,6 @@ public ParquetChunkData(
     this.rows = rows;
     this.parquetWriter = parquetWriter;
     this.output = output;
-    // create a defensive copy of the parameter map because the argument map passed here
-    // may currently be shared across multiple threads.
-    this.metadata = createDefensiveCopy(metadata);
```
sfc-gh-hmadan, line 40: is this fix not needed any more, under high concurrent load?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -263,11 +262,6 @@ boolean hasColumns() {
 
   @Override
   Optional<ParquetChunkData> getSnapshot(final String filePath) {
-    // We insert the filename in the file itself as metadata so that streams can work on replicated
-    // mixed tables. For a more detailed discussion on the topic see SNOW-561447 and
-    // http://go/streams-on-replicated-mixed-tables
-    metadata.put(Constants.PRIMARY_FILE_ID_KEY, StreamingIngestUtils.getShortname(filePath));
```
sfc-gh-hmadan, line 269: since we're removing this, lets also remove filePath from the method signature on the whole callstack starting in FlushService.distributeFlushTasks? AFAIK this was the only reason we were generating the blobPath in the code location pointed out in JIRA.

src/test/java/net/snowflake/ingest/streaming/internal/RowBufferTest.java
```diff
@@ -1730,78 +1719,4 @@ public void testOnErrorAbortRowsWithError() {
     Assert.assertEquals(1, snapshotAbortParquet.size());
     Assert.assertEquals(Arrays.asList("a"), snapshotAbortParquet.get(0));
   }
-
-  @Test
-  public void testParquetChunkMetadataCreationIsThreadSafe() throws InterruptedException {
```
sfc-gh-hmadan, line 1735: lets keep this test around if its testing for thread safety?

sfc-gh-hmadan:
wondering if it is possible to defer filename insertion (into metadata) and do it inside InternalStage.put() ?

I'm working on a very related area - uploading to presigned urls for iceberg - and was thinking through cases where the URL has expired and we want to do a retry with a new URL. With the way things are setup right now, if the presigned token has expired I'll end up redoing the buildAndUpload call in the task that FlushService creates with a new URL each time, which will be more wasteful than just reserializing the metadata chunk when a new filename needs to be used.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetChunkData.java
sfc-gh-tzhang: technically no, as all the other fields in this map will stay the same across all files (for now). But I reverted this back to keep the logic less error-prone in the future.

sfc-gh-tzhang:
> wondering if it is possible to defer filename insertion (into metadata) and do it inside InternalStage.put() ?
> 
> I'm working on a very related area - uploading to presigned urls for iceberg - and was thinking through cases where the URL has expired and we want to do a retry with a new URL. With the way things are setup right now, if the presigned token has expired I'll end up redoing the buildAndUpload call in the task that FlushService creates with a new URL each time, which will be more wasteful than just reserializing the metadata chunk when a new filename needs to be used.

It's possible, but let's discuss this outside of this PR? We want to get this in ASAP and create a new release on top.

sfc-gh-gdoci (approved):
Thanks, fix lgtm. A test to repro the original issue would be good to have.

Re serializeFromParquetWriteBuffers: Good that you bring it up, I'll remove it because we don't plan to enable it.

sfc-gh-azagrebin (commented):
Maybe I am missing something but I think the test is missing for this change. Otherwise LGTM.

It could be at least write/read test in SDK because there is no way to test it before release with our server side tests.
I could be something like this: 
create test buffer -> create flusher from it -> flush to byte[] -> create BdecParquetReader from bytes -> get metadata: `fileReader.getFileMetaData().getKeyValueMetaData()` (`fileReader` needs to become reader's field)

sfc-gh-hmadan (approved):
code change looks good, lets add a test before checking it in!

src/test/java/net/snowflake/ingest/streaming/internal/RowBufferTest.java
sfc-gh-tzhang: added back and compare the column data types instead of file names

sfc-gh-tzhang:
> Maybe I am missing something but I think the test is missing for this change. Otherwise LGTM.
> 
> It could be at least write/read test in SDK because there is no way to test it before release with our server side tests. I could be something like this: create test buffer -> create flusher from it -> flush to byte[] -> create BdecParquetReader from bytes -> get metadata: `fileReader.getFileMetaData().getKeyValueMetaData()` (`ParquetFileReader.fileReader` needs to become a field to get metadata from it)

Thanks Andrey! Took me some time but I'm able to add `testParquetFileNameMetadata` based on your suggestion, PTAL

pom.xml
```diff
@@ -352,6 +352,12 @@
         <version>1.14.9</version>
         <scope>test</scope>
       </dependency>
+      <dependency>
+        <groupId>org.apache.hadoop</groupId>
+        <artifactId>hadoop-mapreduce-client-core</artifactId>
+        <version>${hadoop.version}</version>
+        <scope>test</scope>
+      </dependency>
```
sfc-gh-tzhang, line 360: This is needed for ParquetFileReader to work

PR #825, opened 2024-09-11
Add chunk offset to file id key to make each chunk have a unique key

sfc-gh-psaha:
Also add sdk version to Parquet metadata.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
@@ -10,8 +10,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
-
-import com.google.common.annotations.VisibleForTesting;
 import net.snowflake.ingest.utils.Constants;
 import net.snowflake.ingest.utils.ErrorCode;
 import net.snowflake.ingest.utils.Logging;
@@ -143,7 +141,8 @@ public class ParquetFlusher implements Flusher<ParquetChunkData> {
         chunkMinMaxInsertTimeInMs);
   }
 
-  private static void addFileIdToMetadata(String filePath, long chunkStartOffset, Map<String, String> metadata) {
+  private static void addFileIdToMetadata(
+      String filePath, long chunkStartOffset, Map<String, String> metadata) {
     // We insert the filename in the file itself as metadata so that streams can work on replicated
     // mixed tables. For a more detailed discussion on the topic see SNOW-561447 and
     // http://go/streams-on-replicated-mixed-tables
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
@@ -41,13 +41,13 @@ public class ParquetFlusher implements Flusher<ParquetChunkData> {
 
   @Override
   public SerializationResult serialize(
-      List<ChannelData<ParquetChunkData>> channelsDataPerTable, String filePath)
+          List<ChannelData<ParquetChunkData>> channelsDataPerTable, String filePath, long chunkStartOffset)
       throws IOException {
-    return serializeFromJavaObjects(channelsDataPerTable, filePath);
+    return serializeFromJavaObjects(channelsDataPerTable, filePath, chunkStartOffset);
   }
 
   private SerializationResult serializeFromJavaObjects(
-      List<ChannelData<ParquetChunkData>> channelsDataPerTable, String filePath)
+      List<ChannelData<ParquetChunkData>> channelsDataPerTable, String filePath, long chunkStartOffset)
       throws IOException {
     List<ChannelMetadata> channelsMetadataList = new ArrayList<>();
     long rowCount = 0L;
@@ -116,7 +116,14 @@ public class ParquetFlusher implements Flusher<ParquetChunkData> {
     // We insert the filename in the file itself as metadata so that streams can work on replicated
     // mixed tables. For a more detailed discussion on the topic see SNOW-561447 and
     // http://go/streams-on-replicated-mixed-tables
-    metadata.put(Constants.PRIMARY_FILE_ID_KEY, StreamingIngestUtils.getShortname(filePath));
+    // Using chunk offset as suffix ensures that for interleaved tables, the file
+    // id key is unique for each chunk. Each chunk is logically a separate Parquet file that happens
+    // to be bundled together.
+    String shortName = StreamingIngestUtils.getShortname(filePath);
+    final String[] parts = shortName.split("\\.");
+    metadata.put(
+        Constants.PRIMARY_FILE_ID_KEY,
+        String.format("%s%s.%s", parts[0], chunkStartOffset, parts[1]));
     parquetWriter =
         new BdecParquetWriter(
             mergedData,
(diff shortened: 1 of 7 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -48,12 +48,12 @@ public ParquetFlusher(
 
   @Override
   public SerializationResult serialize(
-      List<ChannelData<ParquetChunkData>> channelsDataPerTable, String filePath)
+      List<ChannelData<ParquetChunkData>> channelsDataPerTable, String filePath, int chunkIndex)
       throws IOException {
     if (enableParquetInternalBuffering) {
       return serializeFromParquetWriteBuffers(channelsDataPerTable, filePath);
```
sfc-gh-psaha, line 54: Why don't we update the `PRIMARY_FILE_ID_KEY` in the `serializeFromParquetWriteBuffers` path?

sfc-gh-psaha: Nvm, I see that Toby has already commented that this is unused.

sfc-gh-gdoci: Right, and here is the PR https://github.com/snowflakedb/snowflake-ingest-java/pull/829 to clean that up.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -116,7 +120,14 @@ private SerializationResult serializeFromJavaObjects(
...
     // http://go/streams-on-replicated-mixed-tables
-    metadata.put(Constants.PRIMARY_FILE_ID_KEY, StreamingIngestUtils.getShortname(filePath));
+    // Using chunk offset as suffix ensures that for interleaved tables, the file
+    // id key is unique for each chunk. Each chunk is logically a separate Parquet file that happens
+    // to be bundled together.
+    String shortName = StreamingIngestUtils.getShortname(filePath);
+    final String[] parts = shortName.split("\\.");
+    metadata.put(
+        Constants.PRIMARY_FILE_ID_KEY,
+        String.format("%s%s.%s", parts[0], chunkStartOffset, parts[1]));
```
sfc-gh-tzhang, line 130: Is there any issue if the does not match with the file name? Is there any server side change this needs to wait before merging? cc: @sfc-gh-kkloudas

sfc-gh-tzhang: And do you think we can do %s_%s.%s?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -74,6 +76,7 @@ public void setupSchema(List<ColumnMetadata> columns) {
     fieldIndex.clear();
     metadata.clear();
     metadata.put("sfVer", "1,1");
+    metadata.put(Constants.SDK_VERSION_KEY, RequestBuilder.DEFAULT_VERSION);
```
sfc-gh-tzhang, line 79: Thanks for adding the SDK version

sfc-gh-tzhang (approved):
Change looks good in SDK, please sync with @sfc-gh-kkloudas to make sure this change won't break replication.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
sfc-gh-psaha: I am trying to match with what Andrey will do on the server side - we don't have to match but would be good to be consistent. Let's have @sfc-gh-azagrebin confirm what he'd do. We can wait for the server side change. And no, it does not need to match the file name at all - it can be a UUID for all change tracking cares about.

sfc-gh-tzhang: And if we change the format, is there a use case that customer needs to join with previous values as part of MERGE?

sfc-gh-psaha: No, this value is supposed to be unique that does not appear in any other files.

sfc-gh-azagrebin: I would also vote for this:
```
And do you think we can do %s_%s.%s?
```
let me also first try the server side change

sfc-gh-azagrebin: just in case to reduce any potential blast radius, I would append only non-zero offset.

sfc-gh-azagrebin: We just discussed it with @sfc-gh-lpapke. We have to add "_" because the name w/o extension ends with a number in our case and we can get into a similar problem that was mentioned for regular partitions:
name_10.bdec  + offset 110 = name_10110.bdec
name_101.bdec  + offset 10 = name_10110.bdec
those 2 are different bdecs but they will be the same PPNs.

sfc-gh-psaha: Good catch. I added an `_` separator before the `.bdec` suffix. PTAL. I have also added a check for 0 chunk offset.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -116,7 +121,19 @@ private SerializationResult serializeFromJavaObjects(
...
+    if (chunkStartOffset == 0) {
+      metadata.put(Constants.PRIMARY_FILE_ID_KEY, StreamingIngestUtils.getShortname(filePath));
+    } else {
+      String shortName = StreamingIngestUtils.getShortname(filePath);
+      final String[] parts = shortName.split("\\.");
+      Preconditions.checkState(parts.length == 2, "Invalid file name format");
+      metadata.put(
+          Constants.PRIMARY_FILE_ID_KEY,
+          String.format("%s_%d.%s", parts[0], chunkStartOffset, parts[1]));
+    }
```
sfc-gh-azagrebin, line 136: nit: I would suggest to factor this out to a separate function
we have this pattern of exploding functions pretty quickly that makes them hard to understand later
especially for new folks

src/test/java/net/snowflake/ingest/streaming/internal/RowBufferTest.java
```diff
@@ -1789,10 +1790,15 @@ public void testParquetFileNameMetadata() throws IOException {
 
     ParquetFlusher flusher = (ParquetFlusher) bufferUnderTest.createFlusher();
     Flusher.SerializationResult result =
-        flusher.serialize(Collections.singletonList(data), filePath);
+        flusher.serialize(Collections.singletonList(data), filePath, 13);
```
sfc-gh-azagrebin, line 1793: nit: zero offset case test

sfc-gh-azagrebin (approved):
Thanks Purujit!

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
sfc-gh-psaha: Done.

src/test/java/net/snowflake/ingest/streaming/internal/RowBufferTest.java
sfc-gh-psaha: Done.

PR #827, opened 2024-09-12
V2.2.2 release with behavior change reverted

sfc-gh-tzhang:


sfc-gh-hmadan:
Do we need a more "param protection" style development pattern in the SDK? So that we can tell customers how to disable a bad update by turning off an internal undocumented parameter after their upgrade?

Reverting is yet another behavior change for someone that starts using the new behavior, so am not sure if we can have code reverts as our primary strategy for behavior change management in the SDK.

sfc-gh-xhuang (approved):

sfc-gh-tzhang:
> Do we need a more "param protection" style development pattern in the SDK? So that we can tell customers how to disable a bad update by turning off an internal undocumented parameter after their upgrade?

I think this makes sense in general, but we haven't have time to do it. And my concern of introducing parameters is the additional work of maintaining them, parameter drift, etc.

sfc-gh-tzhang:
> Reverting is yet another behavior change for someone that starts using the new behavior, so am not sure if we can have code reverts as our primary strategy for behavior change management in the SDK.

This behavior change is not in any of the released SDK versions, we just want to release a patch version with the fix of the corruption you reported (again, thanks for catching it!) exclude this behavior change.

PR #828, opened 2024-09-12
Iceberg Ingestion in CloudStorage mode - Carve out IStorage and ExternalVolume from InternalStage

sfc-gh-hmadan:
This is the first in a sequence of PRs to get the client SDK to work with presigned urls for iceberg tables.

At a high level, it sets things up to have two stacks of storage objects - InternalStageManager and InternalStage (for FDN tables), and ExternalVolumeManager and ExternalVolume (for iceberg tables). Technically the latter can be used for any table that wants to use the presigned-url style of uploading files, but right now we're focused on iceberg.

There is a new interface, IStorage, that both ExternalVolume and InternalStage will implement.

Other changes:
1. Introduce a BlobPath class to carry around the blob path / path with token. We want to be careful to not write the presigned url out to logs etc, or put it into the file metadata, thus we need this structure instead of a simple string
2. minor changes in FlushService ot generate filename only when blobData is empty, this removes the need for decrementBlobSequencer.
3. cleanup the IStorageManager interface to only have whats necessary
4. introduce a TableRef structure instead of passing around fullyQualifiedTableName as a concatenated string and keying off that string in maps.
5. introduce a hidden method (not public) on the ClientFactory that allows setting isIceberg to true when doing local development.


src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -415,11 +411,10 @@ void distributeFlushTasks(Set<String> tablesToFlush) {
             >= this.owningClient.getParameterProvider().getMaxChunksInBlob()) {
           // Create a new blob if the current one already contains max allowed number of chunks
           logger.logInfo(
-              "Max allowed number of chunks in the current blob reached. chunkCount={}"
-                  + " maxChunkCount={} currentBlobPath={}",
+              "Max allowed number of chunks in the current blob reached. chunkCount={} "
```
sfc-gh-alhuang, line 414: Additional space after chunk `chunkCount={}` should be removed?

sfc-gh-alhuang (approved):
Looks good, thanks!

PR #832, opened 2024-09-19
SNOW-1673203 EP info generation for new table format

sfc-gh-alhuang:
This PR contains logic of ep info generation logic as the Iceberg tables requires different EP info than FDN tables. For Iceberg tables, the min/max value come from parquet footer. The NDV and maxLength of string/binary columns are computed by SDK in `bdecParquetWriter`.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/EpInfo.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/EpInfo.java
@@ -1,7 +1,5 @@
 package net.snowflake.ingest.streaming.internal;
 
-import static net.snowflake.ingest.utils.Constants.EP_NDV_UNKNOWN;
-
 import com.fasterxml.jackson.annotation.JsonProperty;
 import java.util.Map;
 import net.snowflake.ingest.utils.ErrorCode;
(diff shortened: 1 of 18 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/EpInfo.java
```diff
@@ -34,14 +32,6 @@ public void verifyEpInfo() {
                 "Null count bigger than total row count on col=%s, nullCount=%s, rowCount=%s",
                 colName, colEp.getNullCount(), rowCount));
       }
```
sfc-gh-hmadan, line 34: undo, lets go back to -1

sfc-gh-hmadan:
This change is fairly non trivial and I'm sceptical on the need for such a change. It's going to worsen CPU usage (due to the repeated iteration over all rows to extract stats later on), worse memory pressure (due to keeping the NDV hashset for each column for each blob), and introduces a lot of unintuitive code paths wherein for the same class (SerializationResult) some fields are populated in FDN mode and others in iceberg mode.

It looks like most of this PR is for NDV and MaxLength calculation, neither of which are P0 to get to GA.

Its unclear why all this bookkeeping can't be done in the RowBufferStats structure, as that'll help with FDN NDV calculations too in future.

Also unclear whether this PR has changes that are needed for structured data types independent of NDV and MaxLength calculations.

Lets chat tomorrow am on what forces our hand to require such a different approach to stats capture and merging.

PR #836, opened 2024-09-20
SNOW-1673203 EP info generation for new table format

sfc-gh-alhuang:
This PR contains logic of ep info generation logic as the Iceberg tables requires different EP info than FDN tables.


sfc-gh-hmadan (approved):
Added a commit in lieu of adding Code review comments :)
Changes lgtm, thanks!

PR #837, opened 2024-09-20
SNOW-1675591 Fill in ExternalVolume and ExternalVolumeManager to do presigned url retrieval + blobname population 

sfc-gh-hmadan:
1. Add a GeneratePresignedUrls API call (new contracts + snowflakeClient API)
2. Enhance ExternalVolume.java to do presigned url file uploads to S3 (GCP and Azure testing pending thus commented out)
3. fix json field names where applicable
4. Add @JsonIgnoreProperties(ignoreUnknown = true) for forward compatiblity with service side evolution to some contracts that were missing this tag


src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolume.java
```diff
@@ -1,9 +1,409 @@
...
+      } catch (InterruptedException e) {
+        // if the thread was interrupted there's nothing we can do about it, definitely shouldn't
+        // continue processing.
+        String message =
+            String.format(
+                "Semaphore acquisition in ExternalVolume.generateUrls was interrupted, "
+                    + "likely because the process is shutting down. TableRef=%s",
+                tableRef);
+        logger.logError(message);
+        throw new SFException(ErrorCode.INTERNAL_ERROR, message);
```
sfc-gh-psaha, line 330: I think we should restore the interrupted flag here in case some higher level adds a retry.

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolume.java
```diff
@@ -1,9 +1,409 @@
...
+    } else {
+      // Since the queue had a non-null entry, there is no way numUrlsInQueue is <=0
+      int remainingUrlsInQueue = this.numUrlsInQueue.decrementAndGet();
+      if (remainingUrlsInQueue <= LOW_WATERMARK_FOR_EARLY_REFRESH) {
+        generate = true;
+        // assert remaininUrlsInQueue >= 0
+      }
+    }
+    if (generate) {
+      // TODO: do this generation on a background thread to allow the current thread to make
```
sfc-gh-psaha, line 295: I feel like the background thread approach would also be simpler - you won't need all the Semaphore business. The bg thread could just keep refreshing in a loop whenever it sees the remaining urls hit a low.

sfc-gh-psaha (commented):
Looks mostly good. Just some minor comments.

src/main/java/net/snowflake/ingest/streaming/internal/ChunkMetadata.java
```diff
@@ -219,13 +219,13 @@ Long getLastInsertTimeInMs() {
   // Snowflake service had a bug that did not allow the client to add new json fields in some
   // contracts; thus these new fields have a NON_DEFAULT attribute.
   @JsonProperty("major_vers")
-  @JsonInclude(JsonInclude.Include.NON_DEFAULT)
```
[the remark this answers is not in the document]
sfc-gh-hmadan: +1, had updated the comment locally was waiting to push changes after one round of review!
here is what i have in comments;
```
  // Snowflake service had a bug that did not allow the client to add new json fields in some
  // contracts; thus these new fields have a NON_NULL attribute. NON_DEFAULT will ignore an explicit
  // zero value, thus NON_NULL is a better fit.
```

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolume.java
```diff
@@ -1,9 +1,409 @@
...
+    SnowflakeStorageClient client =
+        StorageClientFactory.getFactory().createClient(stageInfo, 1, null, null);
+
+    URIBuilder uriBuilder = new URIBuilder(blobPath);
+    HttpPut httpRequest = new HttpPut(uriBuilder.build());
+    httpRequest.setEntity(new ByteArrayEntity(blob));
+
+    addHeadersToHttpRequest(httpRequest, blob, stageInfo, client);
+
+    if (stageInfo.getStageType().equals(StageInfo.StageType.AZURE)) {
```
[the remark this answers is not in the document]
sfc-gh-hmadan: Don't have good local testing for GCP and Azure per the iceberg testing wikis, waiting for changes to hit preprod before i can do gcp and azure testing :|

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolume.java
```diff
@@ -1,9 +1,409 @@
...
+              // AtomicBoolean to avoid retries/sleeps
+              false, // no cookie
+              false, // no url retry query parameters
+              false, // no request_guid
+              true, // retry on HTTP 403
+              true, // no retry
+              new ExecTimeTelemetryData());
+    } catch (Throwable e) {
+      throw e;
+    }
```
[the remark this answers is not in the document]
sfc-gh-hmadan: agreed, removed the catch and throw. have a backlog item to handle transient upload failures and do proper retries.

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolume.java
sfc-gh-hmadan: good catch! fixed.

src/main/java/net/snowflake/ingest/streaming/internal/ExternalVolume.java
sfc-gh-hmadan: Yes, couple reasons I kept this as simple as possible:

1. There can be many many external volumes (one per table), and the right fix there would be to have a small threadpool that multiplexes over all the active external volumes, and potentially even does a batched retrieval instead of one RPC per ExtVol. This is doable, just a matter of work ordering

2. rate matching - some ext vols will need 1 URL every minute, other volumes will need 10 URLs every second; to do a proper fix the low watermark needs to be adaptive instead of the hardcoded 5 right now.

3. Multipart upload support - to do multipart upload, each part requires its' own presigned URL AND needs to be associated with an upload-id; however now we can't cache URLs in advance as we won't know (in advance) how many parts any given file will have and don't have a batch of upload-ids either to quickly choose from.

Of these three, ended up just having a roundabout fix for (2) by allowing concurrent URL retrieval, as soon as there's high pressure on URL retrieval we'll end up caching a lot of URLs and reduce pressure.

sfc-gh-psaha (approved):

PR #839, opened 2024-09-23
Get test coverage for telemetry reporting (was missing) + Fix SnowflakeServiceClient test code to allow per-API overrides 

sfc-gh-hmadan:
As part of bringing in ExternalVolume (presigned url based uploads), I decided to not add test-only code into ExtVol / ExtVolManager. This meant the following:
1. All the httpClient mock logic to return hardcoded responses now needs to handle multiple APIs (client configure AND getPresignedUrls) - this looked very similar to MockSnowflakeServiceClient logic but duplicated 10 times.
2. To fix this, introduced a per-API behavior override mechanism in MockSNowflakeServiceClient; helped me remove hundreds of lines of code that was just setting up the httpclient mock, and reduce how brittle that setup is
3. To do away with isTestMode in InternalStage / InternalStageManager, also did the following two changes:
    1. Make the snowflake service client respond to telemetry upload calls correctly when isTestMode is false
    1. Initialize the telemetry client even when isTestMode is false

Tested all affected unit test classes locally


sfc-gh-alhuang (approved):
Looks good, thanks!

PR #841, opened 2024-09-24
SNOW-1666189 Structured data type support

sfc-gh-alhuang:
Support structured data type for Iceberg table including struct, map, and list.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
@@ -1055,6 +1055,80 @@ class DataValidationUtil {
         insertRowIndex);
   }
 
+  /**
+   * Validate and cast Iceberg struct column to Map<String, Object>. Allowed Java type:
+   *
+   * <ul>
+   *   <li>Map<String, Object>
+   * </ul>
+   *
+   * @param columnName Column name, used in validation error messages
+   * @param input Object to validate and parse
+   * @param insertRowIndex Row index for error reporting
+   * @return Object cast to Map
+   */
+  static Map<String, Object> validateAndParseIcebergStruct(
+      String columnName, Object input, long insertRowIndex) {
+    if (!(input instanceof Map)) {
+      throw typeNotAllowedException(
+          columnName,
+          input.getClass(),
+          "STRUCT",
+          new String[] {"Map<String, Object>"},
+          insertRowIndex);
+    }
+    if (!((Map<?, ?>) input).keySet().stream().allMatch(key -> key instanceof String)) {
+      throw new SFException(
+          ErrorCode.INVALID_FORMAT_ROW,
+          String.format(
+              "Flied name of a struct must be of type String, rowIndex:%d", insertRowIndex));
+    }
+
+    return (Map<String, Object>) input;
+  }
+
+  /**
+   * Validate and parse Iceberg list column to an Iterable. Allowed Java type:
+   *
+   * <ul>
+   *   <li>Iterable
+   * </ul>
+   *
+   * @param columnName Column name, used in validation error messages
+   * @param input Object to validate and parse
+   * @param insertRowIndex Row index for error reporting
+   * @return Object cast to Iterable
+   */
+  static Iterable<Object> validateAndParseIcebergList(
+      String columnName, Object input, long insertRowIndex) {
+    if (!(input instanceof Iterable)) {
+      throw typeNotAllowedException(
+          columnName, input.getClass(), "LIST", new String[] {"Iterable"}, insertRowIndex);
+    }
+    return (Iterable<Object>) input;
+  }
+
+  /**
+   * Validate and parse Iceberg map column to a map. Allowed Java type:
+   *
+   * <ul>
+   *   <li>Map<Object, Object>
+   * </ul>
+   *
+   * @param columnName Column name, used in validation error messages
+   * @param input Object to validate and parse
+   * @param insertRowIndex Row index for error reporting
+   * @return Object cast to Map
+   */
+  static Map<Object, Object> validateAndParseIcebergMap(
+      String columnName, Object input, long insertRowIndex) {
+    if (!(input instanceof Map)) {
+      throw typeNotAllowedException(
+          columnName, input.getClass(), "MAP", new String[] {"Map"}, insertRowIndex);
+    }
+    return (Map<Object, Object>) input;
+  }
+
   static void checkValueInRange(
       BigDecimal bigDecimalValue, int scale, int precision, final long insertRowIndex) {
     BigDecimal comparand =
--- a/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
@@ -1067,7 +1067,7 @@ class DataValidationUtil {
    * @param insertRowIndex Row index for error reporting
    * @return Object cast to Map
    */
-  static Map<String, Object> validateAndParseIcebergStruct(
+  static Map<String, ?> validateAndParseIcebergStruct(
       String columnName, Object input, long insertRowIndex) {
     if (!(input instanceof Map)) {
       throw typeNotAllowedException(
@@ -1077,14 +1077,16 @@ class DataValidationUtil {
           new String[] {"Map<String, Object>"},
           insertRowIndex);
     }
-    if (!((Map<?, ?>) input).keySet().stream().allMatch(key -> key instanceof String)) {
-      throw new SFException(
-          ErrorCode.INVALID_FORMAT_ROW,
-          String.format(
-              "Flied name of a struct must be of type String, rowIndex:%d", insertRowIndex));
+    for (Object key : ((Map<?, ?>) input).keySet()) {
+      if (!(key instanceof String)) {
+        throw new SFException(
+            ErrorCode.INVALID_FORMAT_ROW,
+            String.format(
+                "Field name of a struct must be of type String, rowIndex:%d", insertRowIndex));
+      }
     }
 
-    return (Map<String, Object>) input;
+    return (Map<String, ?>) input;
   }
 
   /**
@@ -1099,13 +1101,13 @@ class DataValidationUtil {
    * @param insertRowIndex Row index for error reporting
    * @return Object cast to Iterable
    */
-  static Iterable<Object> validateAndParseIcebergList(
+  static Iterable<?> validateAndParseIcebergList(
       String columnName, Object input, long insertRowIndex) {
     if (!(input instanceof Iterable)) {
       throw typeNotAllowedException(
           columnName, input.getClass(), "LIST", new String[] {"Iterable"}, insertRowIndex);
     }
-    return (Iterable<Object>) input;
+    return (Iterable<?>) input;
   }
 
   /**
@@ -1120,13 +1122,13 @@ class DataValidationUtil {
    * @param insertRowIndex Row index for error reporting
    * @return Object cast to Map
    */
-  static Map<Object, Object> validateAndParseIcebergMap(
+  static Map<?, ?> validateAndParseIcebergMap(
       String columnName, Object input, long insertRowIndex) {
     if (!(input instanceof Map)) {
       throw typeNotAllowedException(
           columnName, input.getClass(), "MAP", new String[] {"Map"}, insertRowIndex);
     }
-    return (Map<Object, Object>) input;
+    return (Map<?, ?>) input;
   }
 
   static void checkValueInRange(
--- a/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
@@ -1082,7 +1082,8 @@ class DataValidationUtil {
         throw new SFException(
             ErrorCode.INVALID_FORMAT_ROW,
             String.format(
-                "Field name of a struct must be of type String, rowIndex:%d", insertRowIndex));
+                "Field name of struct %s must be of type String, rowIndex:%d",
+                columnName, insertRowIndex));
       }
     }
 
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
@@ -5,19 +5,11 @@
 package net.snowflake.ingest.streaming.internal;
 
 import static net.snowflake.ingest.streaming.internal.BinaryStringUtils.truncateBytesAsHex;
-import static net.snowflake.ingest.utils.Constants.EP_NDV_UNKNOWN;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import java.math.BigInteger;
 import java.util.Objects;
-import org.apache.parquet.column.statistics.BinaryStatistics;
-import org.apache.parquet.column.statistics.BooleanStatistics;
-import org.apache.parquet.column.statistics.DoubleStatistics;
-import org.apache.parquet.column.statistics.FloatStatistics;
-import org.apache.parquet.column.statistics.IntStatistics;
-import org.apache.parquet.column.statistics.LongStatistics;
-import org.apache.parquet.column.statistics.Statistics;
-import org.apache.parquet.schema.LogicalTypeAnnotation;
 
 /** Audit register endpoint/FileColumnPropertyDTO property list. */
 class FileColumnProperties {
@@ -60,6 +52,7 @@ class FileColumnProperties {
 
   FileColumnProperties(RowBufferStats stats, boolean setDefaultValues) {
     this.setColumnOrdinal(stats.getOrdinal());
+    this.setFieldId(stats.getFieldId());
     this.setCollation(stats.getCollationDefinitionString());
     this.setMaxIntValue(
         stats.getCurrentMaxIntValue() == null
@@ -98,38 +91,6 @@ class FileColumnProperties {
     this.setDistinctValues(stats.getDistinctValues());
   }
 
-  FileColumnProperties(int fieldId, Statistics<?> statistics) {
-    this.setColumnOrdinal(fieldId);
-    this.setFieldId(fieldId);
-    this.setNullCount(statistics.getNumNulls());
-    this.setDistinctValues(EP_NDV_UNKNOWN);
-    this.setCollation(null);
-    this.setMaxStrNonCollated(null);
-    this.setMinStrNonCollated(null);
-
-    if (statistics instanceof BooleanStatistics) {
-      this.setMinIntValue(
-          ((BooleanStatistics) statistics).genericGetMin() ? BigInteger.ONE : BigInteger.ZERO);
-      this.setMaxIntValue(
-          ((BooleanStatistics) statistics).genericGetMax() ? BigInteger.ONE : BigInteger.ZERO);
-    } else if (statistics instanceof IntStatistics || statistics instanceof LongStatistics) {
-      this.setMinIntValue(BigInteger.valueOf(((Number) statistics.genericGetMin()).longValue()));
-      this.setMaxIntValue(BigInteger.valueOf(((Number) statistics.genericGetMax()).longValue()));
-    } else if (statistics instanceof FloatStatistics || statistics instanceof DoubleStatistics) {
-      this.setMinRealValue((Double) statistics.genericGetMin());
-      this.setMaxRealValue((Double) statistics.genericGetMax());
-    } else if (statistics instanceof BinaryStatistics) {
-      if (statistics.type().getLogicalTypeAnnotation()
-          instanceof LogicalTypeAnnotation.DecimalLogicalTypeAnnotation) {
-        this.setMinIntValue(new BigInteger(statistics.getMinBytes()));
-        this.setMaxIntValue(new BigInteger(statistics.getMaxBytes()));
-      } else {
-        this.setMinStrValue(truncateBytesAsHex(statistics.getMinBytes(), false));
-        this.setMaxStrValue(truncateBytesAsHex(statistics.getMaxBytes(), true));
-      }
-    }
-  }
-
   @JsonProperty("columnId")
   public int getColumnOrdinal() {
     return columnOrdinal;
@@ -140,6 +101,7 @@ class FileColumnProperties {
   }
 
   @JsonProperty("fieldId")
+  @JsonInclude(JsonInclude.Include.NON_DEFAULT)
   public int getFieldId() {
     return fieldId;
   }
@@ -261,6 +223,7 @@ class FileColumnProperties {
   public String toString() {
     final StringBuilder sb = new StringBuilder("{");
     sb.append("\"columnOrdinal\": ").append(columnOrdinal);
+    sb.append(", \"fieldId\": ").append(fieldId);
     if (minIntValue != null) {
       sb.append(", \"minIntValue\": ").append(minIntValue);
       sb.append(", \"maxIntValue\": ").append(maxIntValue);
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
@@ -1,14 +1,28 @@
+/*
+ * Copyright (c) 2024 Snowflake Computing Inc. All rights reserved.
+ */
+
 package net.snowflake.ingest.streaming.internal;
 
 import static net.snowflake.ingest.streaming.internal.BinaryStringUtils.truncateBytesAsHex;
+import static net.snowflake.ingest.utils.Constants.EP_NDV_UNKNOWN;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import java.math.BigInteger;
 import java.util.Objects;
+import org.apache.parquet.column.statistics.BinaryStatistics;
+import org.apache.parquet.column.statistics.BooleanStatistics;
+import org.apache.parquet.column.statistics.DoubleStatistics;
+import org.apache.parquet.column.statistics.FloatStatistics;
+import org.apache.parquet.column.statistics.IntStatistics;
+import org.apache.parquet.column.statistics.LongStatistics;
+import org.apache.parquet.column.statistics.Statistics;
+import org.apache.parquet.schema.LogicalTypeAnnotation;
 
 /** Audit register endpoint/FileColumnPropertyDTO property list. */
 class FileColumnProperties {
   private int columnOrdinal;
+  private int fieldId;
   private String minStrValue;
 
   private String maxStrValue;
@@ -84,6 +98,38 @@ class FileColumnProperties {
     this.setDistinctValues(stats.getDistinctValues());
   }
 
+  FileColumnProperties(int fieldId, Statistics<?> statistics) {
+    this.setColumnOrdinal(fieldId);
+    this.setFieldId(fieldId);
+    this.setNullCount(statistics.getNumNulls());
+    this.setDistinctValues(EP_NDV_UNKNOWN);
+    this.setCollation(null);
+    this.setMaxStrNonCollated(null);
+    this.setMinStrNonCollated(null);
+
+    if (statistics instanceof BooleanStatistics) {
+      this.setMinIntValue(
+          ((BooleanStatistics) statistics).genericGetMin() ? BigInteger.ONE : BigInteger.ZERO);
+      this.setMaxIntValue(
+          ((BooleanStatistics) statistics).genericGetMax() ? BigInteger.ONE : BigInteger.ZERO);
+    } else if (statistics instanceof IntStatistics || statistics instanceof LongStatistics) {
+      this.setMinIntValue(BigInteger.valueOf(((Number) statistics.genericGetMin()).longValue()));
+      this.setMaxIntValue(BigInteger.valueOf(((Number) statistics.genericGetMax()).longValue()));
+    } else if (statistics instanceof FloatStatistics || statistics instanceof DoubleStatistics) {
+      this.setMinRealValue((Double) statistics.genericGetMin());
+      this.setMaxRealValue((Double) statistics.genericGetMax());
+    } else if (statistics instanceof BinaryStatistics) {
+      if (statistics.type().getLogicalTypeAnnotation()
+          instanceof LogicalTypeAnnotation.DecimalLogicalTypeAnnotation) {
+        this.setMinIntValue(new BigInteger(statistics.getMinBytes()));
+        this.setMaxIntValue(new BigInteger(statistics.getMaxBytes()));
+      } else {
+        this.setMinStrValue(truncateBytesAsHex(statistics.getMinBytes(), false));
+        this.setMaxStrValue(truncateBytesAsHex(statistics.getMaxBytes(), true));
+      }
+    }
+  }
+
   @JsonProperty("columnId")
   public int getColumnOrdinal() {
     return columnOrdinal;
@@ -91,6 +137,15 @@ class FileColumnProperties {
 
   public void setColumnOrdinal(int columnOrdinal) {
     this.columnOrdinal = columnOrdinal;
+  }
+
+  @JsonProperty("fieldId")
+  public int getFieldId() {
+    return fieldId;
+  }
+
+  public void setFieldId(int fieldId) {
+    this.fieldId = fieldId;
   }
 
   // Annotation required in order to have package private fields serialized
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
@@ -14,7 +14,7 @@ import java.util.Objects;
 /** Audit register endpoint/FileColumnPropertyDTO property list. */
 class FileColumnProperties {
   private int columnOrdinal;
-  private int fieldId;
+  private Integer fieldId;
   private String minStrValue;
 
   private String maxStrValue;
@@ -101,12 +101,12 @@ class FileColumnProperties {
   }
 
   @JsonProperty("fieldId")
-  @JsonInclude(JsonInclude.Include.NON_DEFAULT)
-  public int getFieldId() {
+  @JsonInclude(JsonInclude.Include.NON_NULL)
+  public Integer getFieldId() {
     return fieldId;
   }
 
-  public void setFieldId(int fieldId) {
+  public void setFieldId(Integer fieldId) {
     this.fieldId = fieldId;
   }
 
--- a/src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
@@ -11,12 +11,18 @@ import java.math.BigInteger;
 import java.math.RoundingMode;
 import java.nio.charset.StandardCharsets;
 import java.time.ZoneId;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.atomic.AtomicReference;
 import net.snowflake.ingest.utils.Constants;
 import net.snowflake.ingest.utils.ErrorCode;
 import net.snowflake.ingest.utils.SFException;
 import net.snowflake.ingest.utils.Utils;
+import org.apache.parquet.schema.GroupType;
 import org.apache.parquet.schema.LogicalTypeAnnotation;
 import org.apache.parquet.schema.LogicalTypeAnnotation.DateLogicalTypeAnnotation;
 import org.apache.parquet.schema.LogicalTypeAnnotation.DecimalLogicalTypeAnnotation;
@@ -35,7 +41,7 @@ class IcebergParquetValueParser {
    *
    * @param value column value provided by user in a row
    * @param type Parquet column type
-   * @param stats column stats to update
+   * @param statsMap column stats map to update
    * @param defaultTimezone default timezone to use for timestamp parsing
    * @param insertRowsCurrIndex Row index corresponding the row to parse (w.r.t input rows in
    *     insertRows API, and not buffered row)
@@ -44,69 +50,105 @@ class IcebergParquetValueParser {
   static ParquetBufferValue parseColumnValueToParquet(
       Object value,
       Type type,
-      RowBufferStats stats,
+      Map<String, RowBufferStats> statsMap,
       ZoneId defaultTimezone,
       long insertRowsCurrIndex) {
-    Utils.assertNotNull("Parquet column stats", stats);
+    Utils.assertNotNull("Parquet column stats map", statsMap);
+    return parseColumnValueToParquet(
+        value, type, statsMap, defaultTimezone, insertRowsCurrIndex, null, false);
+  }
+
+  private static ParquetBufferValue parseColumnValueToParquet(
+      Object value,
+      Type type,
+      Map<String, RowBufferStats> statsMap,
+      ZoneId defaultTimezone,
+      long insertRowsCurrIndex,
+      String path,
+      boolean isdDescendantsOfRepeatingGroup) {
+    path = (path == null || path.isEmpty()) ? type.getName() : path + "." + type.getName();
     float estimatedParquetSize = 0F;
     if (value != null) {
-      estimatedParquetSize += ParquetBufferValue.DEFINITION_LEVEL_ENCODING_BYTE_LEN;
-      PrimitiveType primitiveType = type.asPrimitiveType();
-      switch (primitiveType.getPrimitiveTypeName()) {
-        case BOOLEAN:
-          int intValue =
-              DataValidationUtil.validateAndParseBoolean(
-                  type.getName(), value, insertRowsCurrIndex);
-          value = intValue > 0;
-          stats.addIntValue(BigInteger.valueOf(intValue));
-          estimatedParquetSize += ParquetBufferValue.BIT_ENCODING_BYTE_LEN;
-          break;
-        case INT32:
-          int intVal = getInt32Value(value, primitiveType, insertRowsCurrIndex);
-          value = intVal;
-          stats.addIntValue(BigInteger.valueOf(intVal));
-          estimatedParquetSize += 4;
-          break;
-        case INT64:
-          long longVal = getInt64Value(value, primitiveType, defaultTimezone, insertRowsCurrIndex);
-          value = longVal;
-          stats.addIntValue(BigInteger.valueOf(longVal));
-          estimatedParquetSize += 8;
-          break;
-        case FLOAT:
-          float floatVal =
-              (float)
-                  DataValidationUtil.validateAndParseReal(
-                      type.getName(), value, insertRowsCurrIndex);
-          value = floatVal;
-          stats.addRealValue((double) floatVal);
-          estimatedParquetSize += 4;
-          break;
-        case DOUBLE:
-          double doubleVal =
-              DataValidationUtil.validateAndParseReal(type.getName(), value, insertRowsCurrIndex);
-          value = doubleVal;
-          stats.addRealValue(doubleVal);
-          estimatedParquetSize += 8;
-          break;
-        case BINARY:
-          byte[] byteVal = getBinaryValue(value, primitiveType, stats, insertRowsCurrIndex);
-          value = byteVal;
-          estimatedParquetSize +=
-              ParquetBufferValue.BYTE_ARRAY_LENGTH_ENCODING_BYTE_LEN + byteVal.length;
-          break;
-        case FIXED_LEN_BYTE_ARRAY:
-          byte[] fixedLenByteArrayVal =
-              getFixedLenByteArrayValue(value, primitiveType, stats, insertRowsCurrIndex);
-          value = fixedLenByteArrayVal;
-          estimatedParquetSize +=
-              ParquetBufferValue.BYTE_ARRAY_LENGTH_ENCODING_BYTE_LEN + fixedLenByteArrayVal.length;
-          break;
-        default:
+      if (type.isPrimitive()) {
+        if (!statsMap.containsKey(path)) {
           throw new SFException(
-              ErrorCode.UNKNOWN_DATA_TYPE,
-              type.getLogicalTypeAnnotation(),
-              primitiveType.getPrimitiveTypeName());
+              ErrorCode.INTERNAL_ERROR,
+              String.format("Stats not found for column: %s", type.getName()));
+        }
+        RowBufferStats stats = statsMap.get(path);
+        estimatedParquetSize += ParquetBufferValue.DEFINITION_LEVEL_ENCODING_BYTE_LEN;
+        estimatedParquetSize +=
+            isdDescendantsOfRepeatingGroup
+                ? ParquetBufferValue.REPETITION_LEVEL_ENCODING_BYTE_LEN
+                : 0;
+        PrimitiveType primitiveType = type.asPrimitiveType();
+        switch (primitiveType.getPrimitiveTypeName()) {
+          case BOOLEAN:
+            int intValue =
+                DataValidationUtil.validateAndParseBoolean(
+                    type.getName(), value, insertRowsCurrIndex);
+            value = intValue > 0;
+            stats.addIntValue(BigInteger.valueOf(intValue));
+            estimatedParquetSize += ParquetBufferValue.BIT_ENCODING_BYTE_LEN;
+            break;
+          case INT32:
+            int intVal = getInt32Value(value, primitiveType, insertRowsCurrIndex);
+            value = intVal;
+            stats.addIntValue(BigInteger.valueOf(intVal));
+            estimatedParquetSize += 4;
+            break;
+          case INT64:
+            long longVal =
+                getInt64Value(value, primitiveType, defaultTimezone, insertRowsCurrIndex);
+            value = longVal;
+            stats.addIntValue(BigInteger.valueOf(longVal));
+            estimatedParquetSize += 8;
+            break;
+          case FLOAT:
+            float floatVal =
+                (float)
+                    DataValidationUtil.validateAndParseReal(
+                        type.getName(), value, insertRowsCurrIndex);
+            value = floatVal;
+            stats.addRealValue((double) floatVal);
+            estimatedParquetSize += 4;
+            break;
+          case DOUBLE:
+            double doubleVal =
+                DataValidationUtil.validateAndParseReal(type.getName(), value, insertRowsCurrIndex);
+            value = doubleVal;
+            stats.addRealValue(doubleVal);
+            estimatedParquetSize += 8;
+            break;
+          case BINARY:
+            byte[] byteVal = getBinaryValue(value, primitiveType, stats, insertRowsCurrIndex);
+            value = byteVal;
+            estimatedParquetSize +=
+                ParquetBufferValue.BYTE_ARRAY_LENGTH_ENCODING_BYTE_LEN + byteVal.length;
+            break;
+          case FIXED_LEN_BYTE_ARRAY:
+            byte[] fixedLenByteArrayVal =
+                getFixedLenByteArrayValue(value, primitiveType, stats, insertRowsCurrIndex);
+            value = fixedLenByteArrayVal;
+            estimatedParquetSize +=
+                ParquetBufferValue.BYTE_ARRAY_LENGTH_ENCODING_BYTE_LEN
+                    + fixedLenByteArrayVal.length;
+            break;
+          default:
+            throw new SFException(
+                ErrorCode.UNKNOWN_DATA_TYPE,
+                type.getLogicalTypeAnnotation(),
+                primitiveType.getPrimitiveTypeName());
+        }
+      } else {
+        return getGroupValue(
+            value,
+            type.asGroupType(),
+            statsMap,
+            defaultTimezone,
+            insertRowsCurrIndex,
+            path,
+            isdDescendantsOfRepeatingGroup);
       }
     }
 
@@ -115,7 +157,14 @@ class IcebergParquetValueParser {
         throw new SFException(
             ErrorCode.INVALID_FORMAT_ROW, type.getName(), "Passed null to non nullable field");
       }
-      stats.incCurrentNullCount();
+      if (type.isPrimitive()) {
+        if (!statsMap.containsKey(path)) {
+          throw new SFException(
+              ErrorCode.INTERNAL_ERROR,
+              String.format("Stats not found for column: %s", type.getName()));
+        }
+        statsMap.get(path).incCurrentNullCount();
+      }
     }
 
     return new ParquetBufferValue(value, estimatedParquetSize);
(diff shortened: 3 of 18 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/ParquetBufferValue.java
```diff
@@ -9,6 +9,8 @@ class ParquetBufferValue {
   // Parquet uses BitPacking to encode boolean, hence 1 bit per value
   public static final float BIT_ENCODING_BYTE_LEN = 1.0f / 8;
 
+  public static final float REPETITION_LEVEL_ENCODING_BYTE_LEN = 2.0f / 8;
```
sfc-gh-hmadan, line 12: add comment on why this is 2.0f/8

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -91,20 +92,46 @@ public void setupSchema(List<ColumnMetadata> columns) {
...
+        }
       }
 
       id++;
     }
     schema = new MessageType(PARQUET_MESSAGE_TYPE_NAME, parquetTypes);
+    if (clientBufferParameters.getIsIcebergMode()) {
+      int ordinal = 0;
+      String prevParentColumnName = "";
+      for (ColumnDescriptor columnDescriptor : schema.getColumns()) {
```
sfc-gh-hmadan, line 114: MInd adding a comment on what this for loop tries to achieve?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -91,20 +92,46 @@ public void setupSchema(List<ColumnMetadata> columns) {
...
+          ordinal++;
+          prevParentColumnName = parentColumnName;
+        }
+        String subColumnName = String.join(".", columnDescriptor.getPath());
+        int fieldId =
+            parentColumnName.equals(subColumnName)
+                ? 0
+                : columnDescriptor.getPrimitiveType().getId().intValue();
+        RowBufferStats stats = new RowBufferStats(subColumnName, null, ordinal, fieldId);
+        this.statsMap.put(subColumnName, stats);
```
sfc-gh-hmadan, line 126: are we overwriting the statsMap entries that the first for loop in this method sets up?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -91,20 +92,46 @@ public void setupSchema(List<ColumnMetadata> columns) {
...
+      int ordinal = 0;
+      String prevParentColumnName = "";
+      for (ColumnDescriptor columnDescriptor : schema.getColumns()) {
+        String parentColumnName = columnDescriptor.getPath()[0];
+        if (!parentColumnName.equals(prevParentColumnName)) {
+          ordinal++;
+          prevParentColumnName = parentColumnName;
+        }
+        String subColumnName = String.join(".", columnDescriptor.getPath());
+        int fieldId =
```
sfc-gh-hmadan, line 121: Is this fieldId and the columnId from the previous for loop meant to be identical?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -91,20 +92,46 @@ public void setupSchema(List<ColumnMetadata> columns) {
...
-          new RowBufferStats(column.getName(), column.getCollation(), column.getOrdinal()));
-
-      if (onErrorOption == OpenChannelRequest.OnErrorOption.ABORT
-          || onErrorOption == OpenChannelRequest.OnErrorOption.SKIP_BATCH) {
-        this.tempStatsMap.put(
+      if (!clientBufferParameters.getIsIcebergMode()) {
+        this.statsMap.put(
             column.getInternalName(),
-            new RowBufferStats(column.getName(), column.getCollation(), column.getOrdinal()));
+            new RowBufferStats(column.getName(), column.getCollation(), column.getOrdinal(), 0));
```
sfc-gh-hmadan, line 98: nit: add comment on what this magic number zero is, like so:
column.getOrdinal, 0 /* argumentName */));

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -161,25 +188,24 @@ private float addRow(
 
     // Create new empty stats just for the current row.
     Map<String, RowBufferStats> forkedStatsMap = new HashMap<>();
+    statsMap.forEach((columnName, stats) -> forkedStatsMap.put(columnName, stats.forkEmpty()));
```
sfc-gh-hmadan, line 191: we don't want to do unnecessary iterations, please keep the old inlined `forkedStatsMap.put` in place.

sfc-gh-hmadan: i see IcebergParquetValueParser needs the whole map, not sure yet why that is, still reading through the changes.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -44,69 +50,105 @@ class IcebergParquetValueParser {
...
+  }
+
+  private static ParquetBufferValue parseColumnValueToParquet(
+      Object value,
+      Type type,
+      Map<String, RowBufferStats> statsMap,
+      ZoneId defaultTimezone,
+      long insertRowsCurrIndex,
+      String path,
+      boolean isdDescendantsOfRepeatingGroup) {
```
sfc-gh-hmadan, line 68: typo: extra d after is

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -115,7 +157,14 @@ static ParquetBufferValue parseColumnValueToParquet(
         throw new SFException(
             ErrorCode.INVALID_FORMAT_ROW, type.getName(), "Passed null to non nullable field");
       }
-      stats.incCurrentNullCount();
+      if (type.isPrimitive()) {
+        if (!statsMap.containsKey(path)) {
+          throw new SFException(
+              ErrorCode.INTERNAL_ERROR,
+              String.format("Stats not found for column: %s", type.getName()));
+        }
```
sfc-gh-hmadan, line 165: move this if !containsKey to before the if (value != null) check instead of duplicating in both if (value == null) and if (value != null) code paths.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -115,7 +157,14 @@ static ParquetBufferValue parseColumnValueToParquet(
         throw new SFException(
```
sfc-gh-hmadan, line 157: is the isRepetition check needed/applicable for primitive types too? If not it should be in an if !isPrimitive block?

We can structure this method as:
```
if (!statsMap.containsKey()) {
    throw;
}

if (primitive) {
    if (!null) {

    } else {

    }
} else {
    if (!null) {
        return getGroupValue();
    } else {
        if (isRepetition) {
            throw
        }
    }
}
```

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
```diff
@@ -131,6 +131,10 @@ public List<Long> getRowCountsFromFooter() {
     return blockRowCounts;
   }
 
+  public List<BlockMetaData> getBlocksMetadata() {
+    return writer.getFooter().getBlocks();
```
sfc-gh-hmadan, line 135: can getFooter be null?
can getBLocks be null?

If either of these is possible please return an Optional<List<BlockMetaData>> instead of expecting the caller to know that the value can be null

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
```diff
@@ -84,6 +98,38 @@ class FileColumnProperties {
     this.setDistinctValues(stats.getDistinctValues());
   }
 
+  FileColumnProperties(int fieldId, Statistics<?> statistics) {
+    this.setColumnOrdinal(fieldId);
+    this.setFieldId(fieldId);
```
sfc-gh-hmadan, line 103: didn't we say fieldId should be 0 for primitive types?

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
```diff
@@ -84,6 +98,38 @@ class FileColumnProperties {
...
+    } else if (statistics instanceof BinaryStatistics) {
+      if (statistics.type().getLogicalTypeAnnotation()
+          instanceof LogicalTypeAnnotation.DecimalLogicalTypeAnnotation) {
+        this.setMinIntValue(new BigInteger(statistics.getMinBytes()));
+        this.setMaxIntValue(new BigInteger(statistics.getMaxBytes()));
+      } else {
+        this.setMinStrValue(truncateBytesAsHex(statistics.getMinBytes(), false));
+        this.setMaxStrValue(truncateBytesAsHex(statistics.getMaxBytes(), true));
+      }
+    }
```
sfc-gh-hmadan, line 130: else throw since we don't know what typed statistics are being presented here? we don't want to silently ignore stats capture for any column.

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
```diff
@@ -84,6 +98,38 @@ class FileColumnProperties {
...
+  }
```
sfc-gh-hmadan, line 131: no setMaxLength call?

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
```diff
@@ -84,6 +98,38 @@ class FileColumnProperties {
     this.setDistinctValues(stats.getDistinctValues());
   }
```
sfc-gh-hmadan, line 100: I see that these two methods are equivalent, trying to understand why we couldn't use RowBufferStats and had to rely on Statistics<?> instead?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -1055,6 +1055,80 @@ static int validateAndParseBoolean(String columnName, Object input, long insertR
...
+   * </ul>
+   *
+   * @param columnName Column name, used in validation error messages
+   * @param input Object to validate and parse
+   * @param insertRowIndex Row index for error reporting
+   * @return Object cast to Iterable
+   */
+  static Iterable<Object> validateAndParseIcebergList(
+      String columnName, Object input, long insertRowIndex) {
+    if (!(input instanceof Iterable)) {
```
sfc-gh-hmadan, line 1104: add a backlog workitem to support native arrays too in addition to Iterables (For example int[]).

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -1055,6 +1055,80 @@ static int validateAndParseBoolean(String columnName, Object input, long insertR
...
+      throw typeNotAllowedException(
+          columnName, input.getClass(), "LIST", new String[] {"Iterable"}, insertRowIndex);
+    }
+    return (Iterable<Object>) input;
```
sfc-gh-hmadan, line 1108: not sure how this works in all cases - I just tried casting Iterable<Integer> to Iterable<Object> locally and got:
`Inconvertible types; cannot cast 'java.lang.Iterable<java.lang.Integer>' to 'java.lang.Iterable<java.lang.Object>'`

sfc-gh-hmadan: At a minimum you need to do `(Iterable<Object>)(Iterable)input` to get around this.

Assuming that's the only way here, there's no reason to use the generic Iterable and might as well use the non-generic Iterable as the return type.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -1055,6 +1055,80 @@ static int validateAndParseBoolean(String columnName, Object input, long insertR
...
+   * @param insertRowIndex Row index for error reporting
+   * @return Object cast to Map
+   */
+  static Map<Object, Object> validateAndParseIcebergMap(
+      String columnName, Object input, long insertRowIndex) {
+    if (!(input instanceof Map)) {
+      throw typeNotAllowedException(
+          columnName, input.getClass(), "MAP", new String[] {"Map"}, insertRowIndex);
+    }
+    return (Map<Object, Object>) input;
```
sfc-gh-hmadan, line 1129: same comment as above on generic vs. non-generic and that this cast will fail.

Are we missing testcases that hit these methods?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -1055,6 +1055,80 @@ static int validateAndParseBoolean(String columnName, Object input, long insertR
...
+   * </ul>
+   *
+   * @param columnName Column name, used in validation error messages
+   * @param input Object to validate and parse
+   * @param insertRowIndex Row index for error reporting
+   * @return Object cast to Map
+   */
+  static Map<Object, Object> validateAndParseIcebergMap(
+      String columnName, Object input, long insertRowIndex) {
+    if (!(input instanceof Map)) {
```
sfc-gh-hmadan, line 1125: add post-GA backlog item to consider supporting any `Set<Map.Entry<K, V>>` object too.

src/main/java/net/snowflake/ingest/streaming/internal/RowBufferStats.java
```diff
@@ -16,6 +16,7 @@
 class RowBufferStats {
 
   private final int ordinal;
+  private final int fieldId;
```
sfc-gh-hmadan, line 19: add comments on what's the difference between ordinal and fieldId, and what's expected values in them for (a) FDN column, (b) iceberg column primitive typed, (c) iceberg column non-primitive typed

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-hmadan: i see the if !iceberg in the section above, nvm

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -91,20 +92,46 @@ public void setupSchema(List<ColumnMetadata> columns) {
...
         addNonNullableFieldName(column.getInternalName());
       }
-      this.statsMap.put(
-          column.getInternalName(),
-          new RowBufferStats(column.getName(), column.getCollation(), column.getOrdinal()));
-
-      if (onErrorOption == OpenChannelRequest.OnErrorOption.ABORT
-          || onErrorOption == OpenChannelRequest.OnErrorOption.SKIP_BATCH) {
-        this.tempStatsMap.put(
+      if (!clientBufferParameters.getIsIcebergMode()) {
```
sfc-gh-hmadan, line 95: add comment on why the stats map tracking is done differently, and whether there's any qualitative differences. For example its column.getInternalName here and columnDescriptor.getPath() in iceberg mode.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -91,20 +92,46 @@ public void setupSchema(List<ColumnMetadata> columns) {
       if (!column.getNullable()) {
```
sfc-gh-hmadan, line 92: 1. the ParquetColumn.columnIndex now doesn't map to either the fieldId or the columnOrdinal, is that correct?
2. How do the IDs flow if I have a table with schema (int, struct, string, array, float, map) ? Are we assuming that all subcolumns are meant to be at the end of the list always?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -91,20 +92,46 @@ public void setupSchema(List<ColumnMetadata> columns) {
...
+              new RowBufferStats(column.getName(), column.getCollation(), column.getOrdinal(), 0));
+        }
       }
 
       id++;
     }
     schema = new MessageType(PARQUET_MESSAGE_TYPE_NAME, parquetTypes);
+    if (clientBufferParameters.getIsIcebergMode()) {
+      int ordinal = 0;
+      String prevParentColumnName = "";
```
sfc-gh-hmadan, line 113: this assumes the list is properly ordered? Pl add a backlog item to validate the schema when we get it from channel_open call and make sure its ordered how we expect it to be over here.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -91,20 +92,46 @@ public void setupSchema(List<ColumnMetadata> columns) {
...
+      for (ColumnDescriptor columnDescriptor : schema.getColumns()) {
+        String parentColumnName = columnDescriptor.getPath()[0];
+        if (!parentColumnName.equals(prevParentColumnName)) {
+          ordinal++;
+          prevParentColumnName = parentColumnName;
+        }
```
sfc-gh-hmadan, line 119: isn't it safer to check `columnDescriptor.getPath().size > 1` ? When > 1 we are in subcol land. 
And then validate that the previous entry's column name is either the parent of the current column, or a sibling, and throw if neither

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -91,20 +92,46 @@ public void setupSchema(List<ColumnMetadata> columns) {
...
+        String subColumnName = String.join(".", columnDescriptor.getPath());
+        int fieldId =
+            parentColumnName.equals(subColumnName)
```
sfc-gh-hmadan, line 122: columnDescriptor.getPath().size == 1 is a cleaner check for whether this is a top level column or a subcolumn.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -91,20 +92,46 @@ public void setupSchema(List<ColumnMetadata> columns) {
...
+                ? 0
+                : columnDescriptor.getPrimitiveType().getId().intValue();
+        RowBufferStats stats = new RowBufferStats(subColumnName, null, ordinal, fieldId);
```
sfc-gh-hmadan, line 125: hardcoding null for the collation - lets chat about this, leaving a comment so i don't forget!

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-hmadan: resolving this.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -1055,6 +1055,80 @@ static int validateAndParseBoolean(String columnName, Object input, long insertR
...
+      throw typeNotAllowedException(
+          columnName,
+          input.getClass(),
+          "STRUCT",
+          new String[] {"Map<String, Object>"},
+          insertRowIndex);
+    }
+    if (!((Map<?, ?>) input).keySet().stream().allMatch(key -> key instanceof String)) {
```
sfc-gh-hmadan, line 1080: pl use plain old for loops instead of streams, for loops are much faster.

Where is the validation that only those keys are in the Map as what's defined on the column schema?

sfc-gh-hmadan: ok i see where it'll make sense (IcebergParquetValueParser), leaving a comment there for the struct schema validation.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -298,4 +347,166 @@ private static int timeUnitToScale(LogicalTypeAnnotation.TimeUnit timeUnit) {
...
+      boolean isdDescendantsOfRepeatingGroup) {
+    Map<String, Object> structVal =
+        DataValidationUtil.validateAndParseIcebergStruct(
+            type.getName(), value, insertRowsCurrIndex);
+    List<Object> listVal = new ArrayList<>(type.getFieldCount());
+    float estimatedParquetSize = 0f;
+    for (int i = 0; i < type.getFieldCount(); i++) {
+      ParquetBufferValue parsedValue =
+          parseColumnValueToParquet(
+              structVal.getOrDefault(type.getFieldName(i), null),
```
sfc-gh-hmadan, line 420: We're missing validation for any unexpected fields being sent in via the map.
Lets do the following:
1. Create a new HashSet(structVal.keySet()) at the beginning of this method
2. For every field that you've validated in this for loop, remove it from this set
3. At the end of the method, the set should be empty. If there's still something left in the set, throw a validation error for unexpected field or whatever the SDK generally throws. With a proper path to the current field of course.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -44,69 +50,105 @@ class IcebergParquetValueParser {
...
-        case FIXED_LEN_BYTE_ARRAY:
-          byte[] fixedLenByteArrayVal =
-              getFixedLenByteArrayValue(value, primitiveType, stats, insertRowsCurrIndex);
-          value = fixedLenByteArrayVal;
-          estimatedParquetSize +=
-              ParquetBufferValue.BYTE_ARRAY_LENGTH_ENCODING_BYTE_LEN + fixedLenByteArrayVal.length;
-          break;
-        default:
+      if (type.isPrimitive()) {
+        if (!statsMap.containsKey(path)) {
```
sfc-gh-hmadan, line 73: collapse the two map hits (across containsKey and get) into one, like so:
```
RowBufferStats stats = statsMap.get();
if (stats == null) {
    throw
}
```

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -44,69 +50,105 @@ class IcebergParquetValueParser {
...
-              type.getLogicalTypeAnnotation(),
-              primitiveType.getPrimitiveTypeName());
+              ErrorCode.INTERNAL_ERROR,
+              String.format("Stats not found for column: %s", type.getName()));
+        }
+        RowBufferStats stats = statsMap.get(path);
+        estimatedParquetSize += ParquetBufferValue.DEFINITION_LEVEL_ENCODING_BYTE_LEN;
+        estimatedParquetSize +=
+            isdDescendantsOfRepeatingGroup
+                ? ParquetBufferValue.REPETITION_LEVEL_ENCODING_BYTE_LEN
```
sfc-gh-hmadan, line 82: is this only needed for primitive typed values? What if its an array of objects, do we not need this constant to be added to estimatedParquetSize for each object of the array?

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -115,7 +157,14 @@ static ParquetBufferValue parseColumnValueToParquet(
...
             ErrorCode.INVALID_FORMAT_ROW, type.getName(), "Passed null to non nullable field");
       }
-      stats.incCurrentNullCount();
+      if (type.isPrimitive()) {
+        if (!statsMap.containsKey(path)) {
+          throw new SFException(
+              ErrorCode.INTERNAL_ERROR,
+              String.format("Stats not found for column: %s", type.getName()));
+        }
+        statsMap.get(path).incCurrentNullCount();
```
sfc-gh-hmadan, line 166: need the same incNullCount call for non-primitive fields too, no?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetBufferValue.java
sfc-gh-alhuang: Added to where the definition level encoding is defined.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: Added.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: No. The columnId for `fieldIndex` is used for internal check only.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: Added.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: Fixed.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: Moved to the beginning.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: The `isRepetition` should be checked for both primitive type and group type. As there might be required primitive column with null value which is not allowed.

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
sfc-gh-alhuang: Yes. Resolved as we no longer need this anymore.

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
sfc-gh-alhuang: The fieldId is set to zero when the `RowbufferStats` is created. Resolve this as this method is no longer needed.

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
sfc-gh-alhuang: Switch to the same code path as FDN. Not using this method anymore.

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
sfc-gh-alhuang: Switch to the same code path as FDN. Not using this method anymore.

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
sfc-gh-alhuang: Switch back to `RowBufferStats`, remove this method.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-alhuang: Converted to Iterable<?>.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-alhuang: Converted to Map<?, ?>. Had a test calling this in `IcebergParquetValueParser.parseMap`.

src/main/java/net/snowflake/ingest/streaming/internal/RowBufferStats.java
sfc-gh-alhuang: Added comments.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: Added comments. Actually streaming to FDN can also use the below logic of setting up Iceberg stats. The `column.getInternalName` is the same as `columnDescriptor.getPath()` as the parquet column was set up using internal name.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: 1. Yes. The `ParquetColumn.columnIndex` is used for internal check and starts from zero. While `columnOrdinal` and `fieldId` starts from one.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: It's ordered as it's generated from depth first search. Switching to get ordinal from schema instead of calculate it.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: Make sense. Fixed.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-alhuang: Modified to for loop.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: Moved the `containsKey` check to the beginning of the method.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: I used [this paper](https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/36632.pdf) as reference to understand repetition level and definition level encoding. And from [this article](https://blog.x.com/engineering/en_us/a/2013/dremel-made-simple-with-parquet), only descendants of a repeating group needs to record repetition level. We only need these on leaf values.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: The EPs from server and parquet footer does not record stats from non-primitive (non-leaf) fields.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: Make sense. Fixed.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: Modified.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -1055,6 +1055,82 @@ static int validateAndParseBoolean(String columnName, Object input, long insertR
...
+          "STRUCT",
+          new String[] {"Map<String, Object>"},
+          insertRowIndex);
+    }
+    for (Object key : ((Map<?, ?>) input).keySet()) {
+      if (!(key instanceof String)) {
+        throw new SFException(
+            ErrorCode.INVALID_FORMAT_ROW,
+            String.format(
+                "Field name of a struct must be of type String, rowIndex:%d", insertRowIndex));
```
sfc-gh-hmadan, line 1085: also log the offending column name, and also the path at which this error was found.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -1055,6 +1055,82 @@ static int validateAndParseBoolean(String columnName, Object input, long insertR
...
+   * @param columnName Column name, used in validation error messages
+   * @param input Object to validate and parse
+   * @param insertRowIndex Row index for error reporting
+   * @return Object cast to Map
+   */
+  static Map<String, ?> validateAndParseIcebergStruct(
+      String columnName, Object input, long insertRowIndex) {
+    if (!(input instanceof Map)) {
+      throw typeNotAllowedException(
+          columnName,
```
sfc-gh-hmadan, line 1074: will columnName here be the full path (for a nested structured object with a failure deep in the hierarchy)

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-alhuang: Added in [jira](https://snowflakecomputing.atlassian.net/browse/SNOW-1690892)

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-alhuang: Added in [jira](https://snowflakecomputing.atlassian.net/browse/SNOW-1690892)

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
```diff
@@ -93,6 +100,16 @@ public void setColumnOrdinal(int columnOrdinal) {
     this.columnOrdinal = columnOrdinal;
   }
 
+  @JsonProperty("fieldId")
+  @JsonInclude(JsonInclude.Include.NON_DEFAULT)
```
sfc-gh-hmadan, line 104: The server validates incoming requests and will expect fieldId to be populated for iceberg tables, even if its 0.
By using NON_DEFAULT we end up seeing a weird behavior where the server cannot differentiate between "null/0 because client intended to not set the value versus null/0 because of the NON_DEFAULT attribute".

Please use NON_NULL and make the field an Integer (so its nullable).
If both RowBufferStats.fieldId and FileColumnProperties.fieldId are nullable, line 55 here can stay as-is. We'll add a conditional to where RowBufferStats is initialized and only pass in a fieldId value for iceberg tables. This way null implies not-set, and not-serialized, whereas non-null implies it was explicitly set.

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
```diff
@@ -206,6 +223,7 @@ void setMaxStrNonCollated(String maxStrNonCollated) {
   public String toString() {
     final StringBuilder sb = new StringBuilder("{");
     sb.append("\"columnOrdinal\": ").append(columnOrdinal);
+    sb.append(", \"fieldId\": ").append(fieldId);
```
sfc-gh-hmadan, line 226: won't this cause immediate IT failures? the server doesn't expect fieldId in the json, so the deserializer is going to barf?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-hmadan: lets chat about this f2f. I see the ParquetColumn.index / this.fieldIndex are used in addRow and it looks like they're meant to only track "top level" fields in terms of what data is sent in by the customer, wherease with your changes they will end up being much larger than necessary.
For instance lets take a table with a single structured object column (C1) with 100 fields (F1-F100), and each field is a structured object with 100 fields each (FF1-FF100).
In addRow you'll get a Map with a single entry (key=C1, value = Map-with-100-entries).
However you'll end up allocating an array with ~100*100 = 10k elements because of this field id business.

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
sfc-gh-alhuang: Changed to `NON_NULL` and Integer.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-alhuang: Passed full path into `DataValidationUtil` in Iceberg mode.

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
sfc-gh-alhuang: I think the `toString` method is used for logging. The deserializer follows the `@JsonProperty` tag.

src/main/java/net/snowflake/ingest/utils/Utils.java
```diff
@@ -411,4 +411,22 @@ public static String getFullyQualifiedChannelName(
...
+
+  /*
+   * Get concat dot path, ignore null or empty string
+   *
+   * @param path the path
+   */
+  public static String concatDotPath(String... path) {
+    StringBuilder sb = new StringBuilder();
+    for (String p : path) {
+      if (!isNullOrEmpty(p)) {
```
sfc-gh-hmadan, line 423: curious why we want to silently ignore a null or empty path element? Shouldn't we be validating this when we receive the schema from service, and throw if any element is null or empty?

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
```diff
@@ -283,7 +282,7 @@ public void prepareForWrite(RecordConsumer recordConsumer) {
 
     @Override
     public void write(List<Object> values) {
-      List<ColumnDescriptor> cols = schema.getColumns();
+      List<Type> cols = schema.getFields();
```
sfc-gh-hmadan, line 285: What's the difference between schema.getColumns and getFields?
Also just to confirm getFields is supposed to return top level columns, and not any subcolumns?

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
```diff
@@ -344,7 +343,19 @@ private void writeValues(List<Object> values, GroupType type) {
                     "Unsupported column type: " + cols.get(i).asPrimitiveType());
             }
           } else {
-            throw new ParquetEncodingException("Unsupported column type: " + cols.get(i));
+            if (cols.get(i).isRepetition(Type.Repetition.REPEATED)) {
+              for (Object o : values) {
+                recordConsumer.startGroup();
+                if (o != null) {
+                  writeValues((List<Object>) o, cols.get(i).asGroupType());
```
sfc-gh-hmadan, line 350: how are we sure that this cast will not fail here?
also applies to the cast in line 356.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -91,20 +94,54 @@ public void setupSchema(List<ColumnMetadata> columns) {
...
+            new RowBufferStats(
+                column.getName(), column.getCollation(), column.getOrdinal(), null /* fieldId */));
+
+        if (onErrorOption == OpenChannelRequest.OnErrorOption.ABORT
+            || onErrorOption == OpenChannelRequest.OnErrorOption.SKIP_BATCH) {
+          this.tempStatsMap.put(
+              column.getInternalName(),
+              new RowBufferStats(
+                  column.getName(),
+                  column.getCollation(),
```
sfc-gh-hmadan, line 110: (lets discuss tom) just like we have column.getCollation here that we pass in correctly, why not do the same thing for iceberg too?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -91,20 +94,54 @@ public void setupSchema(List<ColumnMetadata> columns) {
...
+                  column.getOrdinal(),
+                  null /* fieldId */));
+        }
       }
 
       id++;
     }
     schema = new MessageType(PARQUET_MESSAGE_TYPE_NAME, parquetTypes);
```
sfc-gh-hmadan, line 118: PARQUET_MESSAGE_TYPE_NAME is a constant with the value bdec.

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
sfc-gh-hmadan: I see, that's odd coz we're taking the pain to emit proper json here!

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-hmadan: pl add a comment for this

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
sfc-gh-hmadan: pl add comment

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -1055,6 +1055,82 @@ static int validateAndParseBoolean(String columnName, Object input, long insertR
...
+          input.getClass(),
+          "STRUCT",
+          new String[] {"Map<String, Object>"},
+          insertRowIndex);
+    }
+    for (Object key : ((Map<?, ?>) input).keySet()) {
+      if (!(key instanceof String)) {
```
sfc-gh-hmadan, line 1081: we need to validate that the list of fields being passed in is a subset of the list of fields declared on the schema, no?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -1055,6 +1055,82 @@ static int validateAndParseBoolean(String columnName, Object input, long insertR
...
+   * @param input Object to validate and parse
+   * @param insertRowIndex Row index for error reporting
+   * @return Object cast to Map
+   */
+  static Map<?, ?> validateAndParseIcebergMap(
+      String columnName, Object input, long insertRowIndex) {
+    if (!(input instanceof Map)) {
+      throw typeNotAllowedException(
+          columnName, input.getClass(), "MAP", new String[] {"Map"}, insertRowIndex);
+    }
```
sfc-gh-hmadan, line 1130: need to validate the data type of the key and value of the map too ?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-hmadan: pl checkout the comment here too, on this.fieldIndex.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -91,20 +94,54 @@ public void setupSchema(List<ColumnMetadata> columns) {
...
             column.getInternalName(),
-            new RowBufferStats(column.getName(), column.getCollation(), column.getOrdinal()));
+            new RowBufferStats(
+                column.getName(), column.getCollation(), column.getOrdinal(), null /* fieldId */));
+
+        if (onErrorOption == OpenChannelRequest.OnErrorOption.ABORT
+            || onErrorOption == OpenChannelRequest.OnErrorOption.SKIP_BATCH) {
+          this.tempStatsMap.put(
+              column.getInternalName(),
+              new RowBufferStats(
```
sfc-gh-hmadan, line 108: add a comment on why the same stats object shuld NOT be shared across statsMap and tempStatsMap

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -91,20 +94,54 @@ public void setupSchema(List<ColumnMetadata> columns) {
...
+                  null /* fieldId */));
+        }
       }
 
       id++;
     }
     schema = new MessageType(PARQUET_MESSAGE_TYPE_NAME, parquetTypes);
+    if (clientBufferParameters.getIsIcebergMode()) {
+      /* Iceberg mode requires stats for sub-columns, set them up here. */
+      for (ColumnDescriptor subColumnDescriptor : schema.getColumns()) {
```
sfc-gh-hmadan, line 121: nit: columnDescriptor, since its not just subcolumns that yo're processing here.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -91,20 +94,54 @@ public void setupSchema(List<ColumnMetadata> columns) {
...
+        String subColumnName = concatDotPath(subColumnDescriptor.getPath());
+
+        /* set fieldId to 0 for non-structured columns */
```
sfc-gh-hmadan, line 124: do you mind adding in a big comment block here what are the expected fieldId and ordinal values for each field from the example we had in the conf room?
INT F1,
STRUCT F2 { 
    STRUCT F21 { 
        INT F212 
    }, 
    INT F22 
},
INT F3,
MAP F4 {
    LONG F4K,
    MAP F4V {
        STR F4VK,
        STR F4VV
    }
},
INT F5,
ARRAY F6,
INT F7

sfc-gh-alhuang: Added a comment block with an example to explain `fieldId`.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -91,20 +94,54 @@ public void setupSchema(List<ColumnMetadata> columns) {
...
+        }
       }
 
       id++;
     }
     schema = new MessageType(PARQUET_MESSAGE_TYPE_NAME, parquetTypes);
+    if (clientBufferParameters.getIsIcebergMode()) {
+      /* Iceberg mode requires stats for sub-columns, set them up here. */
+      for (ColumnDescriptor subColumnDescriptor : schema.getColumns()) {
+        String subColumnName = concatDotPath(subColumnDescriptor.getPath());
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Renamed.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: Renamed.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: Added a comment to explain this.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-alhuang: The data type of key and value will be recursively check via `IcebergParquetValueParser` calling `DataValidationUtil`.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-alhuang: Did it [here](https://github.com/snowflakedb/snowflake-ingest-java/blob/a1ab1074f657d8ed515a0b7fc90847f2f66c3955/src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java#L442). The `DataValidationUtil` only validates the data type, range, size etc. Schema related check are kept in `IcebergParquetValueParser`

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: Updated to `schema` under Iceberg mode.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: Discussed offline.

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
sfc-gh-alhuang: Cast to `List<?>`

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
sfc-gh-alhuang: Added comment.

src/main/java/net/snowflake/ingest/utils/Utils.java
sfc-gh-alhuang: Added is null or empty check here.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -44,78 +57,116 @@ class IcebergParquetValueParser {
   static ParquetBufferValue parseColumnValueToParquet(
       Object value,
```
sfc-gh-hmadan, line 58: For next PR: We need to validate the max nesting depth of the object that's passed in, I think snowflake service caps it at 1000.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -298,4 +359,177 @@ private static int timeUnitToScale(LogicalTypeAnnotation.TimeUnit timeUnit) {
...
+          value,
+          type,
+          statsMap,
+          defaultTimezone,
+          insertRowsCurrIndex,
+          path,
+          isDescendantsOfRepeatingGroup);
+    }
+    throw new SFException(
+        ErrorCode.UNKNOWN_DATA_TYPE, logicalTypeAnnotation, type.getClass().getSimpleName());
```
sfc-gh-hmadan, line 408: next PR: can this error have more info, with the rowindex / etc?

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -298,4 +359,177 @@ private static int timeUnitToScale(LogicalTypeAnnotation.TimeUnit timeUnit) {
...
+      ParquetBufferValue parsedValue =
+          parseColumnValueToParquet(
+              structVal.getOrDefault(type.getFieldName(i), null),
+              type.getType(i),
+              statsMap,
+              defaultTimezone,
+              insertRowsCurrIndex,
+              path,
+              isDescendantsOfRepeatingGroup);
+      extraFields.remove(type.getFieldName(i));
```
sfc-gh-hmadan, line 439: structVal.keySet() returns a "view" of the underlying map, and any mutations done on extraFields will reflect on the underlying collection too.
From documentation of hashMap.keyset:
```
   * Returns a {@link Set} view of the keys contained in this map.
     * The set is backed by the map, so changes to the map are
     * reflected in the set, and vice-versa.  If the map is modified
     * while an iteration over the set is in progress (except through
     * the iterator's own {@code remove} operation), the results of
     * the iteration are undefined.  The set supports element removal,
     * which removes the corresponding mapping from the map, via the
     * {@code Iterator.remove}, {@code Set.remove},
     * {@code removeAll}, {@code retainAll}, and {@code clear}
     * operations.  It does not support the {@code add} or {@code addAll}
     * operations.
  ```

  Suggest doing `Set<String> extraFields = new HashSet<>(structVal.keySet())` !

File: src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java [hunk 49]
```diff
@@ -298,4 +359,177 @@ private static int timeUnitToScale(LogicalTypeAnnotation.TimeUnit timeUnit) {
...
+      listVal.add(parsedValue.getValue());
+      estimatedParquetSize += parsedValue.getSize();
+    }
+    if (!extraFields.isEmpty()) {
+      extraFields =
+          extraFields.stream().map(f -> concatDotPath(path, f)).collect(Collectors.toSet());
```
sfc-gh-hmadan, line 445: String extraFieldsStr = extraFields.stream().map().collect(Collectors.joining(","))

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -298,4 +359,177 @@ private static int timeUnitToScale(LogicalTypeAnnotation.TimeUnit timeUnit) {
...
+    Iterable<?> iterableVal =
+        DataValidationUtil.validateAndParseIcebergList(path, value, insertRowsCurrIndex);
+    List<Object> listVal = new ArrayList<>();
+    final AtomicReference<Float> estimatedParquetSize = new AtomicReference<>(0f);
+    iterableVal.forEach(
+        element -> {
+          ParquetBufferValue parsedValue =
+              parseColumnValueToParquet(
+                  element,
+                  type.getType(0).asGroupType().getType(0),
```
sfc-gh-hmadan, line 479: next PR: add a comment on what this does? assume 99% of code readers in this repo do not know much about parquet internals :)

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -298,4 +359,177 @@ private static int timeUnitToScale(LogicalTypeAnnotation.TimeUnit timeUnit) {
...
+      Object value,
+      GroupType type,
+      Map<String, RowBufferStats> statsMap,
+      ZoneId defaultTimezone,
+      final long insertRowsCurrIndex,
+      String path) {
+    Iterable<?> iterableVal =
+        DataValidationUtil.validateAndParseIcebergList(path, value, insertRowsCurrIndex);
+    List<Object> listVal = new ArrayList<>();
+    final AtomicReference<Float> estimatedParquetSize = new AtomicReference<>(0f);
```
sfc-gh-hmadan, line 473: we don't need atomic here because there's no concurrency to manage. It introduces unnecessary volatile reads/writes.
if you do a regular for() loop you won't need the atomicRefrence and can live with regular float?

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -298,4 +359,177 @@ private static int timeUnitToScale(LogicalTypeAnnotation.TimeUnit timeUnit) {
...
+    iterableVal.forEach(
+        element -> {
+          ParquetBufferValue parsedValue =
+              parseColumnValueToParquet(
+                  element,
+                  type.getType(0).asGroupType().getType(0),
+                  statsMap,
+                  defaultTimezone,
+                  insertRowsCurrIndex,
+                  concatDotPath(path, THREE_LEVEL_LIST_GROUP_NAME),
```
sfc-gh-hmadan, line 483: do we do this concat once outside hte for loop instead of doing it repeatedly?

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -298,4 +359,177 @@ private static int timeUnitToScale(LogicalTypeAnnotation.TimeUnit timeUnit) {
...
+      GroupType type,
+      Map<String, RowBufferStats> statsMap,
+      ZoneId defaultTimezone,
+      final long insertRowsCurrIndex,
+      String path,
+      boolean isDescendantsOfRepeatingGroup) {
+    Map<?, ?> mapVal =
+        DataValidationUtil.validateAndParseIcebergMap(path, value, insertRowsCurrIndex);
+    List<Object> listVal = new ArrayList<>();
+    final AtomicReference<Float> estimatedParquetSize = new AtomicReference<>(0f);
```
sfc-gh-hmadan, line 509: same here, can do away with AtomicReference if you don't use a lambda.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -298,4 +359,177 @@ private static int timeUnitToScale(LogicalTypeAnnotation.TimeUnit timeUnit) {
...
+        (k, v) -> {
+          ParquetBufferValue parsedKey =
+              parseColumnValueToParquet(
+                  k,
+                  type.getType(0).asGroupType().getType(0),
+                  statsMap,
+                  defaultTimezone,
+                  insertRowsCurrIndex,
+                  concatDotPath(path, THREE_LEVEL_MAP_GROUP_NAME),
+                  true);
```
sfc-gh-hmadan, line 520: is this hardcoding intentional? add a comment why we're hardcoding true here but passing in isDescendantsOfRepeatingGroup for parsedValue ?

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: Added to [jira](https://snowflakecomputing.atlassian.net/browse/SNOW-1490151).

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: Fixed in #845, thanks for the fix!

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: Modified.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: Added to [jira](https://snowflakecomputing.atlassian.net/browse/SNOW-1490151).

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: Moved concat outside the for loop.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: Changed to for loop.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: both key and value should be hard coded to true. Fixed and modified the test case.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-alhuang: Added columName.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: `this.fieldIndex` is used for row validation with top level column, and the `fieldIndex` only contains top level columns.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -502,35 +494,34 @@ private static ParquetBufferValue get3LevelMapValue(
...
+          parseColumnValueToParquet(
+              entry.getValue(),
+              type.getType(0).asGroupType().getType(1),
+              statsMap,
+              defaultTimezone,
+              insertRowsCurrIndex,
+              mapGroupPath,
+              true);
+      listVal.add(Arrays.asList(parsedKey.getValue(), parsedValue.getValue()));
+      estimatedParquetSize += parsedKey.getSize() + parsedValue.getSize();
```
sfc-gh-hmadan, line 523: Is there no overhead bytes that need to be added in this case? So the parquet file adds a byte[] for the key, and byte[] for the value, but (a) no separator, or (b) no header that writes out the number of bytes of each field?

How are we validating that the estimatedparquetSize is accurate with structured data types?

sfc-gh-alhuang: Afaik parquet use definition & repetition level to determine the structure. The headers only exist for encoding (e.g. bitpack encoding needs 1 bytes header at the start of the run). For estimated parquet size, we only validates this through our assertion test, do you think we should add a test case against the actual generated blob?

sfc-gh-hmadan: yeah lets add this to the testing gaps JIRA, to veirfy how close our estimate gets to actual file size.

sfc-gh-hmadan (approved):

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: Added in [jira](https://snowflakecomputing.atlassian.net/browse/SNOW-1490151).

PR #845, opened 2024-09-29
Set createdon in BlobMetadata and is_iceberg in Drop Channel request

sfc-gh-hmadan:
1. Set createdOn on the blob Metadata, which can be used in the service to populate the FileRegistrationRequest
2. Set is_iceberg in /channels/drop API payload since it gets validated the same way as open channel (where we do have the is_iceberg API available)

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
@@ -142,7 +142,8 @@ class BlobBuilder {
           chunkMetadataBuilder
               .setMajorVersion(Constants.PARQUET_MAJOR_VERSION)
               .setMinorVersion(Constants.PARQUET_MINOR_VERSION)
-              .setCreatedOn(0L)
+              // set createdOn in seconds
+              .setCreatedOn(System.currentTimeMillis() / 1000)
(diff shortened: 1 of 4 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -142,7 +142,8 @@ static <T> Blob constructBlobAndMetadata(
           chunkMetadataBuilder
               .setMajorVersion(Constants.PARQUET_MAJOR_VERSION)
               .setMinorVersion(Constants.PARQUET_MINOR_VERSION)
-              .setCreatedOn(0L)
+              // set createdOn in seconds
+              .setCreatedOn(System.currentTimeMillis() / 1000)
```
sfc-gh-xhuang, line 146: Why do we need this? Blobs already have create time in its property and its cheaper to get it there than to query metadata

sfc-gh-hmadan: For having consistent metadata across XP-based DML inserts into iceberg tables and SDK-based inserts into iceberg tables.

sfc-gh-alhuang (approved):
Looks good, thanks!

PR #846, opened 2024-09-30
Structured Data Type Fix : Don't mutate the collection passed in by the user in InsertRow

sfc-gh-hmadan:
Stacked PR for a fix I found in alec's inprogress changes.

@sfc-gh-alhuang Please add a testcase that uses ImmutableCollections for the structured data type test objects, to catch a regression on this front.


sfc-gh-alhuang (approved):
Thanks for the fix!

PR #848, opened 2024-10-01
Revert "Add chunk offset to file id key to make each chunk have a uni…

sfc-gh-psaha:
…que key (#825)" temporarily to exclude it from the next release.

This reverts commit a107b50cb2c0fc29249a63bc2751e37b2a777e10.


sfc-gh-tzhang (approved):
We don't have to rollback, the release can be on a different branch, but I'm also fine with rolling it back, less work for the triage :)

sfc-gh-xhuang (changes requested):
confirming revert

sfc-gh-xhuang (approved):

PR #849, opened 2024-10-02
SNOW-1707031 hadoop upgrade for vulnerability fix

sfc-gh-japatel:
[SNOW-1707031](https://snowflakecomputing.atlassian.net/browse/SNOW-1707031)

Added few exclusions. Here is how I did it: I looked at the failed command of `mvn clean install` and compared it with `mvn dependency:tree`

Mostly it's about finding which dependencies can be excluded so that the new upgrade which brought in new dependencies are used. 

Other minor fixes for merge gates to pass. Ran locally and it succeeeded
```
mvn install -PcheckShadedContent -DskipTests=true --batch-mode --show-version
```

[SNOW-1707031]: https://snowflakecomputing.atlassian.net/browse/SNOW-1707031?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sfc-gh-xhuang (approved):
Thanks for fixing this!

sfc-gh-tzhang (approved):
LGTM, @sfc-gh-lsembera do we need to check with security about the added license?

sfc-gh-japatel:
> LGTM, @sfc-gh-lsembera do we need to check with security about the added license?

I dont think so, it's existing license we are adding and it's a known one.

sfc-gh-lsembera:
Thanks, @sfc-gh-japatel. The only issue I see here is that it will further inflate the JAR size. Should we try to exclude the new dependencies and see whether tests pass? I don't see why would we the SDK need netty.

PR #850, opened 2024-10-02
Integration data type test for new table format

sfc-gh-alhuang:
1. Created data type ITs for Iceberg table and verified through local VM. This test should be disabled until the sever side change is release.
2. Bug fix for `FileColumnProperties`. The Iceberg mode will fill in default value for corresponding type now if all values in the column are null.
3. Created `AbstractDataTypeTest.testIcebergIngestAndQuery` for multi row testing.


src/test/java/net/snowflake/ingest/streaming/internal/datatypes/AbstractDataTypeTest.java
```diff
@@ -112,6 +122,21 @@ protected String createTable(String dataType) throws SQLException {
...
+
+  protected String createIcebergTable(String dataType) throws SQLException {
+    String tableName = getRandomIdentifier();
+    conn.createStatement()
+        .execute(
+            String.format(
+                "create or replace iceberg table %s (%s string, %s %s) "
+                    + "catalog = 'SNOWFLAKE' "
+                    + "external_volume = 'streaming_ingest' "
+                    + "base_location = 'all_type_col';",
```
sfc-gh-hmadan, line 137: 1. lets use a different base_location for each table.
2. Create a JIRA for how to cleanup all of the data that ITs upload, from the external volume.

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergDataTypeTest.java
```diff
@@ -0,0 +1,441 @@
...
+import net.snowflake.ingest.streaming.SnowflakeStreamingIngestChannel;
+import net.snowflake.ingest.utils.ErrorCode;
+import net.snowflake.ingest.utils.SFException;
+import org.apache.commons.lang3.StringUtils;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Ignore;
+import org.junit.Test;
+
+@Ignore("This test can be enabled after server side Iceberg EP support is released")
```
sfc-gh-hmadan, line 25: Add a backlog JIRA to track adding a different github merge gate for iceberg ITs, we'll then have one merge gate for FDN ITs and another for iceberg ITs.

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergDataTypeTest.java
```diff
@@ -0,0 +1,441 @@
...
+  public void before() throws Exception {
+    super.before(true);
+  }
+
+  @Test
+  public void testBoolean() throws Exception {
+    testIcebergIngestion("boolean", true, new BooleanProvider());
+    testIcebergIngestion("boolean", false, new BooleanProvider());
+    testIcebergIngestion("boolean", 1, true, new BooleanProvider());
+    testIcebergIngestion("boolean", "false", false, new BooleanProvider());
```
sfc-gh-hmadan, line 40: lets also add 0, "true", "1", "0"

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergDataTypeTest.java
```diff
@@ -0,0 +1,441 @@
...
+  private final ObjectMapper objectMapper = new ObjectMapper();
+
+  @Before
+  public void before() throws Exception {
+    super.before(true);
+  }
+
+  @Test
+  public void testBoolean() throws Exception {
+    testIcebergIngestion("boolean", true, new BooleanProvider());
```
sfc-gh-hmadan, line 37: nit: whenever you have time in the next few weeks, lets introduce a class called ITTestcaseBuilder that has methods withDataType, withInputValue, withExpectedValue, withVerificationQuery.

THe provider business is funky as it only supports doing exact value verification with a `where value = {expectedValue}`. I'd much rather write the sql filter directly in the testcase instead of have code generate it, for flexibility.

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergDataTypeTest.java
```diff
@@ -0,0 +1,441 @@
...
+    testIcebergIngestion("boolean", false, new BooleanProvider());
+    testIcebergIngestion("boolean", 1, true, new BooleanProvider());
+    testIcebergIngestion("boolean", "false", false, new BooleanProvider());
+
+    SFException ex =
+        Assert.assertThrows(
```
sfc-gh-hmadan, line 43: pl use fluent assertions (assertj) as they (a) have test code that reads like a story, and (b) generate very well formatted error messages. Assert.assert*() has the most useless error messages when things fail unless every single call to the method has an associated string message that details the current state of whatever was being asserted on.

For instance:
assertThat(ex).isNotNull();
assertThat(ex.getVendorCode).isEqualTo(ErrorCode.INVALID_FORMAT_ROW.getMessageCode());
assertThatCode(() -> lambda).doesNotThrow();
assertThatThrownBy(() -> lambda).hasCause();

For this you'll have to add the assertj library (I added it once and it failed in one of the mergegates, did not have the time to find and fix what happened!). Feel free to remove the org.hamcrest import from this project.

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/AbstractDataTypeTest.java
```diff
@@ -112,6 +122,21 @@ protected String createTable(String dataType) throws SQLException {
             String.format(
                 "create or replace table %s (%s string, %s %s)",
                 tableName, SOURCE_COLUMN_NAME, VALUE_COLUMN_NAME, dataType));
+
+    return tableName;
+  }
+
+  protected String createIcebergTable(String dataType) throws SQLException {
```
sfc-gh-hmadan, line 129: also add an argument for `boolean nullable` and we can make sure the sdk does proper nullability checking for columns marked non-nullable.

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergDataTypeTest.java
```diff
@@ -0,0 +1,441 @@
...
+            SFException.class,
+            () -> testIcebergIngestion("boolean", new Object(), true, new BooleanProvider()));
+    Assert.assertEquals(ErrorCode.INVALID_FORMAT_ROW.getMessageCode(), ex.getVendorCode());
```
sfc-gh-hmadan, line 46: (dupe) lets add nullability tests too - sending in null should work for every column, but when its a non-null column then the sdk should throw.

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergDataTypeTest.java
```diff
@@ -0,0 +1,441 @@
...
+    testIcebergIngestion("int", 0.5f, 1, new IntProvider());
+    testIcebergIngestion("int", "100.4", 100, new IntProvider());
+    testIcebergIngestion("int", new BigDecimal("1000000.09"), 1000000, new IntProvider());
+    testIcebergIngestion("int", Integer.MAX_VALUE, new IntProvider());
+    testIcebergIngestion("int", Integer.MIN_VALUE, new IntProvider());
+
+    SFException ex =
+        Assert.assertThrows(
+            SFException.class,
+            () -> testIcebergIngestion("int", Long.MAX_VALUE, new LongProvider()));
```
sfc-gh-hmadan, line 62: nit: `(Long)Integer.MAX_VALUE + 1` instead of long.maxvalue?

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/AbstractDataTypeTest.java
sfc-gh-alhuang: 1. Set up different base_location for each tests.
2. [JIRA](https://snowflakecomputing.atlassian.net/browse/SNOW-1727545) created.

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergDataTypeTest.java
sfc-gh-alhuang: [jira](https://snowflakecomputing.atlassian.net/browse/SNOW-1727546) created.

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergDataTypeTest.java
sfc-gh-alhuang: Added.

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergDataTypeTest.java
sfc-gh-alhuang: Added to [jira](https://snowflakecomputing.atlassian.net/browse/SNOW-1727532)

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergDataTypeTest.java
sfc-gh-alhuang: Switched to assertj library.

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/AbstractDataTypeTest.java
sfc-gh-alhuang: Added nullable test.

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergDataTypeTest.java
sfc-gh-alhuang: Modified.

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergDataTypeTest.java
sfc-gh-alhuang: Added nullable test for non structured data type. Will fix structured data type null count in next PR.

src/main/java/net/snowflake/ingest/streaming/internal/RowBufferStats.java
```diff
@@ -37,19 +38,58 @@ class RowBufferStats {
   private final String collationDefinitionString;
   /** Display name is required for the registration endpoint */
   private final String columnDisplayName;
+  /** Whether the column is an Iceberg column */
+  private final boolean isIcebergColumn;
+  /** Primitive type of the column, only used for Iceberg columns */
+  private final PrimitiveType primitiveType;
```
sfc-gh-hmadan, line 44: pl update the PR description to explain there's a bug fix in this PR too.

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -649,7 +649,7 @@ static EpInfo buildEpInfoFromStats(
     EpInfo epInfo = new EpInfo(rowCount, new HashMap<>());
```
sfc-gh-hmadan, line 649: pl cleanup the setDefaultValues from the whole callstack and from the parameter provider.

sfc-gh-hmadan: ignore this, the stacked PR I'm sending actually uses this param.

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergDataTypeIT.java
```diff
@@ -0,0 +1,716 @@
...
+import net.snowflake.ingest.utils.ErrorCode;
+import net.snowflake.ingest.utils.SFException;
+import org.apache.commons.lang3.StringUtils;
+import org.assertj.core.api.Assertions;
+import org.junit.Before;
+import org.junit.Ignore;
+import org.junit.Test;
+
+@Ignore("This test can be enabled after server side Iceberg EP support is released")
+public class IcebergDataTypeIT extends AbstractDataTypeTest {
```
sfc-gh-hmadan, line 27: why not split this into multiple classes?

src/main/java/net/snowflake/ingest/streaming/internal/RowBufferStats.java
```diff
@@ -37,19 +38,58 @@ class RowBufferStats {
...
+        null /* primitiveType */);
+  }
+
+  RowBufferStats(
+      String columnDisplayName,
+      String collationDefinitionString,
+      int ordinal,
+      Integer fieldId,
+      boolean isIcebergColumn,
+      PrimitiveType primitiveType) {
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Thanks for the fix!

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergDataTypeIT.java
sfc-gh-alhuang: Split into multiple test classes as FDN data type IT did.

sfc-gh-alhuang:
@sfc-gh-hmadan Confirmed that the string min/max values for Iceberg table EP info also apply the 32 bytes round down/up truncated as FDN did.

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/AbstractDataTypeTest.java
sfc-gh-hmadan: I see nullable was removed, why so?

sfc-gh-alhuang: Removed and use `<datatype> not null` in datatype testing. By doing this we can set nullable on any sub-columns on structured data type.

PR #851, opened 2024-10-03
SNOW-1708577 Parquet V2 support for new table format

sfc-gh-alhuang:
This PR aims to ensure consistency with the Iceberg table scanner and registration on the server side. It includes the following changes:

1. Added support for Parquet V2 (delta encoding) when the schema's `STORAGE_SERIALIZATION_POLICY` is set to `OPTIMIZED.`
2. Enabled dictionary encoding for Iceberg mode.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
@@ -25,6 +25,7 @@ import net.snowflake.ingest.utils.ErrorCode;
 import net.snowflake.ingest.utils.Logging;
 import net.snowflake.ingest.utils.Pair;
 import net.snowflake.ingest.utils.SFException;
+import org.apache.parquet.column.ParquetProperties;
 
 /**
  * The abstract implementation of the buffer in the Streaming Ingest channel that holds the
@@ -668,6 +669,7 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
       ChannelRuntimeState channelRuntimeState,
       ClientBufferParameters clientBufferParameters,
       OffsetTokenVerificationFunction offsetTokenVerificationFunction,
+      ParquetProperties.WriterVersion parquetWriterVersion,
       TelemetryService telemetryService) {
     switch (bdecVersion) {
       case THREE:
@@ -681,6 +683,7 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
                 channelRuntimeState,
                 clientBufferParameters,
                 offsetTokenVerificationFunction,
+                parquetWriterVersion,
                 telemetryService);
       default:
         throw new SFException(
--- a/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
@@ -32,6 +32,7 @@ import net.snowflake.ingest.utils.Cryptor;
 import net.snowflake.ingest.utils.Logging;
 import net.snowflake.ingest.utils.Pair;
 import org.apache.commons.codec.binary.Hex;
+import org.apache.parquet.hadoop.ParquetFileWriter;
 
 /**
  * Build a single blob file that contains file header plus data. The header will be a
@@ -68,8 +69,12 @@ class BlobBuilder {
       List<List<ChannelData<T>>> blobData,
       Constants.BdecVersion bdecVersion,
       InternalParameterProvider internalParameterProvider)
-      throws IOException, NoSuchPaddingException, NoSuchAlgorithmException,
-          InvalidAlgorithmParameterException, InvalidKeyException, IllegalBlockSizeException,
+      throws IOException,
+          NoSuchPaddingException,
+          NoSuchAlgorithmException,
+          InvalidAlgorithmParameterException,
+          InvalidKeyException,
+          IllegalBlockSizeException,
           BadPaddingException {
     List<ChunkMetadata> chunksMetadataList = new ArrayList<>();
     List<byte[]> chunksDataList = new ArrayList<>();
@@ -140,7 +145,7 @@ class BlobBuilder {
 
         if (internalParameterProvider.setIcebergSpecificFieldsInEp()) {
           chunkMetadataBuilder
-              .setMajorVersion(Constants.PARQUET_MAJOR_VERSION)
+              .setMajorVersion(ParquetFileWriter.CURRENT_VERSION)
               .setMinorVersion(Constants.PARQUET_MINOR_VERSION)
               // set createdOn in seconds
               .setCreatedOn(System.currentTimeMillis() / 1000)
--- a/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
@@ -69,12 +69,8 @@ class BlobBuilder {
       List<List<ChannelData<T>>> blobData,
       Constants.BdecVersion bdecVersion,
       InternalParameterProvider internalParameterProvider)
-      throws IOException,
-          NoSuchPaddingException,
-          NoSuchAlgorithmException,
-          InvalidAlgorithmParameterException,
-          InvalidKeyException,
-          IllegalBlockSizeException,
+      throws IOException, NoSuchPaddingException, NoSuchAlgorithmException,
+          InvalidAlgorithmParameterException, InvalidKeyException, IllegalBlockSizeException,
           BadPaddingException {
     List<ChunkMetadata> chunksMetadataList = new ArrayList<>();
     List<byte[]> chunksDataList = new ArrayList<>();
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ClientBufferParameters.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ClientBufferParameters.java
@@ -47,9 +47,7 @@ public class ClientBufferParameters {
     this.isIcebergMode = isIcebergMode;
   }
 
-  /**
-   * @param clientInternal reference to the client object where the relevant parameters are set
-   */
+  /** @param clientInternal reference to the client object where the relevant parameters are set */
   public ClientBufferParameters(SnowflakeStreamingIngestClientInternal clientInternal) {
     this.maxChunkSizeInBytes =
         clientInternal != null
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ClientBufferParameters.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ClientBufferParameters.java
@@ -47,7 +47,9 @@ public class ClientBufferParameters {
     this.isIcebergMode = isIcebergMode;
   }
 
-  /** @param clientInternal reference to the client object where the relevant parameters are set */
+  /**
+   * @param clientInternal reference to the client object where the relevant parameters are set
+   */
   public ClientBufferParameters(SnowflakeStreamingIngestClientInternal clientInternal) {
     this.maxChunkSizeInBytes =
         clientInternal != null
@@ -123,5 +125,9 @@ public class ClientBufferParameters {
 
   public String getParquetMessageTypeName() {
     return isIcebergMode ? PARQUET_MESSAGE_TYPE_NAME : BDEC_PARQUET_MESSAGE_TYPE_NAME;
+  }
+
+  public boolean isEnableDictionaryEncoding() {
+    return isIcebergMode;
   }
 }
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
@@ -394,6 +394,7 @@ public class ParquetRowBuffer extends AbstractRowBuffer<ParquetChunkData> {
         schema,
         clientBufferParameters.getMaxChunkSizeInBytes(),
         clientBufferParameters.getMaxRowGroups(),
-        clientBufferParameters.getBdecParquetCompression());
+        clientBufferParameters.getBdecParquetCompression(),
+        clientBufferParameters.isEnableDictionaryEncoding());
   }
 }
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
@@ -26,6 +26,7 @@ import net.snowflake.ingest.streaming.OpenChannelRequest;
 import net.snowflake.ingest.utils.ErrorCode;
 import net.snowflake.ingest.utils.SFException;
 import org.apache.parquet.column.ColumnDescriptor;
+import org.apache.parquet.column.ParquetProperties;
 import org.apache.parquet.schema.MessageType;
 import org.apache.parquet.schema.PrimitiveType;
 import org.apache.parquet.schema.Type;
@@ -45,6 +46,8 @@ public class ParquetRowBuffer extends AbstractRowBuffer<ParquetChunkData> {
   private final List<List<Object>> data;
   private final List<List<Object>> tempData;
 
+  private final ParquetProperties.WriterVersion parquetWriterVersion;
+
   private MessageType schema;
 
   /** Construct a ParquetRowBuffer object. */
@@ -56,6 +59,7 @@ public class ParquetRowBuffer extends AbstractRowBuffer<ParquetChunkData> {
       ChannelRuntimeState channelRuntimeState,
       ClientBufferParameters clientBufferParameters,
       OffsetTokenVerificationFunction offsetTokenVerificationFunction,
+      ParquetProperties.WriterVersion parquetWriterVersion,
       TelemetryService telemetryService) {
     super(
         onErrorOption,
@@ -70,6 +74,7 @@ public class ParquetRowBuffer extends AbstractRowBuffer<ParquetChunkData> {
     this.metadata = new HashMap<>();
     this.data = new ArrayList<>();
     this.tempData = new ArrayList<>();
+    this.parquetWriterVersion = parquetWriterVersion;
   }
 
   /**
@@ -395,6 +400,7 @@ public class ParquetRowBuffer extends AbstractRowBuffer<ParquetChunkData> {
         clientBufferParameters.getMaxChunkSizeInBytes(),
         clientBufferParameters.getMaxRowGroups(),
         clientBufferParameters.getBdecParquetCompression(),
+        parquetWriterVersion,
         clientBufferParameters.isEnableDictionaryEncoding());
   }
 }
--- a/src/main/java/net/snowflake/ingest/utils/Constants.java
+++ b/src/main/java/net/snowflake/ingest/utils/Constants.java
@@ -79,28 +79,11 @@ public class Constants {
    * otherwise v1.
    */
   public enum IcebergSerializationPolicy {
-    NON_ICEBERG,
     COMPATIBLE,
     OPTIMIZED;
 
-    public static IcebergSerializationPolicy fromName(String name) {
-      if (name == null) {
-        return NON_ICEBERG;
-      }
-      for (IcebergSerializationPolicy e : IcebergSerializationPolicy.values()) {
-        if (e.name().equalsIgnoreCase(name)) {
-          return e;
-        }
-      }
-      throw new IllegalArgumentException(
-          String.format(
-              "Unsupported ICEBERG_SERIALIZATION_POLICY = '%s', allowed values are %s",
-              name, Arrays.asList(IcebergSerializationPolicy.values())));
-    }
-
-    public ParquetProperties.WriterVersion getParquetWriterVersion() {
+    public ParquetProperties.WriterVersion toParquetWriterVersion() {
       switch (this) {
-        case NON_ICEBERG:
         case COMPATIBLE:
           return ParquetProperties.WriterVersion.PARQUET_1_0;
         case OPTIMIZED:
--- a/src/main/java/net/snowflake/ingest/utils/Constants.java
+++ b/src/main/java/net/snowflake/ingest/utils/Constants.java
@@ -5,6 +5,7 @@
 package net.snowflake.ingest.utils;
 
 import java.util.Arrays;
+import org.apache.parquet.column.ParquetProperties;
 import org.apache.parquet.hadoop.metadata.CompressionCodecName;
 
 /** Contains all the constants needed for Streaming Ingest */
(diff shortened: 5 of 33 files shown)


src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
```diff
@@ -45,6 +45,9 @@ public class BdecParquetWriter implements AutoCloseable {
   private final Optional<Integer> maxRowGroups;
   private long rowsWritten = 0;
```
sfc-gh-hmadan, line 47: nit: add the final ones above the non-final ones!

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -89,13 +90,21 @@ static <T> Blob constructBlobAndMetadata(
...
 
         if (internalParameterProvider.getEnableChunkEncryption()) {
           Pair<byte[], Integer> paddedChunk =
               padChunk(serializedChunk.chunkData, Constants.ENCRYPTION_ALGORITHM_BLOCK_SIZE_BYTES);
           byte[] paddedChunkData = paddedChunk.getFirst();
           chunkLength = paddedChunk.getSecond();
 
+          if (internalParameterProvider.getComputeExtendedMetadataSize()) {
+            extendedMetadataSize =
+                Utils.getLittleEndianInt(
```
sfc-gh-hmadan, line 103: Use the BytesUtils methods from the parquet library instead?

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -89,13 +90,21 @@ static <T> Blob constructBlobAndMetadata(
...
+        int extendedMetadataSize = -1;
 
         if (internalParameterProvider.getEnableChunkEncryption()) {
           Pair<byte[], Integer> paddedChunk =
               padChunk(serializedChunk.chunkData, Constants.ENCRYPTION_ALGORITHM_BLOCK_SIZE_BYTES);
           byte[] paddedChunkData = paddedChunk.getFirst();
           chunkLength = paddedChunk.getSecond();
 
+          if (internalParameterProvider.getComputeExtendedMetadataSize()) {
+            extendedMetadataSize =
```
sfc-gh-hmadan, line 102: this is risky in that (a) the behavior of paddedChunkData can change and break the assumption here of no-padding-at-the-tail, (b) no guaranteed way of avoiding a corrupt value from flowing out.

1. lets extract extended metadata size before padding. 
2. verify that the last four bytes are == ParquetFileWriter.MAGIC
3. Is there nothing on ParquetFileWriter that exposes the extendedMetadataSize?

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -89,13 +90,21 @@ static <T> Blob constructBlobAndMetadata(
         final byte[] compressedChunkData;
         final int chunkLength;
         final int compressedChunkDataSize;
+        int extendedMetadataSize = -1;
 
         if (internalParameterProvider.getEnableChunkEncryption()) {
```
sfc-gh-hmadan, line 95: we should do the extendedMetadataSize extraction after line 107 and make it conditional on internalParameterProvider.setIcebergSpecificFieldsInEp

src/main/java/net/snowflake/ingest/utils/Constants.java
```diff
@@ -74,6 +74,9 @@ public class Constants {
   public static final int PARQUET_MAJOR_VERSION = 1;
   public static final int PARQUET_MINOR_VERSION = 0;
 
+  // Length of the magic bytes "PAR1" or "PARE" in Parquet file
```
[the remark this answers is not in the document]
sfc-gh-hmadan: lets use the constant from inside the parquet library instead of from here. Checkout ParquetFileWriter.MAGIC / MAGIC_STR.

src/main/java/net/snowflake/ingest/utils/Constants.java
```diff
@@ -74,6 +74,9 @@ public class Constants {
   public static final int PARQUET_MAJOR_VERSION = 1;
```
sfc-gh-hmadan, line 74: Remove this hardcoded version and read the actual version from one of the parquet writers / parquet config classes ?

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -47,6 +48,10 @@ public class ParameterProvider {
   public static final String ENABLE_NEW_JSON_PARSING_LOGIC =
       "ENABLE_NEW_JSON_PARSING_LOGIC".toLowerCase();
 
+  public static final String PARQUET_WRITER_VERSION = "PARQUET_WRITER_VERSION".toLowerCase();
+  public static final String ENABLE_PARQUET_DICTIONARY_ENCODING =
+      "ENABLE_PARQUET_DICTIONARY_ENCODING".toLowerCase();
```
sfc-gh-hmadan, line 53: leaving it here so we don't forget it : PARQUET_WRITER_VERSION will be dynamic depending on the openchannelresponse

Slso don't make ENABLE_PARQUET_DICTIONARY_ENCODING into a public member. We don't want customers to be exposed to it just yet.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -253,6 +263,16 @@ private void setParameterMap(
...
 
+    this.checkAndUpdate(
+        PARQUET_WRITER_VERSION, PARQUET_WRITER_VERSION_DEFAULT, parameterOverrides, props, false);
+
+    this.checkAndUpdate(
+        ENABLE_PARQUET_DICTIONARY_ENCODING,
+        ENABLE_PARQUET_DICTIONARY_ENCODING_DEFAULT,
+        parameterOverrides,
+        props,
+        false);
```
sfc-gh-hmadan, line 274: nit: add inline comment for argument name

src/main/java/net/snowflake/ingest/utils/Utils.java
```diff
@@ -430,4 +430,18 @@ public static String concatDotPath(String... path) {
...
   }
+
+  /**
+   * Get the little endian int from a byte array with offset
+   *
+   * @param bytes the byte array
+   * @param offset the offset
+   * @return the little endian int
+   */
+  public static int getLittleEndianInt(byte[] bytes, int offset) {
```
sfc-gh-hmadan, line 441: (dupe) remove as we'll use parquet library's BytesUtils

src/main/java/net/snowflake/ingest/streaming/internal/InternalParameterProvider.java
```diff
@@ -31,4 +31,9 @@ boolean setIcebergSpecificFieldsInEp() {
     // in the EP metadata, createdOn, and extendedMetadataSize.
     return isIcebergMode;
   }
+
+  boolean getComputeExtendedMetadataSize() {
```
sfc-gh-hmadan, line 35: no need for a new method here, can reuse setIcebergSpecificFieldsInEp

src/main/java/net/snowflake/ingest/streaming/internal/EpInfo.java
```diff
@@ -22,7 +22,7 @@ class EpInfo {
   }
 
   /** Some basic verification logic to make sure the EP info is correct */
-  public void verifyEpInfo() {
+  public void verifyEpInfo(boolean enableDistinctValues) {
```
sfc-gh-hmadan, line 25: lets add this boolean into the ctor signature instead, as its a property of the columnEps that are already saved to this object's state.

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -647,13 +647,15 @@ public synchronized void close(String name) {
   static EpInfo buildEpInfoFromStats(
       long rowCount, Map<String, RowBufferStats> colStats, boolean setDefaultValues) {
     EpInfo epInfo = new EpInfo(rowCount, new HashMap<>());
+    boolean enableDistinctValues = false;
     for (Map.Entry<String, RowBufferStats> colStat : colStats.entrySet()) {
       RowBufferStats stat = colStat.getValue();
+      enableDistinctValues = stat.isEnableDistinctValue();
```
sfc-gh-hmadan, line 653: this looks odd to me. We are inferring an outer-class characteristic from an inner-object's state.
Lets take in a boolean alongside setDefaultValues

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -102,7 +102,11 @@ public void setupSchema(List<ColumnMetadata> columns) {
         this.statsMap.put(
             column.getInternalName(),
             new RowBufferStats(
-                column.getName(), column.getCollation(), column.getOrdinal(), null /* fieldId */));
+                column.getName(),
+                column.getCollation(),
+                column.getOrdinal(),
+                null /* fieldId */,
+                clientBufferParameters.getIsIcebergMode()));
```
sfc-gh-hmadan, line 109: we are in !getIsICebergMode (see line 100 above), so can just hardcode false here.

sfc-gh-hmadan: also applies to line 124.

src/main/java/net/snowflake/ingest/streaming/internal/RowBufferStats.java
```diff
@@ -38,18 +41,29 @@ class RowBufferStats {
...
+      int ordinal,
+      Integer fieldId,
+      boolean enableDistinctValue) {
     this.columnDisplayName = columnDisplayName;
     this.collationDefinitionString = collationDefinitionString;
     this.ordinal = ordinal;
     this.fieldId = fieldId;
+    this.enableDistinctValue = enableDistinctValue;
+    if (enableDistinctValue) {
+      this.distinctValues = new HashSet<>();
```
sfc-gh-hmadan, line 60: is there any concurrent access to this collection?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -185,14 +190,23 @@ public void setupSchema(List<ColumnMetadata> columns) {
 
         this.statsMap.put(
             columnPath,
-            new RowBufferStats(columnPath, null /* collationDefinitionString */, ordinal, fieldId));
+            new RowBufferStats(
+                columnPath,
+                null /* collationDefinitionString */,
+                ordinal,
+                fieldId,
+                clientBufferParameters.getIsIcebergMode()));
```
sfc-gh-hmadan, line 198: lets add a clientBufferParameters.getEnableDistinctValues() instead of getIsIcebergMode.

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
sfc-gh-hmadan: infact you can read this.clientBufferParameters.enableDistinctValues and don't even need an argument here.

src/main/java/net/snowflake/ingest/streaming/internal/RowBufferStats.java
```diff
@@ -38,18 +41,29 @@ class RowBufferStats {
...
+    }
     reset();
   }
 
   RowBufferStats(String columnDisplayName) {
```
sfc-gh-hmadan, line 65: this ctor is only called by test code. lets add isEnableDistinctValue to the signature, and make all the test classes run with both true and false.

src/main/java/net/snowflake/ingest/streaming/internal/RowBufferStats.java
```diff
@@ -82,12 +97,22 @@ static RowBufferStats getCombinedStats(RowBufferStats left, RowBufferStats right
               "left=%s, right=%s",
               left.getCollationDefinitionString(), right.getCollationDefinitionString()));
     }
+
```
sfc-gh-hmadan, line 100: nice!

src/main/java/net/snowflake/ingest/streaming/internal/RowBufferStats.java
```diff
@@ -131,6 +161,9 @@ void addStrValue(String value) {
 
   void addBinaryValue(byte[] valueBytes) {
     this.setCurrentMaxLength(valueBytes.length);
+    if (enableDistinctValue) {
+      distinctValues.add(new String(Base64.getEncoder().encode(valueBytes)));
```
sfc-gh-hmadan, line 165: ouch this will kill performance - both cpu and memory. If I'm not wrong we are already doing heavy validation of input strings, I'm hoping we can reuse that single pass on the whole string.

Another idea is to keep a `Map<String, Set<Int>>` here for string columns, the map key can be a CRC32 hash and the map value will be a list of row ids that contain a value with that hash. Only in the case where the Set has size > 1 do we look at the actual strings, before serialization. This is non-trivial amount of changes all over though.

src/main/java/net/snowflake/ingest/utils/Constants.java
```diff
@@ -71,9 +72,48 @@ public class Constants {
...
+          String.format(
+              "Unsupported ICEBERG_SERIALIZATION_POLICY = '%s', allowed values are %s",
+              name, Arrays.asList(IcebergSerializationPolicy.values())));
+    }
+
+    public ParquetProperties.WriterVersion getParquetWriterVersion() {
+      switch (this) {
+        case NON_ICEBERG:
+        case COMPATIBLE:
+          return ParquetProperties.WriterVersion.PARQUET_1_0;
```
rockwotj, line 105: Does this mean that non-iceberg tables (which are snowflake managed tables AFAIK) only support parquet v1?

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
sfc-gh-alhuang: Moved to top

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
sfc-gh-alhuang: Changed to `ByteUtils` instead of own func.

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
sfc-gh-alhuang: 1. The `chunkLength` returned by `padChunk` is the length before padding, the offset should be correct.
2. Added check in `Utils.getExtendedMetadataSize`.
3. Afaik the parquet file writer only log this value ([ref](https://github.com/apache/parquet-java/blob/eed26b4ea96881cc543b4233cdddece5d1d608d7/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java#L1918))

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
sfc-gh-alhuang: Moved to `!getEnableChunkEncrpytion() == false` scope and guarded with `setIcebergSpecificFieldsInEp()`

src/main/java/net/snowflake/ingest/utils/Constants.java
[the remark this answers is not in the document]
sfc-gh-alhuang: Yes we only do unencrypted footer which is `PAR1`. Removed this constant and use `ParuquetFileWriter.MAGIC_STR`

src/main/java/net/snowflake/ingest/utils/Constants.java
sfc-gh-alhuang: Use major version from `ParquetFileWirter.CURRENT_VERSION`. Keep minor version as I cannot find one in `ParquetFileWriter`

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-alhuang: Changed to dynamic parquet writer version. Removed `ENABLE_PARQUET_DICTIONARY_ENCODING` , always enable dict encoding in iceberg mode.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-alhuang: Added.

src/main/java/net/snowflake/ingest/utils/Utils.java
sfc-gh-alhuang: Resolved.

src/main/java/net/snowflake/ingest/streaming/internal/InternalParameterProvider.java
sfc-gh-alhuang: Removed.

src/main/java/net/snowflake/ingest/streaming/internal/EpInfo.java
sfc-gh-alhuang: Moved to ctor.

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
sfc-gh-alhuang: Moved ndv flag to parameter. Cannot use `this.clientBufferParameters` as this is a static method called by chained of static methods.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: Switch to `getEnableDistinctValues()`

src/main/java/net/snowflake/ingest/streaming/internal/RowBufferStats.java
sfc-gh-alhuang: The value parser parse every row in a chunk in order. There's no concurrent access within the same `RowBufferStats`

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: Added.

src/main/java/net/snowflake/ingest/streaming/internal/RowBufferStats.java
sfc-gh-alhuang: Added to signature and create parameterized tests.

src/main/java/net/snowflake/ingest/streaming/internal/RowBufferStats.java
sfc-gh-alhuang: Discussed offline. Lets keep NDV disabled by default. When it's enabled, skip the string ndv computation. Added a [JIRA](https://snowflakecomputing.atlassian.net/browse/SNOW-1721193) for this.

src/main/java/net/snowflake/ingest/utils/Constants.java
sfc-gh-alhuang: The server-side scanner for FDN tables supports Parquet V2. This PR is specific to the Iceberg table feature and does not alter the default behavior for streaming to FDN tables.

rockwotj: Thanks!

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -140,7 +141,7 @@ static <T> Blob constructBlobAndMetadata(
 
         if (internalParameterProvider.setIcebergSpecificFieldsInEp()) {
           chunkMetadataBuilder
-              .setMajorVersion(Constants.PARQUET_MAJOR_VERSION)
+              .setMajorVersion(ParquetFileWriter.CURRENT_VERSION)
```
sfc-gh-hmadan, line 144: I'd much rather depend on our own constant than a third-party library's constant, I thought I had left a comment on this but don't see it anywhere :(

sfc-gh-hmadan: ok to take in next PR too, just remove the import whenever you revert this.

src/main/java/net/snowflake/ingest/streaming/internal/ClientBufferParameters.java
```diff
@@ -124,4 +124,8 @@ public Optional<Integer> getMaxRowGroups() {
   public String getParquetMessageTypeName() {
     return isIcebergMode ? PARQUET_MESSAGE_TYPE_NAME : BDEC_PARQUET_MESSAGE_TYPE_NAME;
   }
+
+  public boolean isEnableDictionaryEncoding() {
+    return isIcebergMode;
```
sfc-gh-hmadan, line 129: this *might* be dependent on storage serialization policy too, lets verify. no need to hold up PR.

src/main/java/org/apache/parquet/hadoop/BdecParquetWriter.java
```diff
@@ -43,6 +43,10 @@ public class BdecParquetWriter implements AutoCloseable {
   // Optional cap on the max number of row groups to allow per file, if this is exceeded we'll end
```
sfc-gh-hmadan, line 43: nit: lets rename this file to SnowflakeParquetWriter instead of BdecParquetWriter?!

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/AbstractDataTypeTest.java
```diff
@@ -66,19 +66,52 @@ public abstract class AbstractDataTypeTest {
...
+      String compressionAlgorithm,
+      Constants.IcebergSerializationPolicy serializationPolicy)
+      throws Exception {
     databaseName = String.format("SDK_DATATYPE_COMPATIBILITY_IT_%s", getRandomIdentifier());
     conn = TestUtils.getConnection(true);
     conn.createStatement().execute(String.format("create or replace database %s;", databaseName));
     conn.createStatement().execute(String.format("use database %s;", databaseName));
     conn.createStatement().execute(String.format("use schema %s;", schemaName));
 
+    switch (serializationPolicy) {
```
sfc-gh-hmadan, line 99: nit: if (iceberg) ?

sfc-gh-hmadan (approved):
LGTM, left some minor nits/etc that can be taken with the next PR you have in flight. Lets merge this in to unblock that PR and other PRs that need to go in.

PR #852, opened 2024-10-03
NO-SNOW Remove unnecessary dependencies to reduce jar size

sfc-gh-japatel:
we are not using netty except the netty common, so remove all that I added in for license.


sfc-gh-tzhang (approved):
Looks better now, thanks!

sfc-gh-xhuang (approved):

PR #854, opened 2024-10-04
Fix concurrent null-ref in External Volume Blob path Generation + track URL timeout

sfc-gh-hmadan:
1. Only use a URL if its more than 1 minute away from expiry
2. Handle a nullref in ExternalVolume.generateUrls
3. In case of high concurrent load on the url generation semaphore, only wait to acquire when the queue is null. If this was an optimistic URL fetch there is no need to wait around or the semaphore.
4. fix comment line wrapping.
5. add concurrent test for the generateUrl changes
6. TODO add test for the "dont use urls that are about to expire in the next minute" logic


sfc-gh-alhuang (approved):
Looks good, thanks!

PR #855, opened 2024-10-04
Fix testUtils to not repeatedly do keypair generation

sfc-gh-hmadan:
TestUtils has four unexpected behaviors:
1. When there is no profile.json available, we do keypair generation many many times in every testcase instead of doing it once. Fixed this.
2. The no-profile-json code path ends up hitting snowflake.qa1.int.snowflakecomputing.com from unit tests (should be hitting a mock)
3. If a profile.json is being used for local testing, all unit tests start using it too because TestUtils ends up reading the same profile.json. This is highly unexpected.
4. Profile.json parsing behavior between prod and test codepaths is different. the former is okay with just the URL being specified, the latter forces specification of scheme, host, port, ssl.

Fixing (1) with this PR.

### DIFF
--- a/src/test/java/net/snowflake/ingest/TestUtils.java
+++ b/src/test/java/net/snowflake/ingest/TestUtils.java
@@ -58,7 +58,7 @@ public class TestUtils {
 
   private static final ObjectMapper mapper = new ObjectMapper();
 
-  private static ObjectNode profile = null;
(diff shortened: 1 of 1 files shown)


sfc-gh-alhuang (approved):
Thanks for the fix!

PR #856, opened 2024-10-04
Add (non-public) method to expose iceberg schema for each column of a channel

sfc-gh-hmadan:
For schema evolution of structured data types, Kafka Connector needs a way to discover the schema of the column that had a validation error.
We already have an API on StreamingIngestChannel called getTableSchema() that returns a Map<String, ColumnProperties>.
Am adding a method (currently non-public) to the ColumnProperties type called getIcebergSchema() that returns a spec-compliant JSON of the column's iceberg schema.

We'll make this method public when we make setIsIceberg() public.


sfc-gh-alhuang (approved):
Looks good, thanks!

PR #857, opened 2024-10-05
SNOW-1711460 Exclude avro dependency to fix 853

sfc-gh-xhuang:
I think we don't need the avro dependency and can exclude it to resolve 
https://github.com/snowflakedb/snowflake-ingest-java/issues/853

An alternative is upgrading to 1.6.1 iceberg-core, which has avro version 1.12.0 but if we confirm that we don't need it, it's best to exclude

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -287,6 +287,12 @@
         <groupId>org.apache.iceberg</groupId>
         <artifactId>iceberg-core</artifactId>
         <version>${iceberg.version}</version>
+        <exclusions>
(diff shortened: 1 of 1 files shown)


sfc-gh-asen (approved):

sfc-gh-alhuang (approved):
Thanks for fixing this

sfc-gh-azagrebin:
is this to address SNOW-1711460?
why is this not merged yet?

PR #861, opened 2024-10-11
SNOW-1727532 Set number of values for repeated fields

sfc-gh-alhuang:
This PR includes following change:
1. Set `numberOfValues` for repeated fields (e.g. map, list) and used it in EP file registration.
2. Increment null count for all sub-columns of a null column.
4. Use hash for approximate NDV for string and bytes.
5. Fix NPE when inserting null to a structured column, throwing SF custom exception instead.
6. Added `extendedMetadataSize` and `metadataSize` fields for EP registration.
7. Address comments of #851

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 Snowflake Computing Inc. All rights reserved.
+ * Copyright (c) 2022-2024 Snowflake Computing Inc. All rights reserved.
  */
 
 package net.snowflake.ingest.streaming.internal;
@@ -642,13 +642,17 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
    *
    * @param rowCount: count of rows in the given buffer
    * @param colStats: map of column name to RowBufferStats
-   * @param setAllDefaultValues: whether to set default values for all null fields the EPs
-   *     irrespective of the data type of this column
+   * @param setAllDefaultValues: whether to set default values for all null min/max field in the EPs
+   * @param enableDistinctValuesCount: whether to include valid NDV in the EPs irrespective of the
+   *     data type of this column
    * @return the EPs built from column stats
    */
   static EpInfo buildEpInfoFromStats(
-      long rowCount, Map<String, RowBufferStats> colStats, boolean setAllDefaultValues) {
-    EpInfo epInfo = new EpInfo(rowCount, new HashMap<>());
+      long rowCount,
+      Map<String, RowBufferStats> colStats,
+      boolean setAllDefaultValues,
+      boolean enableDistinctValuesCount) {
+    EpInfo epInfo = new EpInfo(rowCount, new HashMap<>(), enableDistinctValuesCount);
     for (Map.Entry<String, RowBufferStats> colStat : colStats.entrySet()) {
       RowBufferStats stat = colStat.getValue();
       FileColumnProperties dto = new FileColumnProperties(stat, setAllDefaultValues);
--- a/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
@@ -292,6 +292,9 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
   // Temp stats map to use until all the rows are validated
   @VisibleForTesting Map<String, RowBufferStats> tempStatsMap;
 
+  // Map of the column name to the column object, used for null/missing column check
+  protected final Map<String, ParquetColumn> fieldIndex;
+
   // Lock used to protect the buffers from concurrent read/write
   private final Lock flushLock;
 
@@ -352,6 +355,8 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
     // Initialize empty stats
     this.statsMap = new HashMap<>();
     this.tempStatsMap = new HashMap<>();
+
+    this.fieldIndex = new HashMap<>();
   }
 
   /**
@@ -427,7 +432,7 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
     List<String> missingCols = new ArrayList<>();
     for (String columnName : this.nonNullableFieldNames) {
       if (!inputColNamesMap.containsKey(columnName)) {
-        missingCols.add(statsMap.get(columnName).getColumnDisplayName());
+        missingCols.add(fieldIndex.get(columnName).columnMetadata.getName());
       }
     }
 
@@ -447,7 +452,7 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
     for (String columnName : this.nonNullableFieldNames) {
       if (inputColNamesMap.containsKey(columnName)
           && row.get(inputColNamesMap.get(columnName)) == null) {
-        nullValueNotNullCols.add(statsMap.get(columnName).getColumnDisplayName());
+        nullValueNotNullCols.add(fieldIndex.get(columnName).columnMetadata.getName());
       }
     }
 
--- a/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
@@ -30,8 +30,10 @@ import javax.crypto.IllegalBlockSizeException;
 import javax.crypto.NoSuchPaddingException;
 import net.snowflake.ingest.utils.Constants;
 import net.snowflake.ingest.utils.Cryptor;
+import net.snowflake.ingest.utils.ErrorCode;
 import net.snowflake.ingest.utils.Logging;
 import net.snowflake.ingest.utils.Pair;
+import net.snowflake.ingest.utils.SFException;
 import org.apache.commons.codec.binary.Hex;
 
 /**
@@ -141,6 +143,12 @@ class BlobBuilder {
                 .setLastInsertTimeInMs(serializedChunk.chunkMinMaxInsertTimeInMs.getSecond());
 
         if (internalParameterProvider.setIcebergSpecificFieldsInEp()) {
+          if (internalParameterProvider.getEnableChunkEncryption()) {
+            /* metadata size computation only works when encryption and padding is off */
+            throw new SFException(
+                ErrorCode.INTERNAL_ERROR,
+                "Metadata size computation is only supported when encryption is enabled");
+          }
           final long metadataSize = getParquetFooterSize(compressedChunkData);
           final long extendedMetadataSize = serializedChunk.extendedMetadataSize;
           chunkMetadataBuilder
--- a/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
@@ -11,6 +11,7 @@ import static net.snowflake.ingest.utils.Constants.BLOB_FILE_SIZE_SIZE_IN_BYTES;
 import static net.snowflake.ingest.utils.Constants.BLOB_NO_HEADER;
 import static net.snowflake.ingest.utils.Constants.BLOB_TAG_SIZE_IN_BYTES;
 import static net.snowflake.ingest.utils.Constants.BLOB_VERSION_SIZE_IN_BYTES;
+import static net.snowflake.ingest.utils.Utils.getParquetFooterSize;
 import static net.snowflake.ingest.utils.Utils.toByteArray;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
@@ -32,8 +33,6 @@ import net.snowflake.ingest.utils.Cryptor;
 import net.snowflake.ingest.utils.Logging;
 import net.snowflake.ingest.utils.Pair;
 import org.apache.commons.codec.binary.Hex;
-import org.apache.parquet.bytes.BytesUtils;
-import org.apache.parquet.hadoop.ParquetFileWriter;
 
 /**
  * Build a single blob file that contains file header plus data. The header will be a
@@ -91,8 +90,6 @@ class BlobBuilder {
         final byte[] compressedChunkData;
         final int chunkLength;
         final int compressedChunkDataSize;
-        long extendedMetadataSize = -1;
-        long metadataSize = -1;
 
         if (internalParameterProvider.getEnableChunkEncryption()) {
           Pair<byte[], Integer> paddedChunk =
@@ -114,11 +111,6 @@ class BlobBuilder {
           compressedChunkData = serializedChunk.chunkData.toByteArray();
           chunkLength = compressedChunkData.length;
           compressedChunkDataSize = chunkLength;
-
-          if (internalParameterProvider.setIcebergSpecificFieldsInEp()) {
-            metadataSize = getExtendedMetadataSize(compressedChunkData);
-            extendedMetadataSize = serializedChunk.extendedMetadataSize;
-          }
         }
 
         // Compute the md5 of the chunk data
@@ -149,6 +141,8 @@ class BlobBuilder {
                 .setLastInsertTimeInMs(serializedChunk.chunkMinMaxInsertTimeInMs.getSecond());
 
         if (internalParameterProvider.setIcebergSpecificFieldsInEp()) {
+          final long metadataSize = getParquetFooterSize(compressedChunkData);
+          final long extendedMetadataSize = serializedChunk.extendedMetadataSize;
           chunkMetadataBuilder
               .setMajorVersion(Constants.PARQUET_MAJOR_VERSION)
               .setMinorVersion(Constants.PARQUET_MINOR_VERSION)
@@ -300,23 +294,5 @@ class BlobBuilder {
       this.chunksMetadataList = chunksMetadataList;
       this.blobStats = blobStats;
     }
-  }
-
-  /**
-   * Get the metadata size (footer size) from a parquet file
-   *
-   * @param bytes the serialized parquet file
-   * @return the extended metadata size
-   */
-  static long getExtendedMetadataSize(byte[] bytes) throws IOException {
-    final int magicOffset = bytes.length - ParquetFileWriter.MAGIC.length;
-    final int footerSizeOffset = magicOffset - Integer.BYTES;
-    if (footerSizeOffset < 0
-        || !ParquetFileWriter.MAGIC_STR.equals(
-            new String(bytes, magicOffset, ParquetFileWriter.MAGIC.length))) {
-      throw new IllegalArgumentException("Invalid parquet file");
-    }
-
-    return BytesUtils.readIntLittleEndian(bytes, footerSizeOffset);
   }
 }
--- a/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
@@ -31,6 +31,7 @@ import net.snowflake.ingest.utils.Constants;
 import net.snowflake.ingest.utils.Cryptor;
 import net.snowflake.ingest.utils.Logging;
 import net.snowflake.ingest.utils.Pair;
+import net.snowflake.ingest.utils.Utils;
 import org.apache.commons.codec.binary.Hex;
 import org.apache.parquet.hadoop.ParquetFileWriter;
 
@@ -90,6 +91,7 @@ class BlobBuilder {
         final byte[] compressedChunkData;
         final int chunkLength;
         final int compressedChunkDataSize;
+        int extendedMetadataSize = -1;
 
         if (internalParameterProvider.getEnableChunkEncryption()) {
           Pair<byte[], Integer> paddedChunk =
@@ -111,6 +113,10 @@ class BlobBuilder {
           compressedChunkData = serializedChunk.chunkData.toByteArray();
           chunkLength = compressedChunkData.length;
           compressedChunkDataSize = chunkLength;
+
+          if (internalParameterProvider.setIcebergSpecificFieldsInEp()) {
+            extendedMetadataSize = Utils.getExtendedMetadataSize(compressedChunkData, chunkLength);
+          }
         }
 
         // Compute the md5 of the chunk data
@@ -135,7 +141,8 @@ class BlobBuilder {
                     AbstractRowBuffer.buildEpInfoFromStats(
                         serializedChunk.rowCount,
                         serializedChunk.columnEpStatsMapCombined,
-                        internalParameterProvider.setAllDefaultValuesInEp()))
+                        internalParameterProvider.setAllDefaultValuesInEp(),
+                        internalParameterProvider.isEnableDistinctValuesCount()))
                 .setFirstInsertTimeInMs(serializedChunk.chunkMinMaxInsertTimeInMs.getFirst())
                 .setLastInsertTimeInMs(serializedChunk.chunkMinMaxInsertTimeInMs.getSecond());
 
@@ -145,7 +152,7 @@ class BlobBuilder {
               .setMinorVersion(Constants.PARQUET_MINOR_VERSION)
               // set createdOn in seconds
               .setCreatedOn(System.currentTimeMillis() / 1000)
-              .setExtendedMetadataSize(-1L);
+              .setExtendedMetadataSize((long) extendedMetadataSize);
         }
 
         ChunkMetadata chunkMetadata = chunkMetadataBuilder.build();
--- a/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
@@ -31,8 +31,8 @@ import net.snowflake.ingest.utils.Constants;
 import net.snowflake.ingest.utils.Cryptor;
 import net.snowflake.ingest.utils.Logging;
 import net.snowflake.ingest.utils.Pair;
-import net.snowflake.ingest.utils.Utils;
 import org.apache.commons.codec.binary.Hex;
+import org.apache.parquet.bytes.BytesUtils;
 import org.apache.parquet.hadoop.ParquetFileWriter;
 
 /**
@@ -91,7 +91,8 @@ class BlobBuilder {
         final byte[] compressedChunkData;
         final int chunkLength;
         final int compressedChunkDataSize;
-        int extendedMetadataSize = -1;
+        long extendedMetadataSize = -1;
+        long metadataSize = -1;
 
         if (internalParameterProvider.getEnableChunkEncryption()) {
           Pair<byte[], Integer> paddedChunk =
@@ -115,7 +116,8 @@ class BlobBuilder {
           compressedChunkDataSize = chunkLength;
 
           if (internalParameterProvider.setIcebergSpecificFieldsInEp()) {
-            extendedMetadataSize = Utils.getExtendedMetadataSize(compressedChunkData, chunkLength);
+            metadataSize = getExtendedMetadataSize(compressedChunkData);
+            extendedMetadataSize = serializedChunk.extendedMetadataSize;
           }
         }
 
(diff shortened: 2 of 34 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/ClientBufferParameters.java
```diff
@@ -65,14 +73,22 @@ public ClientBufferParameters(SnowflakeStreamingIngestClientInternal clientInter
         clientInternal != null
             ? clientInternal.getParameterProvider().isEnableNewJsonParsingLogic()
             : ParameterProvider.ENABLE_NEW_JSON_PARSING_LOGIC_DEFAULT;
-    this.isIcebergMode =
-        clientInternal != null
-            ? clientInternal.isIcebergMode()
-            : ParameterProvider.IS_ICEBERG_MODE_DEFAULT;
     this.maxRowGroups =
         isIcebergMode
```
sfc-gh-hmadan, line 77: you're reading isIcebergMode before writing to it.

src/main/java/net/snowflake/ingest/streaming/internal/ClientBufferParameters.java
```diff
@@ -125,6 +145,14 @@ public String getParquetMessageTypeName() {
     return isIcebergMode ? PARQUET_MESSAGE_TYPE_NAME : BDEC_PARQUET_MESSAGE_TYPE_NAME;
   }
 
+  public boolean isEnableDistinctValuesCount() {
+    return enableDistinctValuesCount;
+  }
+
+  public boolean isEnableValuesCount() {
+    return enableDistinctValuesCount;
```
sfc-gh-hmadan, line 153: return enableValuesCount?

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
```diff
@@ -284,6 +288,16 @@ void setMaxStrNonCollated(String maxStrNonCollated) {
...
+  long getNumberOfValues() {
+    return numberOfValues;
+  }
+
+  void setNumberOfValues(long numberOfValues) {
+    this.numberOfValues = numberOfValues;
+  }
+
   @Override
   public String toString() {
```
sfc-gh-hmadan, line 302: also add numberOfValues to toString

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
```diff
@@ -284,6 +288,16 @@ void setMaxStrNonCollated(String maxStrNonCollated) {
     this.maxStrNonCollated = maxStrNonCollated;
   }
 
+  @JsonProperty("numberOfValues")
+  @JsonInclude(value = JsonInclude.Include.CUSTOM, valueFilter = IgnoreMinusOneFilter.class)
```
sfc-gh-hmadan, line 292: instead of the customer IgnoreMinusOne logic why not just have  @JsonInclude(JsonInclude.Include.NON_NULL), and let numberOfValues be null by default / null for FDN ingestion?

src/main/java/net/snowflake/ingest/streaming/internal/ClientBufferParameters.java
```diff
@@ -38,13 +42,17 @@ private ClientBufferParameters(
...
     this.maxAllowedRowSizeInBytes = maxAllowedRowSizeInBytes;
     this.bdecParquetCompression = bdecParquetCompression;
     this.enableNewJsonParsingLogic = enableNewJsonParsingLogic;
     this.maxRowGroups = maxRowGroups;
     this.isIcebergMode = isIcebergMode;
+    this.enableDistinctValuesCount = enableDistinctValuesCount;
+    this.enableValuesCount = enableValuesCount;
   }
 
   /** @param clientInternal reference to the client object where the relevant parameters are set */
```
sfc-gh-hmadan, line 58: I've never understood why clientInternal can or should be allowed to be null. Looks like a testcase author leaking their convenience needs into production code.

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -90,6 +91,8 @@ static <T> Blob constructBlobAndMetadata(
         final byte[] compressedChunkData;
         final int chunkLength;
         final int compressedChunkDataSize;
+        long extendedMetadataSize = -1;
+        long metadataSize = -1;
```
sfc-gh-hmadan, line 95: Add Constants.EP_EXTENDED_METDATA_SIZE_UNKNOWN and Constants.EP_METADATA_SIZE_UNKNOWN, just like we have it for NDV_UNKNOWN and NV_UNKNOWN

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -291,4 +301,22 @@ static class Blob {
...
+   * @param bytes the serialized parquet file
+   * @return the extended metadata size
+   */
+  static long getExtendedMetadataSize(byte[] bytes) throws IOException {
+    final int magicOffset = bytes.length - ParquetFileWriter.MAGIC.length;
+    final int footerSizeOffset = magicOffset - Integer.BYTES;
+    if (footerSizeOffset < 0
+        || !ParquetFileWriter.MAGIC_STR.equals(
+            new String(bytes, magicOffset, ParquetFileWriter.MAGIC.length))) {
+      throw new IllegalArgumentException("Invalid parquet file");
```
sfc-gh-hmadan, line 317: also add the various ingredients to the formulas here into the error message. bytes.length, MAGIC.length, footerSizeOffset, new String(bytes, magicOffset, ParquetFileWriter.MAGIC.length), ParquetFileWriter.MAGIC_STR.

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -291,4 +301,22 @@ static class Blob {
...
+   * Get the metadata size (footer size) from a parquet file
+   *
+   * @param bytes the serialized parquet file
+   * @return the extended metadata size
+   */
+  static long getExtendedMetadataSize(byte[] bytes) throws IOException {
+    final int magicOffset = bytes.length - ParquetFileWriter.MAGIC.length;
+    final int footerSizeOffset = magicOffset - Integer.BYTES;
+    if (footerSizeOffset < 0
+        || !ParquetFileWriter.MAGIC_STR.equals(
```
sfc-gh-hmadan, line 315: curious why is this MAGIC_STR comparison being added only now / why is it needed only when calculating ext metadta size?

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -111,6 +114,11 @@ static <T> Blob constructBlobAndMetadata(
           compressedChunkData = serializedChunk.chunkData.toByteArray();
           chunkLength = compressedChunkData.length;
           compressedChunkDataSize = chunkLength;
+
+          if (internalParameterProvider.setIcebergSpecificFieldsInEp()) {
```
sfc-gh-hmadan, line 118: this calculation can happen wholly inside the pre-existing if(setIcebergSpecificFieldsInEp() section at line 151?

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -111,6 +114,11 @@ static <T> Blob constructBlobAndMetadata(
...
+            metadataSize = getExtendedMetadataSize(compressedChunkData);
```
sfc-gh-hmadan, line 119: method name is adding to confusion, lets call it getParquetFooterSize. Also put it in a utils class and not in BlobBuilder.

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -427,7 +432,7 @@ Set<String> verifyInputColumns(
     List<String> missingCols = new ArrayList<>();
     for (String columnName : this.nonNullableFieldNames) {
       if (!inputColNamesMap.containsKey(columnName)) {
-        missingCols.add(statsMap.get(columnName).getColumnDisplayName());
+        missingCols.add(fieldIndex.get(columnName).columnMetadata.getName());
```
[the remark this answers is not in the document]
sfc-gh-hmadan: actually, is there a need to even make this change?

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -447,7 +452,7 @@ Set<String> verifyInputColumns(
     for (String columnName : this.nonNullableFieldNames) {
       if (inputColNamesMap.containsKey(columnName)
           && row.get(inputColNamesMap.get(columnName)) == null) {
-        nullValueNotNullCols.add(statsMap.get(columnName).getColumnDisplayName());
+        nullValueNotNullCols.add(fieldIndex.get(columnName).columnMetadata.getName());
```
sfc-gh-hmadan, line 455: same here, why change this?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -49,6 +48,7 @@ public class ParquetRowBuffer extends AbstractRowBuffer<ParquetChunkData> {
   private final ParquetProperties.WriterVersion parquetWriterVersion;
 
   private MessageType schema;
+  private SubColumnFinder statsFinder;
```
sfc-gh-hmadan, line 51: rename to subColumnFinder

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -113,7 +112,9 @@ public void setupSchema(List<ColumnMetadata> columns) {
                 column.getCollation(),
                 column.getOrdinal(),
                 null /* fieldId */,
-                parquetType.isPrimitive() ? parquetType.asPrimitiveType() : null));
+                parquetType.isPrimitive() ? parquetType.asPrimitiveType() : null,
+                clientBufferParameters.isEnableDistinctValuesCount(),
+                false /* enableValuesCount */));
```
sfc-gh-hmadan, line 117: why not use clientBufferParameters.isEnableValuesCount() here ?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -186,31 +189,46 @@ public void setupSchema(List<ColumnMetadata> columns) {
...
       for (ColumnDescriptor columnDescriptor : schema.getColumns()) {
-        String columnPath = concatDotPath(columnDescriptor.getPath());
+        String[] path = columnDescriptor.getPath();
+        String columnDotPath = concatDotPath(path);
         PrimitiveType primitiveType = columnDescriptor.getPrimitiveType();
+        boolean isInRepeatedGroup =
+            path.length > 1
+                && schema
+                    .getType(Arrays.copyOf(path, path.length - 1))
+                    .isRepetition(Type.Repetition.REPEATED);
```
sfc-gh-hmadan, line 199: is it possible to have Repetition.REPEATED on a path BUT its neither a map key, nor a map value, nor a list element.

Can we do an explicit check for those three situations?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -186,31 +189,46 @@ public void setupSchema(List<ColumnMetadata> columns) {
...
                   ordinal,
                   fieldId,
-                  primitiveType));
+                  primitiveType,
+                  clientBufferParameters.isEnableDistinctValuesCount(),
+                  clientBufferParameters.isEnableValuesCount() && isInRepeatedGroup));
         }
       }
     }
+    statsFinder = new SubColumnFinder(schema);
```
sfc-gh-hmadan, line 231: should only apply to iceberg land, so lets move this inside the previous brace.
Also you'll need a nullcheck in line 341.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -313,9 +336,11 @@ private float addRow(
           RowBufferStats.getCombinedStats(statsMap.get(columnName), forkedColStats.getValue()));
     }
 
-    // Increment null count for column missing in the input map
+    // Increment null count for column and its sub-columns missing in the input map
     for (String columnName : Sets.difference(this.fieldIndex.keySet(), inputColumnNames)) {
-      statsMap.get(columnName).incCurrentNullCount();
+      statsFinder
```
sfc-gh-hmadan, line 341: lets maintain the old behavior for non-iceberg mode.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -313,9 +336,11 @@ private float addRow(
...
+          .getSubColumns(columnName)
+          .forEach(subColumn -> statsMap.get(subColumn).incCurrentNullCount());
```
sfc-gh-hmadan, line 343: nullref if statsMap.get(subColumn) returns null.
can that happen?

src/main/java/net/snowflake/ingest/streaming/internal/ClientBufferParameters.java
sfc-gh-alhuang: Fixed.

src/main/java/net/snowflake/ingest/streaming/internal/ClientBufferParameters.java
sfc-gh-alhuang: Fixed.

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
sfc-gh-alhuang: Added to `toString`

src/main/java/net/snowflake/ingest/streaming/internal/ClientBufferParameters.java
sfc-gh-alhuang: The clientInterTest need this.

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
sfc-gh-alhuang: Added.

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
sfc-gh-alhuang: Added more error message.

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
sfc-gh-alhuang: Addressing this [comment](https://github.com/snowflakedb/snowflake-ingest-java/pull/851#discussion_r1786732455)

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
sfc-gh-alhuang: Moved to line 151.

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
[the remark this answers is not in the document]
sfc-gh-alhuang: If a root column is a structured data type, `statsMap.get(columnName)` will trow NPE as we only stores leaf column stats. Use `fieldIndex` instead for logging to avoid this.

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
sfc-gh-alhuang: Same as [here](https://github.com/snowflakedb/snowflake-ingest-java/pull/861#discussion_r1805498386)

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: Changed to `clientBufferParameters.isEnableValuesCount()`

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: Added check

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: Made `SubColumnFinder` Iceberg mode only and added NPE check.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: No as both `subColumnFinder` and `statsMap` are built from the same schema. Added check here in case.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: Switched to old logic for non-iceberg mode.

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
sfc-gh-alhuang: Switched to `Long` and use`NON_NULL`.

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
sfc-gh-alhuang: Renamed to `getParquetFooterSize` and moved to `BlobBuilder`.

src/main/java/net/snowflake/ingest/utils/Utils.java
```diff
@@ -430,4 +433,31 @@ public static String concatDotPath(String... path) {
...
   }
+
+  /**
+   * Get the footer size (metadata size) of a parquet file
+   *
+   * @param bytes the serialized parquet file
+   * @return the footer size
+   */
+  public static long getParquetFooterSize(byte[] bytes) throws IOException {
+    final int magicOffset = bytes.length - ParquetFileWriter.MAGIC.length;
```
sfc-gh-hmadan, line 444: curious, does this handle both PARE and PAR1 ?
I guess they're both 4 bytes so the length won't change.

src/main/java/net/snowflake/ingest/utils/Utils.java
```diff
@@ -430,4 +433,31 @@ public static String concatDotPath(String... path) {
...
+    if (!ParquetFileWriter.MAGIC_STR.equals(
+        new String(bytes, magicOffset, ParquetFileWriter.MAGIC.length))) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Invalid parquet file. Bad parquet magic, expected=%s, actual=%s.",
+              ParquetFileWriter.MAGIC_STR,
+              new String(bytes, magicOffset, ParquetFileWriter.MAGIC.length)));
+    }
+
+    return BytesUtils.readIntLittleEndian(bytes, footerSizeOffset);
```
sfc-gh-hmadan, line 461: is it always little endian irrespective of parquet V1 and parquet V2? How can we add a test that ensures this method is correct?

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
```diff
@@ -110,6 +114,10 @@ class FileColumnProperties {
     this.setMinStrNonCollated(null);
     this.setNullCount(stats.getCurrentNullCount());
     this.setDistinctValues(stats.getDistinctValues());
```
sfc-gh-hmadan, line 116: we're not doing the if check for NDV, curious why we need to do it for NV?

src/main/java/net/snowflake/ingest/streaming/internal/ClientBufferParameters.java
sfc-gh-hmadan: there's another ctor for testcases to pass in whatever booleans they want, so still unclear why there's a need to pass in a null clientInternal to this ctor.

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -135,17 +135,21 @@ static <T> Blob constructBlobAndMetadata(
...
                         serializedChunk.rowCount,
                         serializedChunk.columnEpStatsMapCombined,
-                        internalParameterProvider.setAllDefaultValuesInEp()))
+                        internalParameterProvider.setAllDefaultValuesInEp(),
+                        internalParameterProvider.isEnableDistinctValuesCount()))
                 .setFirstInsertTimeInMs(serializedChunk.chunkMinMaxInsertTimeInMs.getFirst())
                 .setLastInsertTimeInMs(serializedChunk.chunkMinMaxInsertTimeInMs.getSecond());
 
         if (internalParameterProvider.setIcebergSpecificFieldsInEp()) {
+          final long metadataSize = getParquetFooterSize(compressedChunkData);
```
sfc-gh-hmadan, line 144: if BDEC wants to set parquet footer size, this logic won't work since you're going off the encrypted byte array.
What we really want is to run this off the unencrypted, unpadded byte array, no?

Lets add a comment here for now that this calculation cannot be used when encryption is on. Also add an assert for internalParameterProvider.getEnableChunkEncryption() == false in here.

src/main/java/net/snowflake/ingest/utils/Utils.java
sfc-gh-alhuang: Yes they have the same length. Added check for `PARE`.

src/main/java/net/snowflake/ingest/utils/Utils.java
sfc-gh-alhuang: Yes, the V1 and V2 only affects data page. ([ref](https://github.com/apache/parquet-java/blob/42cf31c0fbe4f000d4ddb1e1092c6634989ea3ca/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java#L1880)). For testing, I think we can use `serializedBytes.length - writer.getColumnIndexReference().getOffset() - metadataSize - ParquetFileWriter.MAGIC.length - footerSizeBytes(4) = extenedMetadataSize`. What do you think?

src/main/java/net/snowflake/ingest/streaming/internal/FileColumnProperties.java
sfc-gh-alhuang: We already have NDV fields in FDN mode and it's always sets to -1, it's safe serialized it. NV is a new fields so adding this check to avoid it populated in FDN mode.

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
sfc-gh-alhuang: Added assertion.

src/main/java/net/snowflake/ingest/utils/Utils.java
sfc-gh-alhuang: Added test for parquet footer size & extendedMetadataSize in `BlobBuilderTest`

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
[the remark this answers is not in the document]
sfc-gh-hmadan: please add this as a code comment.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -113,7 +113,9 @@ public void setupSchema(List<ColumnMetadata> columns) {
                 column.getCollation(),
                 column.getOrdinal(),
                 null /* fieldId */,
-                parquetType.isPrimitive() ? parquetType.asPrimitiveType() : null));
+                parquetType.isPrimitive() ? parquetType.asPrimitiveType() : null,
+                clientBufferParameters.isEnableDistinctValuesCount(),
```
sfc-gh-hmadan, line 117: lets set this to false explicitly until BDEC ingestion decides to start using NDV / NV ?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -313,9 +352,27 @@ private float addRow(
...
     for (String columnName : Sets.difference(this.fieldIndex.keySet(), inputColumnNames)) {
-      statsMap.get(columnName).incCurrentNullCount();
+      if (clientBufferParameters.getIsIcebergMode()) {
+        if (subColumnFinder == null) {
+          throw new SFException(ErrorCode.INTERNAL_ERROR, "SubColumnFinder is not initialized.");
+        }
+        subColumnFinder
+            .getSubColumns(columnName)
+            .forEach(
+                subColumn -> {
```
sfc-gh-hmadan, line 364: nit: you can save four levels of indentation by doing `for (String subColumn: subColumnFilder.getSubColumns(columnName)) {`

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -313,9 +352,27 @@ private float addRow(
...
+                  RowBufferStats stats = statsMap.get(subColumn);
+                  if (stats == null) {
+                    throw new SFException(
+                        ErrorCode.INTERNAL_ERROR,
+                        String.format("Column %s not found in stats map.", subColumn));
+                  }
+                  statsMap.get(subColumn).incCurrentNullCount();
```
sfc-gh-hmadan, line 371: repeated access of concurrent dictionaries for the same key is unnecessary. You can directly do `stats.incCurrentNullCount()` here

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -313,9 +352,27 @@ private float addRow(
           RowBufferStats.getCombinedStats(statsMap.get(columnName), forkedColStats.getValue()));
```
sfc-gh-hmadan, line 352: oh wow this looks like a thread unsafe operation here. RowBuffer.statsMap and tempStatsMap are both constructed as non-concurrent collections, but we don't prohibit concurrent insertRows on one channel!

sfc-gh-hmadan: (nothig to fix for now we'll followup on this separately)

src/main/java/net/snowflake/ingest/streaming/internal/RowBufferStats.java
```diff
@@ -95,13 +129,25 @@ static RowBufferStats getCombinedStats(RowBufferStats left, RowBufferStats right
               "left=%s, right=%s",
               left.getCollationDefinitionString(), right.getCollationDefinitionString()));
     }
+
+    if (left.enableDistinctValuesCount != right.enableDistinctValuesCount) {
```
sfc-gh-hmadan, line 133: need similar chekc for enableValuesCount, no?

sfc-gh-alhuang: Added check.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: fixed.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: All this parameter comes from `InternalParameterProvider` and it's set to false for FDN mode. or do you mean passing `false` as arg right here?

sfc-gh-hmadan: i mean passing false hardcoded since we know FDN doesn't want to do NDV / NV (in fact the service will barf if they're sent over). ok to take in next PR

sfc-gh-hmadan (approved):
LGTM!

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: Modified.

PR #864, opened 2024-10-17
SNOW-1748333 Fix Iceberg decimal type schema parser

sfc-gh-alhuang:
## Summary
This PR addresses the issue where the scanner encounters an error when scanning delta-encoded decimal values with a precision greater than 18.

## Changes
1. Removed `sfVer` metadata in Iceberg mode: This prevents the scanner from applying optimizations specific to Snowflake-written Parquet files, which was the root cause of the issue with scanning.
2. Add tests for multiple random generated decimals.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
@@ -86,7 +86,9 @@ public class ParquetRowBuffer extends AbstractRowBuffer<ParquetChunkData> {
   public void setupSchema(List<ColumnMetadata> columns) {
     fieldIndex.clear();
     metadata.clear();
-    metadata.put("sfVer", "1,1");
+    if (!clientBufferParameters.getIsIcebergMode()) {
+      metadata.put("sfVer", "1,1");
+    }
     List<Type> parquetTypes = new ArrayList<>();
     int id = 1;
 
--- a/src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
+++ b/src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
@@ -4,14 +4,6 @@
 
 package net.snowflake.ingest.utils;
 
-import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.BINARY;
-import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.BOOLEAN;
-import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.DOUBLE;
-import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.FIXED_LEN_BYTE_ARRAY;
-import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.FLOAT;
-import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.INT32;
-import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.INT64;
-
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -20,11 +12,10 @@ import com.google.common.collect.Lists;
 import java.util.Iterator;
 import java.util.List;
 import javax.annotation.Nonnull;
+import org.apache.iceberg.parquet.TypeToMessageType;
 import org.apache.iceberg.types.Type;
 import org.apache.iceberg.types.Types;
 import org.apache.iceberg.util.JsonUtil;
-import org.apache.parquet.schema.GroupType;
-import org.apache.parquet.schema.LogicalTypeAnnotation;
 
 /**
  * This class is used to Iceberg data type (include primitive types and nested types) serialization
@@ -32,9 +23,7 @@ import org.apache.parquet.schema.LogicalTypeAnnotation;
  *
  * <p>This code is modified from
  * GlobalServices/modules/data-lake/datalake-api/src/main/java/com/snowflake/metadata/iceberg
- * /IcebergDataTypeParser.java and <a
- * href="https://github.com/apache/iceberg/blob/main/parquet/src/main/java/org/apache/iceberg/parquet/TypeToMessageType.java">
- * TypeToMessageType.java</a>
+ * /IcebergDataTypeParser.java
  */
 public class IcebergDataTypeParser {
   private static final String TYPE = "type";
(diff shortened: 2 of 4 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -86,7 +86,9 @@ public class ParquetRowBuffer extends AbstractRowBuffer<ParquetChunkData> {
   public void setupSchema(List<ColumnMetadata> columns) {
     fieldIndex.clear();
     metadata.clear();
-    metadata.put("sfVer", "1,1");
+    if (!clientBufferParameters.getIsIcebergMode()) {
+      metadata.put("sfVer", "1,1");
```
sfc-gh-hmadan, line 90: 1. from line 101 it looks like we are adding the logical/physical type info to metadata for iceberg mode too (I thought from our slack conv that we weren't adding it in) ?
2. Can we scope this change to only kick in when serialization policy = optimized?

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
```diff
@@ -44,12 +55,26 @@ public class IcebergDataTypeParser {
...
+      LogicalTypeAnnotation.timestampType(
+          true /* adjusted to UTC */, LogicalTypeAnnotation.TimeUnit.MICROS);
+
+  private static final int DECIMAL_INT32_MAX_DIGITS = 9;
+  private static final int DECIMAL_INT64_MAX_DIGITS = 18;
+  private static final int DECIMAL_MAX_DIGITS = 38;
+  private static final int DECIMAL_MAX_BYTES = 16;
+
   /** Object mapper for this class */
   private static final ObjectMapper MAPPER = new ObjectMapper();
```
sfc-gh-hmadan, line 77: its unclear why you made this change. I see the only difference is in the avro-sanitization of the column name. Is that logic causing issues? 
Copy-pasting code from the iceberg-parquet library will need to consider licensing. IANAL Apache 2.0 requires adding the license as well as a summary of the modifications made while copying their code.

So if we really need to do this, I'd create an explicit TypeToMessageType file, add the proper license attribution at the top, and list what we've modified again in the file header.

My vote is to not do this at all.

If you want to keep the original column name around why not add a new field/property into ParquetTypeInfo, and pass the column name in there from ParquetTypeGenerator.generateColumnParquetTypeInfo ?

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergNumericTypesIT.java
```diff
@@ -26,6 +34,8 @@ public static Object[][] parameters() {
   @Parameterized.Parameter(1)
   public static Constants.IcebergSerializationPolicy icebergSerializationPolicy;
 
+  static final Random generator = new Random(0x5EED);
```
sfc-gh-hmadan, line 37: A fixed seed will imply the exact set of values being tested everytime. Why not have a randomly generated seed and plug that in here.
The reason to explicitly generate a seed and plug it in is - we can now log this seed / print it out in failure messages, and accurately reproduce any failures locally.

If you just do new Random(), reproducing a one-off error will be next to impossible.

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergNumericTypesIT.java
```diff
@@ -368,5 +379,48 @@ public void testDecimalAndQueries() throws Exception {
...
+    for (int i = 0; i < count; i++) {
+      int intPart = generator.nextInt(precision - scale + 1);
+      int floatPart = generator.nextInt(scale + 1);
+      if (intPart == 0 && floatPart == 0) {
+        list.add(null);
+        continue;
+      }
+      list.add(
+          new BigDecimal(
+              RandomStringUtils.randomNumeric(intPart)
```
sfc-gh-hmadan, line 420: this makes it impossible to reproduce failures locally. Why do we need this randomNumeric when we already generated a random number for intpart and floatpart using the seeded RNG in lines 412 and 413?

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergNumericTypesIT.java
sfc-gh-alhuang: Switched to random seed.

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergNumericTypesIT.java
sfc-gh-alhuang: The length of decimal is greater than int/long.max, thus we need string generator for this. Switch to generator with consistent seed.

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
sfc-gh-alhuang: Reverted.

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergNumericTypesIT.java
```diff
@@ -26,9 +36,21 @@ public static Object[][] parameters() {
...
+  private static final Logger logger = LoggerFactory.getLogger(IcebergNumericTypesIT.class);
+  private static Random generator;
+  private static RandomStringGenerator randomStringGenerator;
+
   @Before
   public void before() throws Exception {
     super.beforeIceberg(compressionAlgorithm, icebergSerializationPolicy);
+    long seed = System.currentTimeMillis();
+    logger.info("Random seed: {}", seed);
+    generator = new Random(123);
```
sfc-gh-hmadan, line 48: please use seed instead of 123

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergNumericTypesIT.java
```diff
@@ -368,5 +391,49 @@ public void testDecimalAndQueries() throws Exception {
...
+
+    List<Object> bigDecimals_38_10 = randomBigDecimal(200, 38, 10);
+    testIcebergIngestAndQuery(
+        "decimal(38, 10)",
+        bigDecimals_38_10,
+        "select {columnName} from {tableName}",
+        bigDecimals_38_10);
+  }
+
+  private static List<Object> randomBigDecimal(int count, int precision, int scale) {
```
sfc-gh-hmadan, line 421: not sure i fully understand what this is trying to do (comments will help?). Lets get this merged in to unblock KC folks and I'll follow up with you offline on this method.

sfc-gh-hmadan (approved):
LGTM, test code we'll discuss after this gets merged.

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergNumericTypesIT.java
sfc-gh-alhuang: Use seed instead.

PR #865, opened 2024-10-18
Revert "Revert "Add chunk offset to file id key to make each chunk ha…

sfc-gh-psaha:
…ve a uni… (#848)"

This reverts commit 0930648dfaaefa5abb6fe7aa4dcb83423ac1f11e.


sfc-gh-xhuang (approved):

sfc-gh-tzhang (changes requested):
Let's hold-off the merge until the server side change is in prod, do you know which release it will in?

sfc-gh-psaha:
> Let's hold-off the merge until the server side change is in prod, do you know which release it will in?

It will be in 8.41. What's the risk if we merge now? False alarms in test environments?

sfc-gh-psaha:
> > Let's hold-off the merge until the server side change is in prod, do you know which release it will in?
> 
> It will be in 8.41. What's the risk if we merge now? False alarms in test environments?

The plan is we'd merge this and enable server side detection fix starting from this version. PTAL.

sfc-gh-tzhang (approved):
Please make sure the server side change is enabled before this change is released, thanks!

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -162,6 +159,32 @@ private SerializationResult serializeFromJavaObjects(
         parquetWriter.getExtendedMetadataSize());
   }
 
+  private void addFileIdToMetadata(
+      String filePath, long chunkStartOffset, Map<String, String> metadata) {
+    // We insert the filename in the file itself as metadata so that streams can work on replicated
+    // mixed tables. For a more detailed discussion on the topic see SNOW-561447 and
+    // http://go/streams-on-replicated-mixed-tables
```
sfc-gh-hmadan, line 166: nit: pl add the other golink here too from line 133 in old code

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -162,6 +159,32 @@ private SerializationResult serializeFromJavaObjects(
...
+    // id key is unique for each chunk. Each chunk is logically a separate Parquet file that happens
+    // to be bundled together.
+    if (chunkStartOffset == 0) {
+      metadata.put(
+          enableIcebergStreaming
+              ? Constants.ASSIGNED_FULL_FILE_NAME_KEY
+              : Constants.PRIMARY_FILE_ID_KEY,
+          StreamingIngestUtils.getShortname(filePath));
+    } else {
+      String shortName = StreamingIngestUtils.getShortname(filePath);
```
sfc-gh-hmadan, line 177: lets also assert / validate that enableIcebergStreaming is false when we're in the else path?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -89,6 +91,7 @@ public void setupSchema(List<ColumnMetadata> columns) {
     if (!clientBufferParameters.isEnableIcebergStreaming()) {
       metadata.put("sfVer", "1,1");
     }
+    metadata.put(Constants.SDK_VERSION_KEY, RequestBuilder.DEFAULT_VERSION);
```
sfc-gh-hmadan, line 94: nice, Alec and I had discussed doing precisely this just yesterday!

We're going to have multiple SDK languages in a few months, should we have a slightly more verbose key-value pair here, such as:
userAgent=RequestBuilder.USER_AGENT ?

sfc-gh-hmadan: User-Agent is a very HTTP-ish thing, we could name the key createdBy instead.

sfc-gh-hmadan (approved):

sfc-gh-asen (approved):

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
sfc-gh-psaha: Done.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
sfc-gh-psaha: Done.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-psaha: This current key-value is what XP expects. Let's iterate on it later?

PR #868, opened 2024-10-21
SNOW-1754295 - Start using subscoped tokens for iceberg ingestion

sfc-gh-hmadan:
1. Add logic for invoking new server-side API  (Refresh Table Information)
2. Reuse InternalStage for all the JDBC interop. For this:
    1.  pass in a tableRef into InternalStage
    2. Add a IStorageManager.getRefreshedLocation API that's called by InternalStage
3. Add SubscopedTokenExternalVolumeManager that creates an InternalStage per table, and fetches new credentials via the new API from (1). InternalStage calls storageManager.getRefreshedLocation and doesn't care whether it is operating on an account's internal stage's streaming_ingest subfolder (FDN ingestion) or on a per-session location under the external volume's S3 location.
4. Remove <T> from InternalStage and InternalStageManager

Note that after this change, InternalStage can / should be renamed, did not want to make this PR even bigger thus haven't done that yet.

New tests for SubscopedTokenExtVolManager are pending. Sending this out for review + to see if any merge gate tests barf.


src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeServiceClient.java
```diff
@@ -97,6 +99,27 @@ GeneratePresignedUrlsResponse generatePresignedUrls(GeneratePresignedUrlsRequest
     return response;
   }
 
+  /** Generates a batch of presigned URLs for a table */
```
sfc-gh-psaha, line 102: Is the doc string correct?

src/main/java/net/snowflake/ingest/utils/Constants.java
```diff
@@ -49,12 +49,14 @@ public class Constants {
   public static final int BLOB_FILE_SIZE_SIZE_IN_BYTES = 8;
   public static final int BLOB_CHECKSUM_SIZE_IN_BYTES = 8;
   public static final int BLOB_CHUNK_METADATA_LENGTH_SIZE_IN_BYTES = 4;
-  public static final long THREAD_SHUTDOWN_TIMEOUT_IN_SEC = 300L;
+  public static final long THREAD_SHUTDOWN_TIMEOUT_IN_SEC = 30L;
```
sfc-gh-psaha, line 52: Why is this change needed?

src/main/java/net/snowflake/ingest/streaming/internal/SubscopedTokenExternalVolumeManager.java
sfc-gh-psaha, line 1: Looks like this module is missing unit tests. I think you said you are working on adding them, right?

sfc-gh-hmadan: Correct, adding them while the review is ongoing.

src/main/java/net/snowflake/ingest/utils/Constants.java
sfc-gh-hmadan: Yes accidentally made its way to PR, had removed it locally already, thanks!

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeServiceClient.java
sfc-gh-hmadan: nope, will fix.

src/main/java/net/snowflake/ingest/streaming/internal/SubscopedTokenExternalVolumeManager.java
sfc-gh-hmadan: resolving, tests have been added.

src/main/java/net/snowflake/ingest/streaming/internal/SubscopedTokenExternalVolumeManager.java
```diff
@@ -0,0 +1,230 @@
...
+        logger.logInfo(
+            "Skip registering table since its already been registered with the VolumeManager."
+                + " tableRef=%s",
+            tableRef);
+        return;
+      }
+
+      // get the locationInfo when we know this is the first register call for a given table.
+      // This is done to reduce the unnecessary overload on token generation if a client opens up a
+      // hundred channels at
```
sfc-gh-psaha, line 109: line break

src/main/java/net/snowflake/ingest/streaming/internal/SubscopedTokenExternalVolumeManager.java
```diff
@@ -0,0 +1,230 @@
...
+  public void registerTable(TableRef tableRef) {
+    if (this.externalVolumeMap.containsKey(tableRef.fullyQualifiedName)) {
+      logger.logInfo(
+          "Skip registering table since its already been registered with the VolumeManager."
+              + " tableRef=%s",
+          tableRef);
+      return;
+    }
+
+    // future enhancement - per table locks instead of per-client lock
```
sfc-gh-psaha, line 97: Since you are using a ConcurrentHashMap don't you get per key locks for free anyways if you use `computeIfAbsent` etc.?

src/main/java/net/snowflake/ingest/streaming/internal/SubscopedTokenExternalVolumeManager.java
```diff
@@ -0,0 +1,230 @@
...
+            String.format("fullyQualifiedTableName=%s", tableRef));
+      }
+    }
+  }
+
+  @Override
+  public BlobPath generateBlobPath(String fullyQualifiedTableName) {
+    InternalStage volume = getVolumeSafe(fullyQualifiedTableName);
+
+    // {nullableTableBasePath}/data/streaming_ingest/{figsId}/snow_{volumeHash}_{figsId}_{workerRank}_1_
```
sfc-gh-psaha, line 143: In the non iceberg path I recently added a check to make sure that the deployment id returned when refreshing the location matches the one that the channel was originally created with. This is to prevent a channel from ingesting data into the wrong deployment in replication scenarios. Do we need something like that for iceberg?

src/main/java/net/snowflake/ingest/streaming/internal/SubscopedTokenExternalVolumeManager.java
```diff
@@ -0,0 +1,230 @@
...
+        }
+      }
+
+      return response.getIcebergLocationInfo();
+    } catch (IngestResponseException | IOException e) {
+      throw new SFException(e, ErrorCode.REFRESH_TABLE_INFORMATION_FAILURE, e.getMessage());
+    }
+  }
+
+  private InternalStage getVolumeSafe(String fullyQualifiedTableName) {
```
sfc-gh-psaha, line 219: What is `Safe` about it? Consider renaming.

src/main/java/net/snowflake/ingest/streaming/internal/SubscopedTokenExternalVolumeManager.java
sfc-gh-hmadan: That's correct, i somehow had the assumption that computeIfAbsent does not guarantee singular invocation of the function passed in, just read docs to confirm it does guarantee that. Switched to computeIfAbsent!

src/main/java/net/snowflake/ingest/streaming/internal/SubscopedTokenExternalVolumeManager.java
sfc-gh-hmadan: Yes, already handling that with iceberg. THe workerRank is populated with the deployment id by the service, and is verified in registerBlob() for being correct.So a filepath sent down by deployment 1 cannot be used to register a file in deployment 2.

btw this depl id fix was only needed when doing client side encryption from what i understand, still added it in here for consistent behaviors across iceberg and non-iceberg.

src/main/java/net/snowflake/ingest/streaming/internal/SubscopedTokenExternalVolumeManager.java
sfc-gh-hmadan: I believe this is just a snowflake service side convention to have a method pair for situations where the result can be null.

For instance you can see a lot of `RetVal getField()` and `RetVal getFieldSafe()` method pairs, where the former returns null when field is null, whereas the latter throws if it only has a null to return back.

I prefer an Optional<T> return value to indicate a nullable return type, but this pattern is already imprinted across snowflake so am just carrying it along!


In this case, getVolume() can technically return null thus the *Safe suffix to indicate the method will throw if it can't find an object to return back.

src/main/java/net/snowflake/ingest/streaming/internal/SubscopedTokenExternalVolumeManager.java
sfc-gh-psaha: Thanks. It was not just for encryption - it also prevents some race conditions where blobs are uploaded in one deployment but registered in another and hence gets lost.

sfc-gh-psaha (approved):

PR #869, opened 2024-10-22
SNOW-1757554 Support quoted object fields name

sfc-gh-alhuang:
This PR aims to support quoted object fields name for Iceberg table ingestion. The PR include following change:
* **Preserve original column name in schema**: Decode column name after schema creation to avoid sub-column name mismatch. As we use [TypeToMessageType](https://github.com/apache/iceberg/blob/main/parquet/src/main/java/org/apache/iceberg/parquet/TypeToMessageType.java) to convert from Iceberg schema to parquet schema which include Avro column name encoding for all non-digit/letters characters.
* **Escape dot character in EP info keys**: Escape the dot character in column name of a dot path with backslash. Without this, it's possible for two different columns to have the same dot path. E.g. `("a.a" int, a object(a int))`.
* **Change stats map's key to field id**: The old logic build dot path as key of stats map along with structured data type validation, which might cause performance issue. Remove this logic and use `fieldId` as key instead to avoid string construction. Keep a map of `fieldId -> dotPath` in `subcolumnFinder` for logging purpose.


src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergStructuredIT.java
```diff
@@ -95,59 +91,129 @@ public void testStructuredDataType() throws Exception {
...
+            + "test_test int, "
+            + "test_x5Ftest long, "
+            + "\"obj\\.obj\" float, "
+            + "\"❄️\" double, "
+            + "\"5566\" string, "
+            + "_5566 string, "
+            + "\"_\" fixed(10), "
+            + "_x27_x44_xFE_x0F binary, "
+            + "\"\"\"\" date, "
+            + "\"\\\" string, "
```
sfc-gh-alhuang, line 208: Keep it as String for now as we cannot compare date time value without format. Planning to add a generator & comparator for all types / structured data types.

src/main/java/net/snowflake/ingest/utils/SubColumnFinder.java
```diff
@@ -15,55 +14,67 @@
...
+    }
+    if (!path.isEmpty() && node.getId() != null) {
+      accessMap.put(
+          node.getId().toString(),
+          new SubtreeInfo(startTag, list.size(), concatDotPath(path.toArray(new String[0]))));
     }
-    if (!isNullOrEmpty(dotPath)) {
-      accessMap.put(dotPath, new SubtreeInterval(startTag, list.size()));
+    if (!path.isEmpty()) {
+      path.remove(path.size() - 1);
```
sfc-gh-hmadan, line 77: add code comment on why this remove was called?

src/main/java/net/snowflake/ingest/utils/SubColumnFinder.java
```diff
@@ -15,55 +14,67 @@
...
       for (Type child : node.asGroupType().getFields()) {
-        build(child, dotPath);
+        build(child, path);
       }
     } else {
-      list.add(dotPath);
+      list.add(node.getId().toString());
+    }
+    if (!path.isEmpty() && node.getId() != null) {
+      accessMap.put(
```
sfc-gh-hmadan, line 72: we should change accessMap's type to be Map<ID, SubtreeInfo> instead of String.
I see every place doing ID.toString(), might as well use strong typing to enforce the ID is what comes in and not just any String.

src/main/java/net/snowflake/ingest/utils/SubColumnFinder.java
```diff
@@ -15,55 +14,67 @@
...
 
-    SubtreeInterval(int startTag, int endTag) {
+    SubtreeInfo(int startTag, int endTag, String dotPath) {
       this.startTag = startTag;
       this.endTag = endTag;
+      this.dotPath = dotPath;
     }
   }
 
   private final List<String> list;
```
sfc-gh-hmadan, line 31: add code comments on what is in list, what is in accessMap, similarly for startTag/endTag/SubtreeInfo/all public methods?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -364,12 +364,13 @@ private float addRow(
...
-        for (String subColumn : subColumnFinder.getSubColumns(columnName)) {
-          RowBufferStats stats = statsMap.get(subColumn);
+        for (String subColumnId :
+            subColumnFinder.getSubColumns(fieldIndex.get(columnName).type.getId().toString())) {
+          RowBufferStats stats = statsMap.get(subColumnId);
           if (stats == null) {
             throw new SFException(
                 ErrorCode.INTERNAL_ERROR,
-                String.format("Column %s not found in stats map.", subColumn));
+                String.format("Field id %s not found in stats map.", subColumnId));
```
sfc-gh-hmadan, line 373: can we add the field path to the error message too?

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
```diff
@@ -154,7 +162,14 @@ public static Type getTypeFromJson(@Nonnull JsonNode jsonNode) {
           field.isObject(), "Cannot parse struct field from non-object: %s", field);
 
       int id = JsonUtil.getInt(ID, field);
-      String name = JsonUtil.getString(NAME, field);
+
+      /* TypeToMessageType throws on empty field name, use a backslash to represent it and escape remaining backslash. */
```
sfc-gh-hmadan, line 166: iceberg spec allows empty field name?

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
```diff
@@ -208,4 +223,61 @@ public static Types.MapType mapFromJson(JsonNode json) {
...
+      org.apache.parquet.schema.Type parquetType, Type icebergType, String fieldName) {
+    if (parquetType.isPrimitive() != icebergType.isPrimitiveType()
+        || (!parquetType.isPrimitive()
+            && parquetType.getLogicalTypeAnnotation()
+                == null /* ignore outer layer of map or list */
+            && parquetType.asGroupType().getFieldCount()
+                != icebergType.asNestedType().fields().size())) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Parquet type and Iceberg type mismatch: %s, %s", parquetType, icebergType));
```
sfc-gh-hmadan, line 237: "Parquet type and Iceberg type mismatch. parquetType=%s, icebergType=%s"

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
```diff
@@ -208,4 +223,61 @@ public static Types.MapType mapFromJson(JsonNode json) {
       return Types.MapType.ofOptional(keyId, valueId, keyType, valueType);
     }
   }
+
+  private static org.apache.parquet.schema.Type replaceWithOriginalFieldName(
```
sfc-gh-hmadan, line 227: add method comment on why we're having to do this?

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
```diff
@@ -208,4 +223,61 @@ public static Types.MapType mapFromJson(JsonNode json) {
...
+    }
+    if (parquetType.isPrimitive()) {
+      /* rename field name */
+      return org.apache.parquet.schema.Types.primitive(
+              parquetType.asPrimitiveType().getPrimitiveTypeName(), parquetType.getRepetition())
+          .as(parquetType.asPrimitiveType().getLogicalTypeAnnotation())
+          .id(parquetType.getId().intValue())
+          .length(parquetType.asPrimitiveType().getTypeLength())
+          .named(fieldName);
+    } else {
```
sfc-gh-hmadan, line 247: nit: remove else, the if block already has a return

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
```diff
@@ -208,4 +223,61 @@ public static Types.MapType mapFromJson(JsonNode json) {
...
+      org.apache.parquet.schema.Types.GroupBuilder<org.apache.parquet.schema.GroupType> builder =
+          org.apache.parquet.schema.Types.buildGroup(parquetType.getRepetition());
+      for (org.apache.parquet.schema.Type parquetFieldType :
+          parquetType.asGroupType().getFields()) {
+        if (parquetFieldType.getId() == null) {
```
sfc-gh-hmadan, line 252: do we have unit tests around each of these branches to make sure there's no accidental regressions in the future?

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
```diff
@@ -208,4 +223,61 @@ public static Types.MapType mapFromJson(JsonNode json) {
...
+              replaceWithOriginalFieldName(
+                  parquetFieldType, icebergType, parquetFieldType.getName()));
+        } else {
+          Types.NestedField icebergField =
+              icebergType.asNestedType().field(parquetFieldType.getId().intValue());
+          if (icebergField == null) {
+            throw new IllegalArgumentException(
+                String.format(
+                    "Cannot find Iceberg field with id %d in Iceberg type: %s",
+                    parquetFieldType.getId().intValue(), icebergType));
```
sfc-gh-hmadan, line 264: also log parquetFieldType to the message.

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
```diff
@@ -65,16 +67,21 @@ public static org.apache.parquet.schema.Type parseIcebergDataTypeStringToParquet
       int id,
       String name) {
     Type icebergType = deserializeIcebergType(icebergDataType);
```
sfc-gh-hmadan, line 69: Lets discuss this tomorrow, had an idea that we can explore after this PR goes in / as a future enhancement to reduce complexity :

Do we really need to use parquet types in iceberg ingestion, when there is already an iceberg data type hierarchy that works well?
What if we added an IcebergRowBuffer just like ParquetRowBuffer. IcebergRowBuffer will use iceberg data types, and ParquetRowBuffer will use parquet data types (for BDEC).

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
```diff
@@ -208,4 +223,61 @@ public static Types.MapType mapFromJson(JsonNode json) {
...
+                String.format(
+                    "Cannot find Iceberg field with id %d in Iceberg type: %s",
+                    parquetFieldType.getId().intValue(), icebergType));
+          }
+          builder.addField(
+              replaceWithOriginalFieldName(
+                  parquetFieldType,
+                  icebergField.type(),
+                  icebergField.name().equals(EMPTY_FIELD_CHAR)
+                      ? ""
```
sfc-gh-hmadan, line 271: add a code comment on why we're basically doing a decode from \\ to String.Empty here?

sfc-gh-hmadan: also add testcases around this!

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
```diff
@@ -208,4 +223,61 @@ public static Types.MapType mapFromJson(JsonNode json) {
...
+                      : icebergField
+                          .name()
+                          .replace(EMPTY_FIELD_CHAR + EMPTY_FIELD_CHAR, EMPTY_FIELD_CHAR)));
+        }
+      }
+      if (parquetType.getId() != null) {
+        builder.id(parquetType.getId().intValue());
+      }
```
sfc-gh-hmadan, line 279: else ?

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -75,21 +70,27 @@ private static ParquetBufferValue parseColumnValueToParquet(
...
+    }
+    String id = type.getId().toString();
+
     if (type.isPrimitive()) {
-      if (!statsMap.containsKey(path)) {
+      if (!statsMap.containsKey(id)) {
         throw new SFException(
-            ErrorCode.INTERNAL_ERROR, String.format("Stats not found for column: %s", path));
+            ErrorCode.INTERNAL_ERROR,
+            String.format("Stats not found for fieldId: %s", type.getId()));
```
sfc-gh-hmadan, line 86: nit: id

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergStructuredIT.java
```diff
@@ -58,29 +58,25 @@ public void testStructuredDataType() throws Exception {
...
-            SFException.class,
+    Assertions.assertThatThrownBy(
             () ->
-                assertStructuredDataType("object(a int, b string)", "{\"a\": 1, \"c\": \"test\"}"));
-    Assertions.assertThat(ex)
-        .extracting(SFException::getVendorCode)
+                assertStructuredDataType("object(a int, b string)", "{\"a\": 1, \"c\": \"test\"}"))
+        .isInstanceOf(SFException.class)
+        .extracting("vendorCode")
         .isEqualTo(ErrorCode.INVALID_FORMAT_ROW.getMessageCode());
```
sfc-gh-hmadan, line 80: also validate the string messages in these tests, not just the error code.

src/main/java/net/snowflake/ingest/utils/SubColumnFinder.java
sfc-gh-alhuang: Added.

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
sfc-gh-alhuang: Iceberg spec doesn't explicitly disallow it. Snowflake Iceberg DML supports it.

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
sfc-gh-alhuang: Added

src/main/java/net/snowflake/ingest/utils/SubColumnFinder.java
sfc-gh-alhuang: Added

src/main/java/net/snowflake/ingest/utils/SubColumnFinder.java
sfc-gh-alhuang: Switch to `Type.ID`

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergStructuredIT.java
sfc-gh-alhuang: Added error message validation.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: modified.

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
sfc-gh-alhuang: The middle layer of Iceberg list/map does not have id. We don't want to set it.

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
sfc-gh-alhuang: Added unit test and comments

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
sfc-gh-alhuang: Added

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: Added

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
sfc-gh-alhuang: Updated

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
sfc-gh-alhuang: Added unit test in `IceberDataTypeParserTest` with list/map.

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
sfc-gh-alhuang: Removed

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
sfc-gh-alhuang: The Iceberg types does not have any information about the actual physical type. As long as we are writing parquet I think we always need parquet type.

sfc-gh-hmadan: i see, ok.

src/main/java/net/snowflake/ingest/utils/IcebergDataTypeParser.java
sfc-gh-hmadan: got it, do have an IT test with empty names then, reactivating to make sure you see this!

no need to block this PR for empty field name IT test coverage.

sfc-gh-alhuang: In the IT of this PR. [ref](https://github.com/snowflakedb/snowflake-ingest-java/blob/af46e607320cbb2f8f79402d04d67d2d72c94a90/src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergStructuredIT.java#L180).

src/main/java/net/snowflake/ingest/utils/SubColumnFinder.java
```diff
@@ -15,55 +14,94 @@
...
+  private void build(Type node, List<String> currentPath) {
+    if (currentPath == null) {
       /* Ignore root node type name (bdec or schema) */
-      dotPath = "";
-    } else if (dotPath.isEmpty()) {
-      dotPath = node.getName();
+      currentPath = new ArrayList<>();
     } else {
-      dotPath = concatDotPath(dotPath, node.getName());
+      currentPath.add(node.getName());
```
sfc-gh-hmadan, line 85: Something feels off here. You're passing a mutable list down every recursion level, and then at the end creating a dotPath out of the whole list of levels. 

Lets say you called build(node, null) on a column like so:
OBJECT col1 (OBJECT col11(ARRAY col111), OBJECT col12(MAP col121))

What will the state of the SubColumnFinder be when col1 is added to SubColumnFinder?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -364,12 +364,16 @@ private float addRow(
           throw new SFException(ErrorCode.INTERNAL_ERROR, "SubColumnFinder is not initialized.");
         }
 
-        for (String subColumn : subColumnFinder.getSubColumns(columnName)) {
-          RowBufferStats stats = statsMap.get(subColumn);
+        for (String subColumnId :
+            subColumnFinder.getSubColumns(fieldIndex.get(columnName).type.getId())) {
```
sfc-gh-hmadan, line 368: fieldIndex.get(columnName) can return null, lets handle that and throw explicitly? OK in your next PR.

src/main/java/net/snowflake/ingest/utils/SubColumnFinder.java
```diff
@@ -15,55 +14,94 @@
...
+    final String dotPath;
+
+    SubtreeInfo(int startTag, int endTag, String dotPath) {
       this.startTag = startTag;
       this.endTag = endTag;
+      this.dotPath = dotPath;
     }
   }
 
+  /* A list to store all leaf columns field id in preorder traversal. */
```
sfc-gh-hmadan, line 41: Can the `list` also be List<Type.ID> now instead of List<String>?

sfc-gh-hmadan: can do away with `new Type.ID(Integer.parseInt(subColumnId)) ` in ParquetRowBuffer if yes

sfc-gh-hmadan (approved):

src/main/java/net/snowflake/ingest/utils/SubColumnFinder.java
sfc-gh-alhuang: Yes but we need to do `toString()` every time when accessing `statsMap <String, Stats>`. I think either way works, do you have any preference?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
sfc-gh-alhuang: The `columnName` is from `Set.difference(fieldIndex.keySet(), otherSet)`. I think this should be safe.

PR #871, opened 2024-10-23
SNOW-1760732 Enable Iceberg mode from properties

sfc-gh-alhuang:
The PR includes following changes:
1. Remove `setIceberg` from `StreamingIngestClientFactory`, set Iceberg mode from properties.
2. Renamed `isIceberg` to `enableIcebergStreaming`.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
@@ -163,7 +163,7 @@ public class SnowflakeStreamingIngestClientInternal<T> implements SnowflakeStrea
       Map<String, Object> parameterOverrides) {
     this.parameterProvider = new ParameterProvider(parameterOverrides, prop);
     this.internalParameterProvider =
-        new InternalParameterProvider(parameterProvider.isIcebergMode());
+        new InternalParameterProvider(parameterProvider.isEnableIcebergStreaming());
 
     this.name = name;
     String accountName = accountURL == null ? null : accountURL.getAccount();
@@ -232,7 +232,7 @@ public class SnowflakeStreamingIngestClientInternal<T> implements SnowflakeStrea
     this.snowflakeServiceClient = new SnowflakeServiceClient(this.httpClient, this.requestBuilder);
 
     this.storageManager =
-        parameterProvider.isIcebergMode()
+        parameterProvider.isEnableIcebergStreaming()
             ? new SubscopedTokenExternalVolumeManager(
                 this.role, this.name, this.snowflakeServiceClient)
             : new InternalStageManager(
@@ -342,14 +342,14 @@ public class SnowflakeStreamingIngestClientInternal<T> implements SnowflakeStrea
               request.getTableName(),
               request.getChannelName(),
               Constants.WriteMode.CLOUD_STORAGE,
-              this.parameterProvider.isIcebergMode(),
+              this.parameterProvider.isEnableIcebergStreaming(),
               request.getOffsetToken());
       response = snowflakeServiceClient.openChannel(openChannelRequest);
     } catch (IOException | IngestResponseException e) {
       throw new SFException(e, ErrorCode.OPEN_CHANNEL_FAILURE, e.getMessage());
     }
 
-    if (parameterProvider.isIcebergMode()) {
+    if (parameterProvider.isEnableIcebergStreaming()) {
       if (response.getTableColumns().stream().anyMatch(c -> c.getSourceIcebergDataType() == null)) {
         throw new SFException(
             ErrorCode.INTERNAL_ERROR, "Iceberg table columns must have sourceIcebergDataType set");
@@ -387,7 +387,7 @@ public class SnowflakeStreamingIngestClientInternal<T> implements SnowflakeStrea
             .setDefaultTimezone(request.getDefaultTimezone())
             .setOffsetTokenVerificationFunction(request.getOffsetTokenVerificationFunction())
             .setParquetWriterVersion(
-                parameterProvider.isIcebergMode()
+                parameterProvider.isEnableIcebergStreaming()
                     ? Constants.IcebergSerializationPolicy.valueOf(
                             response.getIcebergSerializationPolicy())
                         .toParquetWriterVersion()
@@ -427,7 +427,7 @@ public class SnowflakeStreamingIngestClientInternal<T> implements SnowflakeStrea
               request.getSchemaName(),
               request.getTableName(),
               request.getChannelName(),
-              this.parameterProvider.isIcebergMode(),
+              this.parameterProvider.isEnableIcebergStreaming(),
               request instanceof DropChannelVersionRequest
                   ? ((DropChannelVersionRequest) request).getClientSequencer()
                   : null);
@@ -587,7 +587,7 @@ public class SnowflakeStreamingIngestClientInternal<T> implements SnowflakeStrea
               this.storageManager.getClientPrefix() + "_" + counter.getAndIncrement(),
               this.role,
               blobs,
-              this.parameterProvider.isIcebergMode());
+              this.parameterProvider.isEnableIcebergStreaming());
       response = snowflakeServiceClient.registerBlob(request, executionCount);
     } catch (IOException | IngestResponseException e) {
       throw new SFException(e, ErrorCode.REGISTER_BLOB_FAILURE, e.getMessage());
--- a/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
@@ -153,7 +153,6 @@ public class SnowflakeStreamingIngestClientInternal<T> implements SnowflakeStrea
    * @param prop connection properties
    * @param httpClient http client for sending request
    * @param isTestMode whether we're under test mode
-   * @param isIcebergMode whether we're streaming to Iceberg tables
    * @param requestBuilder http request builder
    * @param parameterOverrides parameters we override in case we want to set different values
    */
@@ -162,16 +161,16 @@ public class SnowflakeStreamingIngestClientInternal<T> implements SnowflakeStrea
       SnowflakeURL accountURL,
       Properties prop,
       CloseableHttpClient httpClient,
-      boolean isIcebergMode,
       boolean isTestMode,
       RequestBuilder requestBuilder,
       Map<String, Object> parameterOverrides) {
+    this.isIcebergMode =
+        prop != null && Boolean.parseBoolean(prop.getProperty(Constants.STREAMING_ICEBERG));
     this.parameterProvider = new ParameterProvider(parameterOverrides, prop, isIcebergMode);
     this.internalParameterProvider = new InternalParameterProvider(isIcebergMode);
 
     this.name = name;
     String accountName = accountURL == null ? null : accountURL.getAccount();
-    this.isIcebergMode = isIcebergMode;
     this.isTestMode = isTestMode;
     this.httpClient = httpClient == null ? HttpUtil.getHttpClient(accountName) : httpClient;
     this.channelCache = new ChannelCache<>();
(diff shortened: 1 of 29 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -162,16 +161,16 @@ public class SnowflakeStreamingIngestClientInternal<T> implements SnowflakeStrea
       SnowflakeURL accountURL,
       Properties prop,
       CloseableHttpClient httpClient,
-      boolean isIcebergMode,
       boolean isTestMode,
       RequestBuilder requestBuilder,
       Map<String, Object> parameterOverrides) {
+    this.isIcebergMode =
```
sfc-gh-hmadan, line 167: why should isIcebergMode not be treated exactly the same as other parameters, and be handled by ParameterProvider?
Your current fix does not allow `parameterOverrides` to set isIceberg, and forces setting it via config file only.
This makes the isIceberg setting behave differently from every other setting.

Lets also not bunch together all these fixes into one PR, we were doing it to push things through but now that things are in a better state lets please go back to one PR for every logical change.

sfc-gh-alhuang: Make sense, will split this PR. Also I think most parameters in `ParameterProvider` are not intended to be set by the customer in typical use cases. For example `account`, `role`, `authorization_type`, etc. these client level parameters can only be via config.

src/main/java/net/snowflake/ingest/utils/Constants.java
```diff
@@ -33,9 +33,12 @@ public class Constants {
   public static final String OAUTH_CLIENT_SECRET = "oauth_client_secret";
   public static final String OAUTH_REFRESH_TOKEN = "oauth_refresh_token";
   public static final String OAUTH_TOKEN_ENDPOINT = "oauth_token_endpoint";
+  public static final String STREAMING_ICEBERG = "streaming_iceberg";
   public static final String SNOWFLAKE_OAUTH_TOKEN_ENDPOINT = "/oauth/token-request";
   public static final String PRIMARY_FILE_ID_KEY =
       "primaryFileId"; // Don't change, should match Parquet Scanner
+  public static final String FULL_FILL_NAME_KEY =
+      "fullFillNameKey"; // Don't change, should match Parquet Scanner
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Will update. Removed from this PR.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
sfc-gh-alhuang: Discussed offline. Keep it configureable from both props and parameterOverride.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -96,18 +95,23 @@ public class ParameterProvider {
...
    * @param props Properties from profile file
-   * @param isIcebergMode If the provided parameters need to be verified and modified to meet
-   *     Iceberg mode
    */
+  public ParameterProvider(Map<String, Object> parameterOverrides, Properties props) {
+    this.setParameterMap(parameterOverrides, props);
+  }
+
+  /** Constructor for tests */
   public ParameterProvider(
```
sfc-gh-hmadan, line 104: why can tests not pass in a props with isIcebergMode set to true, instead of requiring a special constructor?

sfc-gh-hmadan: that is, IMO we should remove this ctor.

sfc-gh-hmadan: For test convenience, feel free to have a TestUtils.createParameterProvider(isIcebergMode) that ends up passing in a parameters or a props with isIcebergMode set to true.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -152,6 +153,13 @@ private void setParameterMap(
               BUFFER_FLUSH_INTERVAL_IN_MILLIS, MAX_CLIENT_LAG));
     }
 
+    this.checkAndUpdate(
```
sfc-gh-hmadan, line 156: Add comment that this should be the first thing to be set as the final value of this setting can drive the defaulting/logic further on in this method.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -288,7 +296,7 @@ private long getMaxClientLagInMs() {
     Object val =
         this.parameterMap.getOrDefault(
             MAX_CLIENT_LAG,
-            isIcebergMode ? MAX_CLIENT_LAG_ICEBERG_MODE_DEFAULT : MAX_CLIENT_LAG_DEFAULT);
+            isIcebergMode() ? MAX_CLIENT_LAG_ICEBERG_MODE_DEFAULT : MAX_CLIENT_LAG_DEFAULT);
```
sfc-gh-hmadan, line 299: this defaulting already happened in the setParameterMap method, does it need to happen again over here?

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -232,17 +240,17 @@ private void setParameterMap(
 
     this.checkAndUpdate(
         MAX_CLIENT_LAG,
-        isIcebergMode ? MAX_CLIENT_LAG_ICEBERG_MODE_DEFAULT : MAX_CLIENT_LAG_DEFAULT,
+        isIcebergMode() ? MAX_CLIENT_LAG_ICEBERG_MODE_DEFAULT : MAX_CLIENT_LAG_DEFAULT,
```
sfc-gh-hmadan, line 243: do we have a test that makes sure:
1. iceberg mode defaults are what we expect (max chunks = 1 / max client lag = 30 / compression = zstd / etc.)
2. validates that max chunks cannot be set to anything but 1 by the customer

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -468,7 +476,7 @@ public int getMaxChunksInBlob() {
     Object val =
         this.parameterMap.getOrDefault(
             MAX_CHUNKS_IN_BLOB,
-            isIcebergMode ? MAX_CHUNKS_IN_BLOB_ICEBERG_MODE_DEFAULT : MAX_CHUNKS_IN_BLOB_DEFAULT);
+            isIcebergMode() ? MAX_CHUNKS_IN_BLOB_ICEBERG_MODE_DEFAULT : MAX_CHUNKS_IN_BLOB_DEFAULT);
```
sfc-gh-hmadan, line 479: same comment as above, this defaulting was already done, why do all these methods do it again?

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -499,6 +507,12 @@ public boolean isEnableNewJsonParsingLogic() {
     return (val instanceof String) ? Boolean.parseBoolean(val.toString()) : (boolean) val;
   }
 
+  /** @return Whether the client is in Iceberg mode */
+  public boolean isIcebergMode() {
+    Object val = this.parameterMap.getOrDefault(STREAMING_ICEBERG, STREAMING_ICEBERG_DEFAULT);
+    return (val instanceof String) ? Boolean.parseBoolean(val.toString()) : (boolean) val;
```
sfc-gh-hmadan, line 513: The parameter map should be storing validated and parsed values instead of doing it at runtime. This method is called on every row insert and probably multiple times. Such a waste of CPU to do this parseBoolean thosands of times a second!

src/test/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelTest.java
```diff
@@ -89,14 +89,18 @@ public static Object[] isIcebergMode() {
   private SnowflakeStreamingIngestClientInternal<StubChunkData> client;
   private MockSnowflakeServiceClient.ApiOverride apiOverride;
 
+  private Properties prop;
```
sfc-gh-hmadan, line 92: does it really need to be a class member and be part of "test state" ? Add a static method that returns a new Properties object on every call (i see you need props in a testcase down below)

sfc-gh-hmadan: or probably just add a TestUtils.createProps(isIcebergMode) static method and use it everywhere

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-alhuang: Delete ctor and create `createParameterProvider(isIcebergMode)` in TestUtils.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-alhuang: Added.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-alhuang: We have 1 in `ParameterProviderTest.withDefaultValues`, added test for 2.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-alhuang: Added cached value and custom exception for this.

src/test/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelTest.java
sfc-gh-alhuang: Created `createProps`.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-alhuang: Added comment.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-alhuang: This will not be set when both props and parameterOverrides are null.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -79,35 +81,26 @@ public class ParameterProvider {
...
   private final Map<String, Object> parameterMap = new HashMap<>();
 
-  /* Iceberg mode flag */
-  private final boolean isIcebergMode;
-
   // Cached buffer flush interval - avoid parsing each time for quick lookup
   private Long cachedBufferFlushIntervalMs = -1L;
 
+  // Cached enableIcebergStreaming - avoid parsing each time for quick lookup
+  private Boolean cachedEnableIcebergStreaming = null;
```
sfc-gh-hmadan, line 93: curious, why not primitive boolean that's set to false by default?

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -152,6 +142,14 @@ private void setParameterMap(
               BUFFER_FLUSH_INTERVAL_IN_MILLIS, MAX_CLIENT_LAG));
     }
 
+    /* STREAMING_ICEBERG should be the first thing to set as it affects other parameters */
```
sfc-gh-hmadan, line 145: nit: ENABLE_ICEBERG_STREAMING should be ..

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-alhuang: Use null to represent that is not parsed yet. Once parsed will be cached in the `cachedEnableIcebergStreaming` with either true or false.

sfc-gh-hmadan (approved):

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-alhuang: Modified

PR #874, opened 2024-10-25
SNOW-1764301 Iceberg parquet file configuration

sfc-gh-alhuang:
This PR includes following change:
1. Change default compression algorithm to ZSTD when streaming to Iceberg tables.
2. Disable dictionary encoding parameter when use v1 parquet writer which does not support dictionary encoding.


src/main/java/net/snowflake/ingest/streaming/internal/ClientBufferParameters.java
```diff
@@ -89,6 +95,8 @@ public ClientBufferParameters(SnowflakeStreamingIngestClientInternal clientInter
         clientInternal != null
             ? clientInternal.getInternalParameterProvider().isEnableValuesCount()
             : InternalParameterProvider.ENABLE_VALUES_COUNT_DEFAULT;
+    this.enableDictionaryEncoding =
```
sfc-gh-hmadan, line 98: Not very satisfied by how there's not a proliferation of places that control runtime behavior :( 
There's ParameterProvider, internalParameterProvider, clientBuferParameters, and some completely unnecesary defaulting for when clientInternal is null (it should never be not even in tests).

Tried to come up with a better answer but nothing that sticks and is relevant to your current PR.

As a followup whenever you have time i'd suggest first removing the null defaulting, remove the "isTestMode" thing from everywhere (fully doable after a round of changes I did for the ExternalVolume class). Lets see how things look after these two!

sfc-gh-hmadan (approved):

src/main/java/net/snowflake/ingest/streaming/internal/ClientBufferParameters.java
sfc-gh-alhuang: Added [jira](https://snowflakecomputing.atlassian.net/browse/SNOW-1770243). Will merge this after the #874

PR #875, opened 2024-10-27
SNOW-1768252 Ignore getStringForLogging in the requests

sfc-gh-ggeng:
SNOW-1768252 Ignore getStringForLogging in the requests

We introduced the getStringForLogging for SDK side debugging log purpose. However since we do not ignore the property during serialization, we are actually sending the property in the http request entity. This is not entity and would largely increase the payload size of the request. Adding the JsonIgnore to make sure we are not accidentally include the property in the http request


sfc-gh-alhuang (approved):
Looks good. Thanks for the fix!

src/test/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientTest.java
```diff
@@ -414,6 +415,58 @@ public void testDropChannel() throws Exception {
...
+          try {
+            String expectedEntity;
+            if (isIcebergMode) {
+              expectedEntity = String.format(expectEntityFormat, "testPrefix_0", "true");
+            } else {
+              expectedEntity = String.format(expectEntityFormat, "null_0", "false");
+            }
+            String requestEntityString =
+                ServiceResponseHandler.consumeAndReturnResponseEntityAsString(
+                    ((HttpPost) req).getEntity());
```
sfc-gh-hmadan, line 446: Not sure I understand what this test is doing - you're sending a request entity into a method typically used to handle a response entity?
Which means you're not really testing product flow, and really only testing what happens when EntityUtils.toString is called on the request entity.

I think your test can be equally effective with just the following:
1. In executeWithRetries method (that takes IStreamingIngestRequest), extract the code that converts the request into a string into its own public static method (`serializeRequestToString`, say). This code blob i mean:
```
    try {
      payloadInString = objectMapper.writeValueAsString(payload);
    } catch (JsonProcessingException e) {
      throw new SFException(e, ErrorCode.BUILD_REQUEST_FAILURE, message);
    }
```
2. In this testcase, directly call this new method and validate that the response does not have stringForLogging as a key.
3. Undo the change to serviceResponseHandler.

src/test/java/net/snowflake/ingest/streaming/internal/RequestSerializationTest.java
```diff
@@ -0,0 +1,114 @@
...
+    ChannelsStatusRequest request = new ChannelsStatusRequest();
+    request.setChannels(Collections.singletonList(dto));
+    String serializedRequest = StreamingIngestUtils.serializeRequestToString(request, "test");
+    String expectedString =
+        "{\"channels\":[{\"database\":\"db\",\"schema\":\"schemaName\",\"table\":\"tableName\",\"channel_name\":\"channel\",\"client_sequencer\":0}],\"role\":null}";
+    Assert.assertEquals(expectedString, serializedRequest);
+  }
+
+  @Test
+  public void testValidateClientConfigureRequest() {
```
sfc-gh-hmadan, line 107: thanks for being comprehensive and adding a test for each contract!

sfc-gh-hmadan (approved):

PR #877, opened 2024-10-29
[DO NOT MERGE!] SNOW-1176557 Support max lob size to 128MB in client SDK

sfc-gh-ggeng:
SNOW-1176557 Support max lob size increase to 128MB in client SDK

This PR implement the support for max lob size increase to 128MB in client SDK. We shall not merge the PR until the Server side change and fully rollout and stable. This PR also adds tests and demonstrates that the integration with the server side would work.

Other than increase the default value of the max size, there are a few addition changes to support this:

- Since the max lob size has increase significantly, we also need to add the support for customer setting MAX_ALLOWED_ROW_SIZE_IN_BYTES
- We also need to explicitly set the setStreamReadConstraints' max length to make sure that we can serialize/verify the customer input data without error.

We also added a new large lob size example, to show that we could ingest into a four large-lob columns with row size over 400MB in total.

### DIFF
--- a/src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
+++ b/src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
@@ -231,6 +231,13 @@ public class ParameterProvider {
         false /* enforceDefault */);
 
     this.checkAndUpdate(
+        MAX_ALLOWED_ROW_SIZE_IN_BYTES,
+        MAX_ALLOWED_ROW_SIZE_IN_BYTES_DEFAULT,
+        parameterOverrides,
+        props,
+        false /* enforceDefault */);
+
+    this.checkAndUpdate(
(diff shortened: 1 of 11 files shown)


src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -230,6 +230,13 @@ private void setParameterMap(
         props,
         false /* enforceDefault */);
 
+    this.checkAndUpdate(
```
sfc-gh-ggeng, line 233: @sfc-gh-tzhang do you know why in the past we have the parameter here, but the customer can not set it? Is there any explicit reason or just missing implementation? Thanks!

sfc-gh-tzhang:
Closing old PRs, feel free to reopen when it's ready

PR #878, opened 2024-10-29
Add enableIcebergStreaming to telemetry calls

sfc-gh-hmadan:
adding the enableIcebergStreaming boolean flag to all telemetry service calls done by the SDK.
Marking many of the RequestBuilder ctors as Deprecated.


sfc-gh-alhuang (approved):
Looks good, thanks!

PR #880, opened 2024-10-30
SNOW-1774276 Update fileId key in metadata for Iceberg mode

sfc-gh-alhuang:
Update `primaryFileId` to `assignedFullFileName` in metadata when streaming to Iceberg table.

### DIFF
--- a/src/main/java/net/snowflake/ingest/utils/Constants.java
+++ b/src/main/java/net/snowflake/ingest/utils/Constants.java
@@ -36,6 +36,7 @@ public class Constants {
   public static final String SNOWFLAKE_OAUTH_TOKEN_ENDPOINT = "/oauth/token-request";
   public static final String PRIMARY_FILE_ID_KEY =
       "primaryFileId"; // Don't change, should match Parquet Scanner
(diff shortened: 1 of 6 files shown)


src/main/java/net/snowflake/ingest/utils/Constants.java
```diff
@@ -36,6 +36,7 @@ public class Constants {
   public static final String SNOWFLAKE_OAUTH_TOKEN_ENDPOINT = "/oauth/token-request";
   public static final String PRIMARY_FILE_ID_KEY =
       "primaryFileId"; // Don't change, should match Parquet Scanner
+  public static final String ASSIGNED_FULL_FILE_NAME_KEY = "assignedFullFileName";
```
sfc-gh-hmadan, line 39: @sfc-gh-atoth `assignedFullFileName` is what we're embedding into the sdk now, let us know if you have a newer name to embed!

sfc-gh-hmadan (approved):

src/main/java/net/snowflake/ingest/utils/Constants.java
sfc-gh-atoth: Looks good.

sfc-gh-atoth (approved):

PR #888, opened 2024-11-05
SNOW-1787322 Fix InsertError for structured data type

sfc-gh-alhuang:
Currently the `InsertError` doesn't populate extra columns, missing columns, null value for non null columns in `InsertValidationResponse` when ingesting structured data type to Iceberg tables. The PR is fixing this.

We use parquet dot path with escaping dot and back slash character in field name to represent sub-columns. For example, column `x` in `map_col(string, object("a.b" array(object("c\d" object(x int)))))` has a path `MAP_COL.key_value.value.a\.b.list.element.c\\d.x`

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/InsertValidationResponse.java
+++ b/src/main/java/net/snowflake/ingest/streaming/InsertValidationResponse.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Snowflake Computing Inc. All rights reserved.
+ * Copyright (c) 2021-2024 Snowflake Computing Inc. All rights reserved.
  */
 
 package net.snowflake.ingest.streaming;
@@ -115,6 +115,18 @@ public class InsertValidationResponse {
       this.extraColNames = extraColNames;
     }
 
+    /**
+     * Add an extra column name in the input row compared with the table schema
+     *
+     * @param extraColName the extra column name
+     */
+    public void addExtraColName(String extraColName) {
+      if (extraColNames == null) {
+        extraColNames = new ArrayList<>();
+      }
+      extraColNames.add(extraColName);
+    }
+
     /** Get the list of extra column names in the input row compared with the table schema */
     public List<String> getExtraColNames() {
       return extraColNames;
@@ -123,6 +135,18 @@ public class InsertValidationResponse {
     /** Set the missing non-nullable column names in the input row compared with the table schema */
     public void setMissingNotNullColNames(List<String> missingNotNullColNames) {
       this.missingNotNullColNames = missingNotNullColNames;
+    }
+
+    /**
+     * Add a missing non-nullable column name in the input row compared with the table schema
+     *
+     * @param missingNotNullColName the missing non-nullable column name
+     */
+    public void addMissingNotNullColName(String missingNotNullColName) {
+      if (missingNotNullColNames == null) {
+        missingNotNullColNames = new ArrayList<>();
+      }
+      missingNotNullColNames.add(missingNotNullColName);
     }
 
     /**
@@ -139,6 +163,19 @@ public class InsertValidationResponse {
      */
     public void setNullValueForNotNullColNames(List<String> nullValueForNotNullColNames) {
       this.nullValueForNotNullColNames = nullValueForNotNullColNames;
+    }
+
+    /**
+     * Add a name of non-nullable column which have null value in the input row compared with the
+     * table schema
+     *
+     * @param nullValueForNotNullColName the name of non-nullable column which have null value
+     */
+    public void addNullValueForNotNullColName(String nullValueForNotNullColName) {
+      if (nullValueForNotNullColNames == null) {
+        nullValueForNotNullColNames = new ArrayList<>();
+      }
+      nullValueForNotNullColNames.add(nullValueForNotNullColName);
     }
 
     /**
--- a/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
@@ -162,7 +162,12 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
           Set<String> inputColumnNames = verifyInputColumns(row, error, rowIndex);
           rowsSizeInBytes +=
               addRow(
-                  row, rowBuffer.bufferedRowCount, rowBuffer.statsMap, inputColumnNames, rowIndex);
+                  row,
+                  rowBuffer.bufferedRowCount,
+                  rowBuffer.statsMap,
+                  inputColumnNames,
+                  rowIndex,
+                  error);
           rowBuffer.bufferedRowCount++;
         } catch (SFException e) {
           error.setException(e);
@@ -200,7 +205,13 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
       for (Map<String, Object> row : rows) {
         Set<String> inputColumnNames = verifyInputColumns(row, null, tempRowCount);
         tempRowsSizeInBytes +=
-            addTempRow(row, tempRowCount, rowBuffer.tempStatsMap, inputColumnNames, tempRowCount);
+            addTempRow(
+                row,
+                tempRowCount,
+                rowBuffer.tempStatsMap,
+                inputColumnNames,
+                tempRowCount,
+                null /* error */);
         tempRowCount++;
         if ((long) rowBuffer.bufferedRowCount + tempRowCount >= Integer.MAX_VALUE) {
           throw new SFException(ErrorCode.INTERNAL_ERROR, "Row count reaches MAX value");
@@ -249,7 +260,8 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
         try {
           Set<String> inputColumnNames = verifyInputColumns(row, error, rowIndex);
           tempRowsSizeInBytes +=
-              addTempRow(row, tempRowCount, rowBuffer.tempStatsMap, inputColumnNames, rowIndex);
+              addTempRow(
+                  row, tempRowCount, rowBuffer.tempStatsMap, inputColumnNames, rowIndex, error);
           tempRowCount++;
         } catch (SFException e) {
           error.setException(e);
@@ -575,6 +587,7 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
    * @param formattedInputColumnNames list of input column names after formatting
    * @param insertRowIndex Index of the rows given in insertRows API. Not the same as
    *     bufferedRowIndex
+   * @param error
    * @return row size
    */
   abstract float addRow(
@@ -582,7 +595,8 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
       int bufferedRowIndex,
       Map<String, RowBufferStats> statsMap,
       Set<String> formattedInputColumnNames,
-      final long insertRowIndex);
+      final long insertRowIndex,
+      InsertValidationResponse.InsertError error);
 
   /**
    * Add an input row to the temporary row buffer.
@@ -595,6 +609,7 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
    * @param statsMap column stats map
    * @param formattedInputColumnNames list of input column names after formatting
    * @param insertRowIndex index of the row being inserteed from User Input List
+   * @param error
    * @return row size
    */
   abstract float addTempRow(
@@ -602,7 +617,8 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
       int curRowIndex,
       Map<String, RowBufferStats> statsMap,
       Set<String> formattedInputColumnNames,
-      long insertRowIndex);
+      long insertRowIndex,
+      InsertValidationResponse.InsertError error);
 
   /** Move rows from the temporary buffer to the current row buffer. */
   abstract void moveTempRowsToActualBuffer(int tempRowCount);
--- a/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
@@ -211,7 +211,7 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
                 rowBuffer.tempStatsMap,
                 inputColumnNames,
                 tempRowCount,
-                null /* error */);
+                new InsertValidationResponse.InsertError(row, 0) /* dummy error */);
         tempRowCount++;
         if ((long) rowBuffer.bufferedRowCount + tempRowCount >= Integer.MAX_VALUE) {
           throw new SFException(ErrorCode.INTERNAL_ERROR, "Row count reaches MAX value");
@@ -587,7 +587,8 @@ abstract class AbstractRowBuffer<T> implements RowBuffer<T> {
    * @param formattedInputColumnNames list of input column names after formatting
    * @param insertRowIndex Index of the rows given in insertRows API. Not the same as
    *     bufferedRowIndex
-   * @param error
+   * @param error Insert error object, used to populate error details when doing structured data
+   *     type parsing
    * @return row size
    */
   abstract float addRow(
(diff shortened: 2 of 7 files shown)


src/main/java/net/snowflake/ingest/streaming/InsertValidationResponse.java
```diff
@@ -125,6 +137,18 @@ public void setMissingNotNullColNames(List<String> missingNotNullColNames) {
       this.missingNotNullColNames = missingNotNullColNames;
     }
 
+    /**
+     * Add a missing non-nullable column name in the input row compared with the table schema
+     *
+     * @param missingNotNullColName the missing non-nullable column name
+     */
+    public void addMissingNotNullColName(String missingNotNullColName) {
```
sfc-gh-hmadan, line 145: any chance this can be called from multiple threads?

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -200,7 +205,13 @@ public InsertValidationResponse insertRows(
...
         Set<String> inputColumnNames = verifyInputColumns(row, null, tempRowCount);
         tempRowsSizeInBytes +=
-            addTempRow(row, tempRowCount, rowBuffer.tempStatsMap, inputColumnNames, tempRowCount);
+            addTempRow(
+                row,
+                tempRowCount,
+                rowBuffer.tempStatsMap,
+                inputColumnNames,
+                tempRowCount,
+                null /* error */);
```
sfc-gh-hmadan, line 214: the called method now has to know that a null can be passed in, adds to unnecessary code complexity. We should either mark the type as Optional to make it clear a null can be passed in, or just pass in an error even if its not going to be used. The latter is preferable IMO.

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -575,14 +587,16 @@ public ChannelData<T> flush() {
    * @param formattedInputColumnNames list of input column names after formatting
    * @param insertRowIndex Index of the rows given in insertRows API. Not the same as
    *     bufferedRowIndex
+   * @param error
```
sfc-gh-hmadan, line 590: pl fill this in (and also why an error object is needed in addRow)

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -161,12 +173,16 @@ private static ParquetBufferValue parseColumnValueToParquet(
...
             insertRowsCurrIndex,
-            isDescendantsOfRepeatingGroup);
+            isDescendantsOfRepeatingGroup,
+            error);
       }
     }
 
     if (value == null) {
       if (type.isRepetition(Repetition.REQUIRED)) {
+        if (error != null) {
```
sfc-gh-hmadan, line 183: call getDotPath once and reuse it in both line 184 and 191.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -161,12 +173,16 @@ private static ParquetBufferValue parseColumnValueToParquet(
...
+          error.addNullValueForNotNullColName(subColumnFinder.getDotPath(id));
```
sfc-gh-hmadan, line 184: There is a critical gap that we're introducing here. Lets say a map evolves have three columns as nullable. It will take three rounds of insertRow -> get insertError with one field path called out -> alter table -> insert Row again -> get insertError with a second field path called out -> and so on.

Similarly for the other kinds of errors (extra columns / missing columns).

The right way to fix this is to only throw at the end after insertRow has processed all columns and collected all errors in the error object you're passing around. Instead of throwing on the first problem encountered.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -430,28 +449,71 @@ private static ParquetBufferValue getStructValue(
...
       ZoneId defaultTimezone,
       final long insertRowsCurrIndex,
-      boolean isDescendantsOfRepeatingGroup) {
+      boolean isDescendantsOfRepeatingGroup,
+      InsertValidationResponse.InsertError error) {
     Map<String, ?> structVal =
         DataValidationUtil.validateAndParseIcebergStruct(
             subColumnFinder.getDotPath(type.getId().toString()), value, insertRowsCurrIndex);
     Set<String> extraFields = new HashSet<>(structVal.keySet());
+    List<String> missingFields = new ArrayList<>();
```
sfc-gh-hmadan, line 458: also need handling around non-nullable fields checking inside the struct?

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -430,28 +449,71 @@ private static ParquetBufferValue getStructValue(
...
+                type.getType(i),
+                statsMap,
+                subColumnFinder,
+                defaultTimezone,
+                insertRowsCurrIndex,
+                isDescendantsOfRepeatingGroup,
+                error);
+        listVal.add(parsedValue.getValue());
+        estimatedParquetSize += parsedValue.getSize();
+      } else {
```
sfc-gh-hmadan, line 476: this else block is (a) not incrementing estimatedParquetSize, and (b) not adding to listVal in one branch (when required=true). Unclear why you really need this else block?
You can also do the following I think and avoid the unnecessary if-else branching?

```
String fieldName = type.getFieldName(i);
Object val = structVal.getOrDefault(fieldName, null);
ParquetBufferValue parsedValue = parseColumnValueToParquet(val, ....);
listVal.add(parsedValue.getValue());
estParquetSize = ...;
if (type.getType(i).isRepetition(REQUIRED)) {
    missingFields.add(fieldName);
}
```

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -430,28 +449,71 @@ private static ParquetBufferValue getStructValue(
...
+        if (type.getType(i).isRepetition(Repetition.REQUIRED)) {
+          missingFields.add(type.getFieldName(i));
+        } else {
+          listVal.add(null);
+        }
+      }
+    }
+
+    if (!missingFields.isEmpty()) {
+      missingFields.forEach(
```
sfc-gh-hmadan, line 486: please use regular for loop to avoid the unnecessary indentation / lambda!

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -430,28 +449,71 @@ private static ParquetBufferValue getStructValue(
...
+          missingField -> {
+            if (error != null) {
```
sfc-gh-hmadan, line 488: left this comment in another place too - we should mark error as Nonnull and not allow a null error to be passed around.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
```diff
@@ -430,28 +449,71 @@ private static ParquetBufferValue getStructValue(
...
+              List<String> missingFieldPath =
+                  new ArrayList<>(subColumnFinder.getPath(type.getId().toString()));
+              missingFieldPath.add(missingField);
+              error.addMissingNotNullColName(
+                  concatDotPath(missingFieldPath.toArray(new String[0])));
+            }
+          });
+      String missingFieldsStr = missingFields.stream().collect(Collectors.joining(", ", "[", "]"));
+      throw new SFException(
```
sfc-gh-hmadan, line 497: left this comment in another place too - we should not be throwing exceptions on extra cols / missing cols / unexpected-null-value until all of these errors have been collected. Right now we are throwing on the first possible error which is a deviation from current behavior and makes KC's life harder.

src/main/java/net/snowflake/ingest/utils/SubColumnFinder.java
```diff
@@ -52,6 +54,13 @@ public String getDotPath(String id) {
     return accessMap.get(id).dotPath;
   }
 
+  public List<String> getPath(String id) {
```
sfc-gh-hmadan, line 57: you're returning internal state and the callsite can mutate this list accidentally.
Please return an UnmodifiableList from this API.

src/main/java/net/snowflake/ingest/utils/SubColumnFinder.java
```diff
@@ -20,11 +20,13 @@ static class SubtreeInfo {
     final int startTag;
     final int endTag;
     final String dotPath;
+    final List<String> path;
```
sfc-gh-hmadan, line 23: nit: add tests to SubColumnFinderTest too? Feel free to do it with the testing gaps JIRA.

src/main/java/net/snowflake/ingest/streaming/InsertValidationResponse.java
sfc-gh-alhuang: Insert error instance are independent between rows, I think this should be safe.

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
sfc-gh-alhuang: Make sense, passed a dummy error.

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
sfc-gh-alhuang: Added description.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: Called once at the start.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: Deferred all exception after the whole row is parsed. Added a IT of this.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: `IcebergParquetValueParser.parseColumnValueToParquet` will be called if and only if the value exist (could be null). The non-nullable error is handled there.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: The reason to add this else is to distinguish between missing column and column with null value. That is, we cannot use `getOrDefault(fieldName, null)` as this ruined the difference between null value and missing key.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: Changed to native for loop.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-alhuang: Removed null check and passed in dummy error.

src/main/java/net/snowflake/ingest/utils/SubColumnFinder.java
sfc-gh-alhuang: Made it to `UnmodifiableList()`

src/main/java/net/snowflake/ingest/utils/SubColumnFinder.java
sfc-gh-alhuang: Added to test gap [JIRA](https://snowflakecomputing.atlassian.net/browse/SNOW-1490151)

src/main/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParser.java
sfc-gh-hmadan: ah, got it. please add a comment around this.

src/main/java/net/snowflake/ingest/utils/Utils.java
```diff
@@ -417,20 +417,21 @@ public static String getFullyQualifiedChannelName(
...
-      if (isNullOrEmpty(p)) {
-        throw new IllegalArgumentException("Path cannot be null or empty");
+      if (p == null) {
+        throw new IllegalArgumentException("Path cannot be null");
       }
       if (sb.length() > 0) {
         sb.append(".");
       }
-      sb.append(p);
+      sb.append(p.replace("\\", "\\\\").replace(".", "\\."));
```
sfc-gh-hmadan, line 434: does an empty string in p need to also be replaced by something? Have IT for an empty field name?

src/test/java/net/snowflake/ingest/streaming/internal/IcebergParquetValueParserTest.java
```diff
@@ -85,7 +85,7 @@ public void parseValueInt() {
         };
     ParquetBufferValue pv =
         IcebergParquetValueParser.parseColumnValueToParquet(
-            Integer.MAX_VALUE, type, rowBufferStatsMap, mockSubColumnFinder, UTC, 0);
+            Integer.MAX_VALUE, type, rowBufferStatsMap, mockSubColumnFinder, UTC, 0, null);
```
sfc-gh-hmadan, line 88: thanks for not adding a method overload :)

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergStructuredIT.java
```diff
@@ -236,6 +244,180 @@ public void testFieldName() throws Exception {
...
+  }
+
+  @Test
+  public void testMultipleErrors() throws Exception {
+    String tableName =
+        createIcebergTable(
+            "object(k1 int not null, k2 object(k3 int not null, k4 object(k5 int not null) not"
+                + " null) not null) not null");
+    SnowflakeStreamingIngestChannel channel =
+        openChannel(tableName, OpenChannelRequest.OnErrorOption.ABORT);
```
sfc-gh-hmadan, line 384: looks like a pre-existing gap where someone using OnErrorOption.ABORT will not get the same set of information (insertErrors object) as OnErrorOption.CONTINUE ? (unrelated to your PR)

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergStructuredIT.java
```diff
@@ -236,6 +244,180 @@ public void testFieldName() throws Exception {
...
+        openChannel(tableName, OpenChannelRequest.OnErrorOption.CONTINUE);
+    Map<String, Object> row = new HashMap<>();
+    row.put("k2", 1);
+    row.put("k.3", 1);
+    row.put("k\\4", 1);
+    InsertValidationResponse insertValidationResponse =
+        channel.insertRow(createStreamingIngestRow(row), UUID.randomUUID().toString());
+    assertThat(insertValidationResponse.getInsertErrors().size()).isEqualTo(1);
+    assertThat(insertValidationResponse.getInsertErrors().get(0).getExtraColNames())
+        .containsOnly("VALUE.k2", "VALUE.k\\.3", "VALUE.k\\\\4");
```
sfc-gh-hmadan, line 260: FOR KC team: The first VALUE is the default column name used by the test harness, and doesn't have any semantics. This is unlike the key-value/list/element/value words that you'll see in the examples that follow.

sfc-gh-hmadan:
Signing off but please hold off on merging until we get a positive ack from warsaw on this change being sufficient, slack conversation ongoing.

sfc-gh-hmadan (approved):

src/main/java/net/snowflake/ingest/utils/Utils.java
sfc-gh-alhuang: I don't think empty string does not need to be replaced. An empty string does not collision with other dot path. An IT with empty fields was included in previous [PR](https://github.com/snowflakedb/snowflake-ingest-java/pull/869) in `IcebergStructuredIT.testFieldName`.

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergStructuredIT.java
sfc-gh-alhuang: For single row, errors on different OnErrorOption should be the same.

PR #891, opened 2024-11-07
Fix quoted column name for Iceberg primitive type

sfc-gh-alhuang:
For non-structured data type column, server sides perform EP info check which requires the column display name to be exactly the same as the quoted column name. Fix a bug on this and extend existing column name test to run on iceberg as in preprod as well. Disable the Iceberg test now the pass the merge gate.

### DIFF
--- a/src/test/java/net/snowflake/ingest/streaming/internal/it/ColumnNamesIT.java
+++ b/src/test/java/net/snowflake/ingest/streaming/internal/it/ColumnNamesIT.java
@@ -1,3 +1,7 @@
+/*
+ * Copyright (c) 2024 Snowflake Computing Inc. All rights reserved.
+ */
+
 package net.snowflake.ingest.streaming.internal.it;
 
 import java.sql.ResultSet;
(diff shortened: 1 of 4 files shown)


src/test/java/net/snowflake/ingest/streaming/internal/it/ColumnNamesIT.java
```diff
@@ -13,17 +17,45 @@
...
+import org.junit.runners.Parameterized;
 
 public class ColumnNamesIT extends AbstractDataTypeTest {
   private static final int INGEST_VALUE = 1;
 
+  @Parameterized.Parameters(
+      name = "enableIcebergStreaming={0}, compressionAlgorithm={1}, icebergSerializationPolicy={2}")
+  public static Object[][] parameters() {
+    return new Object[][] {
+      // TODO: Enable this after Iceberg testing is set up on sfctest0 GCP / Azure
```
sfc-gh-hmadan, line 34: With our plan to run two IT suites in parallel for iceberg/non-iceberg, how will this class be run in one mode in suite 1 and another mode in suite 2?

src/test/java/net/snowflake/ingest/streaming/internal/it/ColumnNamesIT.java
```diff
@@ -13,17 +17,45 @@
...
+      //      {true, "ZSTD", Constants.IcebergSerializationPolicy.COMPATIBLE},
+      //      {true, "ZSTD", Constants.IcebergSerializationPolicy.OPTIMIZED},
+      {false, "GZIP", null},
+      {false, "ZSTD", null},
```
sfc-gh-hmadan, line 38: why do we need gzip and std testing in this IT? This IT is only validating the column names / EP handling, file compression doesn't need to be tested in this one.

src/test/java/net/snowflake/ingest/streaming/internal/it/ColumnNamesIT.java
```diff
@@ -252,7 +297,12 @@ private void testInsertRowFails(
   private String createSimpleTable(String createTableColumnName) throws SQLException {
     String tableName = "a" + UUID.randomUUID().toString().replace("-", "_");
     String createTableSql =
-        String.format("create table %s (%s int);", tableName, createTableColumnName);
+        String.format(
+            "create %s table %s (%s int) %s",
+            enableIcebergStreaming ? "iceberg" : "",
+            tableName,
+            createTableColumnName,
+            enableIcebergStreaming ? getIcebergTableConfig(tableName) : "");
```
sfc-gh-hmadan, line 305: IMO we should refactor this to have a method that takes in the table name and column schema list only, and adds the proper prefix (iceberg table vs. table) and proper suffix (catalog and external_volume and base_location).

Instead of every create table call having to worry about both these things.

If this PR is good to merge am ok with this test refactoring in the next PR. Please also pick up any deferred comments from the last PR if any!

sfc-gh-hmadan (approved):

src/test/java/net/snowflake/ingest/streaming/internal/it/ColumnNamesIT.java
sfc-gh-alhuang: Will split this into two separate tests in later PR.

src/test/java/net/snowflake/ingest/streaming/internal/it/ColumnNamesIT.java
sfc-gh-alhuang: Will split this into two separate tests in later PR.

PR #898, opened 2024-11-11
Introduce Iceberg streaming merge gate

sfc-gh-alhuang:
Setup merge gates for Iceberg streaming. For now we only supports AWS. Disabled a string test, will add it back after SNOW-1798403 is fixed.

### DIFF
--- a/src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergStringIT.java
+++ b/src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergStringIT.java
@@ -94,13 +94,12 @@ public class IcebergStringIT extends AbstractDataTypeTest {
         Arrays.asList(StringUtils.repeat("a", 33), StringUtils.repeat("*", 3), null, ""),
         "select MAX(LENGTH({columnName})) from {tableName}",
         Arrays.asList(33L));
-    // TODO: Add this back after the following JIRA is fixed
-    // SNOW-1798403 Error when querying maximum values on string columns of managed Iceberg tables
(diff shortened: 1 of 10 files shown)


src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergStringIT.java
```diff
@@ -84,15 +89,18 @@ public void testStringAndQueries() throws Exception {
...
-        "string",
-        Arrays.asList(StringUtils.repeat("a", 16 * 1024 * 1024), null, null, null, "aaa"),
-        "select MAX({columnName}) from {tableName}",
-        Arrays.asList(StringUtils.repeat("a", 16 * 1024 * 1024)));
     testIcebergIngestAndQuery(
         "string",
         Arrays.asList(StringUtils.repeat("a", 33), StringUtils.repeat("*", 3), null, ""),
         "select MAX(LENGTH({columnName})) from {tableName}",
         Arrays.asList(33L));
+    // TODO: Add this back after the following JIRA is fixed
```
sfc-gh-hmadan, line 97: Lets keep the test in with the 16MB-1 hack that you have so it doesn't regress further?

sfc-gh-hmadan (approved):

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergStringIT.java
sfc-gh-alhuang: Added back

PR #906, opened 2024-11-14
SNOW-1806157 Fix field id mismatch during schema evolution for structured data type

sfc-gh-alhuang:
Fix the error when the column ordinal of structured data is out of order. The logic used to assume that the root columns are the first k smallest of whole schema, which is incorrect after schema evolution. Add a map for column ordinal to column to fix this.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
@@ -195,6 +195,11 @@ public class ParquetRowBuffer extends AbstractRowBuffer<ParquetChunkData> {
      *   F7:             ordinal=7, fieldId=0
      */
     if (clientBufferParameters.isEnableIcebergStreaming()) {
+      Map<Integer, ColumnMetadata> ordinalToColumn = new HashMap<>();
+      for (ColumnMetadata column : columns) {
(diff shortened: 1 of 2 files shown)


sfc-gh-hmadan (approved):

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -232,7 +237,15 @@ public void setupSchema(List<ColumnMetadata> columns) {
...
          */
         String columnDisplayName =
-            isPrimitiveColumn ? columns.get(ordinal - 1).getName() : columnDotPath;
+            isPrimitiveColumn
+                ? Optional.ofNullable(ordinalToColumn.get(ordinal))
+                    .orElseThrow(
+                        () ->
+                            new SFException(
+                                ErrorCode.INTERNAL_ERROR,
+                                String.format("Column not found. ordinal=%d.", ordinal)))
```
sfc-gh-hmadan, line 246: also log more info like columnDotPath / columnDescriptor / fieldId ?

sfc-gh-hmadan: github.dev didn't post this comment just realized!

PR #907, opened 2024-11-15
Use role SYSADMIN for testing

sfc-gh-alhuang:
The original testing profile use ACCOUNTADMIN, change it to SYSADMIN.


src/test/java/net/snowflake/ingest/TestUtils.java
```diff
@@ -302,6 +302,7 @@ public static Connection getConnection(boolean isStreamingConnection) throws Exc
     props.put("warehouse", warehouse);
     props.put("client_session_keep_alive", "true");
     props.put("privateKey", privateKey);
+    props.put("role", role);
```
sfc-gh-hmadan, line 306: i'm guessing you've updated the role in the files to be SYSADMIN too..
Did that not require any special permissions in any environment, for instance to grant usage on EXT VOL to role SYSADMIN?

sfc-gh-alhuang: Yes I've grantde permission on those sfctest0 accounts.

sfc-gh-xhuang (approved):

PR #908, opened 2024-11-15
SNOW-1808626 Fix struct type statistic for missing fields

sfc-gh-alhuang:
The original logic didn't calculate null count for missing fields, fix this.

### DIFF
--- a/src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergStructuredIT.java
+++ b/src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergStructuredIT.java
@@ -7,6 +7,7 @@ package net.snowflake.ingest.streaming.internal.datatypes;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergStructuredIT.java
```diff
@@ -464,4 +471,25 @@ private void assertMap(String dataType, Map<?, ?> value) throws Exception {
     JsonNode expectedNode = value == null ? null : objectMapper.valueToTree(value);
     assertThat(actualNode).isEqualTo(expectedNode);
   }
+
+  private void removeNullFields(JsonNode node) {
```
sfc-gh-hmadan, line 475: add comment on why this method is needed?

Also we should have null value comparison too no?

sfc-gh-hmadan (approved):

src/test/java/net/snowflake/ingest/streaming/internal/datatypes/IcebergStructuredIT.java
sfc-gh-alhuang: This method is used to compare {} and {key: null}. As this two values are the same for iceberg struct.

sfc-gh-hmadan: I see. we pass in key=null, but we get back missing-key?

sfc-gh-alhuang: Yes, both are the same.

PR #911, opened 2024-11-19
SNOW-1814204 Use fair lock to ensure granting sequence

sfc-gh-ggeng:
We should make sure that the long waiting thread get the lock. Otherwise, client’s later insertRow request with later offset token may come and get the lock first -- it will get the earlier internal continuation token and insert first. This could cause issues when incident happens and customer trying to replay the data. This may also cause the data ingested not in sequence.

We should use the fair synch lock to make sure that the thread with the longest waiting time (come earlier) wins and get the lock. Doc for the ReentrantLock:  https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/ReentrantLock.html


sfc-gh-xhuang (approved):

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
```diff
@@ -357,7 +357,7 @@ public InsertValidationResponse insertRows(
     this.channelState = channelRuntimeState;
     this.channelFullyQualifiedName = fullyQualifiedChannelName;
     this.nonNullableFieldNames = new HashSet<>();
-    this.flushLock = new ReentrantLock();
+    this.flushLock = new ReentrantLock(true);
```
sfc-gh-tzhang, line 360: Is there any perf implication with this change?

sfc-gh-tzhang (approved):
lgtm, let's make sure the perf implication for normal use cases is minimal

src/main/java/net/snowflake/ingest/streaming/internal/AbstractRowBuffer.java
sfc-gh-ggeng: There could be perf implication main when we are doing the expected things.
When the customer have threads waiting for the lock and a new thread comes to acquire the lock, for unfairness lock, we would directly give the lock to the new thread; while for fairness lock, we would need to wait for the first come thread to wake up --  meaning that the performance implication happens when we are not inserting the data according to the customer's order.

sfc-gh-hmadan:
Please add a test to verify this behavior never regresses?

PR #912, opened 2024-11-19
Make Channel.getIcebergSchema() API public + Stop NDV tracking

sfc-gh-hmadan:
- logging fixes
- make api public for schema evolution in KC
- stop ndv tracking since its causing ingestion to massively slow down.


sfc-gh-alhuang (approved):
Looks good, thanks!

PR #916, opened 2024-11-21
SNOW-1821246 Add parameter for blob upload timeout

sfc-gh-alhuang:
The original blob upload timeout is hard coded as 5 seconds. This might not be feasible for Iceberg table streaming as the 30 seconds client lag in Iceberg streaming might create larger file which requires larger blob upload timeout. Make the `BLOB_UPLOAD_TIMEOUT` constant to configurable parameter and keep the default timeout of FDN table streaming as 5 seconds while have a 20 seconds default for Iceberg table streaming.

### DIFF
--- a/src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
+++ b/src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
@@ -38,6 +38,8 @@ public class ParameterProvider {
   public static final String MAX_CHUNKS_IN_BLOB = "MAX_CHUNKS_IN_BLOB".toLowerCase();
   public static final String MAX_CHUNKS_IN_REGISTRATION_REQUEST =
       "MAX_CHUNKS_IN_REGISTRATION_REQUEST".toLowerCase();
+  public static final String BLOB_UPLOAD_TIMEOUT_IN_SEC =
+      "BLOB_UPLOAD_TIMEOUT_IN_SEC".toLowerCase();
 
   public static final String MAX_CLIENT_LAG = "MAX_CLIENT_LAG".toLowerCase();
 
@@ -62,6 +64,8 @@ public class ParameterProvider {
   public static final long MAX_MEMORY_LIMIT_IN_BYTES_DEFAULT = -1L;
   public static final long MAX_CHANNEL_SIZE_IN_BYTES_DEFAULT = 64 * 1024 * 1024;
   public static final long MAX_CHUNK_SIZE_IN_BYTES_DEFAULT = 256 * 1024 * 1024;
+  public static final int BLOB_UPLOAD_TIMEOUT_IN_SEC_DEFAULT = 5;
+  public static final int BLOB_UPLOAD_TIMEOUT_IN_SEC_ICEBERG_MODE_DEFAULT = 20;
 
   // Lag related parameters
   public static final long MAX_CLIENT_LAG_DEFAULT = 1000; // 1 second
@@ -94,6 +98,9 @@ public class ParameterProvider {
 
   // Cached enableIcebergStreaming - avoid parsing each time for quick lookup
   private Boolean cachedEnableIcebergStreaming = null;
+
+  // Cached blob upload timeout - avoid parsing each time for quick lookup
+  private int cachedBlobUploadTimeoutInSec = -1;
 
   /**
    * Constructor. Takes properties from profile file and properties from client constructor and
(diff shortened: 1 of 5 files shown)


sfc-gh-xhuang (approved):

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -38,6 +38,8 @@ public class ParameterProvider {
   public static final String MAX_CHUNKS_IN_BLOB = "MAX_CHUNKS_IN_BLOB".toLowerCase();
   public static final String MAX_CHUNKS_IN_REGISTRATION_REQUEST =
       "MAX_CHUNKS_IN_REGISTRATION_REQUEST".toLowerCase();
+  public static final String BLOB_UPLOAD_TIMEOUT_IN_SEC =
+      "BLOB_UPLOAD_TIMEOUT_IN_SEC".toLowerCase();
```
sfc-gh-hmadan, line 42: since this is user facing lets make it "in_seconds" instead of "in_sec"

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
```diff
@@ -62,6 +64,8 @@ public class ParameterProvider {
   public static final long MAX_MEMORY_LIMIT_IN_BYTES_DEFAULT = -1L;
   public static final long MAX_CHANNEL_SIZE_IN_BYTES_DEFAULT = 64 * 1024 * 1024;
   public static final long MAX_CHUNK_SIZE_IN_BYTES_DEFAULT = 256 * 1024 * 1024;
+  public static final int BLOB_UPLOAD_TIMEOUT_IN_SEC_DEFAULT = 5;
+  public static final int BLOB_UPLOAD_TIMEOUT_IN_SEC_ICEBERG_MODE_DEFAULT = 20;
```
sfc-gh-hmadan, line 68: 30 as default is good imo.

src/test/java/net/snowflake/ingest/streaming/internal/RegisterServiceTest.java
```diff
@@ -59,7 +58,17 @@ public void teardown() throws Exception {
 
   @Test
   public void testRegisterService() throws ExecutionException, InterruptedException {
-    RegisterService<StubChunkData> rs = new RegisterService<>(null, true);
+    SnowflakeStreamingIngestClientInternal<StubChunkData> client =
+        Mockito.mock(SnowflakeStreamingIngestClientInternal.class);
+    ParameterProvider parameterProvider = Mockito.mock(ParameterProvider.class);
```
sfc-gh-hmadan, line 63: curious why do we need to mock the parameter provider, can we not do away with this unnecessary mocking (not for this PR) ?

sfc-gh-hmadan (approved):

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-alhuang: Fixed.

src/main/java/net/snowflake/ingest/utils/ParameterProvider.java
sfc-gh-alhuang: Changed to 30

src/test/java/net/snowflake/ingest/streaming/internal/RegisterServiceTest.java
sfc-gh-alhuang: We need this as `RegisterServiceTest` is reading blob upload timeout from parameter provider.

sfc-gh-tzhang:
Want to make sure we understand the upload logic correctly, the total wait time is NOT 5 seconds, the total time is BLOB_UPLOAD_TIMEOUT_IN_SEC * BLOB_UPLOAD_MAX_RETRY_COUNT which is 120s, and currently the BLOB_UPLOAD_MAX_RETRY_COUNT is already configurable. Ideally we do not want to exposing more and more configurable parameters even though they're not documented. And please make sure you have read the comment in the upload function
```
        // In order to guarantee fairness between the time spent on waiting blob uploading VS blob
        // registering and make sure the delay on server side commit is relatively small:
        // 1. If no exception and total time is less than or equal to
        // BLOB_UPLOAD_TIMEOUT_IN_SEC * 2, we will wait for all blobs to be uploaded and then
        // register them
        // 2. If no exception and total time is bigger than BLOB_UPLOAD_TIMEOUT_IN_SEC * 2, we
        // will break the waiting and register the processed blob first
        // 3. If hitting non-timeout exception, we will skip the current blob and continue on
        // processing next blob
        // 4. If hitting timeout exception, we will break the waiting if we haven't reached
        // BLOB_UPLOAD_TIMEOUT_IN_SEC * BLOB_UPLOAD_MAX_RETRY_COUNT and register the
        // processed blob first. Otherwise, we will skip the current blob and continue on processing
        // next blob if time out has been reached.
```

sfc-gh-tzhang (changes requested):
Please see my other comment, thanks

sfc-gh-alhuang:
Closed this PR as the customer can configure max retry count.

PR #917, opened 2024-11-22
SNOW-1813772 Enable old IT suites for iceberg tables

sfc-gh-alhuang:
Enable all remaining FDN only ITs for Iceberg table streaming.


src/test/java/net/snowflake/ingest/streaming/internal/it/BigFilesITBase.java
```diff
@@ -17,16 +22,14 @@
...
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 import org.junit.runners.Parameterized.Parameter;
-import org.junit.runners.Parameterized.Parameters;
 
 /** Ingest large amount of rows. */
 @RunWith(Parameterized.class)
-public class StreamingIngestBigFilesIT {
+public abstract class BigFilesITBase {
```
sfc-gh-hmadan, line 32: is this passing?

sfc-gh-alhuang: Yes it's passing with the JDBC uploader. The max chunk size in this test is 19MB.

src/test/java/net/snowflake/ingest/streaming/internal/it/BigFilesITBase.java
```diff
@@ -35,16 +38,15 @@ public class StreamingIngestBigFilesIT {
...
-  @Parameters(name = "{index}: {0}")
-  public static Object[] compressionAlgorithms() {
-    return new Object[] {"GZIP", "ZSTD"};
-  }
+  @Parameter(0)
+  public String compressionAlgorithm;
 
-  @Parameter public String compressionAlgorithm;
+  @Parameter(1)
+  public Constants.IcebergSerializationPolicy icebergSerializationPolicy;
```
sfc-gh-hmadan, line 47: lets remove the runwith parameterized from the base class, and remove the parameter annotations from here too. Makes class design murky.
Let the derived classes have the runwith and the @Parameter, and pass in the values via beforeAll.

I'm also debating the value of running this IT for iceberg when you've already authored a much better large file test thats in my PR rn.

sfc-gh-hmadan: clearly its not exercising multipart upload, if it was iceberg tests would've failed.

sfc-gh-alhuang: Removed this IT and use the on with larger file size for Iceberg streaming.

PR #922, opened 2024-12-02
Send clientname and clientkey with every registerblobs request for metering views

sfc-gh-hmadan:
Iceberg's billing needs the clientname and clientkey at the time of DML Commit. While it can be read via a S3 get-metadata call, its a wasteful approach. This PR optimizes away that S3 call by sending the metadata inline with the registerBlobs API call.

Ran FDN and Iceberg ITs locally against sfctest0 to make sure the json backcompat fix from months ago will allow unknown fields to appear on the payload without causing issues. IT run from this PR's merge gate should be green.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
@@ -181,8 +181,8 @@ class InternalStage implements IStorage {
                 fileTransferMetadataCopy,
                 inStream,
                 proxyProperties,
-                clientPrefix,
                 clientName,
+                clientPrefix,
                 fullFilePath));
       } else {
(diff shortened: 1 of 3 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
```diff
@@ -181,8 +181,8 @@ private Optional<String> putRemote(String fullFilePath, byte[] data, int retryCo
                 fileTransferMetadataCopy,
                 inStream,
                 proxyProperties,
-                clientPrefix,
                 clientName,
+                clientPrefix,
```
sfc-gh-xhuang, line 185: reordered?

sfc-gh-xhuang (approved):

sfc-gh-alhuang (approved):
Looks good, thanks!

PR #929, opened 2025-01-14
[SNOW-1882616] Error out for duplicate keys in variant

sfc-gh-ggeng:
The SDK should not ingest the variant like {"key": 1, "key": 2} otherwise, we would have issues in the XP during select and there is no good mitigation on that.

In this change, we would error out with INVALID_VALUE_ROW for such duplicate cases.

Added Unit test for the change.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
@@ -181,7 +181,12 @@ class DataValidationUtil {
           throw valueFormatNotAllowedException(
               columnName, snowflakeType, "Not a valid JSON: duplicate field", insertRowIndex);
         }
-        throw new SFException(e, ErrorCode.IO_ERROR, "Cannot create JSON Parser or JSON generator");
+        throw new SFException(
+            e,
+            ErrorCode.IO_ERROR,
+            String.format(
+                "Cannot create JSON Parser or JSON generator for column %s of type %s, rowIndex:%d",
+                columnName, snowflakeType, insertRowIndex));
       }
       // We return the minified string from the result writer
       return resultWriter.toString();
--- a/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
@@ -78,7 +78,8 @@ class DataValidationUtil {
 
   private static final ObjectMapper objectMapper = new ObjectMapper();
 
-  private static final JsonFactory factory = new JsonFactory();
+  private static final JsonFactory factory =
+      new JsonFactory().configure(JsonGenerator.Feature.STRICT_DUPLICATE_DETECTION, true);
 
   // The version of Jackson we are using does not support serialization of date objects from the
   // java.time package. Here we define a module with custom java.time serializers. Additionally, we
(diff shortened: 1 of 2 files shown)


src/test/java/net/snowflake/ingest/streaming/internal/DataValidationUtilTest.java
```diff
@@ -856,6 +856,29 @@ public void testValidateAndParseObject() throws Exception {
         () -> validateAndParseObjectNew("COL", Collections.singletonMap("foo", new Object()), 0));
   }
 
+  @Test
+  public void testValidateDuplicateKeys() {
```
sfc-gh-lsembera, line 860: Could we add a test that tests `validateAndParseVariantNew`, as well as `validateAndParseArrayNew` (array of objects with duplicated keys)?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -176,6 +177,10 @@ private static String validateAndParseSemiStructured(
         throw valueFormatNotAllowedException(
             columnName, snowflakeType, "Not a valid JSON", insertRowIndex);
       } catch (IOException e) {
+        if (e.getMessage().contains("Duplicate field")) {
+          throw valueFormatNotAllowedException(
```
sfc-gh-dorlovsky, line 181: Let's 'chain' exceptions like you do below: pass the original `e` exception to the constructor of the one you throw. This will preserve information contained in the original exception.

I wonder if the "Not a valid JSON: duplicate field" message you're adding to the child exception becomes redundant in this case: the original one already has similar information. Keep or edit or delete, up to you.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
```diff
@@ -176,6 +177,10 @@ private static String validateAndParseSemiStructured(
...
+              columnName, snowflakeType, "Not a valid JSON: duplicate field", insertRowIndex);
+        }
         throw new SFException(e, ErrorCode.IO_ERROR, "Cannot create JSON Parser or JSON generator");
```
sfc-gh-dorlovsky, line 184: Nit: would it be useful to add columnName, snowflakeType, and insertRowIndex to this exception too?

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-lsembera: @sfc-gh-dorlovsky We don't want to preserve information here, it may lead to customer data being exposed in logs.

src/main/java/net/snowflake/ingest/streaming/internal/DataValidationUtil.java
sfc-gh-ggeng: Thanks for pointing this out! Added the info

src/test/java/net/snowflake/ingest/streaming/internal/DataValidationUtilTest.java
sfc-gh-ggeng: Thanks for bringing this up. Added!

sfc-gh-tzhang (approved):
lgtm, thanks!

PR #936, opened 2025-02-06
Release 3.1.0

sfc-gh-xhuang:
3.1.0 SDK release

### DIFF
--- a/e2e-jar-test/pom.xml
+++ b/e2e-jar-test/pom.xml
@@ -29,7 +29,7 @@
       <dependency>
         <groupId>net.snowflake</groupId>
         <artifactId>snowflake-ingest-sdk</artifactId>
-        <version>3.0.1</version>
+        <version>3.1.0</version>
       </dependency>
 
       <dependency>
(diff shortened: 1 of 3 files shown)


sfc-gh-skurella:
Hey!

I'm submitting a fix for the SDK which is critical for the Snowfort migration—we lost the ability to inject some data into the SDK when migrating to v3 and need to bring it back now. We brought it up in the streaming ingest sync yesterday.

It'd be great if you could wait until that PR is in so that we can avoid having to create another release shortly after this one.

sfc-gh-azagrebin (changes requested):
our PPN tests are broken for server side
we agreed yesterday that we take 1-2 days more to merge a small refactoring for the next SDK release
to unblock snowfort tests

sfc-gh-tzhang:
@sfc-gh-skurella @sfc-gh-azagrebin This is just merging the version bump change instead of doing actual release, but yeah, it's better to merge this the last to have a cleaner commit history.

sfc-gh-tzhang (approved):

PR #937, opened 2025-02-07
allow passing custom metadata to `ParquetFlusher` for testing

sfc-gh-skurella:
The primary reason for this PR is to allow passing a custom fileId for testing PPN behavior in Snowflake. This unblocks migrating some tests to Snowfort.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
@@ -25,7 +25,6 @@ import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.zip.CRC32;
 import javax.crypto.BadPaddingException;
 import javax.crypto.IllegalBlockSizeException;
@@ -63,8 +62,8 @@ class BlobBuilder {
    * Builds blob.
    *
    * @param filePath Path of the destination file in cloud storage
-   * @param customFileId Allows setting a custom file ID to be embedded for all chunks in storage.
-   *     Used for testing.
+   * @param fileMetadataTestingOverrides Allows setting a custom file ID and SDK version to be
+   *     embedded for all chunks in storage. Used for testing.
    * @param blobData All the data for one blob. Assumes that all ChannelData in the inner List
    *     belongs to the same table. Will error if this is not the case
    * @param bdecVersion version of blob
@@ -72,7 +71,7 @@ class BlobBuilder {
    */
   static <T> Blob constructBlobAndMetadata(
       String filePath,
-      Optional<String> customFileId,
+      FileMetadataTestingOverrides fileMetadataTestingOverrides,
       List<List<ChannelData<T>>> blobData,
       Constants.BdecVersion bdecVersion,
       InternalParameterProvider internalParameterProvider,
@@ -105,7 +104,8 @@ class BlobBuilder {
 
       Flusher<T> flusher = channelsDataPerTable.get(0).createFlusher();
       Flusher.SerializationResult serializedChunk =
-          flusher.serialize(channelsDataPerTable, filePath, curDataSize, customFileId);
+          flusher.serialize(
+              channelsDataPerTable, filePath, curDataSize, fileMetadataTestingOverrides);
 
       if (!serializedChunk.channelsMetadataList.isEmpty()) {
         final byte[] compressedChunkData;
--- a/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
@@ -37,7 +37,6 @@ import net.snowflake.ingest.utils.Logging;
 import net.snowflake.ingest.utils.Pair;
 import net.snowflake.ingest.utils.SFException;
 import org.apache.commons.codec.binary.Hex;
-import org.apache.parquet.Preconditions;
 
 /**
  * Build a single blob file that contains file header plus data. The header will be a
@@ -105,9 +104,8 @@ class BlobBuilder {
                   firstChannelFlushContext.getEncryptionKeyId()));
 
       Flusher<T> flusher = channelsDataPerTable.get(0).createFlusher();
-      String fileId = customFileId.orElse(createFileId(filePath, curDataSize));
       Flusher.SerializationResult serializedChunk =
-          flusher.serialize(channelsDataPerTable, filePath, curDataSize, fileId);
+          flusher.serialize(channelsDataPerTable, filePath, curDataSize, customFileId);
 
       if (!serializedChunk.channelsMetadataList.isEmpty()) {
         final byte[] compressedChunkData;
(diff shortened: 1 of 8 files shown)


sfc-gh-azagrebin:
blocks #936

src/main/java/net/snowflake/ingest/streaming/internal/BlobBuilder.java
```diff
@@ -207,6 +212,19 @@ static <T> Blob constructBlobAndMetadata(
     return new Blob(blobBytes, chunksMetadataList, new BlobStats());
   }
 
+  private static String defaultFileId(String filePath, long chunkStartOffset) {
```
sfc-gh-azagrebin, line 215:
```suggestion
  private static String createFileId(String filePath, long chunkStartOffset) {
```

sfc-gh-azagrebin (approved):

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -160,29 +164,18 @@ private SerializationResult serializeFromJavaObjects(
...
+    if (enableIcebergStreaming) {
       Preconditions.checkState(
-          !enableIcebergStreaming, "Iceberg streaming is not supported with non-zero offsets");
-      String shortName = StreamingIngestUtils.getShortname(filePath);
-      final String[] parts = shortName.split("\\.");
-      Preconditions.checkState(parts.length == 2, "Invalid file name format");
-      metadata.put(
-          Constants.PRIMARY_FILE_ID_KEY,
-          String.format("%s_%d.%s", parts[0], chunkStartOffset, parts[1]));
+              chunkStartOffset == 0, "Iceberg streaming is not supported with non-zero offsets");
```
sfc-gh-psaha, line 175: Can this check move up the stack? Then `chunkStartOffset` won't need to be plumbed through. Ignore it if there are other usage of `chunkStartOffset`.

sfc-gh-psaha (approved):
Approving to unblock but PTAL at the comment.

sfc-gh-psaha:
@sfc-gh-skurella let's also have someone from DT team to take a look - this change looks inoccuous but we have made mistakes in the past.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
sfc-gh-azagrebin: we thought this is specific to Iceberg / Parquet, so we left it in the Parquet flusher

sfc-gh-skurella: I can inline the assertion into `serializeFromJavaObjects`—there will be one less propagation but it'll remain a property of the Parquet flusher. Would you be OK with that?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -160,29 +166,18 @@ private SerializationResult serializeFromJavaObjects(
   }
 
   private void addFileIdToMetadata(
-      String filePath, long chunkStartOffset, Map<String, String> metadata) {
+      String fileId, long chunkStartOffset, Map<String, String> metadata) {
```
sfc-gh-tzhang, line 169: What's the reason of moving fileId construction logic out of this function? We really want to keep the logic as close as possible to the SnowflakeParquetWriter creation to avoid the previous issue we have made, as there could be changes added in between which change the file name, chat with @sfc-gh-azagrebin or see https://docs.google.com/document/d/1Daz_0dxGl3J0olTJEVwK0JHvD3W8oxHEd_K37js-F2E/edit?tab=t.0#heading=h.vnkx2gi28its (second corruption) for more detail

sfc-gh-tzhang (changes requested):
I suggest to keep the fileId construction logic right above the file writer creation to avoid making similar mistakes as before.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
sfc-gh-skurella: Thanks for linking the postmortem—I haven't seen it before and it's super-valuable for building a context.
(I think we should also consider linking them in the relevant tests—it'll make their maintenance easier.)

Re. your question, we discussed the overall approach with @sfc-gh-azagrebin and he suggested that moving the logic out helps with the separation of concerns, because a concrete implementation of a `Flusher` shouldn't necessarily affect row IDs. It resonates with me, but I don't have a strong opinion. @sfc-gh-azagrebin WDYT?

I share your concern about the placement of logic. However, it seems to me that `BlobBuilder` is an adequate place for deciding on the file ID / PPN. If anything, I didn't like splitting the file ID calculation away from the specific headers that it was being written to, but then again creating vs serialising metadata are operations that I can internally justify keeping separate.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
sfc-gh-skurella: Alternatively, I can add an abstract method called e.g. `supportsMultipleTables` to the `Flusher` interface and cross-check that in the `BlobBuilder.constructBlobAndMetadata` with the number of per-table entries in the `blobData` it receives.

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
sfc-gh-azagrebin: @sfc-gh-tzhang 
I also think it is close enough to file creation, the flusher does not do anything multithreaded and does not have to generate the fileId internally. It just serialises the Parquet data including fileId that is not about parquet serialisation but Snowflake internal metadata that belongs to blob building.

sfc-gh-tzhang: > I also think it is close enough to file creation, the flusher does not do anything multithreaded and does not have to generate the fileId internally. There is not reason for file path and id to change, path is already generated here. It just serialises the Parquet data including fileId that is not about parquet serialisation but Snowflake internal metadata that belongs to blob building.

@sfc-gh-skurella @sfc-gh-azagrebin I hear your points and I understand that it complicates the logic, but how would you prevent the same data corruption from happening again? Do you plan to add tests to make sure that fileId will never change between `Builder` and `Flusher`? In the previous corruption, there was no logic in between that changes the fileId at the beginning, then some file split logic got added later by someone else which does not have the full context or assumptions. It just does not make sense to me that we want to increase the risks of data corruption in order to make the testing logic easier, WDYT?

sfc-gh-skurella:
I'll also need to add a way to prevent the SDK from sending an SDK version in Parquet metadata (also for our internal testing purposes)—I'll also push that change to this PR soon.

sfc-gh-tzhang (approved):
lgtm, thanks!

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
sfc-gh-tzhang: > IMHO the issue may still persist in the code regardless of where the logic lives. Since IIUC we're concerned about developers paying extra attention to how distributing chunks across files affects metadata, then I'm quite concerned about:
> 
> Map<String, String> metadata = channelsDataPerTable.get(0).getVectors().metadata;
> in the flusher. (What makes the first chunk special?) I think we could really benefit from an explicit and, ideally, strongly typed reduction/aggregation function instead.

I think this is a valid concern, i think the assumption is that the metadata in all the chunks are the same. Ideally we should have a map at file level. If it requires big refactoring, what about we at least add some verification logic that the metadata should be the same across all the chunks?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetFlusher.java
```diff
@@ -160,29 +169,50 @@ private SerializationResult serializeFromJavaObjects(
...
+              } else {
+                final String[] parts = shortName.split("\\.");
+                Preconditions.checkState(parts.length == 2, "Invalid file name format");
+                return String.format("%s_%d.%s", parts[0], chunkStartOffset, parts[1]);
+              }
+            });
+    metadata.put(key, value);
+  }
+
+  private void overrideSdkVersionInMetadata(
```
sfc-gh-azagrebin, line 206: would it make sense to keep `addFileIdToMetadata` untouched to reduce risk
and do `customFileIdForTesting` in `overrideSdkVersionInMetadata`?
```
customFileIdForTesting.ifPresent(fid ->
  metadata.put(
          enableIcebergStreaming
              ? Constants.ASSIGNED_FULL_FILE_NAME_KEY
              : Constants.PRIMARY_FILE_ID_KEY,
          fid);
);
```

sfc-gh-tzhang (approved):
lgtm, thanks!

Issue #927, opened 2025-01-12
SnowflakeFileTransferAgent throws uploadWithoutConnection exception  


sivankumar86:
val snowflakeIngest = "net.snowflake"              % "snowflake-ingest-sdk" % 3.0.0

Auto refresh warning exception, Could you refresh before expiry ?

Error: 

[ingest-build-upload-thread-0] ERROR net.snowflake.client.jdbc.SnowflakeFileTransferAgent - Exception encountered during file upload in uploadWithoutConnection
net.snowflake.client.jdbc.SnowflakeSQLLoggedException: S3 operation failed: Operation=upload, Error type=Client, Error code=ExpiredToken, Error message=The provided token has expired. (Service: Amazon S3; Status Code: 400; Error Code: ExpiredToken; Request ID: sdfdsfsdfsdf; S3 Extended Request ID: ewerewrewrewrewrewr; Proxy: ), Request id=sdfdsfdsfsdf, Extended request id=sdfdsfdsfdsfdsfdsfdsf6RhWJE2ye55ov1w==
	at net.snowflake.client.jdbc.cloud.storage.SnowflakeS3Client.handleS3Exception(SnowflakeS3Client.java:862)
	at net.snowflake.client.jdbc.cloud.storage.SnowflakeS3Client.upload(SnowflakeS3Client.java:668)
	at net.snowflake.client.jdbc.SnowflakeFileTransferAgent.pushFileToRemoteStore(SnowflakeFileTransferAgent.java:2142)
	at net.snowflake.client.jdbc.SnowflakeFileTransferAgent.uploadWithoutConnection(SnowflakeFileTransferAgent.java:2261)
	at net.snowflake.ingest.streaming.internal.InternalStage.putRemote(InternalStage.java:169)
	at net.snowflake.ingest.streaming.internal.InternalStage.put(InternalStage.java:335)
	at net.snowflake.ingest.streaming.internal.FlushService.upload(FlushService.java:649)
	at net.snowflake.ingest.streaming.internal.FlushService.buildAndUpload(FlushService.java:620)
	at net.snowflake.ingest.streaming.internal.FlushService.lambda$distributeFlushTasks$13(FlushService.java:509)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
[...]

PR #942, opened 2025-02-17
SNOW-1881734 Silence uploadWithoutConnection exception log

sfc-gh-lsembera:
This PR takes advantage of a new feature in JDBC driver 3.22.0, which lets us silence exception log in the JDBC driver in case an upload fails.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
@@ -188,6 +188,7 @@ class InternalStage implements IStorage {
         SnowflakeFileTransferAgent.uploadWithoutConnection(
             SnowflakeFileTransferConfig.Builder.newInstance()
                 .setSnowflakeFileTransferMetadata(fileTransferMetadataCopy)
+                .setSilentException(true)
                 .setUploadStream(inStream)
                 .setRequireCompress(false)
                 .setOcspMode(OCSPMode.FAIL_OPEN)


sfc-gh-lsembera:
cc @sfc-gh-tzhang @sfc-gh-xhuang I realized that upgrading to JDBC 3.22.0 is not enough to silence the exception, we also need to set the flag in the transfer config.

cc @sfc-gh-dprzybysz Could you please check if the usage is correct?

sfc-gh-dprzybysz (approved):

sfc-gh-xhuang (approved):

sfc-gh-japatel:
If we are silencing logs, how would we/customer debug this?

sfc-gh-lsembera:
> If we are silencing logs, how would we/customer debug this?

@sfc-gh-japatel This only silences an [error log statement](https://github.com/snowflakedb/snowflake-jdbc/blob/31d6a9ce16a86efbc6e01fe21c9f5ca75ffd7be7/src/main/java/net/snowflake/client/jdbc/SnowflakeFileTransferAgent.java#L2386) in the JDBC driver, an exception is still being thrown and the caller (=us in this case) needs to deal with it. This error log is causing confusion because it is expected to happen when the cloud storage token expires, in which case we refresh the token, so the error logs are false positives.

sfc-gh-lsembera:
@sfc-gh-tzhang Since you implemented a related logic - avoiding the error log line by proactively refreshing the token after 58 minutes - could you also approve, please?

src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
```diff
@@ -188,6 +188,7 @@ private Optional<String> putRemote(String fullFilePath, byte[] data, int retryCo
         SnowflakeFileTransferAgent.uploadWithoutConnection(
             SnowflakeFileTransferConfig.Builder.newInstance()
                 .setSnowflakeFileTransferMetadata(fileTransferMetadataCopy)
+                .setSilentException(true)
```
sfc-gh-tzhang, line 191: FWIW, this function name is confusing....

sfc-gh-tzhang:
> @sfc-gh-tzhang Since you implemented a related logic - avoiding the error log line by proactively refreshing the token after 58 minutes - could you also approve, please?

Could we remove my logic after my change?

sfc-gh-tzhang (approved):
Let's remove my logic if this fixes the issue, thanks!

sfc-gh-xhuang:
> Let's remove my logic if this fixes the issue, thanks!

I don't think we should remove the logic of proactively refreshing a new token 58 minutes into the hour.

sfc-gh-xhuang:
addresses this: https://github.com/snowflakedb/snowflake-ingest-java/issues/927 ?

sfc-gh-lsembera:
> addresses this: #927 ?

Yes

src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
sfc-gh-lsembera: cc @sfc-gh-dprzybysz

PR #945, opened 2025-02-23
Adds additional docstring information

tjj5036:
There is no information in the Javadocs regarding when an exception can be thrown. This adds the `@throws` annotation to existing methods to indicate that exceptions CAN be thrown in certain scenarios.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
+++ b/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
@@ -9,6 +9,7 @@ import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import javax.annotation.Nullable;
 import net.snowflake.ingest.streaming.internal.ColumnProperties;
+import net.snowflake.ingest.utils.SFException;
 
 /**
  * A logical partition that represents a connection to a single Snowflake table, data will be
@@ -27,7 +28,7 @@ public interface SnowflakeStreamingIngestChannel {
    * Get the fully qualified channel name
    *
    * @return fully qualified name of the channel, in the format of
-   *     dbName.schemaName.tableName.channelName
+   *     `dbName.schemaName.tableName.channelName`
    */
   String getFullyQualifiedName();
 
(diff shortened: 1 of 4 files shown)


sfc-gh-sparkes (approved):

src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
```diff
@@ -244,6 +258,9 @@ public interface SnowflakeStreamingIngestChannel {
    * @param offsetToken offset of given row, used for replay in case of failures. It could be null
    *     if you don't plan on replaying or can't replay
    * @return insert response that possibly contains errors because of insertion failures
+   * @throws net.snowflake.ingest.utils.SFException if the channel is not considered "valid".
```
sfc-gh-rcheng, line 261: nit: should we also @link the exception here?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
```diff
@@ -256,6 +262,14 @@ public CompletableFuture<Void> close() {
     return this.close(false);
   }
 
+  /**
+   * Attempts to close the channel by draining any outstanding data and rejecting further writes to
+   * the channel.
+   *
+   * @param drop if true, the channel will be dropped after all data is successfully committed.
+   * @throws SFException when the channel is invalid or closed, meaning that no new rows can be
```
sfc-gh-rcheng, line 270: nit: lets keep the references to SFException consistent, earlier used `net.snowflake.ingest.utils.SFException`

sfc-gh-rcheng (approved):
thanks, some small nits

src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
```diff
@@ -62,20 +63,33 @@ public interface SnowflakeStreamingIngestChannel {
...
    * Get the fully qualified table name that the channel belongs to
    *
-   * @return fully qualified table name, in the format of dbName.schemaName.tableName
+   * @return fully qualified table name, in the format of `dbName.schemaName.tableName`
    */
   String getFullyQualifiedTableName();
 
-  /** @return a boolean which indicates whether the channel is valid */
+  /**
+   * @return a boolean which indicates whether the channel is valid. Typically, this means whether
```
sfc-gh-ggeng, line 71: Should we mention that this is a "cached" value in the SDK and we actually do not look into the server side to check whether the channel is valid as of now? (although call that out seems awkward...

sfc-gh-ggeng (approved):

src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
tjj5036: I think that this is a fair thing to call out. Or specifically, that server-side invalidations may not immediately propagate to the SDK. Let me add that wording...

src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
tjj5036: Good question - I don't think that's needed when its the first value supplied to `throws` or say `return`?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
tjj5036: Ack

src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
tjj5036: Will leave as-is for now.

sfc-gh-tjones (approved):

PR #947, opened 2025-02-24
SNOW-1257851 Refactor dependencies and imports for snowflake-jdbc-thin

sfc-gh-dhandermann:
SNOW-1257851 Refactors declared dependencies and component imports to support using the `thin` version of the [snowflake-jdbc](https://github.com/snowflakedb/snowflake-jdbc) driver.

This pull request builds on the initial work from #733 and includes additional changes to align with updated library versions and removal of additional unnecessary libraries.

Code changes include replacing a number of shaded package references from Snowflake JDBC with direct references:

- Replaced shaded Apache HTTP Client with standard packages
- Replaced shaded Google Guava with standard packages
- Replaced shaded Apache commons-io with standard packages
- Replaced shaded Jackson JSON with standard packages
- Replaced shaded Gson with Jackson ObjectMapper in OAuthClient class
- Replaced shaded HttpStatusCodes with HttpURLConnection status codes
- Replaced ResourceBundleManager with ResourceBundle in SFException to avoid the need for shaded classes
- Added UnmodifiableMap class to remove commons-collections dependency from Hadoop

Additional dependency adjustments include replacing and removing several libraries:

- Added Google HTTP Client 1.45.0
- Added JCL over SLF4J to replace Commons Logging
- Removed unused direct dependency on Jettison
- Removed usage of netty-common PlatformDependent memory value from Netty property
- Removed unused direct dependency on Commons Configuration 2
- Removed unused direct dependency on Commons Compress
- Excluded unused dependencies from Hadoop Common
- Excluded several libraries from Dependency Convergence checking
- Updated allowed license formats

Dependency version upgrades include a number of shared libraries:

- Updated Apache Commons Codec from 1.17.1 to 1.18.0
- Updated Apache Commons Compress from 1.26.2 to 1.27.1
- Updated Apache Commons Lang3 from 3.14.0 to 3.17.0
- Updated Apache Commons Text from 1.11.0 to 1.13.0
- Updated Apache Hadoop from 3.4.0 to 3.4.1
- Updated Guava from 32.0.1 to 33.3.1
- Updated ThreeTenBP from 1.6.8 to 1.7.0
- Updated Nimbus from 9.37.3 to 9.48
- Updated Jackson from 2.17.2 to 2.18.1
- Updated Google Gson from 2.10.1 to 2.11.0
- Updated Parquet from 1.14.1 to 1.14.4


sfc-gh-dhandermann:
The recent addition of support of Apache Iceberg exhibits tight coupling to classes in the Snowflake JDBC driver, resulting in issues when attempting to run with a shaded version of dependencies such as Jackson and Guava, while also using the unshaded version of the Snowflake JDBC driver.

sfc-gh-tjones (approved):

pom.xml
```diff
@@ -1137,12 +1387,26 @@
             <configuration>
               <artifactSet>
                 <excludes>
-                  <exclude>net.snowflake:snowflake-jdbc</exclude>
+                  <exclude>net.snowflake:snowflake-jdbc-thin</exclude>
+                  <exclude>org.apache.httpcomponents:httpclient</exclude>
```
sfc-gh-alhuang, line 1391: Why this needs to be unshaded?

src/main/java/net/snowflake/ingest/streaming/internal/fileTransferAgent/IcebergAzureClient.java
```diff
@@ -474,4 +477,39 @@ static String FormatStorageExtendedErrorInformation(StorageExtendedErrorInformat
     sb.append("}");
     return sb.toString();
   }
+
+  private static void setSessionlessProxyForAzure(
```
sfc-gh-alhuang, line 481: Is it possible to import `net.snowflake.client.core.HttpUtil;` and use `HttpUtil.setSessionlessProxyForAzure` instead of creating our own?

src/main/java/net/snowflake/ingest/streaming/internal/fileTransferAgent/IcebergS3Client.java
```diff
@@ -555,4 +556,77 @@ private static void handleS3Exception(
       }
     }
   }
+
+  /**
```
sfc-gh-alhuang, line 560: Not familiar about the class definition conflict issue. Does this happens in compile time? Can we use `net.snowflake.client.jdbc.cloud.storage.S3HttpUtil.setSessionlessProxyForS3(...)` in line 146 to avoid adding this?

sfc-gh-alhuang (approved):
Thank you for taking this! Leave some comments regarding the copied code.
Let's check if @sfc-gh-lsembera has any insights on this.

pom.xml
sfc-gh-dhandermann: It is necessary to have Apache HttpClient unshaded because the `TelemetryService` class uses the `TelemetryClient` from the Snowflake JDBC driver. The `TelemetryClient.createSessionlessTelemetry()` method takes an HttpClient instance, so in order to integrate with the thin version of the Snowflake JDBC driver, the unshaded version of the HttpClient library is required.

src/main/java/net/snowflake/ingest/streaming/internal/fileTransferAgent/IcebergAzureClient.java
sfc-gh-dhandermann: Unfortunately not due to shading Microsoft Azure libraries. The `HttpUtil` class in the Snowflake JDBC driver requires the `OperationContext` from the Azure library. Reimplementing that method in this class avoids the tight coupling to the Snowflake JDBC driver and the associated issues with shading the dependency.

src/main/java/net/snowflake/ingest/streaming/internal/fileTransferAgent/IcebergS3Client.java
sfc-gh-dhandermann: Similar to the issue with `HttpUtil` in the `IcebergAzureClient`, attempting to use the `S3HttpUtil` class from the Snowflake JDBC driver causes problems when attempting to support both shaded and unshaded builds of the Ingest SDK due to use of the Amazon S3 class. Implementing this method within this class avoids those problems.

sfc-gh-dhandermann:
> Thank you for taking this! Leave some comments regarding the copied code. Let's check if @sfc-gh-lsembera has any insights on this.

Thanks for the review and feedback @sfc-gh-alhuang!

pom.xml
```diff
@@ -894,21 +1144,19 @@
...
                 <bannedDependencies/>
-                <dependencyConvergence/>
+                <dependencyConvergence>
+                  <excludes>
+                    <exclude>commons-io:commons-io</exclude>
+                    <exclude>commons-codec:commons-codec</exclude>
+                    <exclude>org.apache.avro:avro</exclude>
+                    <exclude>org.apache.commons:commons-compress</exclude>
+                    <exclude>org.apache.commons:commons-lang3</exclude>
+                  </excludes>
```
sfc-gh-tzhang, line 1158: If there're multiple versions exist, would we run into potential class conflict or unexpected behavior?

pom.xml
```diff
@@ -1137,12 +1387,26 @@
             <configuration>
               <artifactSet>
                 <excludes>
-                  <exclude>net.snowflake:snowflake-jdbc</exclude>
+                  <exclude>net.snowflake:snowflake-jdbc-thin</exclude>
```
sfc-gh-tzhang, line 1390: can the jdbc-thin jar be shaded now?

pom.xml
```diff
@@ -1137,12 +1387,26 @@
...
+                  <exclude>org.apache.httpcomponents:httpcore</exclude>
                   <exclude>org.slf4j:slf4j-api</exclude>
+                  <exclude>com.fasterxml.jackson.core:jackson-annotations</exclude>
+                  <exclude>com.fasterxml.jackson.core:jackson-core</exclude>
+                  <exclude>com.fasterxml.jackson.core:jackson-databind</exclude>
+                  <exclude>com.fasterxml.jackson.dataformat:jackson-dataformat-toml</exclude>
+                  <exclude>com.fasterxml.jackson.dataformat:jackson-dataformat-cbor</exclude>
+                  <exclude>com.fasterxml.jackson.datatype:jackson-datatype-jsr310</exclude>
+                  <exclude>com.fasterxml.jackson.datatype:jackson-datatype-jdk8</exclude>
+                  <exclude>com.google.guava:guava</exclude>
```
sfc-gh-tzhang, line 1401: Does this mean customer needs to include all these dependencies now? Is there a way to avoid/minimum this behavior change?

src/main/java/net/snowflake/ingest/streaming/internal/PresignedUrlExternalVolume.java
sfc-gh-tzhang, line 1: empty file?

src/main/java/net/snowflake/ingest/streaming/internal/fileTransferAgent/IcebergS3Client.java
```diff
@@ -555,4 +556,77 @@ private static void handleS3Exception(
...
+        clientConfig.setProxyProtocol(protocolEnum);
+        String logMessage =
+            "Setting sessionless S3 proxy. Host: "
+                + proxyHost
+                + ", port: "
+                + proxyPort
+                + ", non-proxy hosts: "
+                + nonProxyHosts
+                + ", protocol: "
+                + proxyProtocol;
```
sfc-gh-tzhang, line 613: let's use String.format

src/main/java/net/snowflake/ingest/utils/Utils.java
```diff
@@ -354,9 +353,7 @@ public static void showMemory() {
 
     Runtime runtime = Runtime.getRuntime();
     logger.logInfo(
-        "Max direct memory={}, max runtime memory={}, total runtime memory={}, free runtime"
-            + " memory={}",
-        PlatformDependent.maxDirectMemory(),
+        "Max max runtime memory={}, total runtime memory={}, free runtime" + " memory={}",
```
sfc-gh-tzhang, line 356: And any chance we could keep the direct memory?
```suggestion
        "Max runtime memory={}, total runtime memory={}, free runtime" + " memory={}",
```

sfc-gh-tzhang (changes requested):
Thanks David, left some comments, PTAL! Request changes to block the PR from merging to master, let's iterate on the beta version and merge to master after everything is well tested.

pom.xml
```diff
@@ -894,21 +1144,19 @@
...
-                    </dependency>
-                  </dependencies>
                 </banDuplicateClasses>
 
                 <banDuplicatePomDependencyVersions/>
                 <bannedDependencies/>
-                <dependencyConvergence/>
+                <dependencyConvergence>
+                  <excludes>
+                    <exclude>commons-io:commons-io</exclude>
```
sfc-gh-lsembera, line 1153: Can we get rid of these excludes by selectively excluding conflicting transitive dependencies?

pom.xml
sfc-gh-lsembera: +1 on this, can we get rid of these excludes? In particular, `fasterxml` and `guava` are popular libraries that are likely to be included in most customer projects. If even our shaded build requires a specific version, there will be lots of conflicts. 

Why do we need these excludes? Could we have two builds - `shaded` which shades (almost) everything and `unshaded` that shades nothing? 

In the current SDK, [there are three excludes](https://github.com/snowflakedb/snowflake-ingest-java/blob/bcf6a7753031b88e5164d7d6e25d8901bbaac75b/public_pom.xml#L38-L57) from the shaded build - each for a specific reason:

* `snowflake-jdbc` - we wanted to avoid customers who already have `snowflake-jdbc` on classpath to include it twice, which would inflate the distribution size too much.
* `slf4j-api` - safe not to shade as there are no conflicts between versions. 
* `zstd-jni` - cannot be relocated because some resources must be in the root of the JAR, otherwise it doesn't work.

Could you comment on why does each dependency need to be excluded from shading?

pom.xml
sfc-gh-dhandermann: The Apache Commons projects follow fairly strict semantic versioning in general, so minor version updates should not cause problems.

However, I will revisit this in conjunction with evaluating shading of the snowflake-jdbc-thin library itself.

pom.xml
sfc-gh-dhandermann: Great question! A number of the changes included in the initial version of this pull request were the result of attempting to keep the current pattern of a direct dependency on the Snowflake JDBC driver. However, with other restructuring, it seems like it should be possible to shade the JDBC driver, which would obviate the need for direct dependencies on Guava and Jackson. I will evaluate the options and make updates.

src/main/java/net/snowflake/ingest/streaming/internal/PresignedUrlExternalVolume.java
sfc-gh-dhandermann: Thanks for catching this, it looks like the result of some other updates, will remove.

src/main/java/net/snowflake/ingest/utils/Utils.java
sfc-gh-dhandermann: Unfortunately the `PlatformDependent` class comes from Netty, and this is the only direct usage of the Netty library. With the Netty framework including a number of libraries, some of which are specific to particular architectures, adding a direct dependency on Netty does not seem worth it logging. The previous approach built on the shaded copy from the Snowflake JDBC driver, but that is no longer an option.

pom.xml
sfc-gh-dhandermann: That is an option, and would provide the opportunity to specify an explicit version directly. I will look at making those changes.

sfc-gh-dhandermann: Thanks again for highlighting this, I removed the exclusions and reverted this change.

pom.xml
sfc-gh-dhandermann: Thanks for flagging these exclusions. The addition of Jackson and Guava was necessary to continue having an external dependency on `snowflake-jdbc-thin`, due to the use of several classes like `TelemetryClient` within the Ingest SDK.

On further review, I pushed an update that removes these exclusions and also shades `snowflake-jdbc-thin`. Shading `snowflake-jdbc-thin` preserves the encapsulation, limiting the external dependencies to `slf4j-api` and `zstd-jni` as currently defined.

pom.xml
sfc-gh-dhandermann: I pushed an update that also includes `snowflake-jdbc-thin` in the list of shaded dependencies.

src/main/java/net/snowflake/ingest/utils/Utils.java
sfc-gh-dhandermann: Pushed a correction to the log message as noted.

sfc-gh-lsembera:
@sfc-gh-dhandermann Thanks a lot for making the change, the PR LGTM to me with two comments:

- Could you bump the SDK version to an unreleased version and re-run all tests? The problem now is that the e2e-jar-tests [depend on SDK 3.1.1](https://github.com/snowflakedb/snowflake-ingest-java/blob/master/e2e-jar-test/pom.xml#L35), which is an already released version, so the tests are not running against the SDK code from this PR, but with the SDK 3.1.1 code from Maven, so they wouldn't check potential new issues with packaging. @sfc-gh-tzhang To avoid this situation in the future, we need to bump the version in `pom.xml`, `e2e-jar-test/pom.xml` and `RequestBuilder.java` after each SDK release.
- We should also update the [docs about FIPS-compliant](https://github.com/snowflakedb/snowflake-ingest-java?tab=readme-ov-file#using-with-snowflake-jdbc-fips) version of the SDK and the [corresponding tests](https://github.com/snowflakedb/snowflake-ingest-java/blob/master/e2e-jar-test/fips/pom.xml#L45-L48). Previously we were recommending customers to replace `snowflake-jdbc` with `snowflake-jdbc-fips`, but I believe it is not longer required and customers can now continue to use the unshaded SDK build with `snowflake-jdbc-thin` and just replace the regular bouncy castle dependencies with FIPS-compliant ones. This is not blocking this PR, though, and can be done separately.

sfc-gh-lsembera (approved):

sfc-gh-dhandermann:
Thanks @sfc-gh-lsembera! I updated the version to 3.2.0 and also adjusted the README to list `snowflake-jdbc-thin` instead of `snowflake-jdbc` for FIPS.

pom.xml
sfc-gh-dhandermann: Removed these exclusions as part of updates to shade `snowflake-jdbc-thin`.

pom.xml
sfc-gh-dhandermann: Exclusions removed with shading of `snowflake-jdbc-thin`.

sfc-gh-tzhang:
> Thanks @sfc-gh-lsembera! I updated the version to 3.2.0 and also adjusted the README to list `snowflake-jdbc-thin` instead of `snowflake-jdbc` for FIPS.

@sfc-gh-dhandermann Let's revert the version upgrade commit, I already have https://github.com/snowflakedb/snowflake-ingest-java/pull/956 on top of your branch, and we can't use 3.2.0+ until end of next week if you remember what Purujit mentioned in our Slack thread.

sfc-gh-dhandermann:
> > Thanks @sfc-gh-lsembera! I updated the version to 3.2.0 and also adjusted the README to list `snowflake-jdbc-thin` instead of `snowflake-jdbc` for FIPS.
> 
> @sfc-gh-dhandermann Let's revert the version upgrade commit, I already have #956 on top of your branch, and we can't use 3.2.0+ until end of next week if you remember what Purujit mentioned in our Slack thread.

Thanks @sfc-gh-tzhang, I reverted the version number changes.

PR #948, opened 2025-02-24
SNOW-1947905 Fix proxy setting for OAuth http client

sfc-gh-alhuang:
Decoupled the HTTP client for streaming and authentication, as the SDK previously used a single client for both streaming ingestion and OAuth authentication. This change ensures compliance with the customer's requirement to configure separate proxy settings for streaming and authentication.

### DIFF
--- a/src/test/java/net/snowflake/ingest/connection/MockOAuthClient.java
+++ b/src/test/java/net/snowflake/ingest/connection/MockOAuthClient.java
@@ -1,3 +1,7 @@
+/*
+ * Copyright (c) 2025 Snowflake Computing Inc. All rights reserved.
+ */
+
 package net.snowflake.ingest.connection;
 
 import java.util.UUID;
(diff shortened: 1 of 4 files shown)


src/test/java/net/snowflake/ingest/connection/MockOAuthClient.java
```diff
@@ -13,10 +17,7 @@ public class MockOAuthClient extends OAuthClient {
   private int futureRefreshFailCount = 0;
 
   public MockOAuthClient() {
-    super(
-        "ACCOUNT_NAME",
-        new OAuthCredential("CLIENT_ID", "CLIENT_SECRET", "REFRESH_TOKEN"),
-        new URIBuilder());
+    super(new OAuthCredential("CLIENT_ID", "CLIENT_SECRET", "REFRESH_TOKEN"), new URIBuilder());
```
sfc-gh-tzhang, line 20: Do we have any end2end IT tests for OAuth? Let's add one if not

sfc-gh-tzhang (approved):
LGTM, let's add an IT test if we don't have one, thanks!

src/test/java/net/snowflake/ingest/connection/MockOAuthClient.java
sfc-gh-alhuang: The ITs didn't exist in merge gate since the OAuth feature was introduced as the refresh token expire every 90 days. We could add a python script with browser access to fetch refresh token by password in future. Add a [jira](https://snowflakecomputing.atlassian.net/browse/SNOW-1950118) ticket about this.

PR #958, opened 2025-03-05
SNOW-1950183 Resolve Iceberg ingestion filename mismatch between upload and registration

sfc-gh-alhuang:
The SDK has a bug when uploading Iceberg Parquet files at the moment of token expiration. It correctly uploads the file to S3 using a new token and file name but mistakenly registers it with the old file name. This PR ensure the SDK registers the file with the refreshed file name.

Server side [fix](https://github.com/snowflakedb/snowflake/pull/261103) for previous SDK versions.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021-2024 Snowflake Computing Inc. All rights reserved.
+ * Copyright (c) 2021-2025 Snowflake Computing Inc. All rights reserved.
  */
 
 package net.snowflake.ingest.streaming.internal;
@@ -660,13 +660,8 @@ class FlushService<T> {
 
     Timer.Context uploadContext = Utils.createTimerContext(this.owningClient.uploadLatency);
 
-    // The returned etag is non-empty ONLY in case of iceberg uploads. With iceberg files, the XP
-    // scanner expects the
-    // MD5 value to exactly match the etag value in S3. This assumption doesn't hold when multipart
-    // upload kicks in,
-    // causing scan time failures and table corruption. By plugging in the etag value instead of the
-    // md5 value,
-    Optional<String> etag = storage.put(blobPath, blob);
+    // The returned icebergPostUploadMetadata is non-empty ONLY in case of iceberg uploads.
+    Optional<IcebergPostUploadMetadata> icebergPostUploadMetadata = storage.put(blobPath, blob);
 
     if (uploadContext != null) {
       blobStats.setUploadDurationMs(uploadContext);
@@ -677,17 +672,25 @@ class FlushService<T> {
     }
 
     logger.logInfo(
-        "Finish uploading blob={}, size={}, timeInMillis={}, etag={}",
+        "Finish uploading blob={}, size={}, timeInMillis={}, icebergPostUploadMetadata={}",
         blobPath.fileRegistrationPath,
         blob.length,
         System.currentTimeMillis() - startTime,
-        etag);
+        icebergPostUploadMetadata);
 
     // at this point we know for sure if the BDEC file has data for more than one chunk, i.e.
     // spans mixed tables or not
     return BlobMetadata.createBlobMetadata(
-        blobPath.fileRegistrationPath,
-        etag.isPresent() ? etag.get() : BlobBuilder.computeMD5(blob),
+        icebergPostUploadMetadata.isPresent()
+            ? icebergPostUploadMetadata
+                .get()
+                .getRefreshedPath()
+                .orElse(blobPath)
+                .fileRegistrationPath
+            : blobPath.fileRegistrationPath,
+        icebergPostUploadMetadata.isPresent()
+            ? icebergPostUploadMetadata.get().getEtag()
+            : BlobBuilder.computeMD5(blob),
         bdecVersion,
         metadata,
         blobStats,
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
@@ -660,7 +660,7 @@ class FlushService<T> {
 
     Timer.Context uploadContext = Utils.createTimerContext(this.owningClient.uploadLatency);
 
-    // The returned icebergPostUploadMetadata is non-empty ONLY in case of iceberg uploads.
+    // The returned icebergPostUploadMetadata is non-empty if and only if it's iceberg uploads.
     Optional<IcebergPostUploadMetadata> icebergPostUploadMetadata = storage.put(blobPath, blob);
 
     if (uploadContext != null) {
@@ -682,7 +682,7 @@ class FlushService<T> {
     // spans mixed tables or not
     return BlobMetadata.createBlobMetadata(
         icebergPostUploadMetadata
-            .flatMap(IcebergPostUploadMetadata::getRefreshedPath)
+            .map(IcebergPostUploadMetadata::getBlobPath)
             .orElse(blobPath)
             .fileRegistrationPath,
         icebergPostUploadMetadata
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
@@ -681,16 +681,13 @@ class FlushService<T> {
     // at this point we know for sure if the BDEC file has data for more than one chunk, i.e.
     // spans mixed tables or not
     return BlobMetadata.createBlobMetadata(
-        icebergPostUploadMetadata.isPresent()
-            ? icebergPostUploadMetadata
-                .get()
-                .getRefreshedPath()
-                .orElse(blobPath)
-                .fileRegistrationPath
-            : blobPath.fileRegistrationPath,
-        icebergPostUploadMetadata.isPresent()
-            ? icebergPostUploadMetadata.get().getEtag()
-            : BlobBuilder.computeMD5(blob),
+        icebergPostUploadMetadata
+            .flatMap(IcebergPostUploadMetadata::getRefreshedPath)
+            .orElse(blobPath)
+            .fileRegistrationPath,
+        icebergPostUploadMetadata
+            .flatMap(IcebergPostUploadMetadata::getEtag)
+            .orElse(BlobBuilder.computeMD5(blob)),
         bdecVersion,
         metadata,
         blobStats,
--- a/src/main/java/net/snowflake/ingest/streaming/internal/IcebergPostUploadMetadata.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/IcebergPostUploadMetadata.java
@@ -18,7 +18,7 @@ class IcebergPostUploadMetadata {
    * table corruption. By plugging in the etag value instead of the md5 value, we can ensure that
    * the XP scanner can successfully scan the file.
    */
-  private final String etag;
+  private final @Nullable String etag;
 
   /**
    * The updated upload path of the file within the table's external volume. This will be null if
@@ -36,13 +36,13 @@ class IcebergPostUploadMetadata {
    * @param etag The etag of the uploaded file.
    * @param path The updated upload path of the file within the table's external volume.
    */
-  IcebergPostUploadMetadata(String etag, @Nullable BlobPath path) {
+  IcebergPostUploadMetadata(@Nullable String etag, @Nullable BlobPath path) {
     this.etag = etag;
     this.refreshedPath = path;
   }
 
-  String getEtag() {
-    return etag;
+  Optional<String> getEtag() {
+    return Optional.ofNullable(etag);
   }
 
   Optional<BlobPath> getRefreshedPath() {
--- a/src/main/java/net/snowflake/ingest/streaming/internal/IcebergPostUploadMetadata.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/IcebergPostUploadMetadata.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright (c) 2025 Snowflake Computing Inc. All rights reserved.
+ */
+
+package net.snowflake.ingest.streaming.internal;
+
+import java.util.Optional;
+import javax.annotation.Nullable;
+
+/**
+ * Wrapper class containing etag and the final upload path of the parquet file in the table's
+ * external volume.
+ */
+class IcebergPostUploadMetadata {
+  /*
+   * With iceberg files, the XP scanner expects the MD5 value to exactly match the etag value in S3.
+   * This assumption doesn't hold when multipart upload kicks in, causing scan time failures and
+   * table corruption. By plugging in the etag value instead of the md5 value, we can ensure that
+   * the XP scanner can successfully scan the file.
+   */
+  private final String etag;
+
+  /**
+   * The updated upload path of the file within the table's external volume. This will be null if
+   * {@code fileTransferMetadata} is not refreshed during {@link
+   * net.snowflake.ingest.streaming.internal.InternalStage#put}, meaning the final upload path
+   * remains the same as the original path. This is necessary for file registration because the
+   * final upload path may differ from the original one if it's refreshed after the upload thread
+   * has started.
+   */
+  private final @Nullable BlobPath refreshedPath;
+
+  /**
+   * Constructor for IcebergPostUploadMetadata.
+   *
+   * @param etag The etag of the uploaded file.
+   * @param path The updated upload path of the file within the table's external volume.
+   */
+  IcebergPostUploadMetadata(String etag, @Nullable BlobPath path) {
+    this.etag = etag;
+    this.refreshedPath = path;
+  }
+
+  String getEtag() {
+    return etag;
+  }
+
+  Optional<BlobPath> getRefreshedPath() {
+    return Optional.ofNullable(refreshedPath);
+  }
+
+  @Override
+  public String toString() {
+    return String.format("IcebergPostUploadMetadata(etag=%s, path=%s)", etag, refreshedPath);
+  }
+}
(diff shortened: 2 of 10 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
```diff
@@ -136,92 +137,112 @@ class InternalStage implements IStorage {
...
+                  .setOcspMode(OCSPMode.FAIL_OPEN)
+                  .setStreamingIngestClientKey(this.clientPrefix)
+                  .setStreamingIngestClientName(this.clientName)
+                  .setProxyProperties(this.proxyProperties)
+                  .setDestFileName(fullFilePath)
+                  .build());
+          return Optional.empty();
+        }
+      } catch (Exception e) {
+        lastException = e;
```
sfc-gh-dorlovsky, line 225: How about we get rid of the `lastException`, and instead pull `throw new SFException` into this catch block, and do it if this is our last retry? This will also be a more natural way to chain exceptions: in the `catch` block instead of via a temporary variable

src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
```diff
@@ -136,92 +137,112 @@ class InternalStage implements IStorage {
...
+          return Optional.of(
+              new IcebergPostUploadMetadata(
+                  IcebergFileTransferAgent.uploadWithoutConnection(
+                      fileTransferMetadataCopy,
+                      inStream,
+                      proxyProperties,
+                      clientName,
+                      clientPrefix,
+                      fullFilePath),
+                  icebergUploadRefreshedPath));
```
sfc-gh-dorlovsky, line 208: Can we always store the non-null final path in `IcebergPostUploadMetadata` instead of requiring the client to do all the if-null-then-else resolution? That would also simplify comments quite a bit.

sfc-gh-dorlovsky (changes requested):
Looks great, but I have some comments, PTAL

src/test/java/net/snowflake/ingest/streaming/internal/it/SubscopedTokenRefreshIT.java
```diff
@@ -133,18 +107,31 @@ public void testTokenExpire() throws Exception {
...
-    /* Wait for token to expire */
-
-    Thread.sleep((duration + 1) * 1000);
-
-    /* Insert a row to trigger token generation */
-    channel.insertRow(ImmutableMap.of("int_col", 2), "2");
-    TestUtils.waitForOffset(channel, "2");
+    /* Invalidate the token */
+    ((SubscopedTokenExternalVolumeManager) client.getStorageManager())
+        .getStorage(Utils.getFullyQualifiedTableName(database, schema, tableName))
```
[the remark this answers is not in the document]
sfc-gh-alhuang: We need this as the  later `setFileTransferMetadataWithAge` call (or `setEmptyIcebergFileTransferMetadataWithAge` in updated version) is in `InternalStage` but not in `IStorage`. Do you think we should add `setEmptyIcebergFileTransferMetadataWithAge` to `IStorage` and avoid the casting here? I don't have a preference.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeFileTransferMetadataWithAge.java
```diff
@@ -35,4 +37,11 @@ state to record unknown age.
     this.localLocation = localLocation;
     this.timestamp = timestamp;
   }
+
+  // Test only constructor
+  @VisibleForTesting
+  public static SnowflakeFileTransferMetadataWithAge getEmptyIcebergFileTransferMetadataWithAge() {
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Make sense, renamed.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeFileTransferMetadataWithAge.java
```diff
@@ -35,4 +37,11 @@ state to record unknown age.
     this.localLocation = localLocation;
     this.timestamp = timestamp;
   }
+
+  // Test only constructor
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Removed and called first ctor.

src/main/java/net/snowflake/ingest/streaming/internal/IcebergPostUploadMetadata.java
```diff
@@ -0,0 +1,56 @@
...
+ * Copyright (c) 2025 Snowflake Computing Inc. All rights reserved.
+ */
+
+package net.snowflake.ingest.streaming.internal;
+
+import java.util.Optional;
+import javax.annotation.Nullable;
+
+/**
+ * Wrapper class containing etag and the final upload path of the parquet file in the table's
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Removed detail class description, and renamed variable name to `blobPath`.

src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
```diff
@@ -136,92 +137,112 @@ class InternalStage implements IStorage {
...
-        return Optional.ofNullable(
-            IcebergFileTransferAgent.uploadWithoutConnection(
-                fileTransferMetadataCopy,
-                inStream,
-                proxyProperties,
+    while (retryCount <= maxUploadRetries) {
+      try {
+        // Proactively refresh the credential if it's going to expire, to avoid the token expiration
+        // error from JDBC which confuses customer. Also refresh on the first retry.
+        if (Instant.now().isAfter(prevRefresh.plus(refreshDuration)) | shouldRefreshMetadata) {
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Make sense, removed `shouldRefreshMetadata`.

src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
```diff
@@ -390,4 +411,10 @@ static void putLocal(String stageLocation, String fullFilePath, byte[] data) {
   FileLocationInfo getFileLocationInfo() {
     return this.fileLocationInfo;
   }
+
+  @VisibleForTesting
+  public void setFileTransferMetadataWithAge(
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Make sense, added `setIcebergEmptyFileTransferMetadataWithAge`

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -660,13 +660,8 @@ BlobMetadata upload(
...
 
-    // The returned etag is non-empty ONLY in case of iceberg uploads. With iceberg files, the XP
-    // scanner expects the
-    // MD5 value to exactly match the etag value in S3. This assumption doesn't hold when multipart
-    // upload kicks in,
-    // causing scan time failures and table corruption. By plugging in the etag value instead of the
-    // md5 value,
-    Optional<String> etag = storage.put(blobPath, blob);
+    // The returned icebergPostUploadMetadata is non-empty ONLY in case of iceberg uploads.
+    Optional<IcebergPostUploadMetadata> icebergPostUploadMetadata = storage.put(blobPath, blob);
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Updated comment.

src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
```diff
@@ -136,92 +137,112 @@ class InternalStage implements IStorage {
...
+
+          /*
+          Since we can have multiple calls to putRemote in parallel and because the metadata includes the file path
+          we use a copy for the upload to prevent us from using the wrong file path.
+           */
+          fileTransferMetadataCopy =
+              new SnowflakeFileTransferMetadataV1(
+                  fileTransferMetadata.getPresignedUrl(),
+                  fullFilePath,
+                  fileTransferMetadata.getEncryptionMaterial() != null
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Added local var.

src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
```diff
@@ -136,92 +137,112 @@ class InternalStage implements IStorage {
...
+                  .setRequireCompress(false)
+                  .setOcspMode(OCSPMode.FAIL_OPEN)
+                  .setStreamingIngestClientKey(this.clientPrefix)
+                  .setStreamingIngestClientName(this.clientName)
+                  .setProxyProperties(this.proxyProperties)
+                  .setDestFileName(fullFilePath)
+                  .build());
+          return Optional.empty();
+        }
+      } catch (Exception e) {
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Took a look into the exact Exceptions the uploading function throws and looks like it's not well documented. I try to run test to three cloud services and found that `SQLException | IOException | StorageException` are the possible exception that would be thrown when a invalid token is provided. Not sure if this cover all the code path though.

src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
sfc-gh-alhuang: Make sense, changed.

src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
sfc-gh-alhuang: Yes, make it nonnull.

src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
```diff
@@ -136,92 +142,110 @@ class InternalStage implements IStorage {
...
+              fileTransferMetadataWithAge.fileTransferMetadata;
+
+          RemoteStoreFileEncryptionMaterial encryptionMaterial =
+              fileTransferMetadata.getEncryptionMaterial();
+          /*
+          Since we can have multiple calls to putRemote in parallel and because the metadata includes the file path
+          we use a copy for the upload to prevent us from using the wrong file path.
+           */
+          fileTransferMetadataCopy =
+              new SnowflakeFileTransferMetadataV1(
```
sfc-gh-dorlovsky, line 184: Nit: factor out the contents of this 'else' branch to a static helper function. Then this code will look something like
```
fileTransferMetadataCopy = copyFileTransferMetadata(fileTransferMetadataWithAge.fileTransferMetadata, blobPath.uploadPath);
```

`putRemote` function has 100+ lines now, and this will shave off almost 20 of them and make it easier to read. 

Moreover, I think after this change the whole if-then-else construction can be replaced with a ternary ? : operator

src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
```diff
@@ -136,92 +142,110 @@ class InternalStage implements IStorage {
...
+          fileTransferMetadataCopy = this.fetchSignedURL(blobPath.uploadPath);
+        } else {
+          // Set file path to be uploaded
+          SnowflakeFileTransferMetadataV1 fileTransferMetadata =
+              fileTransferMetadataWithAge.fileTransferMetadata;
+
+          RemoteStoreFileEncryptionMaterial encryptionMaterial =
+              fileTransferMetadata.getEncryptionMaterial();
+          /*
+          Since we can have multiple calls to putRemote in parallel and because the metadata includes the file path
```
sfc-gh-dorlovsky, line 180: I know you just copied this comment, but I don't quite understand it.
I think I get the overall idea that since this can be called concurrently, making a copy of something that can change look like a good move, but that's far from real thread safety. Unless there are some assumptions which I don't know, even making a copy of an object that can be modified from multiple threads requires taking locks and I see none.
Can we discuss this?

src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
```diff
@@ -136,92 +142,110 @@ class InternalStage implements IStorage {
...
-      logger.logInfo(
-          "Retrying upload, attempt {}/{} msg: {}, stackTrace:{}",
-          retryCount,
-          maxUploadRetries,
-          e.getMessage(),
-          getStackTrace(e));
-      return this.putRemote(fullFilePath, data, retryCount);
     }
+
+    // Should never reach here
```
sfc-gh-dorlovsky, line 247: Change the loop to `while(true)` and delete these lines

src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
```diff
@@ -349,13 +373,13 @@ SnowflakeFileTransferMetadataV1 fetchSignedURL(String fileName)
...
   /** Upload file to internal stage */
-  public Optional<String> put(BlobPath blobPath, byte[] blob) {
+  public Optional<IcebergPostUploadMetadata> put(BlobPath blobPath, byte[] blob) {
     if (this.isLocalFS()) {
       putLocal(this.fileTransferMetadataWithAge.localLocation, blobPath.fileRegistrationPath, blob);
       return Optional.empty();
     } else {
       try {
-        return putRemote(blobPath.uploadPath, blob, 0);
+        return putRemote(blobPath, blob);
```
sfc-gh-dorlovsky, line 382: As it is implemented now, `putRemote` can only throw SFException so this try-catch block is redundant

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -677,17 +672,22 @@ BlobMetadata upload(
...
-        etag);
+        icebergPostUploadMetadata);
 
     // at this point we know for sure if the BDEC file has data for more than one chunk, i.e.
     // spans mixed tables or not
     return BlobMetadata.createBlobMetadata(
-        blobPath.fileRegistrationPath,
-        etag.isPresent() ? etag.get() : BlobBuilder.computeMD5(blob),
+        icebergPostUploadMetadata
+            .map(IcebergPostUploadMetadata::getBlobPath)
```
sfc-gh-dorlovsky, line 685: This is always present now, right?

src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
```diff
@@ -390,4 +391,43 @@ static void putLocal(String stageLocation, String fullFilePath, byte[] data) {
   FileLocationInfo getFileLocationInfo() {
     return this.fileLocationInfo;
   }
+
+  private SnowflakeFileTransferMetadataV1 copyFileTransferMetadata(
+      SnowflakeFileTransferMetadataV1 fileTransferMetadata, String uploadPath)
+      throws SnowflakeSQLException, IOException {
+    // TODO: SNOW-1969309
+    if (fileTransferMetadata.isForOneFile()) {
+      return this.fetchSignedURL(uploadPath);
```
sfc-gh-dorlovsky, line 400: This branch is not quite copying the `fileTransferMetadata`, that's why I suggested to keep branching in the caller context and only factor out the `else` branch to a helper function. Which can be static in this case.

src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
```diff
@@ -390,4 +391,43 @@ static void putLocal(String stageLocation, String fullFilePath, byte[] data) {
   FileLocationInfo getFileLocationInfo() {
     return this.fileLocationInfo;
   }
+
+  private SnowflakeFileTransferMetadataV1 copyFileTransferMetadata(
+      SnowflakeFileTransferMetadataV1 fileTransferMetadata, String uploadPath)
+      throws SnowflakeSQLException, IOException {
+    // TODO: SNOW-1969309
```
sfc-gh-dorlovsky, line 398: Add a short text comment to TODO, e.g. 'investigate if this code is thread-safe'

src/test/java/net/snowflake/ingest/streaming/internal/it/SubscopedTokenRefreshIT.java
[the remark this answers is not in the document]
sfc-gh-dorlovsky: Sorry, missed this comment.

Didn't notice `SubscopedTokenExternalVolumeManager.getStorage()` returns `InternalStorage` -- C++ doesn't have the idea of covariant return types so I'm not quite used to them.
In this case, how about we cast `IStorage` to `InternalStorage` instead? IIUC it's already exposed for testing, so `SubscopedTokenExternalVolumeManager` doesn't need to be.

src/main/java/net/snowflake/ingest/streaming/internal/fileTransferAgent/IcebergFileTransferAgent.java
```diff
@@ -1,3 +1,7 @@
+/*
```
sfc-gh-dorlovsky, line 1: Revert?

sfc-gh-alhuang: My bad, reverted.

src/test/java/net/snowflake/ingest/streaming/internal/it/SubscopedTokenRefreshIT.java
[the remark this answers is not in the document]
sfc-gh-alhuang: Make sense, casted to `IStorage` and keep `SubscopedTokenExternalVolumeManager` package private.

src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
sfc-gh-alhuang: The `icebergPostUploadMetadata` could still be empty. Use `map` instead of `flatmap` here as we are sure that getBlobPath is not null.

src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
sfc-gh-alhuang: Got it. Moved the branch to call site and make it ternary

src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
sfc-gh-alhuang: Sorry forgot to add this, added.

sfc-gh-dorlovsky (approved):

PR #965, opened 2025-03-19
SNOW-1995342: Exposing flush API as part of the SnowflakeStreamingIngestClient interface

sfc-gh-tzhang:
Exposing flush API as part of the SnowflakeStreamingIngestClient interface, this is still a beta function which may introduce issues if not used in a proper way.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClient.java
+++ b/src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClient.java
@@ -4,8 +4,10 @@
 
 package net.snowflake.ingest.streaming;
 
+import com.google.common.annotations.Beta;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.CompletableFuture;
 
 /**
  * A class that is the starting point for using the Streaming Ingest client APIs, a single client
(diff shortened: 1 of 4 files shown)


src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClient.java
```diff
@@ -58,6 +60,15 @@ public interface SnowflakeStreamingIngestClient extends AutoCloseable {
    */
   void setRefreshToken(String refreshToken);
 
+  /**
+   * Flush all data in memory across all channels to persistent storage and register with a
```
sfc-gh-tjones, line 64: Nit, `and registers them to a Snowflake table`

src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClient.java
```diff
@@ -58,6 +60,15 @@ public interface SnowflakeStreamingIngestClient extends AutoCloseable {
...
+   * Snowflake table. This API is still in beta and not meant for production use.
```
sfc-gh-tjones, line 65: We should clarify why it's not meant for production use. Also, are we convinced that isn't the case?

sfc-gh-tjones (approved):
Small nits regarding the docstring but the change looks good overall. Thanks Toby!

sfc-gh-dhandermann (approved):
Thanks for promoting the `flush()` method as part of the public interface, this looks good!

src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestClient.java
sfc-gh-dhandermann: Concurring with the above question, is it more accurate to indicate that this method is in beta status indicating that it could change, rather than that it is not intended for production use?

sfc-gh-tzhang: Makes sense, let me update

PR #968, opened 2025-03-21
FLOW-1359 Added SnowflakeStreamingIngestChannel#validateRows

sfc-gh-dseweryn:


src/main/java/net/snowflake/ingest/streaming/SnowflakeStreamingIngestChannel.java
```diff
@@ -296,6 +296,12 @@ InsertValidationResponse insertRows(
   InsertValidationResponse insertRows(
       Iterable<Map<String, Object>> rows, @Nullable String offsetToken);
 
+  /**
+   * Validate a batch of rows against the channel, please see {@link
```
sfc-gh-tjones, line 300: Can we add `against the table's current schema`?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -426,6 +399,101 @@ private float addRow(
     return size;
   }
 
+  private static class ParseRowResult {
+    private final float size;
+    private final Map<String, RowBufferStats> forkedStatsMap;
+
+    public ParseRowResult(float size, Map<String, RowBufferStats> forkedStatsMap) {
```
sfc-gh-tjones, line 406: Can we add a docstring for what `forkedStatsMap` is and contains?

src/main/java/net/snowflake/ingest/streaming/internal/ParquetRowBuffer.java
```diff
@@ -426,6 +399,101 @@ private float addRow(
...
+      }
+      if (isActualError(error)) {
+        response.addError(error);
+      }
+      rowIndex++;
+    }
+    return response;
+  }
+
+  private boolean isActualError(InsertValidationResponse.InsertError insertError) {
```
sfc-gh-tjones, line 486: Nit, should we rename this to `isDataError`? Let's make this clear this is specifically about data related errors and not server-side errors.

sfc-gh-tjones (approved):
Change looks good to be but let's validate for Toby to review

sfc-gh-dseweryn:
Thanks for the review, lets wait for Toby. I am also validating with Mark if this API is what we wan't or we should expose the API for a single record at a time

sfc-gh-tzhang (changes requested):
Blocking merge until we conclude that this is absolutely needed

PR #969, opened 2025-03-25
SNOW-2000249: Invalidate channel object when server side responses with invalid channel 

sfc-gh-tzhang:


sfc-gh-aminyaylov (approved):
Thanks!

PR #995, opened 2025-04-15
4.0.0 unshaded

sfc-gh-ggeng:


sfc-gh-tzhang (changes requested):
The branch should be based on the 4.0.0 release. Also take a look at the example, i think you will need to disable a test.

PR #1001, opened 2025-04-16
NO-SNOW Fix merge gate

sfc-gh-alhuang:
Our Github action merge gate keeps failing starting from yesterday due to several errors, this PR fix it.

* Upgrade to `ubuntu-24.04` as the old `ubuntu-20.04` is not supported anymore from yesterday. https://github.com/actions/runner-images/issues/11101
* Upgrade `upgdate_project_version.py` to run in python3. The python2 is not preinstalled in ubuntu 24.04, and cannot be installed via `setup-python` due to python2's deprecation. https://github.com/actions/setup-python/issues/672. Note that this change makes the original comments in `public_pom.xml` unavailable but this doesn't affect the testing results. 
* Upgrade `setup-java@v2` to `setup-java@v4` and specify `cache-dependency-path` as there seems to be some problem with maven caching service on `setup-java@v2` using `action/cache@v3` which is already deprecated https://github.com/actions/toolkit/discussions/1890. See the following error message for reference.
```
Resolved Java 8.0.442+6 from tool-cache
Setting Java 8.0.442+6 as the default

Java configuration:
  Distribution: temurin
  Version: 8.0.442+6
  Path: C:\hostedtoolcache\windows\Java_Temurin-Hotspot_jdk\8.0.442-6\x64

Creating settings.xml with server-id: github
Writing to C:\Users\runneradmin\.m2\settings.xml
Error: Cache service responded with 422
```

### DIFF
--- a/.github/workflows/End2EndTest.yml
+++ b/.github/workflows/End2EndTest.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Checkout Code
         uses: actions/checkout@v2
       - name: Install Java ${{ matrix.java }}
-        uses: actions/setup-java@v3
+        uses: actions/setup-java@v4
         with:
           distribution: temurin
           java-version: ${{ matrix.java }}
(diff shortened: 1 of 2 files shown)


sfc-gh-tzhang (approved):
lgtm

PR #1011, opened 2025-06-11
insert rows api

sfc-gh-dshah:
Added insertRows API

### DIFF
--- a/sdk-wrapper/src/main/java/com/snowflake/ingest/sdk/server/StreamingIngestJavaService.java
+++ b/sdk-wrapper/src/main/java/com/snowflake/ingest/sdk/server/StreamingIngestJavaService.java
@@ -240,7 +240,7 @@ public class StreamingIngestJavaService {
       @Param("offsetToken") String offsetToken)
       throws Exception {
     SnowflakeStreamingIngestChannel channel = getChannelOrThrow(clientId, channelName);
-    
+
     logger.debug(
         "Batch inserting {} rows to channel {}/{}, offsetToken: {}",
         rows.size(),
@@ -250,7 +250,7 @@ public class StreamingIngestJavaService {
 
     InsertValidationResponse response = channel.insertRows(rows, offsetToken);
     if (response.hasErrors()) {
-        throw response.getInsertErrors().get(0).getException();
+      throw response.getInsertErrors().get(0).getException();
     }
 
     return HttpResponse.ofJson(


sdk-wrapper/src/main/java/com/snowflake/ingest/sdk/server/StreamingIngestJavaService.java
```diff
@@ -232,6 +232,32 @@ public HttpResponse insertRow(
...
+  @Post("/clients/{clientId}/channels/{channelName}/insertRows")
+  public HttpResponse insertRows(
+      @Param("clientId") String clientId,
+      @Param("channelName") String channelName,
+      @RequestObject java.util.List<Map<String, Object>> rows,
+      @Param("offsetToken") String offsetToken)
+      throws Exception {
+    SnowflakeStreamingIngestChannel channel = getChannelOrThrow(clientId, channelName);
+
+    // Insert all rows with the same offset token
```
sfc-gh-skarlapudi, line 244: This doesn't look right. Offset tokens are supposed to be incremental.
In v2 for insertRows we have start and end offset tokens. Please check with Toby on what we should do here for v1.

sfc-gh-tzhang: We only have end offset token in V1. In the V2 test, are we passing offset tokens as input or we keep incrementing it inside this function?

sdk-wrapper/src/main/java/com/snowflake/ingest/sdk/server/StreamingIngestJavaService.java
```diff
@@ -232,6 +232,32 @@ public HttpResponse insertRow(
...
+    for (Map<String, Object> row : rows) {
+      InsertValidationResponse response = channel.insertRow(row, offsetToken);
```
sfc-gh-tzhang, line 246: We have an insertRows API, you don't need to do it one row at a time

sdk-wrapper/src/main/java/com/snowflake/ingest/sdk/server/StreamingIngestJavaService.java
```diff
@@ -232,6 +232,32 @@ public HttpResponse insertRow(
...
+      if (response.hasErrors()) {
+        throw response.getInsertErrors().get(0).getException();
```
sfc-gh-tzhang, line 248: When you create the channel, you can specify the ON_ERROR option and if we do ABORT, it will throw the exception for you.

sdk-wrapper/src/main/java/com/snowflake/ingest/sdk/server/StreamingIngestJavaService.java
sfc-gh-dshah: I'm handling it using a JMeter function, following the same approach we used in V2.
The offset token for subsequent requests is derived from the last row count of the previous response.
I've tested this in JMeter and confirmed it works as expected
for e.g
//localhost:61225/clients/client_1749677420542/channels/channel_local_sdk_v1_2chan_20s_20480kbps_3rps_ts_1749677411841_node_7abf6195_4fde_476c_9f70_6227397cdb4e_thread_1/insertRows?offsetToken=0
//localhost:61225/clients/client_1749677420542/channels/channel_local_sdk_v1_2chan_20s_20480kbps_3rps_ts_1749677411841_node_7abf6195_4fde_476c_9f70_6227397cdb4e_thread_0/insertRows?offsetToken=19069
//localhost:61225/clients/client_1749677420542/channels/channel_local_sdk_v1_2chan_20s_20480kbps_3rps_ts_1749677411841_node_7abf6195_4fde_476c_9f70_6227397cdb4e_thread_0/insertRows?offsetToken=38108

sdk-wrapper/src/main/java/com/snowflake/ingest/sdk/server/StreamingIngestJavaService.java
```diff
@@ -232,6 +232,36 @@ public HttpResponse insertRow(
...
+        "Batch inserting {} rows to channel {}/{}, offsetToken: {}",
+        rows.size(),
+        clientId,
+        channelName,
+        offsetToken);
+
+    InsertValidationResponse response = channel.insertRows(rows, offsetToken);
+    if (response.hasErrors()) {
+      throw response.getInsertErrors().get(0).getException();
+    }
```
sfc-gh-tzhang, line 254: Again, we don't need this if the ON_ERROR option is ABORT

sfc-gh-tzhang (approved):

PR #1012, opened 2025-06-11
SNOW-1798717: Add Snowpipe Streaming tests with DT and verify no duplicate PPN

sfc-gh-tzhang:


src/test/java/net/snowflake/ingest/streaming/internal/it/DynamicTablePartitionIT.java
```diff
@@ -0,0 +1,280 @@
...
+      SnowflakeStreamingIngestChannel dynamicTableChannel =
+          client.openChannel(
+              OpenChannelRequest.builder("dynamic_table_direct_channel")
+                  .setDBName(DB_NAME)
+                  .setSchemaName("public")
+                  .setTableName(DYNAMIC_TABLE_NAME)
+                  .setOnErrorOption(OpenChannelRequest.OnErrorOption.ABORT)
+                  .build());
+
+      // If we reach here, the channel was opened successfully, which might be unexpected
```
sfc-gh-ggeng, line 182: Shouldn't we have a defined behavior for whether we should fail on open channel or insert rows?

sfc-gh-ggeng (approved):
LGTM!

PR #1017, opened 2025-06-17
[SNOW-2134338] Add unit tests for http client retry strategy

sfc-gh-aminyaylov:
## Context

Specifically for channelsOpen API, we see the following flow:
1. API handler calls [executeWithRetries()](https://github.com/snowflakedb/snowflake-ingest-java/blob/1cd64137219da10e001b61aa32d509a9f445704c/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeServiceClient.java#L236), which makes the HTTP call and extracts the response
2. HTTP client uses [getServiceUnavailableRetryStrategy()](https://github.com/snowflakedb/snowflake-ingest-java/blob/1cd64137219da10e001b61aa32d509a9f445704c/src/main/java/net/snowflake/ingest/utils/HttpUtil.java#L227-L280) to determine when to retry server errors

The confusion is that step 1 contains [logic to retry internal code `GENERAL_EXCEPTION_RETRY_REQUEST`](https://github.com/snowflakedb/snowflake-ingest-java/blob/1cd64137219da10e001b61aa32d509a9f445704c/src/main/java/net/snowflake/ingest/streaming/internal/StreamingIngestUtils.java#L137-L144). For this code to be reached, the HTTP client has to receive a `200 OK`.  Any non-200 response that isn't handled by the HTTP client will throw an `IngestResponseException`. In other words, for this code to retry a `GENERAL_EXCEPTION_RETRY_REQUEST`, it must receive a `200 OK` with `GENERAL_EXCEPTION_RETRY_REQUEST` internal error code.

However, our server does not do this. Bad responses are send with a [400, 429, or 503](https://github.com/snowflakedb/snowflake/blob/a294c3141c80c5560615d06dd9ad6ea245265e71/GlobalServices/src/main/java/com/snowflake/resources/streaming/StreamingResource.java#L1822-L1832) error code. Specifically for our `channelOpens` case, throttling will always return `429` (legacy `503` with parameter disabled). This error code is retried by the HTTP client retry strategy and this PR verifies this behavior.

### DIFF
--- a/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
+++ b/src/main/java/net/snowflake/ingest/utils/HttpUtil.java
@@ -6,6 +6,7 @@ package net.snowflake.ingest.utils;
 
 import static net.snowflake.ingest.utils.Utils.isNullOrEmpty;
 
+import com.google.common.annotations.VisibleForTesting;
 import java.security.Security;
 import java.time.Duration;
 import java.time.Instant;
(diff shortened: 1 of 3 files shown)


sfc-gh-tzhang (approved):
Left a comment, PTAL, otherwise LGTM, thanks!

sfc-gh-japatel (approved):
lgtm

PR #1021, opened 2025-06-24
Set version to 4.1.1-SNAPSHOT

sfc-gh-lsembera:


src/main/java/net/snowflake/ingest/connection/RequestBuilder.java
```diff
@@ -110,7 +110,7 @@ public class RequestBuilder {
   // Don't change!
   public static final String CLIENT_NAME = "SnowpipeJavaSDK";
 
-  public static final String DEFAULT_VERSION = "4.1.0";
+  public static final String DEFAULT_VERSION = "4.1.1-SNAPSHOT";
```
sfc-gh-tzhang, line 113: I forgot if this will work or not, please verify

sfc-gh-tzhang (approved):
lgtm, thanks!

PR #1022, opened 2025-06-25
SNOW-2174149: Keep retrying on QUEUE_FULL instead of invalidating channels

sfc-gh-tzhang:
We have seen the issue in the O11y case when the chunk queue is full, channel write starts return QUEUE_FULL error, and it causes the channel to be invalidated. Channel invalidation results in a new set of channel reopen, which makes the issue even worse. 

With this PR:
- We will keep retrying on QUEUE_FULL up to roughly 5mins instead of just a few seconds before invalidating the channel.
- Add a limit on the registration queue and throttle insertRows once it reaches a certain size
- Both parameters are configurable in case there is any regressions.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
@@ -762,12 +762,26 @@ class FlushService<T> {
     boolean throttleOnQueuedTasks = buildAndUpload.getQueue().size() > numProcessors;
     if (throttleOnQueuedTasks) {
       logger.logWarn(
-          "Throttled due too many queue flush tasks (probably because of slow uploading speed),"
+          "Throttled due to too many queue flush tasks (probably because of slow uploading speed),"
               + " client={}, buildUploadWorkers stats={}",
           this.owningClient.getName(),
           this.buildUploadWorkers.toString());
     }
     return throttleOnQueuedTasks;
+  }
+
+  /** Throttle if the number of queued registration blobs is bigger than the max allowed */
+  boolean throttleDueToQueuedRegistrationRequests() {
+    int queueSize = registerService.getBlobsList().size();
+    boolean throttleOnQueueSize =
+        queueSize > this.owningClient.getParameterProvider().getMaxRegistrationQueueSize();
+    if (throttleOnQueueSize) {
+      logger.logWarn(
+          "Throttled due to too many queued registration blobs, client={}, size={}",
+          this.owningClient.getName(),
+          queueSize);
+    }
+    return throttleOnQueueSize;
   }
 
   /** Get whether we're running under test mode */
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
@@ -11,6 +11,7 @@ import static net.snowflake.ingest.utils.Constants.THREAD_SHUTDOWN_TIMEOUT_IN_SE
 import static net.snowflake.ingest.utils.Utils.getStackTrace;
 
 import com.codahale.metrics.Timer;
+import com.google.common.annotations.VisibleForTesting;
 import java.io.IOException;
 import java.lang.management.ManagementFactory;
 import java.security.InvalidAlgorithmParameterException;
@@ -787,5 +788,11 @@ class FlushService<T> {
   /** Get whether we're running under test mode */
   boolean isTestMode() {
     return this.isTestMode;
+  }
+
+  /** Get the register service, used for TEST only */
+  @VisibleForTesting
+  RegisterService<T> getRegisterService() {
+    return this.registerService;
   }
 }
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
@@ -771,18 +771,18 @@ class FlushService<T> {
     return throttleOnQueuedTasks;
   }
 
-  /** Throttle if the number of queued registration blobs is bigger than the max allowed */
-  boolean throttleDueToQueuedRegistrationRequests() {
-    int queueSize = registerService.getBlobsList().size();
-    boolean throttleOnQueueSize =
+  /** Check if the number of queued registration blobs is bigger than the max allowed */
+  boolean isMaxRegistrationQueueSizeExceeded() {
+    int queueSize = registerService.getBlobsListSize();
+    boolean isQueueSizeExceeded =
         queueSize > this.owningClient.getParameterProvider().getMaxRegistrationQueueSize();
-    if (throttleOnQueueSize) {
+    if (isQueueSizeExceeded) {
       logger.logWarn(
-          "Throttled due to too many queued registration blobs, client={}, size={}",
+          "The number of queued registration blobs exceeds the max allowed threshold, client={}, size={}",
           this.owningClient.getName(),
           queueSize);
     }
-    return throttleOnQueueSize;
+    return isQueueSizeExceeded;
   }
 
   /** Get whether we're running under test mode */
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
@@ -778,7 +778,8 @@ class FlushService<T> {
         queueSize > this.owningClient.getParameterProvider().getMaxRegistrationQueueSize();
     if (isQueueSizeExceeded) {
       logger.logWarn(
-          "The number of queued registration blobs exceeds the max allowed threshold, client={}, size={}",
+          "The number of queued registration blobs exceeds the max allowed threshold, client={},"
+              + " size={}",
           this.owningClient.getName(),
           queueSize);
     }
--- a/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
@@ -4,6 +4,7 @@
 
 package net.snowflake.ingest.streaming.internal;
 
+import static net.snowflake.ingest.utils.Constants.INSERT_THROTTLE_MAX_RETRY_COUNT;
 import static net.snowflake.ingest.utils.Constants.RESPONSE_SUCCESS;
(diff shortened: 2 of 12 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java
```diff
@@ -762,16 +763,36 @@ boolean throttleDueToQueuedFlushTasks() {
...
+          "Throttled due to too many queue flush tasks (probably because of slow uploading speed),"
               + " client={}, buildUploadWorkers stats={}",
           this.owningClient.getName(),
           this.buildUploadWorkers.toString());
     }
     return throttleOnQueuedTasks;
   }
 
+  /** Throttle if the number of queued registration blobs is bigger than the max allowed */
+  boolean throttleDueToQueuedRegistrationRequests() {
```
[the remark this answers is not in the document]
sfc-gh-tzhang: done, ptal

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -645,10 +645,13 @@ void registerBlobs(List<BlobMetadata> blobs, final int executionCount) {
...
-                                                    == RESPONSE_ERR_GENERAL_EXCEPTION_RETRY_REQUEST)
-                                            && executionCount
-                                                < MAX_STREAMING_INGEST_API_CHANNEL_RETRY) {
+                                                && executionCount
+                                                    < parameterProvider
+                                                        .getMaxChannelWriteRetryCountOnQueueFull())
+                                            || (channelStatus.getStatusCode()
+                                                    == RESPONSE_ERR_GENERAL_EXCEPTION_RETRY_REQUEST
+                                                && executionCount
+                                                    < MAX_STREAMING_INGEST_API_CHANNEL_RETRY)) {
```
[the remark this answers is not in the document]
sfc-gh-tzhang: done

src/test/java/net/snowflake/ingest/streaming/internal/FlushServiceTest.java
```diff
@@ -1317,6 +1319,31 @@ public void testEncryptionDecryption()
...
+    TestContext<List<List<Object>>> testContext = testContextFactory.create();
+    testContext.setParameterOverride(
+        Collections.singletonMap(ParameterProvider.MAX_REGISTRATION_QUEUE_SIZE, 10));
+    FlushService fs = testContext.flushService;
+    Pair<FlushService.BlobData<StubChunkData>, CompletableFuture<BlobMetadata>> blob =
+        new Pair<>(
+            new FlushService.BlobData<>("test", null),
+            CompletableFuture.completedFuture(new BlobMetadata("path", "md5", null, null)));
+
+    int threshold = 10;
```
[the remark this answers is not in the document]
sfc-gh-tzhang: updated to use the variable

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
```diff
@@ -452,15 +451,17 @@ public Map<String, ColumnProperties> getTableSchema() {
...
-    while ((hasLowRuntimeMemory(memoryInfoProvider)
-            || (this.owningClient.getFlushService() != null
-                && this.owningClient.getFlushService().throttleDueToQueuedFlushTasks()))
-        && retry < INSERT_THROTTLE_MAX_RETRY_COUNT) {
+    while (hasLowRuntimeMemory(memoryInfoProvider)
+        || (this.owningClient.getFlushService() != null
+            && (this.owningClient.getFlushService().throttleDueToQueuedFlushTasks()
+                || this.owningClient
+                    .getFlushService()
+                    .throttleDueToQueuedRegistrationRequests()))) {
```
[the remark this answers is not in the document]
sfc-gh-tzhang: which part are you referring to? i don't think it does more than the description.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
```diff
@@ -452,15 +451,17 @@ public Map<String, ColumnProperties> getTableSchema() {
 
   /**
    * Potentially throttles a call to the `insertRow(s)` APIs if needed. Note that throttling applies
-   * to the calling thread by sleeping on the curernt thread and checking whether memory conditions
+   * to the calling thread by sleeping on the current thread and checking whether memory conditions
    * are suitable to accepting the current row.
    */
   void throttleInsertIfNeeded(MemoryInfoProvider memoryInfoProvider) {
     int retry = 0;
```
sfc-gh-ggeng, line 458: It seems that we are now retrying forever, is this the expected change? I'm curious why we want to do that.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
```diff
@@ -452,15 +451,17 @@ public Map<String, ColumnProperties> getTableSchema() {
...
-                && this.owningClient.getFlushService().throttleDueToQueuedFlushTasks()))
-        && retry < INSERT_THROTTLE_MAX_RETRY_COUNT) {
+    // Insert will be throttled if we run into low memory or the queued flush tasks in flush service
+    // exceeds the threshold or the queued registration requests exceeds the threshold.
+    while (hasLowRuntimeMemory(memoryInfoProvider)
+        || (this.owningClient.getFlushService() != null
+            && (this.owningClient.getFlushService().throttleDueToQueuedFlushTasks()
+                || this.owningClient.getFlushService().isMaxRegistrationQueueSizeExceeded()))) {
       try {
         Thread.sleep(insertThrottleIntervalInMs);
```
sfc-gh-ggeng, line 466: nit: Shall we check whether the check is still valid after the each sleep?

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
sfc-gh-ggeng: I'm thinking if we shall return some throttling error back to the customer after some large amount of time instead of retrying forever.

sfc-gh-ggeng: Looking into other retries we have. Ideally the queue should move after 5 minutes, as we would start to invalidate the channels afterwards. I would suggest to add a cap around 10 minutes. If it is longer, I feel it means that something goes wrong.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -643,12 +643,18 @@ void registerBlobs(List<BlobMetadata> blobs, final int executionCount) {
...
-                                                < MAX_STREAMING_INGEST_API_CHANNEL_RETRY) {
+                                                && executionCount
+                                                    < parameterProvider
+                                                        .getMaxChannelWriteRetryCountOnQueueFull();
+                                        boolean retryGenericRetryableException =
+                                            channelStatus.getStatusCode()
+                                                    == RESPONSE_ERR_GENERAL_EXCEPTION_RETRY_REQUEST
+                                                && executionCount
+                                                    < MAX_STREAMING_INGEST_API_CHANNEL_RETRY;
+                                        if (retryQueueFull || retryGenericRetryableException) {
```
sfc-gh-ggeng, line 657: curious, do you know after we invalidate the local channels after the 5 minute wait due to queue full, will the server side reject the next blob request for those channels (as it has gapped sequencer) or it still return RESPONSE_ERR_ENQUEUE_TABLE_CHUNK_QUEUE_FULL -- would the server side check queue full before or after the sequencer check?

Would checking the whether the channels are still valid before we do each upload helping us drain the queue faster (I'm assuming check the local channel validation is a cheap operation)?

sfc-gh-ggeng (approved):
I'm concerning about the infinite sleep/throttling added in this PR, other parts looks good to me!

sfc-gh-ggeng:
By the way, I saw all windows build fails, but feels it is unrelated to your change

```
2025-07-09T02:24:48.5071767Z [INFO] BUILD FAILURE
2025-07-09T02:24:48.5072320Z [INFO] ------------------------------------------------------------------------
2025-07-09T02:24:48.5074346Z [INFO] Total time:  01:23 h
2025-07-09T02:24:48.5077025Z [INFO] Finished at: 2025-07-09T02:24:48Z
2025-07-09T02:24:48.5077683Z [INFO] ------------------------------------------------------------------------
2025-07-09T02:24:48.5091083Z [ERROR] Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:3.6.0:analyze-only (analyze) on project snowflake-ingest-sdk: Execution analyze of goal org.apache.maven.plugins:maven-dependency-plugin:3.6.0:analyze-only failed: Plugin org.apache.maven.plugins:maven-dependency-plugin:3.6.0 or one of its dependencies could not be resolved:
2025-07-09T02:24:48.5093849Z [ERROR] 	Failed to read artifact descriptor for org.apache.maven.doxia:doxia-sink-api:jar:1.11.1
2025-07-09T02:24:48.5095049Z [ERROR] 	Failed to read artifact descriptor for org.apache.maven.reporting:maven-reporting-api:jar:3.1.1
2025-07-09T02:24:48.5096490Z [ERROR] 	Failed to read artifact descriptor for org.apache.maven.reporting:maven-reporting-impl:jar:3.2.0
2025-07-09T02:24:48.5098211Z [ERROR] 	Failed to read artifact descriptor for org.codehaus.plexus:plexus-archiver:jar:4.7.1
2025-07-09T02:24:48.5099364Z [ERROR] 	Failed to read artifact descriptor for org.codehaus.plexus:plexus-io:jar:3.4.1
2025-07-09T02:24:48.5100169Z [ERROR] -> [Help 1]
```

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestChannelInternal.java
sfc-gh-tzhang: not sure if I understand what you mean, we do re-check as part of the while loop

PR #1029, opened 2025-07-15
SNOW-2205584 Refresh TMK in client on channel reopen

sfc-gh-lsembera:
This PR updates the table key in the per-table key map during channel open. The previous code was only updating the client key storage only when server-side TMK refresh is initiated during enqueue.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
@@ -405,6 +405,25 @@ public class SnowflakeStreamingIngestClientInternal<T> implements SnowflakeStrea
                     : ParquetProperties.WriterVersion.PARQUET_1_0)
             .build();
 
+    FullyQualifiedTableName fullyQualifiedTableName =
+        new FullyQualifiedTableName(
+            response.getDBName(), response.getSchemaName(), response.getTableName());
+    Long encryptionKeyId = response.getEncryptionKeyId();
+    String encryptionKey = response.getEncryptionKey();
+    if (encryptionKeyId != null && encryptionKey != null && !encryptionKey.isEmpty()) {
+      // If per table key cache already contains an entry, update it because the cached entry may be
+      // stale, and it takes priority over the key stored in SnowflakeStreamingIngestChannel object.
+      encryptionKeysPerTable.computeIfPresent(
+          fullyQualifiedTableName,
+          (table, key) ->
+              new EncryptionKey(
+                  fullyQualifiedTableName.getDatabaseName(),
+                  fullyQualifiedTableName.getSchemaName(),
+                  fullyQualifiedTableName.getTableName(),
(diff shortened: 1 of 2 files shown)


sfc-gh-bmikaili (approved):
lgtm

sfc-gh-alhuang (approved):
Looks good, thanks for the fix!

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -405,6 +405,25 @@ public SnowflakeStreamingIngestChannelInternal<?> openChannel(OpenChannelRequest
...
 
+    FullyQualifiedTableName fullyQualifiedTableName =
+        new FullyQualifiedTableName(
+            response.getDBName(), response.getSchemaName(), response.getTableName());
+    Long encryptionKeyId = response.getEncryptionKeyId();
+    String encryptionKey = response.getEncryptionKey();
+    if (encryptionKeyId != null && encryptionKey != null && !encryptionKey.isEmpty()) {
+      // If per table key cache already contains an entry, update it because the cached entry may be
+      // stale, and it takes priority over the key stored in SnowflakeStreamingIngestChannel object.
+      encryptionKeysPerTable.computeIfPresent(
```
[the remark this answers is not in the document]
sfc-gh-lsembera: > Is there any concurrency issue in other places that uses this map now?

@sfc-gh-tzhang There should be no concurrency issues because before the SDK starts to flush, it copies keys from this map to a new map, which doesn't change afterwards. This guarantees consistent view of the keys during the flush process. See [here](https://github.com/snowflakedb/snowflake-ingest-java/blob/88138e781580b44ce9e2d8ee0c05603f4b37740a/src/main/java/net/snowflake/ingest/streaming/internal/FlushService.java#L500-L504).

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -405,6 +405,25 @@ public SnowflakeStreamingIngestChannelInternal<?> openChannel(OpenChannelRequest
                     : ParquetProperties.WriterVersion.PARQUET_1_0)
             .build();
 
+    FullyQualifiedTableName fullyQualifiedTableName =
+        new FullyQualifiedTableName(
+            response.getDBName(), response.getSchemaName(), response.getTableName());
+    Long encryptionKeyId = response.getEncryptionKeyId();
+    String encryptionKey = response.getEncryptionKey();
+    if (encryptionKeyId != null && encryptionKey != null && !encryptionKey.isEmpty()) {
```
sfc-gh-tjones, line 413: Can `encryptionKeyId` be null?

sfc-gh-lsembera: It should not be, this is just to be extra safe.

PR #1031, opened 2025-07-23
SNOW-2218309 Fix upload/registration path mismatch in Iceberg ingestion

sfc-gh-alhuang:
### TL;DR

Improved thread safety for file transfer metadata in streaming ingest by using AtomicReference and adding path validation for Iceberg file uploads.

### Fix
This PR makes sure that 
1. `SnowflakeFileTransferMetadataWithAge` and path of `FileLocationInfo` are always in sync, it's stored in an atomic reference to guarantee thread safety.
2. `BlobPath` is regenerated with cached `SnowflakeFileTransferMetadataWithAge` if check doesn't pass and uploaded with the cached metadata.

Ideally, there should be only one call to `generateBlobPath` per upload, and it should be made within the upload thread to ensure credential refresh is handled consistently in Iceberg case.
However, it turns out we're using the pre-generated blob path as a key for logging and submitting to the registration service. Since Iceberg and BDEC ingestion share the same registration service, I choose to stick to path re-generation. Also the original approach works well for BDEC ingest, since there's no need to refresh the blob path in that case.
In my opinion, if more fix is required and differences between Iceberg / BDEC ingestion continue to grow, we should consider splitting the Iceberg registration logic from the BDEC path to better isolate their behavior.

### Background
#958 fixed the upload registration path mismatch problem by regenerated the upload file path right after `refreshSnowflakeMetadata` happens. However, the fix doesn't include the scenario that the refresh could happen in other thread right after the blob path is regenerated, thus causing `FileLocationInfo` and `BlobPath` mismatch. Where `FileLocationInfo` and `SnowflakeFileTransferMetadataWithAge` is already refreshed and `BlobPath` is not. The file transfer agent then concatenated volume path in `FileLocationInfo` and suffix of `BlobPath` thus upload the file to unexpected location.


src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
```diff
@@ -408,19 +417,34 @@ private SnowflakeFileTransferMetadataV1 copyFileTransferMetadata(
...
-    this.fileTransferMetadataWithAge.fileTransferMetadata.getStageInfo().setPresignedUrl("");
-
-    Map<?, ?> stageCredentials =
-        this.fileTransferMetadataWithAge.fileTransferMetadata.getStageInfo().getCredentials();
-    HashMap<String, String> invalidStageCredentials = new HashMap<>();
-    for (Object key : stageCredentials.keySet()) {
-      invalidStageCredentials.put((String) key, "");
+    SnowflakeFileTransferMetadataWithAge snowflakeFileTransferMetadataWithAge =
+        this.metadataRef.get();
+    if (snowflakeFileTransferMetadataWithAge != null) {
```
sfc-gh-tzhang, line 422: when this can be null?

src/main/java/net/snowflake/ingest/utils/Utils.java
```diff
@@ -477,4 +477,19 @@ public static String stripTrailingNulls(String key) {
...
     return end == key.length() ? key : key.substring(0, end);
   }
+
+  /**
+   * Extracts the file name from a path string.
+   *
+   * @param path the file path (e.g., "dir1/dir2/file.txt")
+   * @return the file name (e.g., "file.txt") or null if path is null/empty
+   */
+  public static String extractFileName(String path) {
```
sfc-gh-tzhang, line 487:
```suggestion
  public static String extractFileShortName(String path) {
```

sfc-gh-tzhang (commented):
Left some comments, PTAL, also would be good to get a review from @sfc-gh-dorlovsky if possible

src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
```diff
@@ -150,30 +151,35 @@ private Optional<IcebergPostUploadMetadata> putRemote(BlobPath blobPath, byte[]
...
         /*
-         * Since we can have multiple calls to putRemote in parallel and because the metadata
-         * includes the file path we use a copy for the upload to prevent us from using the wrong
-         * file path.
+         * Since we can have multiple calls to putRemote in parallel which might refresh the metadata (including volume prefix in Iceberg upload)
+         * we need to check if the blob path is up to date. If not, we regenerate the blob path.
          */
+        SnowflakeFileTransferMetadataWithAge snowflakeFileTransferMetadataWithAge =
+            this.metadataRef.get();
+        if (!isBlobPathUpToDate(blobPath, snowflakeFileTransferMetadataWithAge)
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Swapped.

src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
```diff
@@ -255,13 +262,13 @@ synchronized SnowflakeFileTransferMetadataWithAge refreshSnowflakeMetadata(boole
     FileLocationInfo location =
         this.owningManager.getRefreshedLocation(this.tableRef, Optional.empty());
     setFileLocationInfo(location);
-    return this.fileTransferMetadataWithAge;
+    return this.metadataRef.get();
   }
 
-  private synchronized void setFileLocationInfo(FileLocationInfo fileLocationInfo)
+  @VisibleForTesting
+  public synchronized void setFileLocationInfo(FileLocationInfo fileLocationInfo)
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Renamed to `setMetadataRef`

src/test/java/net/snowflake/ingest/streaming/internal/it/SubscopedTokenRefreshIT.java
```diff
@@ -134,6 +138,114 @@ public void testTokenExpire() throws Exception {
...
+     * Create a race condition test with two threads:
+     * Thread 1: Continuously sets new file location info - simulates concurrent metadata refresh
+     * Thread 2: Performs insertions that trigger putRemote calls
+     * This tests the race condition where metadata is refreshed by another thread
+     * after blob path is generated and passed to putRemote. The possibility of no refresh happens is
+     * 1 / numFileLocationInfos^rowCount which is 1 / 5^10 ~ 1e-7.
+     */
+    final int rowCount = 10;
+    final int numFileLocationInfos = 5;
+    final java.util.concurrent.atomic.AtomicBoolean testRunning =
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Modified import.

src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
```diff
@@ -75,8 +76,8 @@ public class InternalStage implements IStorage {
   private final Properties proxyProperties;
   private final boolean useIcebergFileTransferAgent;
 
-  private FileLocationInfo fileLocationInfo;
-  private SnowflakeFileTransferMetadataWithAge fileTransferMetadataWithAge;
+  private final AtomicReference<SnowflakeFileTransferMetadataWithAge> metadataRef =
```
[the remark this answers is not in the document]
sfc-gh-alhuang: Making `SnowflakeFileTransferMetadataWithAge` and `FileLocationInfo` to `AtomicReference`. While non-iceberg ingestion's file location remain unchanged, changing this won't change the behavior of non-iceberg ingestion.

src/main/java/net/snowflake/ingest/utils/Utils.java
sfc-gh-alhuang: Renamed.

src/test/java/net/snowflake/ingest/streaming/internal/SubscopedTokenExternalVolumeManagerTest.java
```diff
@@ -94,7 +94,7 @@ public void testGetStorageWithoutRegister() {
   @Test
   public void testGenerateBlobPathPublic() {
     manager.registerTable(new TableRef("db", "schema", "table"));
-    BlobPath blobPath = manager.generateBlobPath("db.schema.table");
+    BlobPath blobPath = manager.generateBlobPath("db.schema.table", /*pathOverride*/ null);
```
sfc-gh-tzhang, line 97: Do we have any tests for the path override?

src/test/java/net/snowflake/ingest/streaming/internal/it/SubscopedTokenRefreshIT.java
```diff
@@ -134,6 +140,111 @@ public void testTokenExpire() throws Exception {
...
+     * after blob path is generated and passed to putRemote. The possibility of no refresh happens is
+     * 1 / numFileLocationInfos^rowCount which is 1 / 5^10 ~ 1e-7.
+     */
+    final int rowCount = 10;
+    final int numFileLocationInfos = 5;
+    final AtomicBoolean testRunning = new AtomicBoolean(true);
+    final AtomicReference<Exception> threadException = new AtomicReference<>();
+
+    /* Pre-fetch FileLocationInfo objects using getRefreshedLocation logic */
+    final java.util.List<FileLocationInfo> fileLocationInfos = new java.util.ArrayList<>();
```
sfc-gh-tzhang, line 179: move the import to the top

sfc-gh-tzhang (approved):
lgtm, thanks!

src/test/java/net/snowflake/ingest/streaming/internal/it/SubscopedTokenRefreshIT.java
sfc-gh-alhuang: Moved.

PR #1076, opened 2025-09-25
SNOW-2302576 Honor the encryption option set on the external volume

sfc-gh-alhuang:
### Content
Support server side KMS encryption when uploading the file to external volume on AWS and GCP.

### Background
An encryption option can be specified on an external volume on top of a storage bucket. However, SSV1 Iceberg ingestion does not honor the option value because AES_256_SERVER_SIDE_ENCRYPTION (SSE_S3) is hardcoded in the SDK code.

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FileLocationInfo.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FileLocationInfo.java
@@ -65,7 +65,8 @@ public class FileLocationInfo {
   @JsonProperty("encryptionKmsKeyId")
   private String encryptionKmsKeyId;
 
-  String getLocationType() {
+  @VisibleForTesting
+  public String getLocationType() {
     return locationType;
   }
 
@@ -161,7 +162,8 @@ public class FileLocationInfo {
     this.useVirtualUrl = useVirtualUrl;
   }
 
-  String getVolumeEncryptionMode() {
+  @VisibleForTesting
+  public String getVolumeEncryptionMode() {
     return this.volumeEncryptionMode;
   }
 
@@ -169,7 +171,8 @@ public class FileLocationInfo {
     this.volumeEncryptionMode = volumeEncryptionMode;
   }
 
-  String getEncryptionKmsKeyId() {
+  @VisibleForTesting
+  public String getEncryptionKmsKeyId() {
     return this.encryptionKmsKeyId;
   }
 
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FileLocationInfo.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FileLocationInfo.java
@@ -60,7 +60,7 @@ public class FileLocationInfo {
   private boolean useVirtualUrl;
 
   @JsonProperty("volumeEncryptionMode")
-  private String volumeEncryptionMode;
+  private VolumeEncryptionMode volumeEncryptionMode;
 
   @JsonProperty("encryptionKmsKeyId")
   private String encryptionKmsKeyId;
@@ -163,11 +163,11 @@ public class FileLocationInfo {
   }
 
   @VisibleForTesting
-  public String getVolumeEncryptionMode() {
+  public VolumeEncryptionMode getVolumeEncryptionMode() {
     return this.volumeEncryptionMode;
   }
 
-  void setVolumeEncryptionMode(String volumeEncryptionMode) {
+  void setVolumeEncryptionMode(VolumeEncryptionMode volumeEncryptionMode) {
     this.volumeEncryptionMode = volumeEncryptionMode;
   }
 
--- a/src/main/java/net/snowflake/ingest/streaming/internal/FileLocationInfo.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/FileLocationInfo.java
@@ -59,6 +59,12 @@ public class FileLocationInfo {
   @JsonProperty("useVirtualUrl")
   private boolean useVirtualUrl;
 
+  @JsonProperty("volumeEncryptionMode")
+  private String volumeEncryptionMode;
+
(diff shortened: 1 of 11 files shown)


src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
```diff
@@ -75,6 +75,7 @@ public class InternalStage implements IStorage {
   // Proxy parameters that we set while calling the Snowflake JDBC to upload the streams
   private final Properties proxyProperties;
   private final boolean useIcebergFileTransferAgent;
+  private final FileLocationInfo fileLocationInfo;
```
sfc-gh-cqu, line 78: I don't see this new field gets refreshed in `refreshSnowflakeMetadata`. From the public doc, customers can alter the encryption setup of their external volumes. What will happen then? Is the `refreshSnowflakeMetadata` call every 1h sufficient for handling that case? https://docs.snowflake.com/en/sql-reference/sql/alter-external-volume

src/main/java/net/snowflake/ingest/streaming/internal/FileLocationInfo.java
```diff
@@ -154,4 +161,22 @@ boolean getUseVirtualUrl() {
   void setUseVirtualUrl(boolean useVirtualUrl) {
     this.useVirtualUrl = useVirtualUrl;
   }
+
+  @VisibleForTesting
+  public String getVolumeEncryptionMode() {
+    return this.volumeEncryptionMode;
+  }
+
+  void setVolumeEncryptionMode(String volumeEncryptionMode) {
```
sfc-gh-cqu, line 170: Are the settors used anywhere? If not, let's remove them.

src/main/java/net/snowflake/ingest/streaming/internal/fileTransferAgent/IcebergStorageClientFactory.java
```diff
@@ -58,13 +61,15 @@ public IcebergStorageClient createClient(StageInfo stage, int parallel)
             stage.getRegion(),
             stage.getEndPoint(),
             stage.getIsClientSideEncrypted(),
-            useS3RegionalUrl);
+            useS3RegionalUrl,
```
sfc-gh-cqu, line 64: Since client-side encryption is not supported for Iceberg tables, can we add a comment for that? If we want to clean it up, let's file another PR.

sfc-gh-cqu (commented):
Left a question about how we deal with changes with external volume encryption settings. PTAL, thanks.

src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
sfc-gh-alhuang: Yes this is going to have problem with the 1hr refresh rate of `refreshSnowflakeMetadata`, but I don't think we want to refresh metadata to fetch the latest encryption config in this case.

src/main/java/net/snowflake/ingest/streaming/internal/FileLocationInfo.java
sfc-gh-alhuang: Afaik the setters are used for Jackson json serilazation.

src/main/java/net/snowflake/ingest/streaming/internal/InternalStage.java
sfc-gh-cqu: Server-side check resulting in channel invalidation makes sense to me.

src/main/java/net/snowflake/ingest/streaming/internal/fileTransferAgent/IcebergGCSClient.java
```diff
@@ -39,21 +40,29 @@ class IcebergGCSClient implements IcebergStorageClient {
...
-   *                required to decrypt/encrypt content in stage
+   * @param volumeEncryptionMode the volume encryption mode (e.g., "GCS_SSE_KMS")
+   * @param encryptionKmsKeyId the KMS key ID for encryption when using GCS_SSE_KMS
    */
-  public static IcebergGCSClient createSnowflakeGCSClient(StageInfo stage)
+  public static IcebergGCSClient createSnowflakeGCSClient(
+      StageInfo stage, String volumeEncryptionMode, String encryptionKmsKeyId)
       throws SnowflakeSQLException {
     IcebergGCSClient sfGcsClient = new IcebergGCSClient();
+    sfGcsClient.volumeEncryptionMode = volumeEncryptionMode;
```
sfc-gh-ggeng, line 64: Should we verify that when volumeEncryptionMode is set to GCS_SSE_KMS, encryptionKmsKeyId should not be null?

src/main/java/net/snowflake/ingest/streaming/internal/fileTransferAgent/IcebergGCSClient.java
```diff
@@ -278,7 +287,12 @@ private Blob uploadWithDownScopedToken(
             .build();
 
     try {
-      return gcsClient.create(blobInfo, content);
+      // Use KMS encryption if specified, otherwise use default GCS encryption
+      return (GCS_SSE_KMS.equals(this.volumeEncryptionMode)
+              && this.encryptionKmsKeyId != null
+              && !this.encryptionKmsKeyId.isEmpty())
```
sfc-gh-ggeng, line 293: If we assert the encryptionKmsKeyId can not be null/empty if GCS_SSE_KMS is set, we don't need the other two conditions after &&

src/main/java/net/snowflake/ingest/streaming/internal/fileTransferAgent/IcebergGCSClient.java
sfc-gh-alhuang: Added verification.

PR #1078, opened 2025-10-14
[SNOW-2464956] Close client on primary deployment change

sfc-gh-aminyaylov:
## Problem

When the primary deployment URL is updated, the client throws a `CHANNEL_INVALID`​ error. The user responds by re-opening the channel, which succeeds. However, ingestion continues to fail due to the deployment mismatch.

## Solution

Client is automatically closed if it detects the primary deployment has changed, forcing the user to recreate the client.

The following errors indicate a primary deployment reconfiguration and will immediately trigger a client close:

1. **Client-side deployment ID mismatch**: we are uploading to a different bucket location. The bucket location change occurs during periodic storage credential refresh, which calls Snowflake using the current URL.
2. **Server-side** **deployment** **ID** **mismatch**: we are registering a file with the wrong deployment
3. **Server-side** **encryption** **key** **mismatch**: we are registering a file with an invalid encryption key

### DIFF
--- a/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
+++ b/src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
@@ -880,7 +880,7 @@ public class SnowflakeStreamingIngestClientInternal<T> implements SnowflakeStrea
         && ire.getErrorBody().getCode() != null
         && ire.getErrorBody()
             .getCode()
-            .equals(
+            .equalsIgnoreCase(
                 String.valueOf(
                     StreamingIngestResponseCode.ERR_INVALID_ENCRYPTION_KEY.getStatusCode()));
   }


src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -781,38 +846,112 @@ List<BlobMetadata> getRetryBlobs(
...
+  /**
+   * Reconfigure the client when a client-level configuration mismatch is detected.
+   * 1. Invalidates all existing channels
+   * 2. Discards all in-flight work
+   * 3. Reconfigures storage manager and flush service
+   *
+   * @param reason the reason for reconfiguration
+   * @throws SFException if reconfiguration fails (client will be closed)
+   */
+  void reconfigureClient(String reason) {
```
sfc-gh-hmadan: Not convinced that we need to build this self-reconfigure smartness into the SDK. Its totally OK to return a persistent error to the SDK caller that the client is in an invalid state, and they're now expected to close and reopen the client + reopen all channels. Lets chat if you're seeing any strong requirements for going down this road!

sfc-gh-aminyaylov: You'll be happy to learn that's exactly what we're doing: https://app.graphite.dev/github/pr/snowflakedb/snowflake-kafka-connector/1197/SNOW-2193898-Recreate-closed-streaming-ingest-client

sfc-gh-aminyaylov: Also updated this PR to immediately close the client if it detects one of these terminal conditions. PTAL

src/test/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientTest.java
```diff
@@ -1382,4 +1382,169 @@ public void testGetLatestCommittedOffsetTokens() throws Exception {
...
+                + "        \"message\" : \"Deployment ID mismatch\",\n"
+                + "        \"channel\" : \"%s\",\n"
+                + "        \"client_sequencer\" : %d\n"
+                + "      } ]\n"
+                + "    } ]\n"
+                + "  } ]\n"
+                + "}",
+            dbName, schemaName, tableName, channelName, channelSequencer);
+
+    apiOverride.addSerializedJsonOverride(
```
sfc-gh-hmadan: Lets add a true E2E for validating this change works as expected, without any mocks in the picture?
Please check with Alec on how to do so. SDK E2E will only work against prod, by the way.

src/main/java/net/snowflake/ingest/streaming/internal/SnowflakeStreamingIngestClientInternal.java
```diff
@@ -816,6 +829,28 @@ public void close() throws Exception {
...
 
+  /**
+   * Mark the client as closed due to unrecoverable errors. Once closed, all subsequent API calls
+   * will throw CLOSED_CLIENT error.
+   *
+   * @param reason the reason for closure
+   */
+  void closeImmediately(String reason) {
+    logger.logError("Closing client due to terminal error: {}", reason);
+    isClosed = true;
```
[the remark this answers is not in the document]
sfc-gh-aminyaylov: Refactored to call the existing close API without flushing remaining data.

PR #1160, opened 2026-06-25
Upgrade to parquet 1.17.1 + iceberg 1.11.0 on Java 17 (5.0.0)

sfc-gh-miqbal:
## Summary
Upgrades the streaming ingest SDK to **parquet 1.17.1** and **iceberg 1.11.0**, which moves the language level to **Java 17** (iceberg 1.11 ships Java 17 bytecode). Bumps the artifact version to **5.0.0**.

## Changes
- **Versions:** parquet 1.14.4 → 1.17.1, iceberg 1.6.1 → 1.11.0, SDK 4.4.3 → 5.0.0
- **Java 17:** compiler source/target, CI build matrix, and README updated; consumer JDK 8/11 dropped from the `e2e-jar-test` matrix (the artifact is now Java 17)
- **Dependency convergence** (enforcer `RequireUpperBoundDeps`): avro 1.12.1, jackson 2.21.3, slf4j 2.0.17, snappy 1.1.10.7, commons-compress 1.28.0, commons-codec 1.19.0, commons-io 2.20.0
- **Shading / licensing:** relocate `org.locationtech.jts` (pulled in by iceberg) into the shaded namespace and register its EDL 1.0 license; accept the bare `MIT` license name (slf4j 2.0)
- **Tests on Java 17:** run on the classpath (`useModulePath=false`) so the SDK's `org.apache.parquet.*` helper classes resolve under the module system; add the `--add-opens` flags required by PowerMock / JDBC / JWT reflection
- **Build plugins:** align `maven-failsafe-plugin` to 3.2.5 (matches surefire)
- **CI dependency resolution:** route through the Google-hosted Maven Central mirror to avoid `repo.maven.apache.org` rate-limiting on the runners

## Compatibility
Breaking: the SDK is now a **Java 17** artifact — consumers on Java 8/11 must upgrade.

## Notes

Had to route to GCS because Github checks kept getting rate limited when downloading artifacts, if everything looks good I can remove that but checks may fail without that

The few failing checks I think are flaky, e.g. even this job from 4.4.3 release PR has a jacoco error that is similar to the reason one of the checks is failing https://github.com/snowflakedb/snowflake-ingest-java/actions/runs/25461822030/job/74705663103?pr=1152

🤖 Generated with [Claude Code](https://claude.com/claude-code)


.github/workflows/End2EndTest.yml
```diff
@@ -39,7 +39,7 @@ jobs:
       # Don't cancel other jobs in the matrix if one fails
       fail-fast: false
       matrix:
-        java: [ 8 ]
+        java: [ 17 ]
```
sfc-gh-alhuang, line 42: I thought we discussed that parquet 1.17 requires java 11+. Can we use the minimum java for merge gate check?

e2e-jar-test/run_e2e_jar_test.sh
```diff
@@ -2,6 +2,14 @@
 
 set -euo pipefail
 
+# Resolve dependencies via the Google-hosted Maven Central mirror (see
```
sfc-gh-alhuang: @sfc-gh-ggeng I remember you deal with this issue before, any insights here?

e2e-jar-test/run_e2e_jar_test.sh
```diff
@@ -29,12 +37,12 @@ if [[ "${test_type}" == "shaded" ]]; then
   project_version=$(./scripts/get_project_info_from_pom.py pom.xml version)
   ./scripts/update_project_version.py public_pom.xml $project_version > generated_public_pom.xml
 
-  # Build shaded SDK always with Java 8
+  # Build shaded SDK always with Java 17
```
sfc-gh-alhuang, line 32: Ditto, let's use java 11.

pom.xml
```diff
@@ -36,47 +36,49 @@
     <url>https://github.com/snowflakedb/snowflake-ingest-java/tree/master</url>
   </scm>
 
-  <!-- Set our Language Level to Java 8 -->
+  <!-- Set our Language Level to Java 17 -->
   <properties>
     <amazonaws.version>1.12.655</amazonaws.version>
     <apache.httpclient.version>4.5.14</apache.httpclient.version>
     <apache.httpcomponents.version>4.4.16</apache.httpcomponents.version>
+    <argLine/>
```
sfc-gh-alhuang, line 49: Can you add a brief comment why we need these args?

sfc-gh-alhuang (commented):
LGTM overall, question on the minimum java version, PTAL.

.github/workflows/End2EndTest.yml
sfc-gh-miqbal: Iceberg 1.11, and then later upgrade to 1.12 which we are planning, will require java 17

e2e-jar-test/run_e2e_jar_test.sh
sfc-gh-miqbal: Please let me know if I should keep this in this PR, or if you'd like to address this separately. Only added as  a workaround to be able to test

e2e-jar-test/run_e2e_jar_test.sh
sfc-gh-miqbal: Same comment as above with Iceberg 1.11 requiring Java 17

pom.xml
sfc-gh-miqbal: Added a comment above the property in 1ca910ed. Short version: on Java 17 the module system blocks the deep reflection that PowerMock, the snowflake-jdbc driver, and JWT/RSA key handling do into java.base internals (each package maps to a concrete test that was throwing InaccessibleObjectException). It is applied via the surefire/failsafe argLine only, so it is test-only and not in the shipped jar.

.github/workflows/End2EndTest.yml
sfc-gh-alhuang: Sounds good, let's bump to java 17 then. cc @sfc-gh-hmadan @sfc-gh-ggeng

e2e-jar-test/run_e2e_jar_test.sh
sfc-gh-alhuang: Discussed with @sfc-gh-ggeng offline, let's not include this in the PR. The 429 issue can be mitigated by reruns.

.github/workflows/End2EndTest.yml
sfc-gh-ggeng: can we get @sfc-gh-hmadan 's approve on this. I though previously we only want to bump to java 11. I remember we do have quite some customers on java 11

sfc-gh-alhuang: Update: ATM SSv2 minimum support version is java 11+, discussing on this. https://snowflake.slack.com/archives/C08E56WB06N/p1782510903007879

e2e-jar-test/run_e2e_jar_test.sh
sfc-gh-miqbal: have removed

